PR #13, opened 2020-04-07
8242210: [lworld] TestCallingConvention::test36 spuriously fails

TobiHartmann:
Problems:
- G1 pre- and post-barriers used when (un-)packing arguments for the calling convention can call into the runtime which screws up argument registers. Save all registers until JDK-8232094 is fixed in mainline (it's the slow path anyway).
- SignatureStream::as_value_klass triggers a SystemDictionary lookup which acquires the ProtectionDomainSet_lock. When used from fieldDescriptor::print_on_for when some debug printing flags are enabled, this conflicts with the tty_lock. We should simply use get_value_field_klass instead. Also, we should handle null as a vale for non-flattened fields.
- TraceDeoptimization needs to handle re-allocation of the inline type return value.

I've also added a new StressCC option to the ValueTypeTest suite to randomly restrict some compilation to C1 and thereby stress test the calling convention.

---------
### Progress

### Issue
 * [JDK-8242210](https://bugs.openjdk.java.net/browse/JDK-8242210): [lworld] TestCallingConvention::test36 spuriously fails

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/valhalla pull/13/head:pull/13`
`$ git checkout pull/13`


TobiHartmann:
Thanks, Frederic!

PR #24, opened 2020-04-20
allow inline classes to have abstract supers

hseigel:
Plesae review this change to allow inline classes to have super classes that are abstract classes.  This fix does not test that the abstract classes are 'special'.  That will be done in a future release.

Also, fix verification issue with assigning an inline type to an object of one of its super classes.

Tests will be added once javac support for inline types having abstract classes is merged in.

Thanks, Harold

---------
### Progress

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/valhalla pull/24/head:pull/24`
`$ git checkout pull/24`

### DIFF
--- a/src/hotspot/share/classfile/classFileParser.cpp
+++ b/src/hotspot/share/classfile/classFileParser.cpp
@@ -6927,11 +6927,14 @@ void ClassFileParser::post_process_parsed_stream(const ClassFileStream* const st
       return;
     }
 
-    // For a value class, only java/lang/Object is an acceptable super class
+    // For an inline class, only java/lang/Object or special abstract classes
+    // are acceptable super classes.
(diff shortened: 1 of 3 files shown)


hseigel:
Hi Fred,

Thanks for reviewing this change.

I have not created the new CR, but will shortly.

The changes to lines 179-181 should not trigger more class loading than 
today.  Whenever the verifier needs to check if type A is assignable to 
type B, it has to load both types, unless they are the same, or type B 
is java.lang.Object or an interface. However, this load would not be 
needed if the only valid super class for an inline type was 
java.lang.Object.

Thanks, Harold

fparain:
Harold, 

Thank you for the clarifications.

Fred

PR #27, opened 2020-04-27
8243627: [lworld] Co-evolve jdk-valhalla tests along with JDK-8237072

sadayapalam:
Mandy,

May I request you to review these changes to jdk-valhalla tests ? 

of the inline class.

I'll push them after your review and *after* JDK-8237072 itself is pushed.
Thanks in advance.

---------
### Progress

### Issue
 * [JDK-8243627](https://bugs.openjdk.java.net/browse/JDK-8243627): [lworld] Co-evolve jdk-valhalla tests along with JDK-8237072

### Reviewers
 * Mandy Chung ([mchung](@mlchung) - Committer)

### Download
`$ git fetch https://git.openjdk.java.net/valhalla pull/27/head:pull/27`
`$ git checkout pull/27`

### DIFF
--- a/test/jdk/valhalla/valuetypes/ArrayElementVarHandleTest.java
+++ b/test/jdk/valhalla/valuetypes/ArrayElementVarHandleTest.java
@@ -156,7 +156,7 @@ public class ArrayElementVarHandleTest {
             Point.makePoint(100, 200)
     };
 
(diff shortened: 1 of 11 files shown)


mlchung:
Looks good to me.

PR #29, opened 2020-04-27
8243623: [lworld] Syntax and other mechanical changes in langtools tests for JDK-8237072

sadayapalam:
Jim,

May I request you to review these changes to langtools tests that anticipate and
align with JDK-8237072 ? These changes are mechanical in nature and I decided it
is best to separate them from the RFR for JDK-8237072 itself to help create sharper
focus on the important pieces there. (Non-mechanical meaningful test changes with
be included in that RFR itself)

of the inline class.

I'll push them after your review and after JDK-8237072 itself is pushed.
Thanks in advance.

---------
### Progress

### Issue
 * [JDK-8243623](https://bugs.openjdk.java.net/browse/JDK-8243623): [lworld] Syntax and other mechanical changes in langtools tests for JDK-8237072

### Reviewers
 * JimLaskey (no known github.com user name / role)

### Download
`$ git fetch https://git.openjdk.java.net/valhalla pull/29/head:pull/29`
`$ git checkout pull/29`

### DIFF
--- a/test/langtools/tools/javac/valhalla/lworld-values/ArrayCreationWithQuestion.java
+++ b/test/langtools/tools/javac/valhalla/lworld-values/ArrayCreationWithQuestion.java
@@ -24,7 +24,7 @@
 /*
  * @test
  * @bug 8222634
-     * @summary Check array creation with V and V.ref
+ * @summary Check array creation with V and V.ref
  * @modules jdk.compiler/com.sun.tools.javac.util jdk.jdeps/com.sun.tools.javap
  * @compile ArrayCreationWithQuestion.java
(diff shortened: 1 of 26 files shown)


JimLaskey:
test/langtools/tools/javac/valhalla/lworld-values/ArrayCreationWithQuestion.java:27 
    Indentation (tab vs space)
test/langtools/tools/javac/valhalla/lworld-values/CheckQuestionInMessages.java:13
    X.ref[] xa = new Object[10]; should it be X.ref[] xa = new Object.ref[10]; I'm sure it is correct (Object.ref doesn't exist) but there is an inconsistency with old code. Maybe add a comment.

Otherwise, good.

PR #30, opened 2020-04-27
8243625: [lworld] Co-evolve hotspot-compiler tests along with JDK-8237072

sadayapalam:
Tobias,

May I request you to review these changes to hotspot-compiler tests that anticipate
and align with JDK-8237072 ? These changes are mechanical in nature and I decided it
is best to separate them from the RFR for JDK-8237072 itself to help create sharper
focus on the important pieces there. 

of the inline class.

I'll push them after your review and after JDK-8237072 itself is pushed.
Thanks in advance.

---------
### Progress

### Issue
 * [JDK-8243625](https://bugs.openjdk.java.net/browse/JDK-8243625): [lworld] Co-evolve hotspot-compiler tests along with JDK-8237072

### Reviewers
 * Tobias Hartmann ([thartmann](@TobiHartmann) - Committer) ⚠️ Review applies to 035020abf2c49a259eeb294abb05d41672464a50

### Download
`$ git fetch https://git.openjdk.java.net/valhalla pull/30/head:pull/30`
`$ git checkout pull/30`


TobiHartmann:
Here are some more adjustments needed:
[compilerTestFixes.txt](https://github.com/openjdk/valhalla/files/4558969/compilerTestFixes.txt)

All the remaining failures are JIT bugs. I'll work on them next.

PR #32, opened 2020-04-30
8237072: [lworld] Add support for denoting and deriving the reference projection

sadayapalam:
Hello Maurizio & Jim,

    May I request you to please review this patch for valhalla inline types support
under the new scheme ? TIA.

I.   What does this patch do?
II.  What can be skipped in the review? 
III. Recommended order for review.
IV.  Known problems, limitations and deferred tasks.
V.   Report to language designers on open problems.

I. What does this patch do:

    - From every inline class declaration V, derive two classes V and V.ref
    - These two are subtypes at the VM level, but are related (only) by inline
      widening conversion and narrowing conversion at the language level.
    - The two types have the same fields and methods with the same shape and
      accessibility rules.
    - Add source level support for V.ref and V.val notation.
    - Get rid of the experimental support for generics over values. Until we
      figure out the full story with generics and inlines, carrying along this
      experimental support is becoming untenable/unsustainable.
    - Get rid of the experimental support added for noncovariant value arrays
    - Get rid of LW2's "Nullable Projections" including the V? syntax

II. These files can be skipped as they simply revert change and align 
    with origin/master: (roll back V and V? nullable projections of LW2)

    JavacParser.java
    AttrContext.java
    JCTree.java
    Pretty.java
    Printer.java
    RichDiagnosticsFormatter.java
    TaskFactory.java
    TreeCopier.java
    TypePrinter.java

III. Recommended order for review:

    - Type.java, Symbol.java

            Look at the new internal APIs to query/compute various projections.
            Every class type, class symbol, method and field support an API
            to return the "other" projection i.e its doppleganger in the alternate
            universe.

            Most crucially ClassSymbol.referenceProjection()

    - Attr.java:

            Source support for .ref/.val

    - MemberEnter.java:
    - TypeEnter.java:
    - TransTypes.java:

            Co-evolve V.val and V.ref in lock step (also TransValues.java)

    - TransValues.java:

            Fold all V.ref.member accesses to ((V) V.ref).member access.

    - Resolve.java:

            Changes to make sure method look up works OK in a world where values are islands.

    - Types.java:
  
            Implement inline widening/narrowing relationship.

    - ProjectionRelationsTest.java:

            Verify all relationships between V and V.ref and V[] and V.ref[]

    - ClassWriter.java:
            
            Dual class generation scheme with subtyping relationship at the binary/VM level.

    - ClassReader.java:

            Merge the dual classes back, sever the subtyping relationship and make
            them operate as a pair of classes that can convert to each other.

    - New tests:
    
            Most importantly ProjectionRelationsTest.java

    - Other files.

 
IV. There are many known issues that have been deliberately deferred to a later iteration:

    - With Brian's consent I am using a simpler translation strategy than what is
      outlined in the SoV documents. These are good enough for now, but when
      VBC migration is undertaken, will have to enhanced.
    - No support for ref-default and val-default classes yet.
    - No support for class hierarchy sealing.
    - You can't do new V.ref() (V.ref is abstract) although SoV calls for it.
    - Handling of .ref and .val is quick and dirty; Need revisit.
    - The nest mate related attributes are not fully properly emitted as of now
    - Need to verify that attributes from V are not carried over inadvertantly to
      V$ref.class
    - Class hierarchy walking in a world where inline types are islands calls for
      a type switch. I have done this only in places uncovered by existing tests.
      We need to go through various utility methods (similar to what is done in
      Symbol.java and Resolve.java) to make sure these changes are consistently
      applied.
    - Diamond inference with value classes crashes now (implementation creates factory
      methods, projections are missing and need to be created)

V. Problems in the interplay of inlines types with generics and inference:

    Removing the experimental support for generics over values results in several
constructs that used to compile earlier (albeit only with -XDallowGenericsOverValues)
not compiling anymore.

    These expose some issues that feed into the language design. We need to evolve
a short term answer (so that the JDK components tests that are impacted don't get
blocked) and a long term one (the real solution)

Here is a snippet that captures these problems:
```
import java.util.Arrays;

interface I {}

public inline class X implements I {

    int x = 10;

    void problem_01() {
        X [] a1 = new X[0];
        X [] a2 = Arrays.copyOf(a1, a2.length, X[].class);
      
/*
error: incompatible types: inference variable T has incompatible bounds
        X [] a2 = Arrays.copyOf(a1, a2.length, X[].class);
                                         ^
    lower bounds: X,Object
    lower bounds: X$ref
  where T,U are type-variables:
    T extends Object declared in method <T,U>copyOf(U[],int,Class<? extends T[]>)
    U extends Object declared in method <T,U>copyOf(U[],int,Class<? extends T[]>)
1 error
*/
    }

    void foo(Class<? extends I> p) {}

    void problem_02() {
        foo(X.class);
/*
error: incompatible types: Class<X> cannot be converted to Class<? extends I>
        foo(X.class);
*/
    }

    String data() {
        return null;
    }

    // Problem: 3, we infer a stream of X.ref's causing
    // the method reference to not type check.
    void unbound_lookup_with_projected_receiver() {
        X [] a = new X[0];
        Arrays.stream(a).map(X::data).filter(p -> p != null).forEach(System.out::println);
/*
error: incompatible types: invalid method reference
        Arrays.stream(a).map(X::data).filter(p -> p != null).forEach(System.out::println);
                             ^
    method data in class X cannot be applied to given types
      required: no arguments
      found:    X$ref
      reason: actual and formal argument lists differ in length
*/
    }

    void problem_04() {
/*
    test/hotspot/jtreg/runtime/valhalla/valuetypes/UnsafeTest.java:125: warning: [removal] putObject(Object,long,Object) in Unsafe has been deprecated and marked for removal
            U.putObject(v, off_o, List.of("Value1", "Value2", "Value3"));
             ^
/home/srikanth/valhalla/test/valhalla/test/hotspot/jtreg/runtime/valhalla/valuetypes/UnsafeTest.java:127: error: method valueHeaderSize in class Unsafe cannot be applied to given types;
            U.putInt(v, off_v + off_i - U.valueHeaderSize(Value2.class), 999);
                                         ^
  required: Class<V>
  found:    Class<Value2>
  reason: inference variable V has incompatible bounds
    equality constraints: Value2
    lower bounds: Object
  where V is a type-variable:
    V extends Object declared in method <V>valueHeaderSize(Class<V>)
*/
    }

    void problem_05() {
/*
test/hotspot/jtreg/compiler/valhalla/valuetypes/TestIntrinsics.java:119: error: incomparable types: Class<CAP#1> and Class<MyValue1$ref>
        boolean check2 = MyValue1.class.asIndirectType().getSuperclass() == MyValue1.ref.class;
                                                                         ^
  where CAP#1 is a fresh type-variable:
    CAP#1 extends Object super: MyValue1 from capture of ? super MyValue1
*/
    }

    }
}
```

---------
### Progress

### Issue
 * [JDK-8237072](https://bugs.openjdk.java.net/browse/JDK-8237072): [lworld] Add support for denoting and deriving the reference projection

### Reviewers
 * JimLaskey (no known github.com user name / role)

### Download
`$ git fetch https://git.openjdk.java.net/valhalla pull/32/head:pull/32`
`$ git checkout pull/32`


JimLaskey:
src/jdk.compiler/share/classes/com/sun/tools/javac/code/Symbol.java:546
    `outermost' - mix of quotes.
src/jdk.compiler/share/classes/com/sun/tools/javac/code/Symbol.java:1319
    `other' projection: If `this' - mix of quotes.
src/jdk.compiler/share/classes/com/sun/tools/javac/code/Symbol.java:1669 
     // TODO: TYP?, CLINT? - are you tracking?
src/jdk.compiler/share/classes/com/sun/tools/javac/code/Symbol.java:1672
    I haven't seen enough Valhalla code to fully comment but 'r' and 'v' prefixes for 'ref' and 'val' might be too subtle. One lower case letter in an overloaded variable name is visually harder to discern. (Brevity is not the next readers friend.)
src/jdk.compiler/share/classes/com/sun/tools/javac/code/Symbol.java:1688
    I think this.flatname.append('$', this.name.table.names.ref); should be broken out into two? support methods.
src/jdk.compiler/share/classes/com/sun/tools/javac/code/Symbol.java: 1715
    The `other' projection: If `this' - mix of quotes.
src/jdk.compiler/share/classes/com/sun/tools/javac/code/Symbol.java:1805
   Just wonder if this common projection pattern (p ? p.x : null) could be broken out into support methods.
src/jdk.compiler/share/classes/com/sun/tools/javac/code/Symbol.java:1967
    The `other' projection: If `this' - mix of quotes (hate when cut and paste propagates the problem.)
src/jdk.compiler/share/classes/com/sun/tools/javac/code/Type.java:1027
    `other' projection: If `this' - mix of quotes
src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java:1048
   What if isValue(es) and !isValue(et)? Doesn't really follow the pattern of the old code.
src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java:1234
   Same question as above.
src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java:1627
   Tracking unspecified behaviour?
src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java:2194
    Comment indent :-)
src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java:4074
    Naively: What happens if rs.resolveIdent does raise an exception? sym undefined? exception propagated?
src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java:4197
    Similarly to above.
src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Check.java:901
    Tracking tolerate Value.class for now?
src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Resolve.java:420
    Is there a plan to handle projection conversation without modifying the tree?

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Symbol.java
```diff
@@ -426,6 +418,29 @@ public boolean isValue() {
...
+    public Symbol valueProjection() {
+        return null;
+    }
+
+    /**
+     * Return the reference projection IFF `this' happens to be value projection, null
+     * otherwise.
+     */
+    public Symbol referenceProjection() {
+        return null;
```
mcimadamore, line 441: How much would it help to have this return `this` ? It seems to me that there a lot of `if <value> ... referenceProjection()` going around.

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Type.java
```diff
@@ -1161,6 +1182,54 @@ public boolean isReference() {
...
+        }
+
+        @Override
+        public boolean isReferenceProjection() {
+            return tsym != null && tsym.isReferenceProjection();
+        }
+
+        @Override
+        public Type valueProjection() {
+            if (!isReferenceProjection() || projection !=  null)
```
mcimadamore, line 1197: Question here - since you have two symbols, getting a reference projection and value projection shouldn't be as simple as doing `tsym.XYZprojection().type` ? Why do we need to create new types here?

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
```diff
@@ -1771,6 +1756,14 @@ public Boolean visitClassType(ClassType t, Type s) {
                 }
 
                 if (s.hasTag(CLASS) || s.hasTag(ARRAY)) {
+                    if (isValue(t)) {
+                        // (s) Value ? == (s) Value.ref
```
mcimadamore, line 1760: I didn't get these comments

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
```diff
@@ -2202,12 +2191,9 @@ public Type visitClassType(ClassType t, Symbol sym) {
                 if (t.tsym == sym)
                     return t;
 
-                /* For inline types, the wired in super type is j.l.O.
-                   So we need a special check for V <: V?
-                */
-                if (t.tsym == sym.nullFreeTypeSymbol()) {
-                    return new ClassType(t.getEnclosingType(), t.getTypeArguments(), (TypeSymbol)sym, t.getMetadata());
-                }
+               // No man may be an island, but the bell tolls for a value.
```
mcimadamore, line 2194: :-)

src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/ClassReader.java
```diff
@@ -2601,6 +2601,32 @@ private void readClassBuffer(ClassSymbol c) throws IOException {
     }
 
     public void readClassFile(ClassSymbol c) {
+        readClassFileInternal(c);
+        if (c.isValue()) {
```
mcimadamore, line 2605: This approach leads to very compact code - but in javac-land it's a bit odd that we have to prune away members at this late stage. I wonder if some of the pruning could happen earlier (e.g. in TransValues?)

test/langtools/tools/javac/valhalla/lworld-values/InferredValueParameterizationTest.java
```diff
@@ -1,13 +1,40 @@
 /*
- * @test /nodynamiccopyright/
- * @bug 8210346
- * @summary inferred value typed `type arguments' are allowed by Javac even without -XDallowGenericsOverValues
- * @compile/fail/ref=InferredValueParameterizationTest.out -XDrawDiagnostics -XDdev InferredValueParameterizationTest.java
+ * Copyright (c) 2020, Oracle and/or its affiliates. All rights reserved.
```
mcimadamore, line 2: The golden file seems to still be there  - even though this is now a positive test

src/jdk.compiler/share/classes/com/sun/tools/javac/resources/compiler.properties
```diff
@@ -3701,7 +3701,7 @@ compiler.warn.this.exposed.prematurely=\
 
 # 0: type
 compiler.err.generic.parameterization.with.value.type=\
-    Inferred type {0} involves generic parameterization by an inline type; This is allowed only with -XDallowGenericsOverValues
+    Inferred type {0} involves generic parameterization by an inline type
```
mcimadamore, line 3704: Do you have examples when this can happen? Or is this a leftover (in which case I think Check.java could be cleaned up a bit)

test/langtools/tools/javac/valhalla/lworld-values/ProjectionRelationsTest.java
```diff
@@ -0,0 +1,279 @@
+/*
```
mcimadamore, line 1: You might want to look at this test framework:

test/langtools/tools/lib/types/TypeHarness.java

An example of such a test is:

test/langtools/tools/javac/types/PrimitiveConversionTest.java

fparain:
I've tried to reproduce the checkcast issue mentioned by Mauricio with this code:
```
public class Test2 {
    static inline class V {
	int y = 52;
    }

    static void m(V.ref vr) {
	V v = vr;
    }

	m(null);
    }
}
```
But I got an NPE where is was expected:
```
$ ./build/pull32/jdk/bin/java Test2
Exception in thread "main" java.lang.NullPointerException
	at Test2.m(Test2.java:7)
	at Test2.main(Test2.java:11)
```

mcimadamore:
> But I got an NPE where it was expected:

My bad - I realized that the test I put together was only compiling and not executing - sorry for the noise.

mcimadamore:
> // Problem: 3, we infer a stream of X.ref's causing
>     // the method reference to not type check.
>     void unbound_lookup_with_projected_receiver() {
>         X [] a = new X[0];
>         Arrays.stream(a).map(X::data).filter(p -> p != null).forEach(System.out::println);
> /*
> error: incompatible types: invalid method reference
>         Arrays.stream(a).map(X::data).filter(p -> p != null).forEach(System.out::println);
>                              ^
>     method data in class X cannot be applied to given types
>       required: no arguments
>       found:    X$ref
>       reason: actual and formal argument lists differ in length

For this one, I think something that an approach that could be discussed would be to consider the method reference `V::m`, where `V` is an inline class, in the same way as we consider `List::add`. That is, we have `class name::member name`.

In such cases we infer the right parameterization from the context; here it would be nice if we could infer the right val/ref from the context too.

mcimadamore:
Another odd test:

```
static inline class V  {
        int y = 52;

        class Bar { }
        static class Baz { }
    }

        new V.ref().new Bar();
        new V().new Bar();
        V.Baz baz1; 
        V.ref.Baz baz2;
    }
```

This gives:

```
error: cannot find symbol
        new V.ref().new Bar();
                        ^
  symbol:   class Bar
  location: class V$ref
error: cannot find symbol
        V.ref.Baz baz2;
             ^
  symbol:   class Baz
  location: class V$ref
2 errors
```
Now, I don't know what's the status of inner classes and values - but this seems to suggest that member types are not copied from one scope to another (since the nested names are only resolved from the `val` projection.

sadayapalam:
> Overall this looks very good. Note that changes to dot apply cleanly on latest lworld branch, so you might need some rebasing.

Thanks for the review, Maurizio!

(I will look into rebasing, At the time I raised the PR request it was even with openjdk/valhalla)

> I'm impressed at how much we could get done with relatively small set of changes in the core classes. As discussed offline, my mine gripe with the proposed implementation approach is that, I think, it distorts the user model view a little. In my mental model:

[...]

> * to help translation, a single inline class will be translated into a ref/val  classes pair - which means a single inline class will give rise to _two_ runtime types
> 
> In other words, I think that appealing with symmetry with the generic use case, where a single declared class give rise to many types might be cleaner in the long run than having two declared classes (as in the proposed implementation).

I agree this suggestion if it is implementable, has promising upside potential to make the code relatively simpler.

I chose the present approach because:

* some of the state that (naturally) today belong in the class symbol abstraction needs to be distinct between the two projections. (name, flags, type ...)

* Various parts of javac code compare tsyms and assume the types are same if they are same (on the non-parameterized type path)

* there are places where we operates only on symbols and it is not clear we will unambiguously know the symbol ownership in those places.

These may all be solvable or worked around satisfactorily.

I agree that upside potential is promising enough to explore the presently unknown downside potential - JDK-8244227 is raised to explore that, to prototype it so we can make an informed call. 


> This choice surfaces up in a bunch of places where we have relationship which are seemingly defined on symbols (e.g. `Symbol::outermostClass`, and friends) in which we have to always ask first whether the class symbol we are dealing with is a _reference_ class symbol, and, if not, go back to that. This choice also backfires when it comes to keep the members of the two fictional symbols in sync - so there's a number of places where we have to update the scope of one class in the same way we update the scope of the other class.
> 

I disagree with the vocabulary. It is not backfiring! It is additional work we knowingly sign up for :) But yes, it would be good if we don't have to.

I must mention here that the LW2 implementation actually used two ClassSymbol shells that shared the members_field. I had to opt for  a deeper copy to solve some problem I ran into which I am unable to recollect at my advanced geriatric stage :) 
 
> Note: all these things are not blockers - but to me they might be signs that some alternate implementation might be possible.
> 
> The current approach will require adjustments in places where symbols are expected, but something like `V.ref` is found - for instance I put together this fictional example:
> 
> ```
> package a;
> 
> import a.V.ref;
> 
> inline class V {
>    int x = 0;
> }
> ```
> 
> and I got this:
> 
> ```
> error: import requires canonical name for V$ref
> import a.V.ref;
>           ^
> ```
> 
> which seems a bit confusing (more generally I note that the compiler is often emitting V$ref and V$val which should be normalized to the syntax used by developers - e.g. dropping the dollar sign).
> 
I have added this case to JDK-8244229 Thanks!


> I'm afraid that when the time will come and we will start looking at the annotation processing machinery, the choice of having two symbols (hence two `Elements`) for the `val` and `ref` projection might become untenable (since at that point I believe the annotation processor user would expect to get two different `TypeMirror`s for the projections which point to the same underlying declared `Element`). Since `Element` is 1-1 with `Symbol`, this might become a problem.

Hmm. As discussed offline, we may have to use some smoke and mirrors anyways to make them appear as one symbol when dealing with binary classes from classpath.

> On a different topic, I noted something on name resolution:
> 
> ```
> inline class V {
>    int y = 52;
> 
>    static class ref {
>        int x;
>    }
> 
>    void m(V.ref vref) {
>        vref.x = 12;
>    }
> }
> ```
> 
> Which gives:
> 
> ```
> error: cannot find symbol
>        vref.x = 12;
>            ^
>   symbol:   variable x
>   location: variable vref of type V$ref
> ```
> 
> In other words, the compiler is always resolving `.ref` to a projection - which is fine for now, but there's a deeper question as to whether we're ok with this behavior and if we want to even allow defining a nested class called `ref` or `val` inside an inline class.

I have added this test case to JDK-8244229 Thanks!

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Symbol.java
sadayapalam: It could result in cleaner code. I have this one JDK-8244232 to cover it.

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Type.java
sadayapalam: FTR, this is to preserve any parameterizations. The value projection of V.ref<String> is V.val<String> and not V.val<T>

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
sadayapalam: It is saying the predicate is a value type is castable to a type s is the same as the predicate is its reference projection castable to s.

src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/ClassReader.java
sadayapalam: Actually, this code is trying to remove V$ref.class from its owner (the package). The class finder would have from the class path found V.class and V$ref.class and added them to the known classes. When V.class is completed and we find that it is a value class, V.class has a "wrong" super type from javac pov. It need to read V$ref.class and pick up the "real" super type from there. In that process it needs to remove V$ref.class from the package owner.

test/langtools/tools/javac/valhalla/lworld-values/InferredValueParameterizationTest.java
sadayapalam: Thanks, will remove it!

src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/ClassReader.java
mcimadamore: Actually I meant to comment on ClassWriter and clicked on the wrong thing!

src/jdk.compiler/share/classes/com/sun/tools/javac/resources/compiler.properties
sadayapalam: We had a discussion about possibly turning it into a throw. I think let us leave it in for now till the generics story is more evolved and we know where we are heading.

sadayapalam: No we don't know where this code could be exercised. All the inference scenarios are covered in the test case - which is why it had to be changed into positive test - So this could technically go away - only reason is to hold onto it to detect any surprises and to wait for generics story to crystalize.

test/langtools/tools/javac/valhalla/lworld-values/ProjectionRelationsTest.java
sadayapalam: Thanks. This could come in handy for future. Appreciated.

sadayapalam:
> Another odd test:
> 
> ```
> static inline class V  {
>         int y = 52;
> 
>         class Bar { }
>         static class Baz { }
>     }
> 
>     public static void main(String[] args) {
>         new V.ref().new Bar();
>         new V().new Bar();
>         V.Baz baz1; 
>         V.ref.Baz baz2;
>     }
> ```
> 
> This gives:
> 
> ```
> error: cannot find symbol
>         new V.ref().new Bar();
>                         ^
>   symbol:   class Bar
>   location: class V$ref
> error: cannot find symbol
>         V.ref.Baz baz2;
>              ^
>   symbol:   class Baz
>   location: class V$ref
> 2 errors
> ```

This is acknowledged as a known TODO in 
com.sun.tools.javac.code.Symbol.ClassSymbol#referenceProjection

I have raised JDK-8244233to cover it,

src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/ClassReader.java
sadayapalam: Actually we don't prune anything, We simply don't write the methods. But carry along the symbols as we must because there could be other files that are in the pipeline, (from same TU or otherwise)

sadayapalam:
> ```
> src/jdk.compiler/share/classes/com/sun/tools/javac/code/Symbol.java:546
>     `outermost' - mix of quotes.
> src/jdk.compiler/share/classes/com/sun/tools/javac/code/Symbol.java:1319
>     `other' projection: If `this' - mix of quotes.
> src/jdk.compiler/share/classes/com/sun/tools/javac/code/Symbol.java: 1715
>     The `other' projection: If `this' - mix of quotes.
> src/jdk.compiler/share/classes/com/sun/tools/javac/code/Symbol.java:1967
>     The `other' projection: If `this' - mix of quotes (hate when cut and paste propagates the problem.)
> src/jdk.compiler/share/classes/com/sun/tools/javac/code/Type.java:1027
>     `other' projection: If `this' - mix of quotes

I have cleaned up these, but this is a common coding/commenting convention in Javac. See that in the same files Symbol.java & Type.java, there are about 20 other unrelated places where quotes in this `style' exist :)

> src/jdk.compiler/share/classes/com/sun/tools/javac/code/Symbol.java:1669 
>      // TODO: TYP?, CLINT? - are you tracking?

Yes: JDK-8244233

> src/jdk.compiler/share/classes/com/sun/tools/javac/code/Symbol.java:1672
>     I haven't seen enough Valhalla code to fully comment but 'r' and 'v' prefixes for 'ref' and 'val' might be too subtle. One lower case letter in an overloaded variable name is visually harder to discern. (Brevity is not the next readers friend.)

Agreed it is poor choice. Fixed by expanding to ref* and val*

> src/jdk.compiler/share/classes/com/sun/tools/javac/code/Symbol.java:1688
>     I think this.flatname.append('$', this.name.table.names.ref); should be broken out into two? support methods.
> src/jdk.compiler/share/classes/com/sun/tools/javac/code/Symbol.java:1805
>    Just wonder if this common projection pattern (p ? p.x : null) could be broken out into support methods.

I have left these as is for now, as it is not clear what exactly is being suggested. If the original code is not clear enough, do outline what you have in mind. TIA

> src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java:1048
>    What if isValue(es) and !isValue(et)? Doesn't really follow the pattern of the old code.
> src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java:1234
>    Same question as above.

Basically, we are checking here if T[] <: S[]. This is true if T.ref <: S. Example of such a check being is Point [] <: Object []

If the polarity is reversed and we are checking if Object[] <: Point [], we would come up with
    es = Point
    et =  Object

and return isSubtype(object, Point) 

which will return false which is indeed the right answer for if Object[] <: Point []

> src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java:1627
>    Tracking unspecified behaviour?

Two clarifications: (1) there is no "behavior" here. it is only a comment I have left in for our reference as we are going to hit it the moment we are going to add support for inline type arguments. (2) And because we are going to hit it most definitely as soon as we start working on generics support that there is no ticket for that per se. 
 
> src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java:2194
>     Comment indent :-)

Fixed.

> src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java:4074
>     Naively: What happens if rs.resolveIdent does raise an exception? sym undefined? exception propagated?
> src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java:4197
>     Similarly to above.

The top level exception handling really happens in com.sun.tools.javac.main.Main#compile(java.lang.String[], com.sun.tools.javac.util.Context)
in the various catch blocks between lines 330-356. 


> src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Check.java:901
>     Tracking tolerate Value.class for now?

I have fixed the comment to just say: "tolerate Value.class" dropping the "for now" - The comment can be confusing. We should *always* tolerate and not report error on Value.class - thanks for catching this. (the "for now" originally meant "may be there are other such cases we will discover we may have need to support also" but we have't found anything else that must be tolerated)

> src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Resolve.java:420
>     Is there a plan to handle projection conversation without modifying the tree?
> ```

No, this may not be doable - it is the intrinsic/organic cost of saying inline types are islands (i.e they have no supertypes of any kind in the lang model including jlO and any declared super types). The best that can be done is to ensure (a) the swicheroo happens for the narrowest window needed (b) the tree is restored in a finally block

Many many thanks for the review Jim!

PR #34, opened 2020-05-04
JDK-8243204 add class load time checking of abstract super classes of…

hseigel:
Please review this chagne.  It adds checks of super classes of inline classes.  It checks that an inline class's super class is  either java.lang.Object or an abstract class.  And, if abstract, checks that the super, and its super, etc. do not have instance fields, nor synchronized instance methods, or non-empty constructors, nor constructors with args.

The fix was tested with tiers 1 - 3.

---------
### Progress

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/valhalla pull/34/head:pull/34`
`$ git checkout pull/34`


fparain:
Shouldn't `ClassFileParser::check_super_of_inline_type()` also check that no super classes implement the `java.lang.IdentityObject` interface?

hseigel:
The update to the pull request implements John's "bit A and bit B" idea posted in the java-valhalla slack channel.  Instead of A and B, the flags are named invalid_inline_super and invalid_identity_super.  In this pull request, invalid_identity_super exists but is never set nor read.  That is for a future change.

Flag invalid_inline_super is set to TRUE in a class's InstanceKlass, if, at class load time, the class is determined to be an invalid super class for an inline type.

Checking that the class's constructor is abstract will be done at a future date once there is compiler support.

src/hotspot/share/classfile/classFileParser.cpp
```diff
@@ -6874,6 +6898,38 @@ void ClassFileParser::parse_stream(const ClassFileStream* const stream,
...
+      } else if (method->name() == vmSymbols::object_initializer_name()) {
+        if (method->signature() != vmSymbols::void_method_signature() ||
+            !method->is_vanilla_constructor()) {
+          return true;
+        }
+      }
+    }
+  }
+  return false;
+}
```
fparain, line 6931: It seems this method could be made static.

src/hotspot/share/classfile/classFileParser.cpp
```diff
@@ -6874,6 +6898,38 @@ void ClassFileParser::parse_stream(const ClassFileStream* const stream,
   // all bytes in stream read and parsed
 }
 
+// Return true if the specified class is not a valid super class for an inline type.
+// A valid super class for an inline type is abstract, has no instance fields,
+// does not implement interface java.lang.IdentityObject (checked elsewhere), has
+// an empty body-less no-arg constructor, and no synchronized instance methods.
```
fparain, line 6904: The comment should be more explicit about the fact that it is performing a local test, without considering super types. The final decision taking account super types being performed in `fill_instance_klass()`.

hseigel:
Today's updated change improves the comment for function is_invalid_super_for_inline_type() and makes it a static function.  Also, it moves the InstanceKlass invalid_inline_super and invalid_identity_super flags into misc_flags.

src/hotspot/share/classfile/classFileParser.cpp
hseigel: The latest change updates this comment.

src/hotspot/share/classfile/classFileParser.cpp
hseigel: Done.

src/hotspot/share/oops/instanceKlass.hpp
```diff
@@ -258,6 +258,9 @@ class InstanceKlass: public Klass {
   // _misc_flags.
   bool            _is_marked_dependent;  // used for marking during flushing and deoptimization
 
+  bool _invalid_inline_super;   // if true, invalid super type for an inline type.
+  bool _invalid_identity_super; // if true, invalid super type for an identity type.
```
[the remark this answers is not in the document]
hseigel: Done.  Thanks for the suggestion!

PR #35, opened 2020-05-06
8244414: [lworld] Migrate Valhalla micro benchmarks suite to not use V? syntax

sadayapalam:
Description of changes:

Hi Maruizio,

    Could you please review these assorted changes that block the build of Valhalla micro
benchmarks ? This is a culumative fix for JDK-8244414 and JDK-8244458

    You could skip a review of the benchmark source files themselves - these have only mechanical changes from V? to V.ref except for one case where there was a usage of Node?<K,V> with Node actually being an identity class. We don't support the transformed Node.ref<K,V> for identity classes, so I had to drop the '?'

Problems fixed:

    (i) Resolve.java: Diamond inference did not work for inline type constructors: Basically, the (static factory method equivalent) constructors synthesized during diamond inference did not have their reference projection counterparts and had to be paired up with them.

    (ii) Types.java: LUB computation in the presence of mixed values and non-values should first apply widening conversion.

    (iii) Enter.java: When we recompile files due to annotation processing rounds, discard
any prior state carried on by the reference projection so as to keep it in sync with its peer.
(A separate test will be added separately - for now building microbenchmarks is the test for this)

---------
### Progress

### Issue
 * [JDK-8244414](https://bugs.openjdk.java.net/browse/JDK-8244414): [lworld] Migrate Valhalla micro benchmarks suite to not use V? syntax

### Reviewers
 * Maurizio Cimadamore ([mcimadamore](@mcimadamore) - Committer)

### Download
`$ git fetch https://git.openjdk.java.net/valhalla pull/35/head:pull/35`
`$ git checkout pull/35`

### DIFF
--- a/src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
+++ b/src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
@@ -4000,6 +4000,22 @@ public class Types {
 
         int[] kinds = new int[ts.length];
 
+        boolean haveValues = false;
+        boolean haveRefs = false;
+        for (int i = 0 ; i < ts.length ; i++) {
+            if (ts[i].isValue())
+                haveValues = true;
+            else
+                haveRefs = true;
+        }
+        if (haveRefs && haveValues) {
+            System.arraycopy(ts, 0, ts = new Type[ts.length], 0, ts.length);
+            for (int i = 0; i < ts.length; i++) {
+                if (ts[i].isValue())
+                    ts[i] = ts[i].referenceProjection();
+            }
+        }
+
         int boundkind = UNKNOWN_BOUND;
         for (int i = 0 ; i < ts.length ; i++) {
             Type t = ts[i];
--- a/test/langtools/tools/javac/valhalla/lworld-values/LubWithInlines.java
+++ b/test/langtools/tools/javac/valhalla/lworld-values/LubWithInlines.java
@@ -26,28 +26,24 @@
 /*
  * @test
  * @bug 8244458 8244414
- * @summary Diamond inference does not work with value classes
- * @run main InlineDiamondTest
+ * @summary Check that javac does not crash while computing LUB involving values.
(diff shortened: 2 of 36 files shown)


test/langtools/tools/javac/valhalla/lworld-values/LubWithInlines.java
```diff
@@ -0,0 +1,53 @@
...
+ */
+
+/*
+ * @test
+ * @bug 8244458 8244414
+ * @summary Diamond inference does not work with value classes
+ * @run main InlineDiamondTest
+ */
+
+public class InlineDiamondTest<E> {
```
mcimadamore, line 33: This test seems the same as InlineDiamondTest...

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
```diff
@@ -4000,6 +4000,22 @@ public Type lub(Type... ts) {
 
         int[] kinds = new int[ts.length];
 
+        boolean haveValues = false;
```
mcimadamore, line 4003: If the goal is to fix the behavior of conditional expression typing, I think the fix should happen outside lub(), more specifically in Attr::condType

test/langtools/tools/javac/valhalla/lworld-values/LubWithInlines.java
sadayapalam: Hmmm. I have no idea how this garbling came about. It did exist in proper form some point as I can see from my backups! Anyway thanks for catching this.!

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
sadayapalam: Hmm. - to me it looks more natural to push this down to the service. But I am open to bubbling this up to the client. With your consent, I will combine this with the fix for JDK-8244513

PR #36, opened 2020-05-06
8244519: [lworld] C2 compilation fails with 'monitors must always exist for synchronized methods'

TobiHartmann:
C2 incorrectly assumes that the receiver of a synchronized method of an abstract class can be an inline type and therefore adds a runtime check to the locking code (see 'can_be_value_type' check in PhaseMacroExpand::expand_lock_node). While processing the JVMState of an uncommon trap emitted for that runtime check, we assert because there is no monitor (yet) although the method is synchronized. 

I've modified ciInstanceKlass::can_be_value_klass accordingly and added an to catch similar problems earlier. I've also removed some dead code.

This fix includes the is_vanilla_constructor() fix from JDK-8243204.

---------
### Progress

### Issue
 * [JDK-8244519](https://bugs.openjdk.java.net/browse/JDK-8244519): [lworld] C2 compilation fails with "monitors must always exist for synchronized methods"

### Download
`$ git fetch https://git.openjdk.java.net/valhalla pull/36/head:pull/36`
`$ git checkout pull/36`


TobiHartmann:
Thanks Roland!

PR #50, opened 2020-05-18
8242612: [lworld] Javac should not expressly encode the new super interface types in class files

sadayapalam:
Jim, Could you please review this change ? TIA.

Changes:

    - Withdraw the new top type InlineObject (per design)
    - Do not encode the new top types in class files as this is noisy and
      very disruptive and let the static compiler and the dynamic environments
      inject these types as needed. (since classes compiled with legacy compilers
      will not encode these top types anyway the VM has to resort to injection,
      why not resort to injection for everything is the motivation)
    - Adjust some code that is surprised by super interfaces being empty.

More background is available in https://bugs.openjdk.java.net/browse/JDK-8242612

---------
### Progress

### Issue
 * [JDK-8242612](https://bugs.openjdk.java.net/browse/JDK-8242612): [lworld] Javac should not expressly encode the new super interface types in class files ⚠️ Title mismatch between PR and JBS.

### Reviewers
 * Jim Laskey ([jlaskey](@JimLaskey) - no project role) ⚠️ Review applies to ecb41ea498231c1fb44aa7fd7d2dcd600537a9f3

### Download
`$ git fetch https://git.openjdk.java.net/valhalla pull/50/head:pull/50`
`$ git checkout pull/50`


src/jdk.compiler/share/classes/com/sun/tools/javac/code/Symtab.java
```diff
@@ -274,7 +271,13 @@ public VarSymbol getClassField(Type type, Types types) {
                        I1, I2, ... In, V.class is typed to be Class<? extends Object & I1 & I2 .. & In>
                     */
                     if (type.isValue()) {
-                        Type it = types.makeIntersectionType(((ClassType)type).interfaces_field, true);
+                        ClassType ct = (ClassType) type;
+                        Type it;
+                        if (ct.interfaces_field == null || ct.interfaces_field.isEmpty()) {
```
JimLaskey, line 276: I'm assuming that you are using this to indicate no top interfaces. Is this safe (non-hackish)?

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java
```diff
@@ -2520,9 +2520,17 @@ Type adjustMethodReturnType(Symbol msym, Type qualifierType, Name methodName, Li
                 // as a special case, x.getClass() has type Class<? extends |X|>
                 // Temporary treatment for inline class: Given an inline class V that implements
                 // I1, I2, ... In, v.getClass() is typed to be Class<? extends Object & I1 & I2 .. & In>
-                Type wcb = qualifierType.isValue()
-                              ? types.makeIntersectionType(((ClassType) qualifierType).interfaces_field, true)
-                              : types.erasure(qualifierType);
+                Type wcb;
+                if (qualifierType.isValue()) {
+                    ClassType ct = (ClassType) qualifierType;
+                    if (ct.interfaces_field == null || ct.interfaces_field.isEmpty()) {
```
JimLaskey, line 2526: Same - maybe define a predicate for this.

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Symtab.java
sadayapalam: Yes, it is safe. Basically earlier on all value classes had at least one interface (InlineObject) and possibly more. Now the magic top type InlineObject going away, values may have no super interfaces at all and this would upset makeIntersectionType which first asserts that it has not been called with an empty list of types.

PR #51, opened 2020-05-18
8245216: [lworld] The JVM should inject the IdentityObject interface to types which need it

fparain:
Please review these changes for JDK-8245216.

This code makes the JVM to inject the IdentityObject interface when needed.
The details are in the CR:
https://bugs.openjdk.java.net/browse/JDK-8245216

Thank you,

Fred

---------
### Progress

### Issue
 * [JDK-8245216](https://bugs.openjdk.java.net/browse/JDK-8245216): [lworld] The JVM should inject the IdentityObject interface to types which need it ⚠️ Title mismatch between PR and JBS.

### Reviewers
 * Harold Seigel ([hseigel](@hseigel) - Committer) ⚠️ Review applies to 00cfcb6c42c94186f3786a3b807cfe0daac67f87

### Download
`$ git fetch https://git.openjdk.java.net/valhalla pull/51/head:pull/51`
`$ git checkout pull/51`

### DIFF
--- a/src/hotspot/share/classfile/verificationType.cpp
+++ b/src/hotspot/share/classfile/verificationType.cpp
@@ -64,7 +64,8 @@ bool VerificationType::resolve_and_check_assignability(InstanceKlass* klass, Sym
     // Otherwise, we treat interfaces as java.lang.Object.
     return !from_is_array ||
       this_class == SystemDictionary::Cloneable_klass() ||
-      this_class == SystemDictionary::Serializable_klass();
+      this_class == SystemDictionary::Serializable_klass() ||
+      this_class == SystemDictionary::IdentityObject_klass();
   } else if (from_is_object) {
     Klass* from_class = SystemDictionary::resolve_or_fail(
         from_name, Handle(THREAD, klass->class_loader()),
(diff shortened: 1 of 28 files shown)


fparain:
Got some crashes during testing, investigating...

sadayapalam:
This patch addresses the two langtools failure it should. Thanks!

However, hotspot/jtreg/runtime/valhalla/valuetypes/testSupers/TestSuperClasses.java is still referencing withdrawn classes and fails. Are you able to include a fix for that too ? 

I looked into preparing a patch, but not familiar with jcod files...

fparain:
The PR has bee updated with the following changes:
  - fix JVMTI class file reconstruction to exclude injected interfaces
  - fix java.lang.Class and TestSupers tests
  - fix JVMCI implementation and test

fparain:
Stress testing shown an issue with metaspace allocation/deallocation.
Withdrawing the PR until the issue has been solved.

src/hotspot/share/classfile/verificationType.cpp
```diff
@@ -64,7 +64,8 @@ bool VerificationType::resolve_and_check_assignability(InstanceKlass* klass, Sym
     // Otherwise, we treat interfaces as java.lang.Object.
     return !from_is_array ||
       this_class == SystemDictionary::Cloneable_klass() ||
```
hseigel, line 66: Can you update the comment at line 63 to include java.lang.IdentityObject?

fparain:
Thank you Harold. I've updated the comment in verificationType.cpp in the last commit.

Fred

PR #55, opened 2020-05-26
8245729: [lworld] Adjust JIT code to reference projection changes

TobiHartmann:
Changes:
- Removed all array storage properties related code
- Now using array tags in layout helper to determine flattened/null-free properties
- Removed handling of inline/indirect mirror
- Removed _asPrimaryType/_asIndirectType intrinsics and removed usages of corresponding API from tests
- Refactored loop unswitching code and added more tests
- Removed all special casing in the typesystem that was necessary to support [V <: [V? with the klass for [V not being equal to the klass for [V?

The code that clears the property bits when loading the klass pointer from the metadata is still in but marked with: "// TODO remove clear_prop_bits bits stuff once the runtime does not set it anymore"

---------
### Progress

### Issue
 * [JDK-8245729](https://bugs.openjdk.java.net/browse/JDK-8245729): [lworld] Adjust JIT code to reference projection changes

### Download
`$ git fetch https://git.openjdk.java.net/valhalla pull/55/head:pull/55`
`$ git checkout pull/55`


TobiHartmann:
Thanks Roland!

PR #64, opened 2020-06-05
Rename value type -XX options to inline...

hseigel:
Rename hotspot -XX value type related options to have 'inline type' in their names instead of 'value type'.  Also, changed associated comments.
The changes were tested with mach5 tiers 1-3.

---------
### Progress

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/valhalla pull/64/head:pull/64`
`$ git checkout pull/64`

### DIFF
--- a/src/hotspot/share/opto/callnode.cpp
+++ b/src/hotspot/share/opto/callnode.cpp
@@ -734,7 +734,7 @@ Node *CallNode::match(const ProjNode *proj, const Matcher *match, const RegMask*
     } else {
       // The Call may return multiple values (value type fields): we
       // create one projection per returned values.
-      assert(con <= TypeFunc::Parms+1 || ValueTypeReturnedAsFields, "only for multi value return");
+      assert(con <= TypeFunc::Parms+1 || InlineTypeReturnedAsFields, "only for multi value return");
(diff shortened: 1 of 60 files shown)


src/hotspot/share/opto/callnode.cpp
```diff
@@ -734,7 +734,7 @@ Node *CallNode::match(const ProjNode *proj, const Matcher *match, const RegMask*
     } else {
       // The Call may return multiple values (value type fields): we
```
fparain, line 735: Should the comment say "inline fields" instead of "value type fields"?

src/hotspot/share/runtime/arguments.cpp
```diff
@@ -4139,8 +4139,8 @@ jint Arguments::apply_ergo() {
   }
   if (!EnableValhalla || (is_interpreter_only() && !is_dumping_archive())) {
     // Disable calling convention optimizations if value types are not supported
```
fparain, line 4141: "if value types are not supported" -> "if inline types are not supported"

hseigel:
Thanks Fred!  I'll change the places you pointed out in the next set of rename changes.

PR #77, opened 2020-06-10
8247357: Flattenable field concept needs some cleanup

fparain:
Please review these changes cleaning up the flattenable field concept.
The concept has evolved with time and now all fields with an inline type are by definition flattenable, so the need to have a "flattenable bit" on the side is gone. The changeset contains a mix of renaming and code cleaning.
The changes don't include JIT code, which would be fix in a follow-up patch.

Thank you,

Fred

---------
### Progress

### Issue
 * [JDK-8247357](https://bugs.openjdk.java.net/browse/JDK-8247357): Flattenable field concept needs some cleanup

### Reviewers
 * Harold Seigel ([hseigel](@hseigel) - Committer) ⚠️ Review applies to 5b49d2d7434e24e6d4d965fd009c02d981200593

### Download
`$ git fetch https://git.openjdk.java.net/valhalla pull/77/head:pull/77`
`$ git checkout pull/77`


MrSimms:
+1

> 
> Just wondering if "is_inlined()" wouldn't be an option as well? "Allocated" sounds like there is
> some heap allocation for that field going on.
> 
> Best regards,
> Tobias
>

fparain:
OK, after another round of renaming:
 - "is_inline_type"  is used to test if the type of a field is an inline type or not
 - "is_inlined" is used to test if the field has been flattened in the layout of its container

Both expression are close, and can easily be found together with grep.

I've tried to fix all the comments to align them with the new names of methods and fields.

The term "flattened" is still use for arrays.

fparain:
Harold,

Thank you for reviewing this. I've fixed the typo and changed back the value of JVM_ACC_FIELD_INLINED (it doesn't make a difference right now, but there's no reason to change it).

Fred

PR #83, opened 2020-06-15
rename verifier related value-types, rename JVM_SIGNATURE_VALUETYPE, …

hseigel:
This pull requiest renames JVM_SIGNATURE_VALUETYPE to JVM_SIGNATURE_INLINETYPE, renames verifier related value_type items to inline_type, and has a few other small changes.

---------
### Progress

### Reviewers
 * David Simms ([dsimms](@MrSimms) - Committer)

### Download
`$ git fetch https://git.openjdk.java.net/valhalla pull/83/head:pull/83`
`$ git checkout pull/83`


src/hotspot/share/classfile/verificationType.hpp
```diff
@@ -222,7 +222,7 @@ class VerificationType {
     return ((_u._data & Category2_2nd) == Category2_2nd);
   }
   bool is_reference_check() const { return _u._data == ReferenceQuery; }
-  bool is_valuetype_check() const { return _u._data == ValueTypeQuery; }
+  bool is_inlinetype_check() const { return _u._data == InlineTypeQuery; }
```
fparain, line 225: is_inlinetype_check() -> is_inline_type_check() ?

src/hotspot/share/classfile/verificationType.hpp
```diff
@@ -242,11 +242,11 @@ class VerificationType {
   bool is_double_array() const { return is_x_array(JVM_SIGNATURE_DOUBLE); }
   bool is_object_array() const { return is_x_array(JVM_SIGNATURE_CLASS); }
   bool is_array_array() const { return is_x_array(JVM_SIGNATURE_ARRAY); }
-  bool is_value_array() const { return is_x_array(JVM_SIGNATURE_VALUETYPE); }
+  bool is_inline_array() const { return is_x_array(JVM_SIGNATURE_INLINE_TYPE); }
```
fparain, line 245: The term "inline" alone can be misleading (is it about the element being an inline type, is it about the elements of the array being inlined in the array, is it about the array being inlined in something else, etc).
Here, the check is about the kind of elements stored in the array, so I would suggest "is_inline_type_array()".

src/hotspot/share/classfile/verificationType.hpp
```diff
@@ -263,10 +263,10 @@ class VerificationType {
     return VerificationType(is_long() ? Long_2nd : Double_2nd);
   }
 
-  static VerificationType change_ref_to_valuetype(VerificationType ref) {
+  static VerificationType change_ref_to_inlinetype(VerificationType ref) {
```
fparain, line 266: change_ref_to_inlinetype() -> change_ref_to_inline_type() ?

src/hotspot/share/classfile/verifier.cpp
```diff
@@ -588,9 +588,9 @@ void ErrorContext::stackmap_details(outputStream* ss, const Method* method) cons
 
 // Methods in ClassVerifier
 
-VerificationType reference_or_valuetype(InstanceKlass* klass) {
+VerificationType reference_or_inlinetype(InstanceKlass* klass) {
```
fparain, line 591: reference_or_inlinetype() -> reference_or_inline_type() ?

PR #95, opened 2020-06-22
8248003: [lworld] [lw3] VM crashes when classes with inline type fields are loaded from CDS archive

fparain:
Please review this changeset fixing several issues in CDS related to inline type metadata removal and restoration.

Thank you,

Fred

---------
### Progress

### Issue
 * [JDK-8248003](https://bugs.openjdk.java.net/browse/JDK-8248003): [lworld] [lw3] VM crashes when classes with inline type fields are loaded from CDS archive

### Download
`$ git fetch https://git.openjdk.java.net/valhalla pull/95/head:pull/95`
`$ git checkout pull/95`


fparain:
The fix has been updated according to Ioi's review, and a new test has been added with class redefinition.

Thank you,

Fred

fparain:
Ioi,

Thank you for the second review. I've updated the comment as you suggested.

Fred.

PR #101, opened 2020-06-25
8248251 don't all java.lang.IdentityObject to be redefined or retrans…

hseigel:
Add check to prevent interface java.lang.IdentityObject from being redefined or retransformed.  And, add test case.

---------
### Progress

### Issue
 * [JDK-8248251](https://bugs.openjdk.java.net/browse/JDK-8248251): test IsModifiableClassAgent.java fails trying to retransform java.lang.IdentityObject ⚠️ Title mismatch between PR and JBS.

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/valhalla pull/101/head:pull/101`
`$ git checkout pull/101`

### DIFF
--- a/src/hotspot/share/prims/jvmtiRedefineClasses.cpp
+++ b/src/hotspot/share/prims/jvmtiRedefineClasses.cpp
@@ -292,6 +292,11 @@ bool VM_RedefineClasses::is_modifiable_class(oop klass_mirror) {
     return false;
   }
 
+  // Cannot redefine or retransform interface java.lang.IdentityObject.
+  if (k->name() == vmSymbols::java_lang_IdentityObject()) {
+    return false;
+  }
+
   // Cannot redefine or retransform a hidden or an unsafe anonymous class.
   if (InstanceKlass::cast(k)->is_hidden() ||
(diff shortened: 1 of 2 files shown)


hseigel:
Hi Fred,
Do we need another test besides the test case that was added to test/jdk/java/lang/instrument/IsModifiableClassAgent.java ?
Thanks, Harold

fparain:
Sorry, I miss-read the second file. java.lang.IdentityObject is part of the well-known classes, so it is always loaded, so it covered by the test. Ignore my comment.

Thank you,

Fred

hseigel:
You can redefine / retransform java.lang.Object.

forax:
i'm trying to understand why you can not redefine IdentityObject but your can redefine Object ?

hseigel:
One reason is that interface java.lang.IdentityObject contains no functions so there's nothing to redefine.

forax:
But you don't need a special case for that ?

hseigel:
Our initial decision was that it did not make sense to allow java.lang.IdentityObject to be retransformed.  Would you prefer that a JBS issue be entered to review this decision?

PR #102, opened 2020-07-01
JDK-8247663: [lworld] LambdaMetaFactory should allow an inline value projection type to convert to its reference projection type 

mlchung:
This patch fixes `LambdaMetaFactory` to allow inline value projection type to convert to its reference projection type.

`Class::isAssignableFrom` and `Class::asSubclass` are the APIs to determine if a given type can be safely cast to this Class.

-  Class::isAssignableFrom tests if a given type can be converted to this Class via a widening reference conversion.
- Class::asSubclass tests if a given type can be converted to this Class via a narrowing conversion.

The current thought is to keep  `Class::isAssignableFrom` and `Class::asSubclass` to respect the class file hierarchy and a cast is needed when converting from `V.val` to `V.ref`.   `LambdaMetaFactory` and other frameworks will need to extend the type adaptability check to support inline types.

I considered adding `Class::isConvertibleFrom` method but we should wait for more use cases proving the need of such convenience method.  This check can be done by calling `Class::referenceType` and `Class::valueType`.

---------
### Progress

### Issue
 * [JDK-8247663](https://bugs.openjdk.java.net/browse/JDK-8247663): [lworld] LambdaMetaFactory should allow an inline value projection type to convert to its reference projection type 

### Reviewers
 * Roger Riggs ([rriggs](@RogerRiggs) - Committer) ⚠️ Review applies to 1cde585aedafb87a99136622f229ccbe948ecc5d

### Download
`$ git fetch https://git.openjdk.java.net/valhalla pull/102/head:pull/102`
`$ git checkout pull/102`


test/jdk/valhalla/valuetypes/InlineTypeConversionTest.java
```diff
@@ -0,0 +1,159 @@
...
+
+    @Test
+    public static void inlineNarrowing() throws Throwable {
+        MethodHandles.Lookup lookup = MethodHandles.lookup();
+        MethodHandle mh = lookup.findStatic(InlineTypeConversionTest.class, "widen", methodType(Value.ref.class, Value.class));
+        Object v = mh.invoke(VALUE);
+        assertTrue(v == null);
+        try {
+            Object v1 = mh.invoke((Object)null);
+            assertTrue(false);
```
RogerRiggs, line 87: Isn't this just Assert.fail("message");

src/java.base/share/classes/java/lang/invoke/AbstractValidatingLambdaMetafactory.java
```diff
@@ -366,12 +366,56 @@ private boolean isAdaptableTo(Class<?> fromType, Class<?> toType, boolean strict
...
+     * if {@code toType} is the {@linkplain Class#valueType() value projection type}
+     * of {@code fromType}.
+     * <p>
+     * Otherwise, this method returns {@code false}.
+     *
+     * @param     fromType the {@code Class} object to be converted from
+     * @param     toType the {@code Class} object to be converted to
+     * @return    {@code true} if {@code fromType} can be converted to {@code toType}
+     */
+    private boolean canConvert(Class<?> fromType, Class<?> toType) {
```
RogerRiggs, line 398: Can this be a static method?

src/java.base/share/classes/java/lang/invoke/AbstractValidatingLambdaMetafactory.java
```diff
@@ -366,12 +366,56 @@ private boolean isAdaptableTo(Class<?> fromType, Class<?> toType, boolean strict
...
+        if (toType.isAssignableFrom(fromType)) {
+            return true;
+        }
+
```
RogerRiggs, line 402: Since fromType.valueType is an Optional<Class<?>> the code could be rearranged to use is Present() and not double check for inline so many times.
If it is a performance issue it should be refactored to do the complete check of one case before the other.

test/jdk/valhalla/valuetypes/InlineTypeConversionTest.java
mlchung: Good suggestion.

src/java.base/share/classes/java/lang/invoke/AbstractValidatingLambdaMetafactory.java
mlchung: Yes.  OTOH, this is consistent with the existing convention with other similar methods such as `isAdaptableTo` which are not static.

src/java.base/share/classes/java/lang/invoke/AbstractValidatingLambdaMetafactory.java
mlchung: I simplified the check.  I also uncovered a bug in the spec for `Class::valueType` and `Class::referenceType` and included a spec fix in this patch.

mlchung: @RogerRiggs can you check out the delta patch:
   https://github.com/openjdk/valhalla/pull/102/commits/b295ac571e78c12e8bb80a4568aab0d79cdd3d06

RogerRiggs: Looks good. No more comments.

test/jdk/valhalla/valuetypes/InlineTypeConversionTest.java
```diff
@@ -0,0 +1,159 @@
...
+    public static void inlineNarrowing() throws Throwable {
+        MethodHandles.Lookup lookup = MethodHandles.lookup();
+        MethodHandle mh = lookup.findStatic(InlineTypeConversionTest.class, "widen", methodType(Value.ref.class, Value.class));
+        Object v = mh.invoke(VALUE);
+        assertTrue(v == null);
+        try {
+            Object v1 = mh.invoke((Object)null);
+            fail("Expected NullPointerException but not thrown");
+        } catch (NullPointerException e) {
+            e.printStackTrace();
```
lalithsuresh, line 89: Needn't print stack trace here or below after invoking mh2.

mlchung: It's not needed and it's left-over from my debugging.  I'll clean it up when we touch this file.

PR #104, opened 2020-07-02
8248163: jdb crashes when invoking "lock" command on inline class

fparain:
Please review this patch which fix an issue with the JVMTI method GetObjectMonitorUsage() and inline types.
The behavior of the method when called on an inline type instance is to return a valid result and not throwing an exception.
The result will always show that the monitor is not locked and has never been locked.

A unit test has been added to cover this case.

Thank you,

Fred

---------
### Progress

### Issue
 * [JDK-8248163](https://bugs.openjdk.java.net/browse/JDK-8248163): jdb crashes when invoking "lock" command on inline class

### Reviewers
 * Harold Seigel ([hseigel](@hseigel) - Committer) ⚠️ Review applies to 897ee98e09d94f1bcc236dc5d1612ce820bc3c34

### Download
`$ git fetch https://git.openjdk.java.net/valhalla pull/104/head:pull/104`
`$ git checkout pull/104`


fparain:
Harold,

Thank you for your review.
I've added a comment and fixed the copyright years format.

Regards,

Fred

PR #105, opened 2020-07-06
8242402: multianewarray is missing checks on the bottom class

fparain:
Please review this patch which adds a missing check in array class resolution.

When an array class is resolved and the bottom type is not a primitive type, the envelope of the bottom type signature is removed in order to extract the class name and load it. Currently, there's no check that the kind of the loaded class matches the enveloped stripped from the signature.

The patch adds this check and includes an unit test.

Tested on all all platforms with mach5, tiers 1 to 3.

Thank you,

Fred

---------
### Progress

### Issue
 * [JDK-8242402](https://bugs.openjdk.java.net/browse/JDK-8242402): multianewarray is missing checks on the bottom class

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/valhalla pull/105/head:pull/105`
`$ git checkout pull/105`

### DIFF
--- a/src/hotspot/share/classfile/systemDictionary.cpp
+++ b/src/hotspot/share/classfile/systemDictionary.cpp
@@ -359,6 +359,10 @@ Klass* SystemDictionary::resolve_array_class_or_null(Symbol* class_name,
                                                          protection_domain,
                                                          CHECK_NULL);
     if (k != NULL) {
+      if ((class_name->is_Q_array_signature() && !k->is_inline_klass()) ||
+          (!class_name->is_Q_array_signature() && k->is_inline_klass())) {
+            THROW_NULL(vmSymbols::java_lang_IncompatibleClassChangeError());
+          }
       k = k->array_klass(ndims, CHECK_NULL);
     }
   } else {
--- a/src/hotspot/share/classfile/systemDictionary.cpp
+++ b/src/hotspot/share/classfile/systemDictionary.cpp
@@ -361,7 +361,7 @@ Klass* SystemDictionary::resolve_array_class_or_null(Symbol* class_name,
     if (k != NULL) {
       if ((class_name->is_Q_array_signature() && !k->is_inline_klass()) ||
           (!class_name->is_Q_array_signature() && k->is_inline_klass())) {
-            THROW_NULL(vmSymbols::java_lang_IncompatibleClassChangeError());
+            THROW_MSG_NULL(vmSymbols::java_lang_IncompatibleClassChangeError(), "L/Q mismatch on bottom type");
           }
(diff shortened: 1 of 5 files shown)


src/hotspot/share/classfile/systemDictionary.cpp
```diff
@@ -359,6 +359,10 @@ Klass* SystemDictionary::resolve_array_class_or_null(Symbol* class_name,
                                                          protection_domain,
                                                          CHECK_NULL);
     if (k != NULL) {
+      if ((class_name->is_Q_array_signature() && !k->is_inline_klass()) ||
+          (!class_name->is_Q_array_signature() && k->is_inline_klass())) {
+            THROW_NULL(vmSymbols::java_lang_IncompatibleClassChangeError());
+          }
```
hseigel, line 365: Is it worthwhile using THROW_MSG_NULL to provide a helpful message with the exception?

test/hotspot/jtreg/runtime/valhalla/valuetypes/MultiANewArrayTest/MultiANewArrayTypeCheck.jcod
```diff
@@ -0,0 +1,120 @@
+class MultiANewArrayTypeCheck {
```
hseigel, line 1: I think a copyright notice is needed here.  Also, could you add a comment containing Java pseudo-code that shows what this file is doing?

src/hotspot/share/classfile/systemDictionary.cpp
fparain: Something like this?

`THROW_MSG_NULL(vmSymbols::java_lang_IncompatibleClassChangeError(), "L/Q mismatch on bottom type");`

Fred

hseigel: Yes. That looks good.  Thanks.

test/hotspot/jtreg/runtime/valhalla/valuetypes/MultiANewArrayTest/MultiANewArrayTypeCheck.jcod
fparain: I've added the missing copyrights and a comment explaining the content of the jcod file.

Thank you,

Fred

hseigel: Looks Good.

fparain:
Harold,

Thank you for your review.

Fred

PR #108, opened 2020-07-14
8249396: JNI function IsSameObject should apply the substitutatibility test for inline types

fparain:
Please review this patch which fix the implementation of the JNI function IsSameObject by performing a substitutability test when two references to inline type instances are passed in argument.

Passed tests tiers 1 to 3 with Mach5.

Thank you,

Fred


src/hotspot/share/runtime/jniHandles.cpp
```diff
@@ -325,6 +326,40 @@ bool JNIHandles::current_thread_in_native() {
...
+          // If it is an error, just let it propagate
+          // If it is an exception, wrap it into an InternalError
+          if (!PENDING_EXCEPTION->is_a(SystemDictionary::Error_klass())) {
+            Handle e(THREAD, PENDING_EXCEPTION);
+            CLEAR_PENDING_EXCEPTION;
+            THROW_MSG_CAUSE_(vmSymbols::java_lang_InternalError(), "Internal error in substitutability test", e, false);
+          }
+        }
+        ret = result.get_jboolean();
+    }
```
hseigel, line 357: indentation issue:  could lines 338 - 356 should be shifted 2 spaces to the left.

fparain: Fixed

fparain:
Harold,

Thank you for reviewing.
I've added a comment to JNIHandles::is_same_object(), let me know if it is explicit enough.

Regards,

Fred

src/hotspot/share/runtime/jniHandles.cpp
hseigel: The changes look good!  Thanks for making them.

PR #109, opened 2020-07-15
8249257 Rename ValueKlass to InlineKlass

hseigel:
Please review this tedious change.  It renames class ValueKlass to InlineKlass and renames its fields and methods.  It does not rename ValueArrayKlass.  That will be a future change.

Also, this change does not rename things defined in gc or jit source files.

The change was tested with tiers 1 and 2 on Mac, Windows, and Linux x64, and tiers 3-5 on Linux x64.

Thanks, Harold

---------
### Progress

### Issue
 * [JDK-8249257](https://bugs.openjdk.java.net/browse/JDK-8249257): Rename ValueKlass to InlineKlass

### Reviewers
 * Frederic Parain ([fparain](@fparain) - Committer) ⚠️ Review applies to e832a5c4df356abd91c25255571dccbad6202b2e

### Download
`$ git fetch https://git.openjdk.java.net/valhalla pull/109/head:pull/109`
`$ git checkout pull/109`

### DIFF
--- a/src/hotspot/share/memory/oopFactory.cpp
+++ b/src/hotspot/share/memory/oopFactory.cpp
@@ -39,7 +39,6 @@
 #include "oops/oop.inline.hpp"
 #include "oops/typeArrayKlass.hpp"
 #include "oops/typeArrayOop.inline.hpp"
-#include "oops/inlineKlass.hpp"
 #include "oops/valueArrayKlass.hpp"
 #include "oops/valueArrayOop.inline.hpp"
 #include "oops/valueArrayOop.hpp"
--- a/src/hotspot/share/memory/oopFactory.cpp
+++ b/src/hotspot/share/memory/oopFactory.cpp
@@ -39,7 +39,7 @@
 #include "oops/oop.inline.hpp"
 #include "oops/typeArrayKlass.hpp"
 #include "oops/typeArrayOop.inline.hpp"
-#include "oops/valueKlass.hpp"
+#include "oops/inlineKlass.hpp"
 #include "oops/valueArrayKlass.hpp"
 #include "oops/valueArrayOop.inline.hpp"
 #include "oops/valueArrayOop.hpp"
--- a/src/hotspot/share/oops/inlineKlass.hpp
+++ b/src/hotspot/share/oops/inlineKlass.hpp
@@ -186,9 +186,8 @@ class InlineKlass: public InstanceKlass {
   // Casting from Klass*
   static InlineKlass* cast(Klass* k);
 
-  // Use this to return the size of an instance in heap words
-  // Implementation is currently simple because all inline types are allocated
-  // in Java heap like Java objects.
+  // Use this to return the size of an instance in heap words.
+  // Note that this size only applies to heap allocated stand-alone instances.
   virtual int size_helper() const {
     return layout_helper_to_size_helper(layout_helper());
   }
--- a/src/hotspot/share/oops/inlineKlass.hpp
+++ b/src/hotspot/share/oops/inlineKlass.hpp
@@ -22,87 +22,87 @@
  *
  */
 
-#ifndef SHARE_VM_OOPS_VALUEKLASS_HPP
-#define SHARE_VM_OOPS_VALUEKLASS_HPP
+#ifndef SHARE_VM_OOPS_INLINEKLASS_HPP
+#define SHARE_VM_OOPS_INLINEKLASS_HPP
 
 #include "classfile/javaClasses.hpp"
 #include "oops/instanceKlass.hpp"
 #include "oops/method.hpp"
 //#include "oops/oop.inline.hpp"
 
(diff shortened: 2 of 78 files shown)


src/hotspot/share/memory/oopFactory.cpp
```diff
@@ -39,7 +39,7 @@
 #include "oops/oop.inline.hpp"
 #include "oops/typeArrayKlass.hpp"
 #include "oops/typeArrayOop.inline.hpp"
-#include "oops/valueKlass.hpp"
+#include "oops/inlineKlass.hpp"
```
fparain, line 42: Do we really need this include?
InlineKlass is not used directly in this file, ValueArrayKlass is used and valueArrayKlass.hpp includes inlineKlass.hpp.

src/hotspot/share/oops/inlineKlass.hpp
```diff
@@ -184,10 +184,10 @@ class ValueKlass: public InstanceKlass {
   bool is_inline_klass_slow() const        { return true; }
 
   // Casting from Klass*
-  static ValueKlass* cast(Klass* k);
+  static InlineKlass* cast(Klass* k);
 
   // Use this to return the size of an instance in heap words
-  // Implementation is currently simple because all value types are allocated
+  // Implementation is currently simple because all inline types are allocated
```
fparain, line 190: This comment might need an update, because now, not all inline types are allocated in the Java heap.
May be just specifying this size only applies to heap allocated standalone instances.

src/hotspot/share/opto/escape.cpp
```diff
@@ -3243,12 +3243,12 @@ void ConnectionGraph::split_unique_types(GrowableArray<Node *>  &alloc_worklist,
...
       } else if (use->Opcode() == Op_Return) {
         assert(_compile->tf()->returns_value_type_as_fields(), "must return a value type");
-        // Get ValueKlass by removing the tag bit from the metadata pointer
+        // Get InlineKlass by removing the tag bit from the metadata pointer
         Node* klass = use->in(TypeFunc::Parms);
         intptr_t ptr = igvn->type(klass)->isa_rawptr()->get_con();
         clear_nth_bit(ptr, 0);
         assert(Metaspace::contains((void*)ptr), "should be klass");
-        assert(((ValueKlass*)ptr)->contains_oops(), "returned value type must contain a reference field");
+        assert(((InlineKlass*)ptr)->contains_oops(), "returned value type must contain a reference field");
```
fparain, line 3251: returned value type -> returned inline type

src/hotspot/share/runtime/deoptimization.cpp
```diff
@@ -1395,10 +1393,10 @@ static int reassign_fields_by_klass(InstanceKlass* klass, frame* fr, RegisterMap
 
 // restore fields of an eliminated value type array
```
fparain, line 1394: value type array -> inline type array

src/hotspot/share/runtime/deoptimization.cpp
```diff
@@ -1395,10 +1393,10 @@ static int reassign_fields_by_klass(InstanceKlass* klass, frame* fr, RegisterMap
...
 void Deoptimization::reassign_value_array_elements(frame* fr, RegisterMap* reg_map, ObjectValue* sv, valueArrayOop obj, ValueArrayKlass* vak, TRAPS) {
-  ValueKlass* vk = vak->element_klass();
+  InlineKlass* vk = vak->element_klass();
   assert(vk->flatten_array(), "should only be used for flattened value type arrays");
```
fparain, line 1397: flattened value type array -> flattened inline type array

src/hotspot/share/runtime/sharedRuntime.cpp
```diff
@@ -2835,7 +2835,7 @@ CodeOffsets::Entries CompiledEntrySignature::c1_value_ro_entry_type() const {
 void CompiledEntrySignature::compute_calling_conventions() {
   // Get the (non-scalarized) signature and check for value type arguments
```
fparain, line 2836: value type arguments -> inline type arguments

src/hotspot/share/runtime/signature.hpp
```diff
@@ -573,7 +573,7 @@ typedef GrowableArrayFilterIterator<SigEntry, SigEntryFilter> ExtendedSignature;
 
 // Used for adapter generation. One SigEntry is used per element of
 // the signature of the method. Value type arguments are treated
```
fparain, line 575: Value type argument -> Inline type argument

fparain:
Harold,

Thank you for doing this tedious but necessary renaming job.
Looks good to me, just a few comments, mostly about fixing comments (no need for another review).

Regards,

Fred

hseigel:
Thanks Fred!  I'll fix the issues that you found before integrating the change.
Harold

PR #110, opened 2020-07-16
8249613: [lworld] Use initialization free allocation when copying a value

fparain:
Please review this patch which removes some unnecessary zeroing of heap allocated inline type instances.

Passed tests tier 1 to 3 with Mach5.

Thank you,

Fred

---------
### Progress

### Issue
 * [JDK-8249613](https://bugs.openjdk.java.net/browse/JDK-8249613): [lworld] Use initialization free allocation when copying a value

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/valhalla pull/110/head:pull/110`
`$ git checkout pull/110`


fparain:
Thanks Harold!

Fred

PR #113, opened 2020-07-20
More renaming from value type to inline type

fparain:
Please review this small patch with more renaming of symbols from "value type" to "inline type".

Thank you,

Fred

---------
### Progress

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/valhalla pull/113/head:pull/113`
`$ git checkout pull/113`


fparain:
Thanks Harold!

Fred

PR #114, opened 2020-07-21
JDK-8234364:  [lworld] ValueBootstrapMethods::isSubstitutable0 should do null check and type check

mlchung:
NPE was only reproducible with LW2 but not in LW3.   This patch adds null check and type check in `ValueBootstrapMethods::isSubstitutable0` to ensure the invoker is called only when the parameters are non-null and of the same type.

---------
### Progress

### Issue
 * [JDK-8234364](https://bugs.openjdk.java.net/browse/JDK-8234364): [lworld] ValueBootstrapMethods::isSubstitutable0 should do null check and type check

### Download
`$ git fetch https://git.openjdk.java.net/valhalla pull/114/head:pull/114`
`$ git checkout pull/114`


RogerRiggs:
Looks ok.

PR #115, opened 2020-07-21
jdk-8249555 Rename valueArray* source files to flatArray*

hseigel:
Please review this change which renames the appropriate hotspot runtime source files, methods, fields, etc. from "valueArray*" to "flatArray*".  The change does not rename any gc or jit source files.

The change was tested with tiers 1 and 2 on Windows, Mac, and Linux x64, and tiers 3-5 on Linux x64.

Thanks! Harold

---------
### Progress

### Reviewers
 * Frederic Parain ([fparain](@fparain) - Committer) ⚠️ Review applies to 39a92a9fd9a6daac3367c1e7d48cb53445e4ea31

### Download
`$ git fetch https://git.openjdk.java.net/valhalla pull/115/head:pull/115`
`$ git checkout pull/115`


src/hotspot/cpu/aarch64/templateTable_aarch64.cpp
```diff
@@ -810,7 +810,7 @@ void TemplateTable::aaload()
   // r0: array
   // r1: index
   index_check(r0, r1); // leaves index in r1, kills rscratch1
-  if (ValueArrayFlatten) {
+  if (FlatArrayFlatten) {
```
fparain, line 813: FlatArrayFlatten is not very explicit.
What about UseFlatArray?

src/hotspot/share/ci/ciTypeFlow.hpp
```diff
@@ -342,7 +342,7 @@ class ciTypeFlow : public ResourceObj {
       ciType* array = pop_value();
       if (array == null_type())  return NULL;
       // Value type arrays may contain oop or flattened representation
-      assert(array->is_obj_array_klass() || (ValueArrayFlatten && array->is_value_array_klass()),
+      assert(array->is_obj_array_klass() || (FlatArrayFlatten && array->is_value_array_klass()),
```
fparain, line 345: Shouldn't is_value_array_klass() be is_flat_array_klass()?

src/hotspot/share/runtime/globals.hpp
```diff
@@ -763,7 +763,7 @@ const size_t minimumSymbolTableSize = 1024;
   notproduct(bool, PrintInlineLayout, false,                                \
           "Print field layout for each inline type")                        \
                                                                             \
-  notproduct(bool, PrintInlineArrayLayout, false,                           \
+  notproduct(bool, PrintFlatArrayLayout, false,                             \
           "Print array layout for each inline type array")                  \
                                                                             \
   product(intx, InlineArrayElemMaxFlatSize, -1,                             \
```
fparain, line 769: InlineArrayElementMaxFlatSize -> FlatArrayElementMaxSize ?

src/hotspot/share/ci/ciTypeFlow.hpp
hseigel: I think is_value_array_klass() is correct because array is a ciType*, and I did not rename ci related functions.

src/hotspot/cpu/aarch64/templateTable_aarch64.cpp
hseigel: Sure.  That's a better name.

src/hotspot/share/runtime/globals.hpp
hseigel: How about if I change InlineArrayElementMaxFlatSize and InlineArrayElemMaxFlatOops in the next change because changing them affects a lot of tests and would make this change even bigger?

hseigel:
Thanks Fred for reviewing these changes!

PR #121, opened 2020-07-28
8250664: [lworld] Rename 'value type' to 'inline type'

TobiHartmann:
Renamed all inline type specific occurrences of "value" to "inline" in JIT code and tests.

I did not rename:
- Variable names "vt" (for value type) and "vk" (for value klass) because alternative naming "it" and "ik" is already used for iterators and instance klasses.
- "MyValue" inline class names in the tests because these are user defined "values".

---------
### Progress

### Issue
 * [JDK-8250664](https://bugs.openjdk.java.net/browse/JDK-8250664): [lworld] Rename "value type" to "inline type" ⚠️ Title mismatch between PR and JBS.

### Reviewers
 * Harold Seigel ([hseigel](@hseigel) - Committer) ⚠️ Review applies to c72a557822a8655a022911ac8f6df2afed0f026f

### Download
`$ git fetch https://git.openjdk.java.net/valhalla pull/121/head:pull/121`
`$ git checkout pull/121`


TobiHartmann:
Thanks Harold, good catch! I'll fix these as well before pushing (found some more minor issues).

PR #128, opened 2020-07-31
8250234: [lworld] [lw3] jdk/valhalla/valuetypes/SubstitutabilityTest.java fails with C1

fparain:
Please review this small changes to fix a bug in C1 transformations which cause an acmp involving inline types to be translated into a pointer comparison instead of a substitutability test.
The patch also includes a few fixes in C1 debugging features.

Thank you.

Fred

---------
### Progress

### Issue
 * [JDK-8250234](https://bugs.openjdk.java.net/browse/JDK-8250234): [lworld] [lw3] jdk/valhalla/valuetypes/SubstitutabilityTest.java fails with C1

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/valhalla pull/128/head:pull/128`
`$ git checkout pull/128`

### DIFF
--- a/src/hotspot/share/c1/c1_Instruction.hpp
+++ b/src/hotspot/share/c1/c1_Instruction.hpp
@@ -2136,8 +2136,8 @@ LEAF(If, BlockEnd)
     s->append(fsux);
     set_sux(s);
     if (!_substitutability_check) {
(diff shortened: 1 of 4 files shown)


fparain:
Thanks Tobias!

Fred

PR #136, opened 2020-08-05
8244975: [lworld] Investigate if VM can implement Object::hashCode by calling ValueBootstrapMethods for inline types

MrSimms:
Use ValueBootstrapMethods for Object::hashCode()

---------
### Progress

### Issue
 * [JDK-8244975](https://bugs.openjdk.java.net/browse/JDK-8244975): [lworld] Investigate if VM can implement Object::hashCode by calling ValueBootstrapMethods for inline types

### Reviewers
 * Srikanth Adayapalam ([sadayapalam](@sadayapalam) - Committer) ⚠️ Review applies to c9f96b1d43dac628e6ea1110e45f54b024e9ee82
 * Mandy Chung ([mchung](@mlchung) - Committer)

### Download
`$ git fetch https://git.openjdk.java.net/valhalla pull/136/head:pull/136`
`$ git checkout pull/136`

### DIFF
--- a/src/hotspot/share/runtime/globals.hpp
+++ b/src/hotspot/share/runtime/globals.hpp
@@ -2471,6 +2471,9 @@ const size_t minimumSymbolTableSize = 1024;
   product(bool, EnableValhalla, true,                                       \
           "Enable experimental Valhalla features")                          \
                                                                             \
+  product(bool, InlineTypeIHashCodeViaMethod, true,                         \
+         "Use generated hashCode for identity hash code")                   \
+                                                                            \
   product_pd(bool, InlineTypePassFieldsAsArgs,                              \
           "Pass each inline type field as an argument at calls")            \
                                                                             \
--- a/src/hotspot/share/runtime/globals.hpp
+++ b/src/hotspot/share/runtime/globals.hpp
@@ -2471,9 +2471,6 @@ const size_t minimumSymbolTableSize = 1024;
   product(bool, EnableValhalla, true,                                       \
           "Enable experimental Valhalla features")                          \
                                                                             \
-  product(bool, InlineTypeIHashCodeViaMethod, true,                         \
-         "Use generated hashCode for identity hash code")                   \
-                                                                            \
   product_pd(bool, InlineTypePassFieldsAsArgs,                              \
           "Pass each inline type field as an argument at calls")            \
                                                                             \
--- a/src/java.base/share/classes/java/lang/invoke/ValueBootstrapMethods.java
+++ b/src/java.base/share/classes/java/lang/invoke/ValueBootstrapMethods.java
@@ -193,6 +193,36 @@ public final class ValueBootstrapMethods {
                                                instanceFalse));
         }
 
+        static MethodHandle inlineTypeHashCode(Class<?> type) {
+            assert type.isInlineClass();
+            MethodHandle target = dropArguments(constant(int.class, SALT), 0, type);
+            MethodHandle cls = dropArguments(constant(Class.class, type),0, type);
+            MethodHandle classHashCode = filterReturnValue(cls, hashCodeForType(Class.class));
+            MethodHandle combiner = filterArguments(HASH_COMBINER, 0, target, classHashCode);
+            // int v = SALT * 31 + type.hashCode();
+            MethodHandle init = permuteArguments(combiner, target.type(), 0, 0);
+            MethodHandles.Lookup lookup = new MethodHandles.Lookup(type);
+            MethodHandle[] getters = MethodHandleBuilder.getters(lookup);
+            MethodHandle iterations = dropArguments(constant(int.class, getters.length), 0, type);
+            MethodHandle[] hashers = new MethodHandle[getters.length];
+            for (int i=0; i < getters.length; i++) {
+                MethodHandle getter = getters[i];
+                // For inline type or reference type, this calls Objects::hashCode.
+                // If the instance is of inline type and the hashCode method is not
+                // overridden, VM will call invokeHashCode to compute the
+                // hash code.
+                MethodHandle hasher = hashCodeForType(getter.type().returnType());
+                hashers[i] = filterReturnValue(getter, hasher);
+            }
+
+            // for (int i=0; i < getters.length; i++) {
+            //   v = computeHash(v, i, a);
+            // }
+            MethodHandle body = COMPUTE_HASH.bindTo(hashers)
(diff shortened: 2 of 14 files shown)


src/java.base/share/classes/java/lang/invoke/ValueBootstrapMethods.java
```diff
@@ -458,6 +458,35 @@ private static LinkageError newLinkageError(Throwable e) {
...
+        }
+    }
+
+    static <T> MethodHandle hashCodeInvoker(Class<T> type) {
+        return HASHCODE_METHOD_HANDLES.get(type);
+    }
+
+    private static ClassValue<MethodHandle> HASHCODE_METHOD_HANDLES = new ClassValue<>() {
+        @Override protected MethodHandle computeValue(Class<?> type) {
+            return MethodHandleBuilder.hashCodeForType(type);
```
mlchung, line 486: @MrSimms thanks for working on this.

I expect this to call the existing `hashCodeInvoker` if the type is an inline type.   `hashCodeForType` is the method handle invoking `java.util.Objects::hashCode`.   The existing `hashCodeInvoker` should be changed to take `Class<?> type` once javac no longer generates `equals` and `hashCode` method for inline type.

Maybe I should take this patch and make the coordinated changes??

src/hotspot/share/runtime/globals.hpp
```diff
@@ -2471,6 +2471,9 @@ const size_t minimumSymbolTableSize = 1024;
   product(bool, EnableValhalla, true,                                       \
           "Enable experimental Valhalla features")                          \
                                                                             \
+  product(bool, InlineTypeIHashCodeViaMethod, true,                         \
```
mlchung, line 2474: Do you anticipate this VM flag is needed?

src/java.base/share/classes/java/lang/invoke/ValueBootstrapMethods.java
```diff
@@ -458,6 +458,35 @@ private static LinkageError newLinkageError(Throwable e) {
...
+            return (int) hashCodeInvoker(type).invoke(o);
+        } catch (Error|RuntimeException e) {
+            throw e;
+        } catch (Throwable e) {
+            if (VERBOSE) e.printStackTrace();
+            throw new InternalError(e);
+        }
+    }
+
+    static <T> MethodHandle hashCodeInvoker(Class<T> type) {
```
forax, line 480: the type parameter is not needed, can be:
static MethodHandle hashCodeInvoker(Class<?> type) {

src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/Gen.java
```diff
@@ -1125,19 +1115,7 @@ private void synthesizeValueMethod(JCMethodDecl methodDecl) {
                     indyType,
                     List.nil().toArray(new LoadableConstant[0]));
 
-
             switch (methodDecl.name.toString()) {
```
sadayapalam, line 1118: This switch and the case "toString" should be deleted since it is completely unnecessary. Earlier switch would ensure that the only way control can flow into this point is when the method concerned is toString() so the check is redundant

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/TypeEnter.java
```diff
@@ -1176,16 +1176,9 @@ private void addValueMembers(JCClassDecl tree, Env<AttrContext> env) {
                         switch (parameterTypes.size()) {
```
sadayapalam, line 1176: I think this switch should be replaced with if (parameterTypes.size() == 0) - with only one case it reads off to use a switch.

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/TypeEnter.java
```diff
@@ -1200,36 +1193,6 @@ else if (name.equals("toString"))
...
-                                List.nil(),
-                                List.of(make.VarDef(make.Modifiers(PARAMETER), names.fromString("o"), make.Ident(names.fromString("Object")), null )),
-                                List.nil(), // thrown
-                                body,
-                                null);
-                memberEnter.memberEnter(equals, env);
-                tree.defs = tree.defs.append(equals);
-            }
-
             if (requireToString) {
```
sadayapalam, line 1196: The two statements preceding if (requireToString) viz
```
make.at(tree.pos);
            // Make a body comprising { throw new RuntimeException(""Internal error: This method must have been replaced by javac"); }
            JCBlock body = make.Block(Flags.SYNTHETIC, List.of(make.Throw(
                    make.NewClass(null,
                            null,
                            make.Ident(names.fromString("RuntimeException")),
                            List.of(make.Literal(CLASS, "Internal error: This method must have been replaced by javac")),
                            null))));

```
should be folded into the if (requireToString) itself for better readability and encapsulation

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/TypeEnter.java
MrSimms: Applied comments, thanks !

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/TypeEnter.java
MrSimms: Applied comments, thanks !

sadayapalam:
There will be follow on work taken up independently for JDK-8244560 - There is still some discussion that needs to happen for the right behavior when an inline type also happens to be a record, but that can be tackled separately as a javac work item

MrSimms:
> +1 for javac changes, assuming tests are all green

Looks like I did break something:
`ValueModifierTest.java:16:28: compiler.err.override.meth: (compiler.misc.cant.override: toString(), compiler.misc.anonymous.class: ValueModifierTest$3, toString(), value), final`  

rather than .out file: 

`ValueModifierTest.java:16:28: compiler.err.override.meth: (compiler.misc.cant.override: hashCode(), compiler.misc.anonymous.class: ValueModifierTest$3, hashCode(), value), final`

Not sure I completely understand the test...ahh, yeah I do, no longer generate hashCode, but toString was

mlchung:
This looks good to me.  I made a small clean up on javac `Gen::synthesizeValueMethod` method to remove the switch statement.

PR #137, opened 2020-08-05
8251046: [lworld] [lw3] C1 should avoid heap allocations in withfield when possible

fparain:
Please review this patch which reduces the number of heap allocations when executing the withfield bytecode with C1.
The CR linked below contains a more detailed description of the optimization:

CR: https://bugs.openjdk.java.net/browse/JDK-8251046

A rudimentary benchmark shows encouraging numbers:
Measuring inline type creation time with -XX:TieredStopAtLevel=1
Inline class Point has two int fields.
Inline class Rectangle has four int fields.

Without the optimization:
Benchmark                                Mode  Samples   Score  Score error  Units
o.s.MyBenchmark.testPointCreation        avgt      200   7.573        0.027  ns/op
o.s.MyBenchmark.testRectangleCreation    avgt      200  21.387        0.067  ns/op

With the optimization:
Benchmark                                Mode  Samples  Score  Score error  Units
o.s.MyBenchmark.testPointCreation        avgt      200  5.284        0.014  ns/op
o.s.MyBenchmark.testRectangleCreation    avgt      200  7.750        0.082  ns/op

Tested manually and with Mach5, tiers 1 to 3 (with Tobias' changes enabling more C1 testing).

Thank you,

Fred

---------
### Progress

### Issue
 * [JDK-8251046](https://bugs.openjdk.java.net/browse/JDK-8251046): [lworld] [lw3] C1 should avoid heap allocations in withfield when possible

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/valhalla pull/137/head:pull/137`
`$ git checkout pull/137`

### DIFF
--- a/src/hotspot/share/c1/c1_GraphBuilder.cpp
+++ b/src/hotspot/share/c1/c1_GraphBuilder.cpp
@@ -957,6 +957,9 @@ void GraphBuilder::load_local(ValueType* type, int index) {
 void GraphBuilder::store_local(ValueType* type, int index) {
   Value x = pop(type);
   store_local(state(), x, index);
+  if (x->as_NewInlineTypeInstance() != NULL) {
+    x->as_NewInlineTypeInstance()->set_local_index(index);
+  }
 }
 
 
@@ -986,6 +989,9 @@ void GraphBuilder::store_local(ValueStack* state, Value x, int index) {
 
   x->set_local_index(index);
   state->store_local(index, round_fp(x));
+  if (x->as_NewInlineTypeInstance() != NULL) {
+    x->as_NewInlineTypeInstance()->set_local_index(index);
+  }
 }
 
 
@@ -1085,6 +1091,7 @@ void GraphBuilder::store_indexed(BasicType type) {
 
 }
 
+#define UPDATE_LARVA(x) if (x != NULL && x->as_NewInlineTypeInstance() != NULL) { x->as_NewInlineTypeInstance()->set_not_larva_anymore(); }
 
 void GraphBuilder::stack_op(Bytecodes::Code code) {
   switch (code) {
@@ -1099,6 +1106,7 @@ void GraphBuilder::stack_op(Bytecodes::Code code) {
       break;
     case Bytecodes::_dup:
       { Value w = state()->raw_pop();
+        UPDATE_LARVA(w);
         state()->raw_push(w);
         state()->raw_push(w);
       }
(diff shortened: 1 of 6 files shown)


src/hotspot/share/c1/c1_Instruction.cpp
```diff
@@ -419,6 +419,40 @@ void BlockBegin::state_values_do(ValueVisitor* f) {
 }
 
 
+  StoreField::StoreField(Value obj, int offset, ciField* field, Value value, bool is_static,
```
TobiHartmann, line 422: Indentation should be fixed.

src/hotspot/share/c1/c1_Instruction.cpp
```diff
@@ -419,6 +419,40 @@ void BlockBegin::state_values_do(ValueVisitor* f) {
...
+  AssertValues assert_value;
+  values_do(&assert_value);
+#endif
+    pin();
+    if (value->as_NewInlineTypeInstance() != NULL) {
+      value->as_NewInlineTypeInstance()->set_not_larva_anymore();
+    }
+  }
+
+  StoreIndexed::StoreIndexed(Value array, Value index, Value length, BasicType elt_type, Value value, ValueStack* state_before,
```
TobiHartmann, line 438: Indentation should be fixed.

src/hotspot/share/c1/c1_GraphBuilder.cpp
```diff
@@ -1085,27 +1100,31 @@ void GraphBuilder::store_indexed(BasicType type) {
 
 }
 
-
 void GraphBuilder::stack_op(Bytecodes::Code code) {
   switch (code) {
     case Bytecodes::_pop:
-      { state()->raw_pop();
+      { Value w = state()->raw_pop();
+        NewInlineTypeInstance::update_stack_count(w);
```
TobiHartmann, line 1107: Wouldn't it be cleaner to use virtual calls on the Value class here?

src/hotspot/share/c1/c1_Instruction.hpp
```diff
@@ -1420,11 +1409,31 @@ LEAF(NewInlineTypeInstance, StateSplit)
...
-        _is_optimizable_for_withfield = false;
+        set_not_larva_anymore();
       }
     }
   }
-  virtual bool is_optimizable_for_withfield() const {  return _is_optimizable_for_withfield; }
+
+  virtual bool in_larval_state() const { return _in_larval_state; }
+  virtual void set_not_larva_anymore() {
+    _in_larval_state = false; }
```
TobiHartmann, line 1419: No newline needed after "{" or add newline before "}" as well.

fparain:
Hi Tobias!
Thank you for your feedback.
Code formatting has been fixed.
Adding virtual calls to the Value class forced to add null checks everywhere, so instead, updates of the inline type larvae are now performed in methods of the GraphBuilder class, encapsulating the null check.

Regards,

Fred

fparain:
Thanks Tobias!

Fred

PR #152, opened 2020-08-18
JDK-8251986: [lworld] implement Class::valueType and Class::referenceType in Java

mlchung:
Now that javac generates the reference projection class as a sealed type, 
this patch reimplements  `Class::valueType` and `Class::referenceType` in Java. 
VM no longer has any knowledge of the value projection type and reference projection type.

This patch includes a temporary workaround to disable the preview feature check
for sealed types because inline types are not (yet) a preview feature in valhalla repo.
This workaround should be removed when it's time to make inline class a preview feature.

---------
### Progress

### Issue
 * [JDK-8251986](https://bugs.openjdk.java.net/browse/JDK-8251986): [lworld] implement Class::valueType and Class::referenceType in Java

### Reviewers
 * Roger Riggs ([rriggs](@RogerRiggs) - Committer)

### Download
`$ git fetch https://git.openjdk.java.net/valhalla pull/152/head:pull/152`
`$ git checkout pull/152`


src/java.base/share/classes/java/lang/Class.java
```diff
@@ -593,14 +598,80 @@ public boolean isInlineClass() {
...
+        if (isInlineClass()) {
+            Class<?> superClass = getSuperclass();
+            System.out.println("superclass " + superClass);
+            if (superClass != Object.class && superClass.isReferenceProjectionType()) {
+                projectionTypes = new Class<?>[] { this, superClass };
+            } else {
+                projectionTypes = new Class<?>[] { this };
+            }
+        } else if (isReferenceProjectionType()) {
+            projectionTypes = new Class<?>[] { valueProjectionType(), this };
```
RogerRiggs, line 640: It seems odd that isReferenceProjectType() uses projectionTypes before it is initialized (or is it re-initialized).
(Though there is a check for null).

src/java.base/share/classes/java/lang/Class.java
```diff
@@ -593,14 +598,80 @@ public boolean isInlineClass() {
...
+        // an inline type must be of the same package.
+        String[] subclassNames = getPermittedSubclasses0();
+        System.out.println(getName() + " permits " + Arrays.toString(subclassNames));
+        if (subclassNames.length == 1) {
+            String cn = subclassNames[0].replace('/', '.');
+            int index = cn.lastIndexOf('.');
+            String pn = index > 0 ? cn.substring(0, index) : "";
+            if (pn.equals(getPackageName())) {
+                try {
+                    Class<?> valType = Class.forName(cn, false, getClassLoader());
```
RogerRiggs, line 674: The lookup by name implies that hidden classes can not be value classes (because they cannot be resolved by name).

src/java.base/share/classes/java/lang/Class.java
```diff
@@ -593,14 +598,80 @@ public boolean isInlineClass() {
      *         inline class; an empty {@link Optional} otherwise
      * @since Valhalla
      */
-    public Optional<Class<T>> referenceType() {
-        return valType != null ? Optional.ofNullable(refType) : Optional.of(this);
+    public Optional<Class<?>> referenceType() {
+        if (isPrimitive() || isInterface() || isArray())
+            return Optional.empty();
+        ensureProjectionTypesInited();
+        System.out.println(getName() + " " + Arrays.toString(projectionTypes));
```
RogerRiggs: Remove the println.

src/java.base/share/classes/java/lang/Class.java
```diff
@@ -576,8 +576,13 @@ public boolean isInlineClass() {
...
      * @since Valhalla
      */
-    public Optional<Class<T>> valueType() {
-        return Optional.ofNullable(valType);
+    public Optional<Class<?>> valueType() {
+        if (isPrimitive() || isInterface() || isArray())
+            return Optional.empty();
+
+        ensureProjectionTypesInited();
+        System.out.println(getName() + " " + Arrays.toString(projectionTypes));
```
RogerRiggs, line 587: println?

src/java.base/share/classes/java/lang/Class.java
```diff
@@ -593,14 +598,80 @@ public boolean isInlineClass() {
...
+        return valueProjectionType() != null;
+    }
+
+    private Class<?> valueProjectionType() {
+        // A reference projection type must be a sealed abstract class
+        // that permits the inline projection type to extend.
+        // The inline projection type and reference projection type for
+        // an inline type must be of the same package.
+        String[] subclassNames = getPermittedSubclasses0();
+        System.out.println(getName() + " permits " + Arrays.toString(subclassNames));
```
RogerRiggs: Debug println().

src/java.base/share/classes/java/lang/Class.java
```diff
@@ -593,14 +598,80 @@ public boolean isInlineClass() {
...
+        if (subclassNames.length == 1) {
+            String cn = subclassNames[0].replace('/', '.');
```
RogerRiggs, line 669: String.regionMatches() could be used to compare the substrings without making copies and it should be ok to compare the package names without translating '/' to '.'.

src/java.base/share/classes/java/lang/Class.java
```diff
@@ -593,14 +598,80 @@ public boolean isInlineClass() {
...
+        if (projectionTypes != null) {
+            return projectionTypes.length == 2 && projectionTypes[1] == this;
+        }
+
+        int mods = getModifiers();
+        if (!Modifier.isAbstract(mods)) {
+            return false;
+        }
+
+        return valueProjectionType() != null;
```
RogerRiggs, line 659: This path computed the valueProjecttionType and then discards it to be computed again on line 631.

src/java.base/share/classes/java/lang/Class.java
mlchung: An inline hidden class cannot be named and hence it can't be listed in `PermittedSubclasses`.  In other words, it can't do like the language translation to translate an inline class `V` to have a pair `V$ref` and `V$val extends V$ref`.

src/java.base/share/classes/java/lang/Class.java
mlchung: Thanks for catching that.

src/java.base/share/classes/java/lang/Class.java
mlchung: Leftover debugging code.  Will remove

src/java.base/share/classes/java/lang/Class.java
mlchung: I don't think this path is performance critical.  We can look into this if performance turns out to be an issue.

src/java.base/share/classes/java/lang/Class.java
mlchung: getPackageName() returns a "."-separated package name (not the internal form). I will leave it as is.

src/java.base/share/classes/java/lang/Class.java
mlchung: I have redone the logic and incorporated your suggestion in `getProjectionType` method.

PR #154, opened 2020-08-19
8252040: [lworld] Execute compiler unit tests in random order

TobiHartmann:
- Profiling of callee methods depends on execution order of unit tests. Execute tests in random order to increase coverage.
- At double field to test inline type
- Verify that expected NPEs are always thrown
- Fixed a test bug in TestLWorldProfiling.java (array element is not re-initialized and therefore NULL)
- Refactoring and removal of old, non-applicable TODOs

---------
### Progress

### Issue
 * [JDK-8252040](https://bugs.openjdk.java.net/browse/JDK-8252040): [lworld] Execute compiler unit tests in random order

### Download
`$ git fetch https://git.openjdk.java.net/valhalla pull/154/head:pull/154`
`$ git checkout pull/154`


TobiHartmann:
Executing tests in random order triggered at least 4 new bugs that I'll fix before pushing this change.

PR #155, opened 2020-08-20
JDK-8252061  [lword] Support Object::toString for inline type

mlchung:
This adds the `Object::toString` support for inline classes.  This will replace the need
for javac to generate the `toString` method to call the bootstrap method.

JDK-8206077 will revisit the overlap of `ValueBootstrapMethods` and `ObjectMethods`.

This patch also removes the private `isSubstituable0` method and makes sure 
that `isSubstitutable` does not have `acmp` instruction and can be invoked by VM.

---------
### Progress

### Issue
 * [JDK-8252061](https://bugs.openjdk.java.net/browse/JDK-8252061): [lword] Support Object::toString for inline type

### Reviewers
 * Srikanth Adayapalam ([sadayapalam](@sadayapalam) - Committer)
 * Roger Riggs ([rriggs](@RogerRiggs) - Committer)

### Download
`$ git fetch https://git.openjdk.java.net/valhalla pull/155/head:pull/155`
`$ git checkout pull/155`


rose00:
A couple of comments:

1. In order for this to perform well, the code for each generated toString method needs to be JIT-ted, and to be reached quickly (or even inlined) via ClassValue::get.  Optimization work (to be done later) should probably consider [JDK-8238260](https://bugs.openjdk.java.net/browse/JDK-8238260).

2. We should try to handle Record subtypes the same way.  Putting funky bytecodes into classfiles, even using indy, is worse than allowing Object::toString (and equals and hashCode) to vector through a specified, customized method, using ClassValue::get.

3. I think it is reasonable to have special intrinsic logic in the JIT to hurry along the previous two steps, although in the long run the JVM should have a good story (perhaps using ClassValue, perhaps not) for auto-expanded method bodies.

sadayapalam (approved):
+1 for javac changes

PR #156, opened 2020-08-20
8251527: CTW: C2 (Shenandoah) compilation fails with SEGV due to unhandled catchproj == NULL

rwestrel:
In PhaseIdealLoop::flatten_array_element_type_check(), when pushing:

(LoadKlass (AddP (CastPP ...

through a Phi, the control of CastPP clone is set to region->in(i) but
the AddP clone control is set to get_ctrl(phi->in(i)) which can be
above region->in(i) while the AddP is below the CastPP. Fix is to use
region->in(i) for all clones.

---------
### Progress

### Issue
 * [JDK-8251527](https://bugs.openjdk.java.net/browse/JDK-8251527): CTW: C2 (Shenandoah) compilation fails with SEGV due to unhandled catchproj == NULL

### Download
`$ git fetch https://git.openjdk.java.net/valhalla pull/156/head:pull/156`
`$ git checkout pull/156`


rwestrel:
Used incorrect bug id. Closing it,

PR #167, opened 2020-08-26
8229897: [lworld] C1 should avoid allocation when reading a field from a flattened field

fparain:
Please review this patch which removes temporary heap allocations when reading a field from a flattened field.

During the HIR graph construction, C1 detects sequences of consecutive getfield bytecodes involving flattened fields and optimizes them with a single LoadField or NewInlineTypeInstance directly to the offset of the last nested field.

Performance improvement on a simple benchmark

    static inline class Point {
        int i = 0, j = 0;
    }
    static inline class Rectangle {
        Point p0 = new Point(), p1 = new Point();
    }
    static class NamedRectangle {
        Rectangle rect = new Rectangle();
        String name;
    }
    static NamedRectangle nr = new NamedRectangle();
    @Benchmark
    public void testGetfieldChain1(Blackhole blackhole) {
        int i = nr.rect.p0.j;                                                              // reads two flattened fields before reading an int
        blackhole.consume(i);
    }

    @Benchmark
    public void testGetfieldChain2(Blackhole blackhole) {
        Point p = nr.rect.p1;                                                          // reads a flattened field inside a flattened field
        blackhole.consume(p);
    }

Without the optimization:
Benchmark                             Mode  Samples  Score  Score error  Units
o.s.MyBenchmark.testGetfieldChain1    avgt      200  8.696        0.024  ns/op
o.s.MyBenchmark.testGetfieldChain2    avgt      200  8.969        0.027  ns/op

With the optimization:
Benchmark                             Mode  Samples  Score  Score error  Units
o.s.MyBenchmark.testGetfieldChain1    avgt      200  2.134        0.019  ns/op
o.s.MyBenchmark.testGetfieldChain2    avgt      200  5.150        0.017  ns/op

Thank you,

Fred

---------
### Progress

### Issue
 * [JDK-8229897](https://bugs.openjdk.java.net/browse/JDK-8229897): [lworld] C1 should avoid allocation when reading a field from a flattened field

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/valhalla pull/167/head:pull/167`
`$ git checkout pull/167`


TobiHartmann (approved):
Looks good to me!

fparain:
Thanks Tobias!

Fred

PR #178, opened 2020-09-04
8252506: [lworld] Multiple issues with C2's arraycopy intrinsic

TobiHartmann:
This patch fixes multiple issues in C2's arraycopy/copyof/clone intrinsics:
- Flat inline type arrays containing oops might be copied without GC barriers.
- Missing membarrier in copyOf code at macro expansion
- We can often avoid flat/null-free checks by relying on the src <: dst subtype check
- We often know at parse time that a flat array does not contain oops and should make use of that information
- And more.. Hopefully the comments in the code are self-explanatory

Other changes:
- generic_arraycopy stub needs to handle flat/null-free inline type arrays
  - Bail if src is flat or dst is flat/null-free
  - Primitive array verification code is broken because array tag contains more bits
  - The "Load layout helper" comment needs adjustment as well but I'll wait with that until we've moved bits to the mark word
- C1 arraycopy intrinsic does not need to check arguments to generic_arraycopy stub (moved checks below)
- Added asserts and runtime verification code to catch issues earlier
- Added lots of test for all issues I've found and new IR matching rules to verify that all arraycopy optimizations work as expected
- Some refactoring

I've identified some remaining optimization opportunities and marked them with "TODO 8251971" in code and tests.

---------
### Progress

### Issue
 * [JDK-8252506](https://bugs.openjdk.java.net/browse/JDK-8252506): [lworld] Multiple issues with C2's arraycopy intrinsic

### Download
`$ git fetch https://git.openjdk.java.net/valhalla pull/178/head:pull/178`
`$ git checkout pull/178`

### DIFF
--- a/src/hotspot/cpu/x86/c1_LIRAssembler_x86.cpp
+++ b/src/hotspot/cpu/x86/c1_LIRAssembler_x86.cpp
@@ -3302,14 +3302,6 @@ void LIR_Assembler::emit_arraycopy(LIR_OpArrayCopy* op) {
     return;
   }
 
-  if (flags & LIR_OpArrayCopy::src_inlinetype_check) {
-    arraycopy_inlinetype_check(src, tmp, stub, false, (flags & LIR_OpArrayCopy::src_null_check));
-  }
-
-  if (flags & LIR_OpArrayCopy::dst_inlinetype_check) {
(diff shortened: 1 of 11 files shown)


TobiHartmann:
Thanks Roland!

PR #187, opened 2020-09-14
8253113: [lworld] [lw3] C1 should avoid copying element of flattened arrays when reading a sub-element

fparain:
Please review these changes optimizing access to flattened arrays in C1.

The optimization avoids unnecessary copies of intermediate values when the code accesses a sub-element of a flattened array.

For instance, with the following code:

inline class Point { int x = 0, y = 0; }
void foo() {
    Point[] array = new Point[10];
    int i = array[1].x;
}

C1 used to create a new instance of Point, copy the content from the array, then reads the x field from this new instance.
With the optimization, C1 directly accesses the x field from the flattened array, and makes no heap allocation.

A simple benchmark on the "int i = array[1].x;" line gives these results:

baseline:
Benchmark                          Mode  Samples  Score  Score error  Units
o.s.MyBenchmark.testArrayReads     avgt      200  4.250        0.011  ns/op

optimized:
Benchmark                          Mode  Samples  Score  Score error  Units
o.s.MyBenchmark.testArrayReads     avgt      200  2.228        0.004  ns/op

Thank you,

Fred

---------
### Progress

### Issue
 * [JDK-8253113](https://bugs.openjdk.java.net/browse/JDK-8253113): [lworld] [lw3] C1 should avoid copying element of flattened arrays when reading a sub-element

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/valhalla pull/187/head:pull/187`
`$ git checkout pull/187`

### DIFF
--- a/src/hotspot/share/c1/c1_LIRGenerator.cpp
+++ b/src/hotspot/share/c1/c1_LIRGenerator.cpp
@@ -1595,20 +1595,8 @@ class TempResolvedAddress: public Instruction {
   virtual const char* name() const  { return "TempResolvedAddress"; }
 };
 
-void LIRGenerator::access_sub_element(LIRItem& array, LIRItem& index, LIR_Opr& result, ciField* field, int sub_offset) {
-  assert(field != NULL, "Need a subelement type specified");
-  // Find the starting address of the source (inside the array)
-  ciType* array_type = array.value()->declared_type();
-  ciFlatArrayKlass* flat_array_klass = array_type->as_flat_array_klass();
-  assert(flat_array_klass->is_loaded(), "must be");
-
-  ciInlineKlass* elem_klass = NULL;
-  int array_header_size = flat_array_klass->array_header_in_bytes();
-  int shift = flat_array_klass->log2_element_size();
-
-  BasicType subelt_type = field->type()->basic_type();
-
-  #ifndef _LP64
+LIR_Opr LIRGenerator::get_index_in_register(LIRItem& index, int shift) {
+#ifndef _LP64
   LIR_Opr index_op = new_register(T_INT);
   // FIXME -- on 32-bit, the shift below can overflow, so we need to check that
   // the top (shift+1) bits of index_op must be zero, or
@@ -1619,6 +1607,7 @@ void LIRGenerator::access_sub_element(LIRItem& array, LIRItem& index, LIR_Opr& r
   } else {
     __ shift_left(index_op, shift, index.result());
   }
+  return index_op;
 #else
   LIR_Opr index_op = new_register(T_LONG);
   if (index.result()->is_constant()) {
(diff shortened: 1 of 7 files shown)


src/hotspot/share/c1/c1_LIRGenerator.cpp
```diff
@@ -1595,13 +1595,84 @@ class TempResolvedAddress: public Instruction {
...
-void LIRGenerator::access_flattened_array(bool is_load, LIRItem& array, LIRItem& index, LIRItem& obj_item) {
+void LIRGenerator::access_sub_element(LIRItem& array, LIRItem& index, LIR_Opr& result, ciField* field, int sub_offset) {
+  assert(field != NULL, "Need a subelement type specified");
   // Find the starting address of the source (inside the array)
   ciType* array_type = array.value()->declared_type();
   ciFlatArrayKlass* flat_array_klass = array_type->as_flat_array_klass();
   assert(flat_array_klass->is_loaded(), "must be");
 
-  ciInlineKlass* elem_klass = flat_array_klass->element_klass()->as_inline_klass();
+  ciInlineKlass* elem_klass = NULL;
```
TobiHartmann, line 1605: Unused.

src/hotspot/share/c1/c1_LIRGenerator.cpp
```diff
@@ -1595,13 +1595,84 @@ class TempResolvedAddress: public Instruction {
...
+  int array_header_size = flat_array_klass->array_header_in_bytes();
+  int shift = flat_array_klass->log2_element_size();
+
+  BasicType subelt_type = field->type()->basic_type();
+
+  #ifndef _LP64
```
TobiHartmann, line 1611: Wrong indentation.

src/hotspot/share/c1/c1_LIRGenerator.cpp
```diff
@@ -1595,13 +1595,84 @@ class TempResolvedAddress: public Instruction {
   virtual const char* name() const  { return "TempResolvedAddress"; }
 };
 
-void LIRGenerator::access_flattened_array(bool is_load, LIRItem& array, LIRItem& index, LIRItem& obj_item) {
+void LIRGenerator::access_sub_element(LIRItem& array, LIRItem& index, LIR_Opr& result, ciField* field, int sub_offset) {
```
TobiHartmann, line 1598: There is quite some code duplication here with access_flattened_array, can we factor it out?

TobiHartmann (approved):
Looks good to me, added some minor comments.

src/hotspot/share/c1/c1_LIRGenerator.cpp
fparain: Fixed

src/hotspot/share/c1/c1_LIRGenerator.cpp
fparain: Fixed

TobiHartmann (approved):
Looks good, thanks for making these changes!

src/hotspot/share/c1/c1_LIRGenerator.cpp
fparain: Code has been refactored, and the common code in access_flattened_array() and access_sub_element() is now shared in the new method get_and_load_element_address().

src/hotspot/share/c1/c1_LIRGenerator.cpp
```diff
@@ -1631,12 +1629,58 @@ void LIRGenerator::access_flattened_array(bool is_load, LIRItem& array, LIRItem&
   LIR_Opr elm_op = new_pointer_register();
   LIR_Address* elm_address = new LIR_Address(array.result(), index_op, array_header_size, T_ADDRESS);
   __ leal(LIR_OprFact::address(elm_address), elm_op);
+  return elm_op;
+}
+
+void LIRGenerator::access_sub_element(LIRItem& array, LIRItem& index, LIR_Opr& result, ciField* field, int sub_offset) {
+  assert(field != NULL, "Need a subelement type specified");
+
+// Find the starting address of the source (inside the array)
```
TobiHartmann, line 1638: Indentation is wrong.

src/hotspot/share/c1/c1_LIRGenerator.cpp
```diff
@@ -1631,12 +1629,58 @@ void LIRGenerator::access_flattened_array(bool is_load, LIRItem& array, LIRItem&
...
+    __ branch_destination(L_end->label());
+    set_in_conditional_code(false);
+  }
+}
+
+void LIRGenerator::access_flattened_array(bool is_load, LIRItem& array, LIRItem& index, LIRItem& obj_item,
+                                          ciField* field, int sub_offset) {
+  assert(sub_offset == 0 || field != NULL, "Sanity check");
+
+// Find the starting address of the source (inside the array)
```
TobiHartmann, line 1670: Indentation is wrong.

TobiHartmann (approved):
Nice refactoring. Looks good to me.

fparain:
Thanks Tobias for your review.
I've fixed the wrong indentations.

Fred

TobiHartmann (approved):

PR #191, opened 2020-09-16
8253218 ClassFileParser hits assert(klass->access_flags().is_inline_t…

hseigel:
Please review this fix for JDK-8253218.  The fix throws a ClassFormatError exception if a class file, with an old class file version, contains a Q signature.

The fix was tested with tiers 1-2 on Windows, Linux x64, and MacOS, and tiers 3-5 on Linux x64.

---------
### Progress

### Issue
 * [JDK-8253218](https://bugs.openjdk.java.net/browse/JDK-8253218): [lworld] ClassFileParser hits assert(klass->access_flags().is_inline_type()) failed: Value type expected ⚠️ Title mismatch between PR and JBS.

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/valhalla pull/191/head:pull/191`
`$ git checkout pull/191`

### DIFF
--- a/src/hotspot/share/classfile/classFileParser.cpp
+++ b/src/hotspot/share/classfile/classFileParser.cpp
@@ -4725,6 +4725,7 @@ void ClassFileParser::verify_legal_class_modifiers(jint flags, TRAPS) const {
       "Class modifier ACC_INLINE in class %s requires option -XX:+EnableValhalla",
       _class_name->as_C_string()
     );
+    return;
   }
 
   if (!_need_verify) { return; }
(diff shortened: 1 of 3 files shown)


src/hotspot/share/classfile/classFileParser.cpp
```diff
@@ -5299,6 +5299,9 @@ void ClassFileParser::verify_legal_field_signature(const Symbol* name,
                                                    const Symbol* signature,
                                                    TRAPS) const {
   if (!_need_verify) { return; }
+  if (!supports_inline_types() && signature->is_Q_signature()) {
```
fparain, line 5302: Shouldn't the test also include a check for Q-arrays?  (signature->is_Q_array_signature()) ?

hseigel: Yes, it should check for Q-arrays.  I'll add the check.  Thanks!

fparain (approved):
Looks good to me.

Fred

PR #195, opened 2020-09-23
8253511: [lworld] ClassFileParser does not handle field type mismatch

fparain:
Please review this change with replaces an assert with the throwing of an IncompatibleClassChangeError when a pre-loaded class is not an inline type. This case was correctly handled for static fields, but not for non-static fields.

Thank you,

Fred

---------
### Progress

### Issue
 * [JDK-8253511](https://bugs.openjdk.java.net/browse/JDK-8253511): [lworld] ClassFileParser does not handle field type mismatch

### Reviewers
 * [Harold Seigel](https://openjdk.java.net/census#hseigel) (@hseigel - Committer)

### Download
`$ git fetch https://git.openjdk.java.net/valhalla pull/195/head:pull/195`
`$ git checkout pull/195`

### DIFF
--- a/src/hotspot/share/classfile/classFileParser.cpp
+++ b/src/hotspot/share/classfile/classFileParser.cpp
@@ -6596,7 +6596,14 @@ void ClassFileParser::post_process_parsed_stream(const ClassFileStream* const st
           Handle(THREAD, _loader_data->class_loader()),
           _protection_domain, true, CHECK);
       assert(klass != NULL, "Sanity check");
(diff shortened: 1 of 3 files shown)


hseigel (approved):
The changes look good!  You could add code to the test's exception handler that looks at err.getMessage() to check that ICE got thrown for the expected reason.

fparain:
Harold,

Thank you for the review, I've updated the test, checking that the error message is the expected one.

Fred

PR #203, opened 2020-09-29
8253724: [lworld] C1 compilation hits assert: Empty inline type access should be removed

fparain:
Please review these changes fixing handling the handling of empty inline types in C1.
The changeset includes two new unit tests, but they are currently disabled because they are causing failures with C2 which has issues with empty inline types too (C2 will be fixed with a different patch).

Thank you,

Fred

---------
### Progress

### Issue
 * [JDK-8253724](https://bugs.openjdk.java.net/browse/JDK-8253724): [lworld] C1 compilation hits assert: Empty inline type access should be removed

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/valhalla pull/203/head:pull/203`
`$ git checkout pull/203`


TobiHartmann (approved):
Looks good! I'll re-enable these tests once I've fixed C2.

fparain:
Tobias,

Thank you for the review.

Fred

PR #206, opened 2020-10-01
8253873: jdi, jvmti, and jdwp tests fail because of unexpected number of interfaces

hseigel:
Please review these changes to fix tier4 and tier5 tests that started failing because injected interface IdentityObject is no longer hidden.  Tiers 4 and 5 were run on linux X64, Macosx, and Windows to check that the tests passed.

---------
### Progress

### Issue
 * [JDK-8253873](https://bugs.openjdk.java.net/browse/JDK-8253873): jdi, jvmti, and jdwp tests fail because of unexpected number of interfaces

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/valhalla pull/206/head:pull/206`
`$ git checkout pull/206`

### DIFF
--- a/test/hotspot/jtreg/vmTestbase/nsk/jdi/ClassType/allInterfaces/allinterfaces001.java
+++ b/test/hotspot/jtreg/vmTestbase/nsk/jdi/ClassType/allInterfaces/allinterfaces001.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2001, 2018, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2001, 2020, Oracle and/or its affiliates. All rights reserved.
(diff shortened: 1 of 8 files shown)


fparain (approved):
Harold,

Thank you for fixing this.
Overall, changes look good, but it would be good to update the comments because in the current form, it looks like there's a mismatch between the comments and the code.

For instance, one test has this comment:
    "// Class1forCheck - 1 direct and 1 indirect interfaces"
but the code has been updated to:
     "  if (iface1list.size() != 3) {"

Someone not familiar with the IdentityObject injection would wonder where this "3" comes from,
as the comment suggest the correct answer is "2".

Regards,

Fred

hseigel:
Hi Fred,
Thanks for reviewing this change.  Please see updated commit containing additional comments about interface IdentityObject.

Thanks, Harold

fparain (approved):
Looks good to me.
Thank you for the updated comments.

Fred

PR #210, opened 2020-10-05
8254022: [lworld] [type-restrictions] Initial support for RestrictedField

fparain:
[Second attempt, trying to do it on the right branch this time]

Please review these changes adding an initial support for RestrictedField.

The patch includes:
 - recognizing and processing the new attribute during class file parsing
 - creating new meta-data to store both the erased type and the restricted type
 - fixing all intermediate method dealing with field types (FieldInfo, FieldStream, etc)
 - fixing field resolution to take into the two possible signatures of the fields
 - adding checkcast against the restricted type in the interpreter implementation of putfield

The patch doesn't include:
 - propagating new meta-data to CI
 - adding new type checks in JIT generated code

The main take away of this first patch is that the VM mostly cares about the sharp type (from the
RestrictedField attribute), and the only place where it needs knowledge of the erased type is for
field resolution. This is why the implementation of the RestrictedField in the VM is inverted compared
to the content of the class file: the traditional FieldInfo is filled with the sharp type from the RestrictedField
attribute, while the original erased type is moved to a new variable section of InstanceKlass.

Note: the patch includes a serious clean up of the FieldInfo data structure. It used to store the allocation
type of the field, but this information is not used anymore after the replacement of the old field layout
code by the field layout builder. Removing the allocation type from the data structure simplifies it a lot.

Thank you,

Fred

---------
### Progress

### Issue
 * [JDK-8254022](https://bugs.openjdk.java.net/browse/JDK-8254022): [lworld] [type-restrictions] Initial support for RestrictedField

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/valhalla pull/210/head:pull/210`
`$ git checkout pull/210`

### DIFF
--- a/src/hotspot/cpu/x86/templateTable_x86.cpp
+++ b/src/hotspot/cpu/x86/templateTable_x86.cpp
@@ -3365,6 +3365,22 @@ void TemplateTable::putfield_or_static(int byte_no, bool is_static, RewriteContr
   // volatile_barrier(Assembler::Membar_mask_bits(Assembler::LoadStore |
   //                                              Assembler::StoreStore));
 
+
+  Label notRestricted;
+  __ movl(rdx, flags);
+  __ shrl(rdx, ConstantPoolCacheEntry::has_restricted_type);
+  __ andl(rdx, 0x1);
+  __ testl(rdx, rdx);
+  __ jcc(Assembler::zero, notRestricted);
+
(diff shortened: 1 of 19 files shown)


MrSimms (approved):
It looks good. Thanks for explanation on the fieldInfo clean up (described in PR body).

fparain:
David,

Thank you for the review.

Fred

PR #221, opened 2020-10-09
8254271: Development to deprecate wrapper class constructors for removal

RogerRiggs:
The public constructors of the following classes have been deprecated since SE 9, and should be deprecated for removal in anticipation of these classes becoming inline classes.

java.lang.Boolean, Byte, Short, Integer, Long, Float, Double, and Character.

Existing uses of the constructors are converted to use the .valueOf(xxx) methods in classes and tests.

---------
### Progress

### Testing

| Build | ✔️ (5/5 passed) | ✔️ (2/2 passed) | ✔️ (2/2 passed) |
| Test (tier1) | ❌ (1/9 failed) | ❌ (1/9 failed) | ❌ (1/9 failed) |

**Failed test tasks**
- [Linux x64 (langtools/tier1)](https://github.com/RogerRiggs/valhalla/runs/1232836396)
- [Windows x64 (langtools/tier1)](https://github.com/RogerRiggs/valhalla/runs/1232933286)
- [macOS x64 (langtools/tier1)](https://github.com/RogerRiggs/valhalla/runs/1232787026)

### Issue
 * [JDK-8254271](https://bugs.openjdk.java.net/browse/JDK-8254271): Development to deprecate wrapper class constructors for removal

### Reviewers
 * [Mandy Chung](https://openjdk.java.net/census#mchung) (@mlchung - **Reviewer**)

### Download
`$ git fetch https://git.openjdk.java.net/valhalla pull/221/head:pull/221`
`$ git checkout pull/221`

### DIFF
--- a/src/java.base/share/classes/java/lang/invoke/MemberName.java
+++ b/src/java.base/share/classes/java/lang/invoke/MemberName.java
@@ -773,7 +773,7 @@ final class MemberName implements Member, Cloneable {
     }
 
     @Override
-    @SuppressWarnings("deprecation")
+    @SuppressWarnings({"deprecation", "removal"})
     public int hashCode() {
         // Avoid autoboxing getReferenceKind(), since this is used early and will force
         // early initialization of Byte$ByteCache
--- a/src/java.xml/share/classes/com/sun/org/apache/xml/internal/serialize/HTMLdtd.java
+++ b/src/java.xml/share/classes/com/sun/org/apache/xml/internal/serialize/HTMLdtd.java
@@ -438,8 +438,8 @@ public final class HTMLdtd
     private static void defineEntity( String name, char value )
     {
         if ( _byName.get( name ) == null ) {
-            _byName.put( name, new Integer( value ) );
-            _byChar.put( new Integer( value ), name );
+            _byName.put( name, Integer.valueOf( value ) );
+            _byChar.put( Integer.valueOf( value ), name );
         }
     }
 
--- a/test/jdk/java/lang/WrappersTest.java
+++ b/test/jdk/java/lang/WrappersTest.java
@@ -1,0 +1,62 @@
+/*
+ * Copyright (c) 2020, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+
+import org.testng.annotations.Test;
+
+import java.lang.reflect.Constructor;
+import java.util.List;
+
+import static org.testng.Assert.*;
+
+/*
+ * @test
+ * @bug 8252180
+ * @summary Test the primitive wrappers constructors are deprecated for removal
+ * @run testng WrappersTest
+ */
+
+@Test
+public class WrappersTest {
+
+    @Test
+    void checkForDeprecated() {
+        List<Class<?>> classes =
+                List.of(Byte.class,
+                        Short.class,
+                        Integer.class,
+                        Long.class,
+                        Float.class,
+                        Double.class,
+                        Character.class,
+                        Boolean.class);
+        for (Class<?> cl : classes) {
+            for (Constructor<?> cons : cl.getConstructors()) {
+                Deprecated dep = cons.getAnnotation(Deprecated.class);
+                assertNotNull(dep, "Missing @Deprecated annotation");
+                System.out.println(cons + ": " + dep);
+                assertTrue(dep.forRemoval(), cl.toString() + " deprecated for removal: ");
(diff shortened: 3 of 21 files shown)


dansmithcode:
Should change issue number to 8254271, which is targeted to 'repo-valhalla'. (I'm actually not sure what happens if you push with 8252180, which is targeted to 'tbd'...)

RogerRiggs:
> Should change issue number to 8254271, which is targeted to 'repo-valhalla'. (I'm actually not sure what happens if you push with 8252180, which is targeted to 'tbd'...)

The fields in the bug are just hints for people.  The actual push using git is specific to the repo and branch its targeted to.
The bugs are updated after the fact by 'hgupdater'.

src/java.base/share/classes/java/lang/invoke/MemberName.java
```diff
@@ -773,7 +773,7 @@ public MemberName getDefinition() {
     }
 
     @Override
-    @SuppressWarnings("deprecation")
+    @SuppressWarnings({"deprecation", "removal"})
```
dansmithcode, line 776: What's going to happen to these warning suppressions when we actually remove (make private) the constructors? Is the intent to just defer that problem until it happens?

dansmithcode:
> > Should change issue number to 8254271, which is targeted to 'repo-valhalla'. (I'm actually not sure what happens if you push with 8252180, which is targeted to 'tbd'...)
> 
> The fields in the bug are just hints for people. The actual push using git is specific to the repo and branch its targeted to.
> The bugs are updated after the fact by 'hgupdater'.

Understood.

I'm saying it needs to be configured so that 'hgupdater' leaves 8252180 open and closes 8254271. (And I'm unsure of exactly how things will go wrong otherwise. I know sometimes hgupdater will create backport issues...)

src/java.base/share/classes/java/lang/invoke/MemberName.java
RogerRiggs: When the code is changed to not use the constructors, the warning (deprecation and remove) should be removed.
Some of the code is upstream and requires more changes.

test/jdk/java/lang/WrappersTest.java
```diff
@@ -0,0 +1,62 @@
+/*
```
jddarcy, line 1: We have historically not found it necessary to write tests to verify this kind of change and have instead relied on signature testing.

src/java.xml/share/classes/com/sun/org/apache/xml/internal/serialize/HTMLdtd.java
```diff
@@ -438,8 +438,8 @@ private static void initialize()
     private static void defineEntity( String name, char value )
     {
         if ( _byName.get( name ) == null ) {
-            _byName.put( name, new Integer( value ) );
```
jddarcy, line 441: Please make sure @JoeWang-Java reviews the XML changes; there are complications in some areas where the master code is upstream of the code in the JDK.

test/jdk/java/lang/WrappersTest.java
mlchung: +1. 

FWIW.  You have also verified when you get the warnings from JDK build and needs to suppressed it.

src/java.xml/share/classes/com/sun/org/apache/xml/internal/serialize/HTMLdtd.java
mlchung: JAXP was subsumed by JSR 379 in Java SE 9.

src/java.base/share/classes/java/lang/invoke/MemberName.java
dansmithcode: Yeah, Mandy replied to the question I should have asked: for all the suppressed "removal" points, what is the plan to fix the offending code? If not now, when? ("Later" may be a reasonable answer, but fleshing out the followup tasks would be helpful.)

RogerRiggs: The comment in MemberName makes it clear that `hashCode` method may be called *before* the cache is setup and `valueOf` uses the cache.

test/jdk/java/lang/WrappersTest.java
RogerRiggs: The test can be removed but was a double check that all of the constructors of the named classes had been modified.

src/java.base/share/classes/java/lang/invoke/MemberName.java
RogerRiggs: Its up to the Graal folks to replace their code. In some cases, they are depending on Identity semantics of new Integer().

dansmithcode:
Clarification on integration behavior: 

The model is that each pull to `valhalla` or `jdk` needs a separate bugid. That's the workflow we need to follow unless we can flesh out and get tooling support for an alternative model.

I'm asking that 8252180 be reserved for the `jdk` pull and for the CSR. And that all pulls to `valhalla` use bugids of subtasks of that issue, such as 8254271.

Robin says if you push with bugid 8252180 (which is targeted to 'tbd'), the issue will be closed, disrupting its use for the `jdk` pull and CSR. He will consider adding a guard that warns about the mismatch of JBS target and PR target to prevent this sort of accident.

RogerRiggs:
/csr remove

The CSR for this change will be made against 8252180

src/java.base/share/classes/java/lang/invoke/MemberName.java
mlchung: Thanks for the clarification.   Can you file a JBS issue to Graal to make change in its upstream project?

mlchung (approved):

RogerRiggs:
Please review the CSR for 8252180, to be used for the integration

CSR 8254324: [JEP 390] Deprecate wrapper class constructors for removal

dansmithcode:
Have you considered rephrasing the reference to constructors in the `valueOf` methods?:

```
     * If a new {@code Byte} instance is not required, this method
     * should generally be used in preference to the constructor
     * {@link #Byte(byte)}
```

The message now is "Don't use the constructors! Don't try to get a new instance!"

RogerRiggs:
> Have you considered rephrasing the reference to constructors in the `valueOf` methods?:
> 
> ```
>      * If a new {@code Byte} instance is not required, this method
>      * should generally be used in preference to the constructor
>      * {@link #Byte(byte)}
> ```
> 
> The message now is "Don't use the constructors! Don't try to get a new instance!"

I think "if ... is not required" covers the case already. The behavior of Integer is not changing.
The valueOf() method is already the recommendation.

PR #222, opened 2020-10-09
8254275: [valhalla/jep390] Revise "value-based class" & apply to wrappers

dansmithcode:
Polishing the specification of "value-based class" to align with requirements of inline classes, allow classes (like Integer) with deprecated constructors, and clarify expectations for clients.

Update cross-references from existing value-based classes and the primitive wrapper classes.

Full docs build: http://cr.openjdk.java.net/~dlsmith/8254275/8254275-20201124/api/index.html

---------
### Progress

### Testing

|     | Linux x64 | Linux x86 | Windows x64 | macOS x64 |
| --- | ----- | ----- | ----- | ----- |
| Build | ⏳ (6/6 running) | ⏳ (2/2 running) | ⏳ (2/2 running) | ⏳ (2/2 running) |

### Issue
 * [JDK-8254275](https://bugs.openjdk.java.net/browse/JDK-8254275): [valhalla/jep390] Revise "value-based class" & apply to wrappers

### Reviewers
 * [Mandy Chung](https://openjdk.java.net/census#mchung) (@mlchung - **Reviewer**) ⚠️ Review applies to 3288f433513d55a45cefa2bb1bd8278b5788b6b5
 * [Roger Riggs](https://openjdk.java.net/census#rriggs) (@RogerRiggs - **Reviewer**)

### Download
`$ git fetch https://git.openjdk.java.net/valhalla pull/222/head:pull/222`
`$ git checkout pull/222`


src/java.base/share/classes/java/lang/doc-files/ValueBased.html
```diff
@@ -35,32 +35,37 @@ <h1 id="ValueBased">{@index "Value-based Classes"}</h1>
...
+        code of an instance, or synchronization on an instances's intrinsic lock;</li>
     <li>have implementations of <code>equals</code>,
         <code>hashCode</code>, and <code>toString</code> which are computed
         solely from the instance's state and not from its identity or the state
         of any other object or variable;</li>
-    <li>make no use of identity-sensitive operations such as reference
-        equality (<code>==</code>) between instances, identity hash code of
-        instances, or synchronization on an instances's intrinsic lock;</li>
-    <li>are considered equal solely based on <code>equals()</code>, not
-        based on reference equality (<code>==</code>);</li>
```
dansmithcode, line 48: I removed this bullet because I don't think it really means anything. We already said classes are required not to expose identity through their methods, and to override 'equals'

src/java.base/share/classes/java/lang/doc-files/ValueBased.html
```diff
@@ -35,32 +35,37 @@ <h1 id="ValueBased">{@index "Value-based Classes"}</h1>
 <code>java.time.LocalDateTime</code>, are <em>value-based</em>.  Instances of a
 value-based class:
 <ul>
-    <li>are final and immutable (though may contain references to mutable
+    <li>are final</li>
+    <li>have only final instance fields (though these may contain references to mutable
         objects);</li>
+    <li>extend either <code>Object</code> or a hierarchy of abstract classes that
+        declare no instance fields and have no instance initialization logic</li>
```
dansmithcode, line 42: This is new, and is relevant only to suitability for inline class migration. But it's true for all classes that claim to be value-based, so seems like a reasonable thing to state, and a good way to encourage people who care about this concept to structure their API in a migration-friendly way.

mlchung (approved):
This looks fine to me.

RogerRiggs (changes requested):
The existing value-based classes do not adhere to the new statements.
These changes are re-writing the spec after the fact.
Existing classes do not have to be final and can extend anything they want.

dansmithcode:
> The existing value-based classes do not adhere to the new statements.
> These changes are re-writing the spec after the fact.
> Existing classes do not have to be final and can extend anything they want.

Yes, this is a deceptively-small-looking change to the spec for a large number of classes.

You'll have to clarify which classes you're concerned about, though. I just verified that all of the following classes are both `final` and extend an appropriate superclass:

KeyValueHolder, MapN, Map1, SetN, Set12, ListN, List12, Optional, OptionalDouble, OptionalInt, OptionalLong, OffsetTime, ZonedDateTime, Duration, Instant, LocalDateTime, LocalTime, YearMonth, Year, MonthDay, OffsetDateTime, ZoneRegion, ZoneOffset, MinguoDate, HijrahDate, JapaneseDate, ThaiBuddhistDate, ProcessHandleImpl, Runtime.Version

(I stopped before digging into the ConstantDesc and foreign classes...)

RogerRiggs:
I was most concerned about the primitive wrapper classes, that are not currently specified to be value based (currently do not have references to ValueBased.html or similar statements).  The existing classes wrapper classes have some explicit identity requirements in the JLS.
Adding the @ValueBased annotation, with a javadoc reference to the updated ValueBased.html would be a specification change.
For JEP 390, the intention is to enable warnings, not to change the specification. 
So either the @ValueBased annotation is weaker than a spec change or the link to ValueBased.html must be a future aspirational statement, not a link to a current specification requirement.

mlchung:
> For JEP 390, the intention is to enable warnings, not to change the specification.
So either the @ValueBased annotation is weaker than a spec change or the link to ValueBased.html must be a future aspirational statement, not a link to a current specification requirement.

I guess you may be concerning about the potential compatibility risks by this proposed spec change.  Of course JEP 390 can make spec change if discussed and agreed.    There is no behavioral change to the primitive wrapper classes except the warnings are emitted if `synchronized` on these wrapper objects. 

@RogerRiggs can you clarify more what you are concerned about?

RogerRiggs:
Yes, the compatibility concern is foremost. We don't intend to change the behavior (yet) so the exact language used to describe what is being changed is important.  A simple assertion that Integer is ValueBased breaks compatibility.
The JLS requirement for reference equality for small values is at odds with ValueBased.

RogerRiggs:
Byte, Short, Integer, and Long all have caches for small values.

dansmithcode:
Yes, understood. I was using Integer as an example, but the above should apply to all of those classes.

dansmithcode:
New revision: I updated the definition, especially with respect to `==`, and applied to boilerplate to the wrapper classes. I realized the changed definition requires changing the boilerplate, so I've done so and fixed up all the references.

Please be hyper-critical of the boilerplate text, as this appears in lots of places—it's worth getting it just right.

dansmithcode:
I added a link to a docs build in the summary.

src/java.base/share/classes/java/lang/doc-files/ValueBased.html
```diff
@@ -31,36 +31,42 @@
...
+        are equal according to <code>equals()</code> produces no visible change in
+        the behavior of the class's methods;
     </li>
+    <li>perform no synchronization on an instance's intrinsic lock;</li>
+    <li>do not have (or have deprecated any) accessible constructors;</li>
+    <li>may support instance creation through factory methods that do <em>not</em>
+        promise a unique identity for each invocation—in particular, each factory
+        method must allow for the possibility that if two independently-produced
+        instances are equal according to <code>equals()</code>, they may also be
+        equal according to <code>==</code>;
```
RogerRiggs, line 55: Is using "may" a way to allow contradiction of the previous requirements?
Otherwise, in a spec, any statement using 'may' can be ignored or removed, it does not specify any behavior either required or disallowed.

src/java.base/share/classes/java/lang/doc-files/ValueBased.html
```diff
@@ -31,36 +31,42 @@
...
+<p>When two instances of a value-based class are `equal`, a program should not attempt
+    to distinguish between their identities, whether directly via reference
     equality or indirectly via an appeal to synchronization, identity hashing,
-    serialization, or any other identity-sensitive mechanism.  Use of such
-    identity-sensitive operations on instances of value-based classes may have
-    unpredictable effects and should be avoided.</p>
+    serialization, or any other identity-sensitive mechanism.</p>
+<p>Synchronization on instances of value-based classes is strongly discouraged,
+    because the programmer cannot usually guarantee unique ownership of the
+    associated lock.</p>
```
RogerRiggs, line 70: I don't think the phrase: 
`because the programmer cannot usually guarantee unique ownership of the associated lock.`
adds anything and is obscure.  Until the semantics of object change, there is no change in the synchronization behavior.
For Valhalla value-based instances, there is explicitly no (implicitly) associated lock, so synchronization is illegal.

RogerRiggs (commented):
In the text referring to ValueBase.html in many classes:

Is the "," misplaced.  Should it read:
"not use instances for synchronization or unpredictable behavior may occur."

If the comment in each class was a simple reference to ValueBased, it would be easier to maintain and there would be less duplication between lots of classes and the explanation in ValueBased.html.

src/java.base/share/classes/java/lang/doc-files/ValueBased.html
```diff
@@ -31,36 +31,42 @@
 <body>
 <h1 id="ValueBased">{@index "Value-based Classes"}</h1>
 
-Some classes, such as <code>java.util.Optional</code> and
-<code>java.time.LocalDateTime</code>, are <em>value-based</em>.  Instances of a
-value-based class:
+Some classes, such as <code>java.lang.Integer</code> and
+<code>java.time.LocalDateTime</code>, are <em>value-based</em>.
```
jodastephen, line 35: `LocalDate` is a better examplar than `LocalDateTime` because `LocalDate` is much more widely used.

jodastephen:
Can I suggest some additions?

* `java.time.temporal.ValueRange` - meets the criteria AFAICT

* `java.time.format.DecimalStyle` - meets the criteria AFAICT despite implementation internal cache of instances

* `java.time.Clock` subclasses - abstract base class equals/hashCode looks hostile, but JDK impls like `SystemClock` meet the criteria AFAICT

* `java.time.chrono.AbstractChronology` subclases - all five subclasses meet the criteria AFAICT

I don't think `java.time.temporal.WeekFields` meets the criteria as the factory promises "same instance", although it is conceptually value-based.

`java.util.Currency` is conceptually value-based, but equals/hashCode isn't suitable at present - I think this should be worked on, as currency is a key value-based concept.

dansmithcode:
> In the text referring to ValueBase.html in many classes:
> 
> Is the "," misplaced. Should it read:
> "not use instances for synchronization or unpredictable behavior may occur."

Grammatically, it's a compound sentence. Two statements: 1) programmers should do some stuff; 2) otherwise, unpredictable behavior. The comma helps to communicate where the "parentheses" go: (1a && 1b) || 2.

But I think you may be concerned that the "unpredictable behavior" only applies to the synchronization part of (1), not the "interchangeable" part? It's my intent that it applies to both. If you try to distinguish between instances with ==, or do synchronization, behavior will be unpredictable (because it's possible things that were != yesterday will be == today, per the rules about factories).

> If the comment in each class was a simple reference to ValueBased, it would be easier to maintain and there would be less duplication between lots of classes and the explanation in ValueBased.html.

Yes. This approach of scattering boilerplate throughout the API is not optimized for maintainability. But that's okay, because change will not be frequent, and meanwhile it's optimized for getting readers' attention instead. I would worry that if we stripped out all the boilerplate and just had a link, most readers wouldn't follow the link or appreciate what it meant.

src/java.base/share/classes/java/lang/doc-files/ValueBased.html
```diff
@@ -31,36 +31,42 @@
...
 <ul>
-    <li>are final and immutable (though may contain references to mutable
-        objects);</li>
-    <li>have implementations of <code>equals</code>,
+    <li>declare only final instance fields (though these may contain references
+        to mutable objects);</li>
+    <li>declare implementations of <code>equals</code>,
         <code>hashCode</code>, and <code>toString</code> which are computed
         solely from the instance's state and not from its identity or the state
         of any other object or variable;</li>
```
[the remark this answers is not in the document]
dansmithcode: Perhaps "not from its identity or the state of any other object or variable that is not part of the instance's state"?

src/java.base/share/classes/java/lang/doc-files/ValueBased.html
```diff
@@ -31,36 +31,42 @@
...
-    <li>are <em>freely substitutable</em> when equal, meaning that interchanging
-        any two instances <code>x</code> and <code>y</code> that are equal
-        according to <code>equals()</code> in any computation or method
-        invocation should produce no visible change in behavior.
+    <li>have instances that are <em>freely substitutable</em> when equal, meaning
+        that interchanging any two instances <code>x</code> and <code>y</code> that
+        are equal according to <code>equals()</code> produces no visible change in
+        the behavior of the class's methods;
     </li>
+    <li>perform no synchronization on an instance's intrinsic lock;</li>
```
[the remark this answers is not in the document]
dansmithcode: "perform no synchronization using an instance's monitor" (aligns with terminology in JLS 17.1)

src/java.base/share/classes/java/lang/doc-files/ValueBased.html
```diff
@@ -31,36 +31,42 @@
...
+    <li>do not have (or have deprecated any) accessible constructors;</li>
+    <li>may support instance creation through factory methods that do <em>not</em>
+        promise a unique identity for each invocation—in particular, each factory
```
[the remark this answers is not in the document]
dansmithcode: "Not allowed to promise a unique identity" based on what? Which other bullet in this list? There's the assertion about being "freely substitutable", but that doesn't mean a factory method can't promise to create a new instance on every invocation.

src/java.base/share/classes/java/lang/doc-files/ValueBased.html
dansmithcode: What I'm going for is "may declare factory methods, and if they do then...". But you're right, it doesn't work as written. I'll rephrase. (Why not require a factory method? Because a class could also just define a handful of named instances, enum-style, and not expose any factory-like behavior. Boolean is almost an example of this, although it does have a 'valueOf' method, too.)

src/java.base/share/classes/java/lang/doc-files/ValueBased.html
dansmithcode: The point is that if you can't prove you have a unique identity (per the factory method clause), then you can't guarantee that you have exclusive control over the monitor, so synchronization is dangerous. A variant of the text discouraging synchronization was there before any design discussions about primitive object semantics.

I added "usually" because of the possibility that someone can prove that a field value they provide is unique (according to 'equals') and thus can prove that the value-based class instance is also unique. But when I think about actual examples of value-based classes, that seems quite unlikely, and it would probably be clearer to remove the "usually".

src/java.base/share/classes/java/lang/doc-files/ValueBased.html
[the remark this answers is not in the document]
dansmithcode: Eh, the "any other object or variable" clause is too strong. What about a `toString` based on `System.lineSeparator()` or something like that? Seems like it's trying too hard to prohibit some kind of workaround to mutable state. But I think "computed solely" already communicates that pretty clearly.

How about just: "which are computed solely from the values of the class's instance fields (and properties of the objects they reference), not from the instance's identity"

src/java.base/share/classes/java/lang/doc-files/ValueBased.html
```diff
@@ -31,36 +31,40 @@
...
-        invocation should produce no visible change in behavior.
-    </li>
+        solely from the values of the class's instance fields (and properties of
+        the objects they reference), not from the instance's identity</li>
+    <li>have instances that are <em>freely substitutable</em> when equal, meaning
+        that interchanging any two instances <code>x</code> and <code>y</code> that
+        are equal according to <code>equals()</code> produces no visible change in
+        the behavior of the class's methods;</li>
+    <li>perform no synchronization using an instance's monitor;</li>
+    <li>do not have (or have deprecated any) accessible constructors;</li>
```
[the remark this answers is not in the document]
dansmithcode: The idea behind all the instance creation restrictions is this: clients of value-based classes should never be promised a unique, private identity associated with any instances they are given. By designing the API in this way, we discourage clients from inappropriate dependencies on identity, and pave the way for the classes to more smoothly migrate to be primitive classes someday.

Yes, primitive classes can have public constructors. But if a *migrated* primitive class has a public constructor, its clients will face binary and behavioral incompatibilities when the migration happens. If clients are using factories that don't promise unique identities, and aren't doing risky synchronization on objects they don't uniquely control, they will encounter neither of those incompatibilities.

The "or have deprecated any" rule allows for classes like the wrapper classes that do have accessible constructors, but have used deprecation to discourage clients from using them. Deprecation is a strong enough signal that when clients face future incompatibilities, they will have been sufficiently warned.

src/java.base/share/classes/java/lang/doc-files/ValueBased.html
```diff
@@ -31,36 +31,40 @@
...
+    <li>do not provide any other instance creation mechanism that promises
+        a unique identity on each method call&mdash;in particular, any factory
```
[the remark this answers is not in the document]
dansmithcode: Hmm. That's annoying. In what context? Worked okay when I did a `make docs`...

Are character entities *supposed* to work? Or is the coding convention to stick strictly to ASCII?

src/java.base/share/classes/java/lang/doc-files/ValueBased.html
```diff
@@ -31,36 +31,40 @@
...
+        method must allow for the possibility that if two independently-produced
+        instances are equal according to <code>equals()</code>, they may also be
+        equal according to <code>==</code>;</li>
+    <li>are final;</li>
```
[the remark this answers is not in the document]
dansmithcode: The way this has been handled until now is to say "all implementing classes should be value-based" in the abstract class's (or factory method's) javadoc. In terms of prose, that seems to get the job done, without needing to complicate the definition of *value-based class*.

For the annotation, yes the intent is that it will be applied both to value-based classes and to abstract classes/interfaces that require all implementing classes to be value-based.

src/java.base/share/classes/java/lang/doc-files/ValueBased.html
```diff
@@ -31,36 +31,40 @@
...
+<p>When two instances of a value-based class are `equal`, a program should not attempt
+    to distinguish between their identities, whether directly via reference
     equality or indirectly via an appeal to synchronization, identity hashing,
-    serialization, or any other identity-sensitive mechanism.  Use of such
-    identity-sensitive operations on instances of value-based classes may have
-    unpredictable effects and should be avoided.</p>
+    serialization, or any other identity-sensitive mechanism.</p>
+<p>Synchronization on instances of value-based classes is strongly discouraged,
+    because the programmer cannot guarantee exclusive ownership of the
+    associated monitor.</p>
```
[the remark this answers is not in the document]
dansmithcode: I'll revise to clarify that the bulleted list is about properties of the class declaration, while the subsequent sentences are constraints on/advice to clients. (Is there a better way to talk about clients than "a program should not..."? "Programmer", "client", "user" all kind of work, but I don't love any of them...) With that distinction in mind, I don't think the client stuff works very well until we first define what we mean by "value-based class".

I did remove what seemed to me like a redundant bullet from the original list (see my earlier comment). With what's left, each one is trying to say something distinct, although maybe some rephrasing in certain places could help.

The `final` class and field restrictions were combined before, but I split them because they're two very different constraints—one about extension, the other about immutability. (It's not ideal that Java uses the same keyword for both properties.) Actually, though, it works pretty well to combine the final class restriction with the superclass restriction, leaving one bullet all about subclassing. I'll do that.

The bullets that talk about equality are saying 1) the class needs an appropriate equals/hashCode/toString; 2) the class's instance methods don't have distinct behaviors for instances that are `equals`; 3) the class's factories do not promise distinct identities for instances that are `equals`. Those are three distinct things to say; I'm not seeing a clear way to combine them.

Synchronization: forget primitive classes. This is advice to current clients of value-based classes. And the advice is: don't synchronize, because you can't be sure someone else isn't doing the same on the same object. This is advice that is relevant to current clients without asking them to imagine a future in which the Object & monitor model has changed. Yet the implication is the same: don't do it.

dansmithcode:
Pushed some updates; here's an updated docs build: http://cr.openjdk.java.net/~dlsmith/8254275/8254275-20201016/api/index.html

dansmithcode:
> Can I suggest some additions?

Thanks for the suggestions, I've initiated a discussion about them. For this issue & JEP, we'll restrict ourselves to classes that already claim to be value-based (plus the wrappers). But it would be worthwhile to separately explore applying the term to additional Java SE classes. It can be handled as an RFE for each of the relevant APIs.

src/java.base/share/classes/java/lang/doc-files/ValueBased.html
```diff
@@ -31,36 +31,39 @@
...
+    <li>the class does not provide any other instance creation mechanism that promises
+        a unique identity on each method call&mdash;in particular, any factory
+        method must allow for the possibility that if two independently-produced
+        instances are equal according to <code>equals()</code>, they may also be
+        equal according to <code>==</code>;</li>
+    <li>the class is final, and extends either <code>Object</code> or a hierarchy of
+        abstract classes that declare no instance fields or instance initializers
+        and whose constructors are empty.</li>
+    <li>
 </ul>
```
RogerRiggs, line 60: Perhaps as an intro to the following points to make it clear these are about what a program should and should not do.
Should use equals(); should use explicit synchronization using lock objects or instances that are not value-based classes, etc.
`Programs should not attempt to distinguish the identities of value based class instances, otherwise the result may be unpredictable.`

mlchung (approved):

src/java.base/share/classes/java/lang/doc-files/ValueBased.html
```diff
@@ -31,36 +31,39 @@
...
-        based on reference equality (<code>==</code>);</li>
-    <li>do not have accessible constructors, but are instead instantiated
-        through factory methods which make no commitment as to the identity
-        of returned instances;</li>
-    <li>are <em>freely substitutable</em> when equal, meaning that interchanging
-        any two instances <code>x</code> and <code>y</code> that are equal
-        according to <code>equals()</code> in any computation or method
-        invocation should produce no visible change in behavior.
-    </li>
+        solely from the values of the class's instance fields (and properties of
```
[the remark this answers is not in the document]
dansmithcode: "members" would be the idiomatic Java word, I suppose

src/java.base/share/classes/java/lang/doc-files/ValueBased.html
```diff
@@ -31,36 +31,39 @@
...
+        the objects they reference), not from the instance's identity;</li>
+    <li>the class's methods treat instances as <em>freely substitutable</em>
+        when equal, meaning that interchanging any two instances <code>x</code> and
+        <code>y</code> that are equal according to <code>equals()</code> produces no
+        visible change in the behavior of the class's methods;</li>
+    <li>the class performs no synchronization using an instance's monitor;</li>
+    <li>the class does not declare (or has deprecated any) accessible constructors;</li>
+    <li>the class does not provide any other instance creation mechanism that promises
```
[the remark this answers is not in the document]
dansmithcode: Agree, I can drop "other".

The intent is that there's a restriction on a factory method's *contract*. I can make that explicit: "any factory method's contract must allow for the possibility..."

src/java.base/share/classes/java/lang/doc-files/ValueBased.html
dansmithcode: I'm not totally following. This is a revision to the "When two instances" paragraph? Can you propose an alternative phrasing for the entire paragraph?

RogerRiggs: These two paragraphs are fine, they express the negative, not what a developer should do.

RogerRiggs (approved):

src/java.base/share/classes/java/lang/doc-files/ValueBased.html
```diff
@@ -31,36 +31,39 @@
...
+Some classes, such as <code>java.lang.Integer</code> and
+<code>java.time.LocalDate</code>, are <em>value-based</em>.
+A value-based class has the following properties:
 <ul>
-    <li>are final and immutable (though may contain references to mutable
-        objects);</li>
-    <li>have implementations of <code>equals</code>,
+    <li>the class declares only final instance fields (though these may contain references
+        to mutable objects);</li>
+    <li>the class declares implementations of <code>equals</code>,
```
RogerRiggs, line 40: Does saying the class "declares" the methods apply to Records and other classes that have provided or generated methods for equals, hashcode, and toString?  Must the class explicitly declare the methods to meet the criteria?

Perhaps "the implementations of equals, hashCode, and tostring use only the instance's state...".

dansmithcode: Generated fields and methods are "implicitly declared". But I suppose acceptable implementations could be inherited from a superclass, if the superclass isn't Object. I'll fix.

RogerRiggs (approved):
Looks good, thanks for the updates.

PR #231, opened 2020-10-20
    8255054: Fix test broken by incorrect removal of java.lang.ValueBased

RogerRiggs:
Restore java.lang.ValueBased that was incorrectly removed by 8254957.


RogerRiggs:
The commits have the wrong contents. Withdrawing.

PR #237, opened 2020-10-26
8254274: Development to add 'lint' warning for @ValueBased classes

sadayapalam:
New javac lint mode (-Xlint:synchronize) to warn about client side as well JDK internal attempts to synchronize on instances of value based classes

---------
### Progress

### Testing

| Build | ✔️ (5/5 passed) | ✔️ (2/2 passed) | ✔️ (2/2 passed) |
| Test (tier1) | ✔️ (9/9 passed) | ✔️ (9/9 passed) | ✔️ (9/9 passed) |

### Issue
 * [JDK-8254274](https://bugs.openjdk.java.net/browse/JDK-8254274): Development to add 'lint' warning for @ValueBased classes

### Reviewers
 * [Dan Smith](https://openjdk.java.net/census#dlsmith) (@dansmithcode - Committer) ⚠️ Review applies to 1a660823bae70898f77188f9d1c563a5d6ee2301
 * [Jim Laskey](https://openjdk.java.net/census#jlaskey) (@JimLaskey - **Reviewer**)

### Download
`$ git fetch https://git.openjdk.java.net/valhalla pull/237/head:pull/237`
`$ git checkout pull/237`


sadayapalam:
A few questions for reviewers:

(1) Do we want the lint mode to be on my default ? Presently it is off and has to be expressly opted into.
(2) Do we want to not use up a bit for VALUE_BASED flag and simply walk the annotations to determine if a type is value based ??

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java
```diff
@@ -2418,6 +2421,40 @@ public void visitApply(JCMethodInvocation tree) {
 
             chk.checkRefTypes(tree.typeargs, typeargtypes);
 
+            if (localEnv.info.lint.isEnabled(LintCategory.SYNCHRONIZE)) {
```
dansmithcode, line 2424: Looks like you're checking calls to wait/notify? I don't think that's necessary—there will already be a warning at the `synchronized` statement; if that doesn't happen, it's malformed code that will fail at runtime (attempting to wait without owning the lock). In practice, I think we just end up with lots of redundant warnings.

src/jdk.compiler/share/classes/com/sun/tools/javac/resources/javac.properties
```diff
@@ -261,6 +261,9 @@ javac.opt.Xlint.desc.varargs=\
 javac.opt.Xlint.desc.preview=\
     Warn about use of preview language features
```
dansmithcode, line 262: Old code, I know, but the convention appears to be that these descriptions should end in a period. (Same comment for both `javac.opt.Xlint.desc.varargs` and `javac.opt.Xlint.desc.preview`.)

dansmithcode:
> (1) Do we want the lint mode to be on my default ? Presently it is off and has to be expressly opted into.

Given that it's already been established that there's an ad hoc set of on-by-default lint warnings, yeah, I think this one makes sense to put in that set. (Inventing a new concept of on-by-default lint warnings is something I'd be more hesitant about.)

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Lint.java
```diff
@@ -282,6 +282,11 @@ public String toString() {
          */
         STATIC("static"),
 
+        /**
+         * Warn about synchronization attempts on instances of @ValueBased classes.
+         */
+        SYNCHRONIZE("synchronize"),
```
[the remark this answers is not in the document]
sadayapalam: Thanks, I prefer the latter and will opt for it (synchronization)

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java
sadayapalam: Sounds good. I agree this is redundant and noisy.

src/jdk.compiler/share/classes/com/sun/tools/javac/resources/javac.properties
sadayapalam: OK, will do.

sadayapalam:
I have pushed additional changes to incorporate review comments:

    - The lint mode's token is "synchronization" (instead of "synchronize")
    - This lint mode is on by default
    - Redundant warnings on wait/notify calls are removed.
    - We don't burn up a flag bit to mark value based classes. We simply walk the declaration annotations on a class to discover the ones that are @ValueBased
    - Ensure descriptions (in the neighbourhood) in javac.properties end with a period for consistency's sake. 
    - Streamline and simplify tests.

dansmithcode (approved):

src/jdk.compiler/share/classes/com/sun/tools/javac/resources/javac.properties
```diff
@@ -256,10 +256,13 @@ javac.opt.Xlint.desc.unchecked=\
...
 javac.opt.Xlint.desc.varargs=\
-    Warn about potentially unsafe vararg methods
+    Warn about potentially unsafe vararg methods.
 
 javac.opt.Xlint.desc.preview=\
-    Warn about use of preview language features
+    Warn about use of preview language features.
+
+javac.opt.Xlint.desc.synchronization=\
+    Warn about synchronization attempts on instances of @ValueBased classes.
```
RogerRiggs, line 265: The annotation java.internal.ValueBased is purely internal and should not show up in error messages.
The reference to value based should be to the ValueBased.html or just be a lower case value-based.

sadayapalam: Good point. Fixed two places to say value-based consistently since this is the language mentioned by ValueBased.html.

JimLaskey (approved):
Wondering if keyword SYNCHRONIZATION is too broad, suggest VALUE_ SYNCHRONIZATION. Would it make sense to have a test for a class.ref type value?

sadayapalam:
> Wondering if keyword SYNCHRONIZATION is too broad, suggest VALUE_ SYNCHRONIZATION. Would it make sense to have a test for a class.ref type value?

class.ref is a valhalla'ism - but the present work is targetting JDK16 as a precursor/preparation for Valhalla. (even though the integration first happens to jep390 branch of Valhalla repo as a temporary home). So the .ref syntax is really not available to test/need not be tested.

As for the keyword, I am comfortable with it, but I will make it a point to solicit input at the next stage of review. (The accumulated cross component changes will get to a RFR when targetting jdk16)

PR #274, opened 2020-11-20
8252182: [JEP 390] Diagnose synchronization on @ValueBased classes

lfoltan:
Please review the following changes to add JVM support for @ValueBased class annotation.

---------
### Progress

### Testing

|     | Linux x64 | Linux x86 | Windows x64 | macOS x64 |
| --- | ----- | ----- | ----- | ----- |
| Build | ⏳ (4/6 running) | ⏳ (2/2 running) | ⏳ (2/2 running) | ⏳ (2/2 running) |

### Issue
 * [JDK-8252182](https://bugs.openjdk.java.net/browse/JDK-8252182): [JEP 390] Diagnose synchronization on @ValueBased classes

### Reviewers
 * [Frederic Parain](https://openjdk.java.net/census#fparain) (@fparain - Committer) ⚠️ Review applies to 0247b9f228c9f29d151d0a0a664d675bc40aaf58
 * [Harold Seigel](https://openjdk.java.net/census#hseigel) (@hseigel - **Reviewer**)

### Download
`$ git fetch https://git.openjdk.java.net/valhalla pull/274/head:pull/274`
`$ git checkout pull/274`


fparain (approved):
Looks good to me.
Just one comment: The ValueBased annotation is ignored for non-privileged classes, this would prevent users from testing their own code with the DiagnoseSyncOnValueBasedClasses flag. Is there a particular reason to not enforce the ValueBased annotation on user code?

Fred

hseigel (commented):
The changes look good!  Can the check for "DiagnoseSyncOnPrimitveWrappers != 0" be removed from lines 330 and 442?  I don't think "obj->klass()->is_value_based()" will be TRUE unless "DiagnoseSyncOnPrimitiveWrappers != 0".

Perhaps do an assert instead?

RogerRiggs:
> Looks good to me.
> Just one comment: The ValueBased annotation is ignored for non-privileged classes, this would prevent users from testing their own code with the DiagnoseSyncOnValueBasedClasses flag. Is there a particular reason to not enforce the ValueBased annotation on user code?
> 
> Fred

The jdk.internal.ValueBased annotation is internal to the base module and only defined for use by the JDK implementation as per JEP 390. It is intended to indicate that an application or library is using a system class instance incorrectly.

lfoltan:
Thank you Harold, I have made your suggested changes to runtime/synchronizer.cpp

lfoltan:
Thank you Fred for the review.

hseigel (approved):
Changes look good.  Thanks for doing this!

PR #288, opened 2020-12-02
8257600: [type-restrictions] Implement RestrictedField in C1

fparain:
Please review theses changes adding RestrictedField support to C1.
Tests have been added using the annotation added by JDK-8255856.
Tests revealed some issues with the support of RestrictedField in the interpreter, fixes are included in this patch.
The changes also includes some renaming after discussions with Dan.

Thank you,

Fred

---------
### Progress

### Issue
 * [JDK-8257600](https://bugs.openjdk.java.net/browse/JDK-8257600): [type-restrictions] Implement RestrictedField in C1

### Reviewers
 * [Lois Foltan](https://openjdk.java.net/census#lfoltan) (@lfoltan - Committer) ⚠️ Review applies to 21ac75622686d5305247551b83a5bf5562f6d8e2
 * [Tobias Hartmann](https://openjdk.java.net/census#thartmann) (@TobiHartmann - Committer) ⚠️ Review applies to 21ac75622686d5305247551b83a5bf5562f6d8e2

### Download
`$ git fetch https://git.openjdk.java.net/valhalla pull/288/head:pull/288`
`$ git checkout pull/288`


forax:
Hi,
the requirement that a restricted type must be an inline type is too strong,
the restricted type should be able to be at least the same as the field descriptor too otherwise you can not using it in a template class if the argument type is not an inline class.

And for other languages than Java that may want to support full reification, it's better to support any subtypes.

src/hotspot/share/classfile/classFileParser.cpp
```diff
@@ -6751,6 +6768,20 @@ void ClassFileParser::post_process_parsed_stream(const ClassFileStream* const st
...
                     InstanceKlass::cast(klass)->external_name()));
       }
+      if (fs.has_restricted_type()) {
+        // descriptor type supposed to be a super type of the restricted type, so after the pre-loading
+        // of the restricted type above, the descriptor type should be loaded at this point
+        Symbol* descriptor_name = _cp->symbol_at(_descriptor_signature_info->at(fs.index()));
+        Klass* desc_klass = SystemDictionary::resolve_or_null(descriptor_name,
+                                                   Handle(THREAD, _loader_data->class_loader()),
+                                                   _protection_domain, CHECK);
+        if (desc_klass == NULL || !klass->is_subtype_of(desc_klass)) {
```
lfoltan, line 6778: Consider if using a ResolvingSignatureStream & a call to ResolvingSignatureStream::as_klass() might be a more JVM consistent way to obtain the Klass from the descriptor signature.

lfoltan (approved):
The interpreter and class file parsing changes look good.  I left a few minor comments to consider.  I did not review in depth the c1 changes.
Thanks,
Lois

TobiHartmann (approved):
Compiler changes look good to me!

fparain:
Lois, Tobias,

thank you for your reviews.
Lois' comments have been addressed in the latest commit.

fparain:
Remi,

The current goal of the type restrictions branch is to measure the impact and the cost of type restrictions checks, so features are being added incrementally (not a full model yet). The current goal is not to provide a support for full reification, and some models being proposed on top of type restrictions are relying on the fact that type restrictions can only be applied to primitive classes.

For the first phase of this exploration, we should stick to the simple model until we reach the first mile-stone: full support of RestrictedField and RestrictedMethod in the interpreter, C1 and C2.

By the time we implement this first step, we hope we'll have a better understanding on how to evolve the model, relaxing constraints, keeping them or modifying them.

Regards,

Fred

forax:
Hi Frédéric,
nevermind, in the meantime, i can patch the attribute name so if it's not a record the attribute itself will not be recognized

PR #301, opened 2020-12-11
8228634: [lworld] ciField::will_link() returns incorrect result for the withfield bytecode

TobiHartmann:
In `GraphBuilder::withfield`, `ciField::will_link()` can return false even if the holder klass is loaded (see bug comments). In that case we should not deoptimize but patch the store instruction that writes the new field value. That is required even if the field offset is known because the field could not be accessible/available and we would need to throw an exception during patching. The load/store instructions emitted by `copy_inline_content` to initialize the fields of the new inline type buffer copy should never require patching: We know their offsets (because the holder is loaded) and we don't need to check access restrictions.

Unrelated changes:
- Array loads should not be delayed if the subsequent getfield requires patching (see changes to `GraphBuilder::load_indexed`)
- Replaced the `WithField` and `DefaultValue` IR nodes by a `Deoptimize` node
- Refactoring and removal of dead code

Thanks,
Tobias

---------
### Progress

### Issue
 * [JDK-8228634](https://bugs.openjdk.java.net/browse/JDK-8228634): [lworld] ciField::will_link() returns incorrect result for the withfield bytecode

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/valhalla pull/301/head:pull/301`
`$ git checkout pull/301`

### DIFF
--- a/src/hotspot/share/c1/c1_Canonicalizer.cpp
+++ b/src/hotspot/share/c1/c1_Canonicalizer.cpp
@@ -644,8 +644,7 @@ void Canonicalizer::do_NewInlineTypeInstance(NewInlineTypeInstance* x) {}
 void Canonicalizer::do_NewTypeArray   (NewTypeArray*    x) {}
 void Canonicalizer::do_NewObjectArray (NewObjectArray*  x) {}
 void Canonicalizer::do_NewMultiArray  (NewMultiArray*   x) {}
-void Canonicalizer::do_WithField      (WithField*       x) {}
-void Canonicalizer::do_DefaultValue   (DefaultValue*    x) {}
+void Canonicalizer::do_Deoptimize     (Deoptimize*      x) {}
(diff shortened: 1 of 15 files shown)


fparain (approved):
Hi Tobias,

With the discussion we had off-line about nestmates, changes look good to me.

Suggestion for a future improvement: the Deoptimize node in the HIR is a very nice improvement. However, if it looks like a general purpose node in the HIR, the LIRGenerator has a hard coded deopt reason for it: "Reason_unloaded". It would be nice to have a more flexible Deoptmize node where the reason could be specified at the HIR level.

Thank you,

Fred

TobiHartmann:
Thanks for the review Fred!

Yes, I agree that the Deoptimize node should be more generic. As we've discussed off-thread, we can address this once more reasons for deoptimization need to be supported (probably with the restricted fields prototype).

Thanks,
Tobias

PR #311, opened 2021-01-22
8260283: [lworld] C1's EliminateFieldAccess optimization fails with "wrong types"

TobiHartmann:
C1's delayed field access optimization ([JDK-8229897](https://bugs.openjdk.java.net/browse/JDK-8229897)) does not update the field when updating a `LoadField`. The `EliminateFieldAccess` optimization then uses the wrong field (offset) to look up the loaded value in the `MemoryBuffer`. We hit an assert because the field types do not match.

Thanks,
Tobias

---------
### Progress

### Issue
 * [JDK-8260283](https://bugs.openjdk.java.net/browse/JDK-8260283): [lworld] C1's EliminateFieldAccess optimization fails with "wrong types"

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/valhalla pull/311/head:pull/311`
`$ git checkout pull/311`


fparain (approved):
Looks good to me.

Fred

TobiHartmann:
Thanks Fred!

PR #343, opened 2021-02-19
8262025: [lworld] C2 should optimize acmp of the same inline type

TobiHartmann:
I've noticed that a simple `val == val` where `val` is an inline type is not optimized by C2. The reason is that we are buffering the inline type operands before applying optimizations. The fix is to simply check for equality before that.

Best regards,
Tobias

---------
### Progress

### Issue
 * [JDK-8262025](https://bugs.openjdk.java.net/browse/JDK-8262025): [lworld] C2 should optimize acmp of the same inline type

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/valhalla pull/343/head:pull/343`
`$ git checkout pull/343`

### DIFF
--- a/src/hotspot/share/opto/parse2.cpp
+++ b/src/hotspot/share/opto/parse2.cpp
@@ -2176,6 +2176,12 @@ void Parse::do_acmp(BoolTest::mask btest, Node* left, Node* right) {
     return;
   }
 
+  // Check for equality before potentially allocating
+  if (left == right) {
+    do_if(btest, makecon(TypeInt::CC_EQ));
+    return;
+  }
+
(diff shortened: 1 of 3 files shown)


TobiHartmann:
I've noticed that with the fix, the `Ifacmp.java` test started to time out due to an endless loop in the `testAlot` method. It seems that this is due to C2 now being able to fully remove the loop (because it can prove that `a == b`) and therefore also the safepoint that would trigger a GC. I've added a call to `System.gc`. @MrSimms does that seem reasonable to you?

MrSimms (approved):
Runtime bits look good.

TobiHartmann:
Thanks for the review!

PR #345, opened 2021-02-22
8262128: [lworld] C1's ValueNumbering optimization does not correctly handle delayed accesses

TobiHartmann:
C1's ValueNumbering optimization replaces an array+field load `array[0].intArray` by another non-equivalent array+field load `array[0].floatArray` because it does not handle delayed accesses. The hash value of `LoadIndexed` needs to take delayed accesses into account.

Best regards,
Tobias

---------
### Progress

### Issue
 * [JDK-8262128](https://bugs.openjdk.java.net/browse/JDK-8262128): [lworld] C1's ValueNumbering optimization does not correctly handle delayed accesses

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/valhalla pull/345/head:pull/345`
`$ git checkout pull/345`

### DIFF
--- a/src/hotspot/share/c1/c1_Instruction.cpp
+++ b/src/hotspot/share/c1/c1_Instruction.cpp
@@ -265,6 +265,14 @@ ciType* LoadIndexed::declared_type() const {
   return ak->element_type();
 }
 
+intx LoadIndexed::hash_inline_access() const {
+  if (delayed() != NULL) {
+    return HASH2(_vt, delayed()->offset());
+  } else {
+    return HASH1(_vt);
+  }
+}
+
 bool StoreIndexed::is_exact_flattened_array_store() const {
   if (array()->is_loaded_flattened_array() && value()->as_Constant() == NULL && value()->declared_type() != NULL) {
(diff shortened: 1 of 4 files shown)


fparain:
Hi Tobias,

Thank you for fixing this.
However, the fix seems to have an issue. The hash_inline_access() method encapsulates the additional offset from the delayed access into a hash value, this encapsulation prevents the is_equal(Value v) method (from the the HASHING macro) to see all individual arguments of the hash function and could cause the method return an incorrect result.

Fred

TobiHartmann:
Hi Fred,

thanks for looking at this, you are right.

Best regards,
Tobias

TobiHartmann:
As we've discussed offline, I've disabled ValueNumbering for delayed accesses for now and filed [https://bugs.openjdk.java.net/browse/JDK-8262136](JDK-8262136]) to re-enable it later.

fparain (approved):
Looks good to me.

Fred

TobiHartmann:
Thanks, Fred!

PR #348, opened 2021-02-23
8262231: [lworld] C2 compilation fails with assert "user must call transfer_exceptions_into_jvms"

TobiHartmann:
When running with `-XX:PerMethodSpecTrapLimit=0 -XX:PerMethodTrapLimit=0`, we hit an assert because the unsafe access intrinsic adds an exception to the `GraphKit` state but then bails out and therefore does not transfer that exception into the `JVMState`. In general, we should only modify the graph once we are sure we won't bail out from intrinsification. I've added an assert that catches this even without any VM flags and fixed all affected intrinsics.

Best regards,
Tobias

---------
### Progress

### Issue
 * [JDK-8262231](https://bugs.openjdk.java.net/browse/JDK-8262231): [lworld] C2 compilation fails with assert "user must call transfer_exceptions_into_jvms"

### Download
`$ git fetch https://git.openjdk.java.net/valhalla pull/348/head:pull/348`
`$ git checkout pull/348`

### DIFF
--- a/src/hotspot/share/opto/library_call.cpp
+++ b/src/hotspot/share/opto/library_call.cpp
@@ -134,16 +134,7 @@ JVMState* LibraryIntrinsic::generate(JVMState* jvms) {
     return kit.transfer_exceptions_into_jvms();
   }
   // The intrinsic bailed out
-#ifdef ASSERT
-  if (ctrl != kit.control()) {
-    ctrl->dump(0);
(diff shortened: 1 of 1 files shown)


TobiHartmann:
I've removed the assert because it triggers issues in mainline code. Will add the assert back into mainline after fixing issues there.

PR #361, opened 2021-03-04
8262287: [lworld] C2 compilation fails with assert "should have been removed from the graph"

rwestrel:
hook_memory_on_init() sets up memory edges so stores can be captured
at an allocation. To do that, it adds extra edges for each field after
and before the InitializeNode. This also happens for flat arrays
eventhough until Compile::adjust_flattened_array_access_aliases() runs
there's a single slice for all fields of a flat array element. The
extra edges added after the InitalizeNode can cause PhiNodes to be
created for each fields of the array element which confuses
Compile::adjust_flattened_array_access_aliases() (because it expects
no Phi for individual fields until it runs). I don't think the memory
edges out of the InitializeNode are needed as
Compile::adjust_flattened_array_access_aliases() should create
them. So I propose not adding them.

---------
### Progress

### Issue
 * [JDK-8262287](https://bugs.openjdk.java.net/browse/JDK-8262287): [lworld] C2 compilation fails with assert "should have been removed from the graph"

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/valhalla pull/361/head:pull/361`
`$ git checkout pull/361`

### DIFF
--- a/src/hotspot/share/opto/cfgnode.cpp
+++ b/src/hotspot/share/opto/cfgnode.cpp
@@ -1064,10 +1064,14 @@ void PhiNode::verify_adr_type(bool recursive) const {
   if (Node::in_dump())               return;  // muzzle asserts when printing
 
   assert((_type == Type::MEMORY) == (_adr_type != NULL), "adr_type for memory phis only");
+  // Flat array element shouldn't get their own memory slice until flattened_accesses_share_alias is cleared.
+  // It could be the graph has no loads/stores and flattened_accesses_share_alias is never cleared. EA could still
+  // creates per element Phis but that wouldn't be a problem as there are no memory accesses for that array.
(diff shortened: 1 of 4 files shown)


TobiHartmann (commented):
Looks good to me. Some comments wouldn't hurt though :) and could you please add the regression test? You can simply add it to `jtreg/compiler/valhalla/inlinetypes/TestGenerated.java` where the generated tests live.

rwestrel:
> Looks good to me. Some comments wouldn't hurt though :) and could you please add the regression test? You can simply add it to `jtreg/compiler/valhalla/inlinetypes/TestGenerated.java` where the generated tests live.

Thanks for the review. I updated the change accordingly. Sorry the force push but after merging github included extra changed and rebasing was the quickest way I found to fix this.

TobiHartmann (approved):
Looks good, thanks for making these changes.

PR #353, opened 2021-02-25
8258038: [AARCH64] [lworld] Fix inline type implementation

nick-arm:
This patch updates the original AArch64 lworld port from December 2019 to match the x86 C1/C2/interpreter changes since then. It doesn't support InlineTypePassFieldsAsArgs or InlineTypeReturnedAsFields: this can be added later.

Known issues:

* compiler/valhalla/inlinetypes/TestLWorld test9 fails with an internal C2 error:

```
  #  Internal Error (/home/nicgas01/valhalla/src/hotspot/share/opto/buildOopMap.cpp:360), pid=2011, tid=2025
  #  assert(false) failed: there should be a oop in OopMap instead of a live raw oop at safepoint
```

  I haven't investigated this but it doesn't seem to be related to any of the platform specific code. I'll make a separate JBS issue for it.

* compiler/valhalla/inlinetypes/TestArrays.java and TestNullableArrays.java fail some sub-tests with:

```
  stderr: [OpenJDK 64-Bit Server VM warning: InlineTypePassFieldsAsArgs is not supported on this platform
  OpenJDK 64-Bit Server VM warning: InlineTypeReturnedAsFields is not supported on this platform
  Exception in thread "main" java.lang.RuntimeException: Graph for 'TestNullableArrays::test1' contains different number of match nodes (expected 1 but got 0):
  : expected 1 to equal 0
  	at jdk.test.lib.Asserts.fail(Asserts.java:594)
  	at jdk.test.lib.Asserts.assertEquals(Asserts.java:205)
  	at jdk.test.lib.Asserts.assertEQ(Asserts.java:178)
  	at compiler.valhalla.inlinetypes.InlineTypeTest.parseOutput(InlineTypeTest.java:600)
  	at compiler.valhalla.inlinetypes.InlineTypeTest.execute_vm(InlineTypeTest.java:482)
  	at compiler.valhalla.inlinetypes.InlineTypeTest.run(InlineTypeTest.java:436)
  	at compiler.valhalla.inlinetypes.TestNullableArrays.main(TestNullableArrays.java:58)
  ]
```

  I think this is related to not implementing passing/returning fields.

* compiler/valhalla/inlinetypes/TestBufferTearing.java fails intermittently. I haven't investigated this and will make a separate JBS issue for it.

This patch includes two changes that I also submitted to openjdk/jdk:
* https://github.com/openjdk/jdk/commit/f7e0a09802f74
* https://github.com/openjdk/jdk/commit/be67aaabe63a

---------
[...]

PR #362, opened 2021-03-05
8263067: [lworld] Improve buffer instance allocation in MacroAssembler::store_inline_type_fields_to_buf

kelthuzadx:
TLAB allocation in MacroAssembler::store_inline_type_fields_to_buf looks duplicated with MacroAssembler::tlab_allocate, we might use existing MacroAssembler::tlab_allocate instead of duplicated ones. It will generate safer and faster code. (See details on JBS). All jtreg tests are passed with slowdebug mode. When UseTLAB is disabled, we can also try to allocate the buffer object in eden space.

Thanks!
Yang

---------
### Progress

### Issue
 * [JDK-8263067](https://bugs.openjdk.java.net/browse/JDK-8263067): [lworld] Improve buffer instance allocation in MacroAssembler::store_inline_type_fields_to_buf

### Reviewers
 * [Tobias Hartmann](https://openjdk.java.net/census#thartmann) (@TobiHartmann - Committer) ⚠️ Review applies to 90e3b7c92a448b9a1570e042c2b27472823a85e0

### Download
`$ git fetch https://git.openjdk.java.net/valhalla pull/362/head:pull/362`
`$ git checkout pull/362`

### DIFF
--- a/src/hotspot/cpu/x86/macroAssembler_x86.cpp
+++ b/src/hotspot/cpu/x86/macroAssembler_x86.cpp
@@ -5412,60 +5412,62 @@ int MacroAssembler::store_inline_type_fields_to_buf(ciInlineKlass* vk, bool from
   int call_offset = -1;
 
 #ifdef _LP64
+  // The following code is similar to allocate_instance but has some slightly differences,
+  // e.g. object size is always not zero, sometimes it's constant; storing klass ptr after
+  // allocating is not necessary if vk != NULL, etc. allocate_instance can not aware of these.
   Label slow_case;
-
-  // Try to allocate a new buffered inline type (from the heap)
-  if (UseTLAB) {
-    // FIXME -- for smaller code, the inline allocation (and the slow case) should be moved inside the pack handler.
-    if (vk != NULL) {
-      // Called from C1, where the return type is statically known.
-      movptr(rbx, (intptr_t)vk->get_InlineKlass());
-      jint lh = vk->layout_helper();
-      assert(lh != Klass::_lh_neutral_value, "inline class in return type must have been resolved");
-      movl(r14, lh);
+  // 1. Try to allocate a new buffered inline instance either from TLAB or eden space
+  mov(rscratch1, rax); // save rax for slow_case since *_allocate may corrupts it when allocation failed
+  if (vk != NULL) {
+    // Called from C1, where the return type is statically known.
+    movptr(rbx, (intptr_t)vk->get_InlineKlass());
+    jint obj_size = vk->layout_helper();
+    assert(obj_size != Klass::_lh_neutral_value, "inline class in return type must have been resolved");
+    if (UseTLAB) {
+      tlab_allocate(r15_thread, rax, noreg, obj_size, r13, r14, slow_case);
     } else {
-      // Call from interpreter. RAX contains ((the InlineKlass* of the return type) | 0x01)
-      mov(rbx, rax);
-      andptr(rbx, -2);
-      movl(r14, Address(rbx, Klass::layout_helper_offset()));
+      eden_allocate(r15_thread, rax, noreg, obj_size, r13, slow_case);
     }
-
-    movptr(r13, Address(r15_thread, in_bytes(JavaThread::tlab_top_offset())));
-    lea(r14, Address(r13, r14, Address::times_1));
-    cmpptr(r14, Address(r15_thread, in_bytes(JavaThread::tlab_end_offset())));
-    jcc(Assembler::above, slow_case);
-    movptr(Address(r15_thread, in_bytes(JavaThread::tlab_top_offset())), r14);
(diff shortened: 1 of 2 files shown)


TobiHartmann:
Hi Yang,

I had this on my To-Do list as well, thanks for fixing! I'm just wondering if we shouldn't also attempt an allocation in eden if `UseTLAB` is disabled? Similar to what code in `MacroAssembler::allocate_instance` does.

Thanks,
Tobias

kelthuzadx:
Hi Tobias,

I think it's reasonable. We can create such a buffered instance by `MacroAssembler::allocate_instance`(maybe with some modifications). What do you think? If you think it's a feasible way, I can also try to tweak the implementation, keeping using the same buffer allocation mechanism on  aarch64.

Thanks!
Yang

TobiHartmann:
Yes, I think it would be nice to factor the common code out of `MacroAssembler::allocate_instance` (we don't need the zeroing and some of the checks) and use it for buffer allocation.

kelthuzadx:
Now it will try to allocate a new buffered inline instance either from TLAB or eden space. I tried to build and change its counterpart on aarch64 but failed. Fixing these errors may conflict with #353, so I keep it unchanged until #353 merged.

I've noticed a [FIXME comment](https://github.com/openjdk/valhalla/pull/362/files#diff-5d34d63ceb6d72dd9124399b7628bf498b4dec22ea5c5edc0426e80ee6e3bae4R5456) that hope to inline buffer packing in place, it would reduce a call instruction and do all stuff(buffer allocation+packing) in fastpath, but on the other hand, it increases the compiled code size and the final code may be duplicated with pack handler in SharedRuntime::generate_buffered_inline_type_adapter. I'm not sure which is more attractive to us, or keep it unchanged? Looking forward to your inputs.

Thanks!
Yang

src/hotspot/cpu/x86/macroAssembler_x86.cpp
```diff
@@ -5412,60 +5412,62 @@ int MacroAssembler::store_inline_type_fields_to_buf(ciInlineKlass* vk, bool from
   int call_offset = -1;
 
 #ifdef _LP64
+  // The following code is similar to allocate_instance but has some slightly differences,
```
TobiHartmann, line 5415: `slightly differences` -> `slight differences`

src/hotspot/cpu/x86/macroAssembler_x86.cpp
```diff
@@ -5412,60 +5412,62 @@ int MacroAssembler::store_inline_type_fields_to_buf(ciInlineKlass* vk, bool from
...
+  // e.g. object size is always not zero, sometimes it's constant; storing klass ptr after
+  // allocating is not necessary if vk != NULL, etc. allocate_instance can not aware of these.
```
TobiHartmann, line 5417: `allocate_instance can not aware of these` -> `allocate_instance is not aware of these`

src/hotspot/cpu/x86/macroAssembler_x86.cpp
```diff
@@ -5412,60 +5412,62 @@ int MacroAssembler::store_inline_type_fields_to_buf(ciInlineKlass* vk, bool from
...
+      mov(r13, rbx);
     }
     Register tmp_store_klass = LP64_ONLY(rscratch1) NOT_LP64(noreg);
-    store_klass(r13, rbx, tmp_store_klass);  // klass
-
-    // We have our new buffered inline type, initialize its fields with an inline class specific handler
+    store_klass(buffer_obj, rbx, tmp_store_klass);
+    // 3. Initialize its fields with an inline class specific handler
     if (vk != NULL) {
       // FIXME -- do the packing in-line to avoid the runtime call
```
TobiHartmann, line 5456: This can be removed. I don't think it pays off.

src/hotspot/cpu/x86/macroAssembler_x86.cpp
```diff
@@ -5412,60 +5412,62 @@ int MacroAssembler::store_inline_type_fields_to_buf(ciInlineKlass* vk, bool from
...
-  if (UseTLAB) {
-    // FIXME -- for smaller code, the inline allocation (and the slow case) should be moved inside the pack handler.
-    if (vk != NULL) {
-      // Called from C1, where the return type is statically known.
-      movptr(rbx, (intptr_t)vk->get_InlineKlass());
-      jint lh = vk->layout_helper();
-      assert(lh != Klass::_lh_neutral_value, "inline class in return type must have been resolved");
-      movl(r14, lh);
+  // 1. Try to allocate a new buffered inline instance either from TLAB or eden space
+  mov(rscratch1, rax); // save rax for slow_case since *_allocate may corrupts it when allocation failed
```
TobiHartmann, line 5420: `corrupts it` -> `corrupt it`

TobiHartmann (approved):
Found some little typos in the comments, otherwise the changes look good to me. I think the `FIXME` you've mentioned can be removed.

Best regards,
Tobias

kelthuzadx:
Thank you @TobiHartmann !

/integrate

PR #364, opened 2021-03-07
javac support for Parametric VM, first stab

vicente-romero-oracle:
This patch is the initial support for Parametric VM in javac. This is the first iteration of the prototype, so still a work in progress. In order to test the new code users need to pass the hidden option: `-XDsupportParametricVM` to javac. See the examples below for an illustration of the bytecode the compiler can generate now. Lets assume we have the following class annotated with the `@Parametric` annotation:

```
import java.lang.annotation.*;

@Parametric(id="PAIR")
class Pair<X, Y> {
    @Parametric(id="PAIR")
    X fst;
    @Parametric(id="PAIR")
    Y snd;

    Pair(X fst, Y snd) {
        this.fst = fst;
        this.snd = snd;
    }

    Pair<Y, X> swap() {
        return new Pair<>(snd, fst);
    }

    @Parametric(id="make", kind=ParamKind.METHOD_ONLY)
    static <U, W> Pair<U, W> make(U fst, W snd) {
        return new Pair<>(fst, snd);
    }

    @Parametric(id="setFst", kind=ParamKind.METHOD_AND_CLASS)
    <Z> Pair<Z, Y> setFst(Z newFst) {
        return null;
    }
}
```

After compilation this class will contain the following new entries in the constant pool:

```
#22 = Parameter          CLASS:0        // CLASS:0
#37 = Parameter          METHOD_ONLY:0  // METHOD_ONLY:0
#44 = Parameter          METHOD_AND_CLASS:0 // METHOD_AND_CLASS:0
```
also there will be several elements with the new attribute `Parametric` which will appear in both fields, the `Pair` class per-se and methods: `make` and `setFst`.

Let's now assume we define a Client willing to use class `Pair`, it will need to define values to bound to the "holes" defined in class `Pair` using the `Parametric` annotation. For this class Client should use annotations `@LinkageClass` and or `@LinkageMethod` depending on the kind of hole in class `Pair` it is intending to bound to. Let's define class `Client` as:

```
import java.lang.annotation.*;

class Client {
    @LinkageClass("linkageClass")
    void linkClassParams() {
        Pair<String, Integer> psi = new Pair<>("first", 2);
    }

    @LinkageClass("linkageClass")
    void linkClassParamsFieldAccess(Pair<String, Integer> psi) {
        psi.fst = "field";
    }


    @LinkageMethod("linkageMethod1")
    void linkMethodParams(Pair<String, Integer> psi) {
        Pair<String, Integer> pair = Pair.make("hello", 1);
    }

    @LinkageClass("linkageClass")
    @LinkageMethod("linkageMethod2")
    void linkClassAndMethodParams(Pair<String, Integer> psi) {
        Pair<String, Integer> pss = psi.setFst("b");
    }
}
```

again compiling this class with the hidden option `-XDsupportParametricVM` we can find these new entries in the CP:

```
   #7 = Linkage            #8:#9          // linkageClass:Pair
   #8 = String             #10            // linkageClass
   #9 = Class              #11            // Pair
```

this Linkage CP entry is referred from method: `linkClassParams` as:

```
  void linkClassParams();
    descriptor: ()V
    flags: (0x0000)
    Code:
      stack=4, locals=2, args_size=1
         0: new           #7                  // Linkage linkageClass:Pair
```

also:

```
  #25 = Linkage            #8:#26         // linkageClass:Pair.fst:Ljava/lang/Object;
  #26 = Fieldref           #7.#27         // Pair.fst:Ljava/lang/Object;
  #27 = NameAndType        #28:#29        // fst:Ljava/lang/Object;
```

referred from method: `linkClassParamsFieldAccess` the bytecode looks like:

```
  void linkClassParamsFieldAccess(Pair<java.lang.String, java.lang.Integer>);
    descriptor: (LPair;)V
    flags: (0x0000)
    Code:
      stack=2, locals=2, args_size=2
         0: aload_1
         1: ldc           #23                 // String field
         3: putfield      #25                 // Linkage linkageClass:Pair.fst:Ljava/lang/Object;
         6: return
```

also:

```
  #32 = Linkage            #33:#34        // linkageMethod1:Pair.make:(Ljava/lang/Object;Ljava/lang/Object;)LPair;
  #33 = String             #35            // linkageMethod1
  #34 = Methodref          #9.#36         // Pair.make:(Ljava/lang/Object;Ljava/lang/Object;)LPair;
```

this time referred by method: `linkMethodParams` with bytecode:

```
  void linkMethodParams(Pair<java.lang.String, java.lang.Integer>);
    descriptor: (LPair;)V
    flags: (0x0000)
    Code:
      stack=2, locals=3, args_size=2
         0: ldc           #30                 // String hello
         2: iconst_1
         3: invokestatic  #14                 // Method java/lang/Integer.valueOf:(I)Ljava/lang/Integer;
         6: invokestatic  #32                 // Linkage linkageMethod1:Pair.make:(Ljava/lang/Object;Ljava/lang/Object;)LPair;
         9: astore_2
        10: return
```

and finally:

```
  #41 = Linkage            #42:#43        // linkageMethod2:Pair.setFst:(Ljava/lang/Object;)LPair;
  #42 = String             #44            // linkageMethod2
  #43 = Methodref          #7.#45         // Pair.setFst:(Ljava/lang/Object;)LPair;
```

in this last example it can be seen that Methodref at #43 is also pointing to #7 which, as commented above, is another Linkage_info entry in the CP.

TIA, for the feedback and comments,
Vicente

---------
### Progress

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/valhalla pull/364/head:pull/364`
`$ git checkout pull/364`

### DIFF
--- a/src/java.base/share/classes/java/lang/annotation/Parametric.java
+++ b/src/java.base/share/classes/java/lang/annotation/Parametric.java
@@ -1,0 +1,43 @@
+/*
+ * Copyright (c) 2021, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+package java.lang.annotation;
+
+/** Annotates elements that are to be parameterized
+ */
+@Target({ElementType.TYPE, ElementType.FIELD, ElementType.METHOD, ElementType.CONSTRUCTOR})
+public @interface Parametric {
+    /**
+     * the id
+     * @return the id
+     */
+    String id();
+
+    /**
+     * the kind
+     * @return the kind
+     */
+    ParamKind kind() default ParamKind.CLASS;
+}
--- a/src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/ClassWriter.java
+++ b/src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/ClassWriter.java
@@ -108,6 +108,8 @@ public class ClassWriter extends ClassFile {
 
     private Check check;
 
+    private boolean supportParametricVM;
+
     /**
      * If true, class files will be written in module-specific subdirectories
      * of the CLASS_OUTPUT location.
@@ -192,6 +194,8 @@ public class ClassWriter extends ClassFile {
             dumpInnerClassModifiers = modifierFlags.indexOf('i') != -1;
             dumpMethodModifiers = modifierFlags.indexOf('m') != -1;
         }
+
+        supportParametricVM = options.isSet("supportParametricVM");
     }
 
     public void addExtraAttributes(ToIntFunction<Symbol> addExtraAttributes) {
@@ -371,6 +375,10 @@ public class ClassWriter extends ClassFile {
             int alenIdx = writeAttr(names.Signature);
             databuf.appendChar(poolWriter.putSignature(sym));
             endAttr(alenIdx);
+            acount++;
+        }
+        if (supportParametricVM && sym.attribute(syms.parametricType.tsym) != null) {
+            writeParametricAttribute(sym);
             acount++;
         }
         acount += writeJavaAnnotations(sym.getRawAttributes());
@@ -858,6 +866,26 @@ public class ClassWriter extends ClassFile {
             acount += writeMemberAttrs(v, true);
             endAttrs(acountIdx, acount);
         }
+        endAttr(alenIdx);
+        return 1;
+    }
+
+    int writeParametricAttribute(Symbol sym) {
+        int alenIdx = writeAttr(names.Parametric);
+        int kind = 1; // the default
+        Attribute.Compound c = sym.attribute(syms.parametricType.tsym);
+        Attribute value = c.member(names.kind);
+        if (value != null && value instanceof Attribute.Enum) {
+            Name kindName = ((Attribute.Enum)value).value.name;
+            kind = switch (kindName.toString()) {
+                case "CLASS" -> 1;
+                case "METHOD_ONLY" -> 2;
+                case "METHOD_AND_CLASS" -> 3;
+                default -> throw new AssertionError("unexpected kind");
+            };
+        }
+        databuf.appendChar(poolWriter.putParameter(kind));
+        databuf.appendChar(0);
         endAttr(alenIdx);
         return 1;
     }
@@ -1707,6 +1735,11 @@ public class ClassWriter extends ClassFile {
 
         if (!poolWriter.innerClasses.isEmpty()) {
             writeInnerClasses();
+            acount++;
+        }
+
+        if (supportParametricVM && c.attribute(syms.parametricType.tsym) != null) {
+            writeParametricAttribute(c);
             acount++;
         }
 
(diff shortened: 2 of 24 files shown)


src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/PoolWriter.java
```diff
@@ -124,16 +164,100 @@ int putClass(Type t) {
...
+            return currentMethSymbol != null && (currentMethSymbol.attribute(syms.linkageMethodType.tsym) != null ||
+                    currentMethSymbol.attribute(syms.linkageClassType.tsym) != null);
+        }
+
+        /* this method will try to generate a Linkage that will eventually be written into the constant pool
+         * if it can't find all the needed information it will bail out, and return -1, to inform the caller
+         * to generate the corresponding legacy entry. If successful it will return the index of the corresponding
+         * Linkage_info constant in the constant pool
+         */
+        int generateLinkage(Object s, Attribute.Compound parametricAnno) {
```
mcimadamore, line 201: This can probably be simplified, by having this method return the constant to be added - e.g.
```
Pool.Constant wrapWithLinkageIfNeeded(...)
```
And then let clients call `writeIfNeeded`, so that if flow is also more consistent with other pool methods.

mcimadamore (commented):
Overall looks solid - I see that that there's no ClassReader support, which is probably ok given that we rely on (declaration) annotations to do the job.

src/java.base/share/classes/java/lang/annotation/Parametric.java
```diff
@@ -0,0 +1,43 @@
...
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+package java.lang.annotation;
+
+/** Annotates elements that are to be parameterized
+ */
+@Target({ElementType.TYPE, ElementType.FIELD, ElementType.METHOD, ElementType.CONSTRUCTOR})
+public @interface Parametric {
```
mcimadamore, line 31: Since we depend on this during code generation, should this has runtime retention? Otherwise, with separate compilation, I don't think we're able to see that a class is parameteric (if we're using it from a different client) ?

vicente-romero-oracle:
> Tried out on few examples, and it seems to work as expected - javap support working too. Well done!

thanks for the review!

src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/ClassWriter.java
```diff
@@ -862,6 +870,26 @@ int writeRecordAttribute(ClassSymbol csym) {
...
+        if (value != null && value instanceof Attribute.Enum) {
+            Name kindName = ((Attribute.Enum)value).value.name;
+            kind = switch (kindName.toString()) {
+                case "CLASS" -> 1;
+                case "METHOD_ONLY" -> 2;
+                case "METHOD_AND_CLASS" -> 3;
+                default -> throw new AssertionError("unexpected kind");
+            };
+        }
+        databuf.appendChar(poolWriter.putParameter(kind));
```
mcimadamore, line 887: There is something fishy here: note that parameter have an "id" - which can be reused by other entries. That is, you can have 5 different @Parameteric annotations, referring to the same parameter ID. Wouldn't this create duplicate parameters in the constant pool?

As a quick hack - we could pretend that the parameter CP entry had a "name" so that, instead of doing `putParameter`, you do `putParameter(name)` and poolWriter will use existing uniqueness logic to determine if a new parameter has to be emitted.

src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/PoolConstant.java
```diff
@@ -221,4 +222,57 @@ public Object poolKey(Types types) {
...
+        }
+
+        @Override
+        public int poolTag() {
+            return ClassFile.CONSTANT_Parameter;
+        }
+
+        @Override
+        public Object poolKey(Types types) {
+            return Integer.valueOf(kind);
```
mcimadamore, line 244: This is not good enough - as it means that you can have only one parameter of each kind.
See also my other related comment on this topic. I think this Parameter entry should have a "fake" String id, so that we can give them a name, and distinguish between them; then you can just use the String id as a pool key.

mcimadamore (commented):
I realized that the logic for adding parameter entries is not 100% correct - added some comments.

vicente-romero-oracle:
Hi Maurizio, I have added another commit (5b6afe9)[https://github.com/openjdk/valhalla/pull/364/commits/5b6afe9f90d1a90867a925ad3e443a20462f8524] that, I hope, fixes the issue with the Parameter CP you found, thanks for your comments

src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/PoolConstant.java
```diff
@@ -228,20 +228,26 @@ public Object poolKey(Types types) {
      */
     final class Parameter implements PoolConstant {
 
+        String id;
         final int kind;
 
         Parameter(int kind) {
```
forax, line 234: Why do you create a Parameter without an id ?

mcimadamore: That is during class reading - where we don't care about the "id" anymore. The "id" is just a "fiction" when writing, to help us generating multiple CP holes (possibly with the same kind).

mcimadamore (approved):
Looks good to me!

src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/PoolConstant.java
forax: Ok, i did not think about that part of the process, and you do not do any roundtrip, reader - >  javac -> gen,
thanks

vicente-romero-oracle:
> Looks good to me!

thanks!

src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/PoolConstant.java
vicente-romero-oracle: right roundtrips shouldn't happen, this is why we have this duality on the constructors

PR #368, opened 2021-03-15
8263568: [lworld] Fix residual reference to 'value'

kelthuzadx:
javap tool is still outputting primitive type class modifiers as `value` class, it looks like we should tweak this.

(Actually, I'm wondering what's the meaning of 'native' for fields that annotated with ACC_PRIMITIVE in flagToModifier? Should we remove it and change the `case 0x100` to `case ACC_PRIMITIVE` so that IDE will help us find all references if we want to do some code refactor in the future?)

Thanks,
Yang

---------
### Progress

### Issue
 * [JDK-8263568](https://bugs.openjdk.java.net/browse/JDK-8263568): [lworld] Fix residual reference to 'value'

### Reviewers
 * [Roger Riggs](https://openjdk.java.net/census#rriggs) (@RogerRiggs - Committer) ⚠️ Review applies to 5640dc868b3fc521cc1498b9fe9b8f0d373f9150

### Download
To checkout this PR locally:
`$ git fetch https://git.openjdk.java.net/valhalla pull/368/head:pull/368`
`$ git checkout pull/368`

To update a local copy of the PR:
`$ git checkout pull/368`
`$ git pull https://git.openjdk.java.net/valhalla pull/368/head`

### DIFF
--- a/src/jdk.jdeps/share/classes/com/sun/tools/classfile/AccessFlags.java
+++ b/src/jdk.jdeps/share/classes/com/sun/tools/classfile/AccessFlags.java
@@ -205,6 +205,7 @@ public class AccessFlags {
             case ACC_VOLATILE:
                 return "volatile";
             case 0x100:
+                // ACC_NATIVE or ACC_PRIMITIVE
                 return (t == Kind.Class || t == Kind.InnerClass) ? "primitive" : "native";
             case ACC_ABSTRACT:
                 return "abstract";
--- a/src/jdk.jdeps/share/classes/com/sun/tools/classfile/AccessFlags.java
+++ b/src/jdk.jdeps/share/classes/com/sun/tools/classfile/AccessFlags.java
@@ -205,7 +205,7 @@ public class AccessFlags {
             case ACC_VOLATILE:
                 return "volatile";
             case 0x100:
-                return (t == Kind.Class || t == Kind.InnerClass) ? "value" : t == Kind.Field ? "__Flattenable" : "native";
+                return (t == Kind.Class || t == Kind.InnerClass) ? "primitive" : "native";
             case ACC_ABSTRACT:
                 return "abstract";
             case ACC_STRICT:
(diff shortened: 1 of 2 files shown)


sadayapalam:
Thanks for the patch, I will review it tomorrow and help take it forward

RogerRiggs (approved):
Thanks for fixing the oversight.

sadayapalam:
A clarification on the Access flag encoding value 0x100:

On jdk mainline 0x100 is ACC_NATIVE used to flag native *methods*. (does not apply to or make sense for fields)

On Valhalla, this flag is overloaded to mean ACC_PRIMITIVE for classes and till https://bugs.openjdk.java.net/browse/JDK-8212792 pulled the plug was also overloaded for ACC_FLATTENABLE to flag fields that carried the flattenability hint.

Post https://bugs.openjdk.java.net/browse/JDK-8212792,  ACC_FLATTENABLE is gone and we have only two overloads
ACC_NATIVE and ACC_PRIMITIVE.

So the proposed patch looks good, but the 0x100 cannot be replaced with ACC_PRIMITIVE without some loss of readability.

sadayapalam:
(There is some loss of readability already - we should add a comment next to the case 0x100: to call out the variants. That is better than inventing a name like ACC_NATIVE_OR_PRIMITIVE - I will follow up on that)

sadayapalam:
Hello!

Can you please make these two changes ?

(1) This patch is "failing" one langtools test - test/langtools/tools/javac/valhalla/lworld-values/QTypeTest.java
The fix is simple, to replace value with primitive in the javap output verified by that test

(2) Could you add a comment next to the case 0x100: // ACC_NATIVE or ACC_PRIMITIVE 

Thanks!

kelthuzadx:
> Hello!
> 
> Can you please make these two changes ?
> 
> (1) This patch is "failing" one langtools test - test/langtools/tools/javac/valhalla/lworld-values/QTypeTest.java
> The fix is simple, to replace value with primitive in the javap output verified by that test
> 
> (2) Could you add a comment next to the case 0x100: // ACC_NATIVE or ACC_PRIMITIVE
> 
> Thanks!

I've changed the test case and add this comment. 

Thanks~
Yang

sadayapalam (approved):
Changes look good. Thanks!

sadayapalam:
Hello Yang, I think the process is for you to issue an integrate command followed by a sponsor command by me. Thanks for looking into this.

kelthuzadx:
Sure, thanks @RogerRiggs @sadayapalam for the review!

/integrate

PR #369, opened 2021-03-19
Make "PrimitiveParameterizedClass.default" a poly expression.

jespersm:
Make .default a separate node type in the parser.

## Issue
[JDK-8211914](https://bugs.openjdk.java.net/browse/JDK-8211914): [lworld] Javac should support type inference for default value creation

Note: The Linux x86 builds in GitHub actions seem to fail with something completely unrelated to these changes.

---------
### Progress

### Integration blocker
&nbsp;⚠️ Whitespace errors (failed with the updated jcheck configuration)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/369/head:pull/369` \
`$ git checkout pull/369`

`$ git checkout pull/369` \
`$ git pull https://git.openjdk.org/valhalla.git pull/369/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 369`

`$ git pr show -t 369`

</details>

<a href="https://git.openjdk.org/valhalla/pull/369.diff">https://git.openjdk.org/valhalla/pull/369.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/369#issuecomment-802621811)


mcimadamore (commented):
Nice work - it's a very nice cleanup of the code base! I like how the code that was previously scattered in several places now can just move where it belongs.

Functionality-wise, I'm not sure this supports poly expressions yet - e.g. I see that you set the polyKind tag correctly, but I'm afraid that this along is not sufficient to support stuff  like:

```
List<String> ls = List.default;
```

or:

```
void m(List<String> ls) { ... }

m(List.default)
```

I think you might be led to believe that the current impl is doing the right thing - but my feeling is that javac might be tricking you: if the expected type is e.g. `List<String>` but the qualifier of the default expression is `List`, the compiler will check (as per checkId) that List <: List<String> - which is true by subtyping conversion. I don't think there is any inference going on, in other words. You might try debugging and looking at the type of the expression that comes out of checkId.

Anyway, even if this works and I did miss something, I don't see how this approach can scale to default expression passed as method arguments - given that in that case there is no `pt()`. To add support for that you need to add another case/node in `ArgumentAttr` (see what's been done for anonymous inner classes - this should be simpler as there's no constructor call).

In other words, you want Attr.visitApply to treat your default expression method argument in a special way - by creating what we call a "deferred type" - that is a type that is not fully inferred until _after_ overload resolution (when you DO know the pt()).

mcimadamore:
So, I debugged what happens in case where we have

`List<String> ls = List.default`

What I missed is that we create a variable symbol for `default` which has the right type (from the expected type) - so the expression is correctly typed as `List<String>`. But this seems to happen "just because" the compiler detects that the expected type has same base class as the default expression type. Then, the check performed by `checkId` is always vacuously satisfied, given that we're passing in a synthetic variable whose type is the same as the expected one.

Counter example:

```
interface Foo<X extends Number> extends List<X> { }
...
List<String> ls = Foo.default; // still ok, type is `Foo`
```

This example is more complex, for two reasons:

* in this case, Foo is a sub-interface, so the trick of saying "if base class is the same just use the expected type" doesn't work. In fact the compiler types this with `Foo` - which is a raw type
* in this particular instance, the code should not even be allowed - because the constraint on the `Foo` type parameter are incompatible with those of the expected type List<String> (X must extend Number).

All this stuff needs to work correctly if we want to claim that default expressions are true poly expressions.

jespersm:
@mcimadamore : I've opened https://bugs.openjdk.java.net/browse/JDK-8263900 for the refactoring alone. I'm not totally confident that all the checks in Attr are syntactically possible, and should likely be turned into assertions.

So this PR is temporarily on hold, awaiting merge of #370

sadayapalam:
Hello Jesper, this latest patch is problematic. it does not properly treat Foo.default as a poly expression. As Maurizio's earlier review comments points out: 

```
I don't see how this approach can scale to default expression passed as method arguments - given that in that case there is no pt(). To add support for that you need to add another case/node in ArgumentAttr (see what's been done for anonymous inner classes - this should be simpler as there's no constructor call).

In other words, you want Attr.visitApply to treat your default expression method argument in a special way - by creating what we call a "deferred type" - that is a type that is not fully inferred until after overload resolution (when you DO know the pt()).
```

I don't see the relevant code changes in your change set - i.e there is no changes in ArgumentAttr {} - I would have expected to see a visitDefaultValue method there.

As Maurizio further pointed out earlier, in Attr.visitApply, right after arguments are attributed, the type of Foo.default as a method argument should be a deferred type . If I debug the uploaded patch, the type of Foo.default is actually Foo<T>

The problem is masked by there being only negative tests in your patch.

If I compile the following:

```
public primitive class X<T> {

    
    static void m(X<String> ls) { }

        m(new X<>()); // OK
        m(X<String>.default); // OK.
        m(X.default);   // OK, BUT get an error here!
    }

}

```

at the line with the .default usage I get:
```
 error: incompatible types: X<T> cannot be converted to X<String>
        m(X.default);
           ^
  where T is a type-variable:
    T extends Object declared in class X
Note: Some messages have been simplified; recompile with -Xdiags:verbose to get full output
1 error
``` 

This shows that there is no real inference happening there.

(when a poly type argument is attributed without a target type in context, it should be typed to be DeferredType, when considered against overload candidates the positional parameter type becomes the pt() and the deferred type can now be typed to a concrete type)

sadayapalam:
A way to see this in more detail is to debug through the Attr.visitApply call corresponding to the method invocation

```
m(new X<>(), X.default);
```

in this test case:

```
public primitive class X<T> {

    
    static void m(X<String> ls, X<String> ls2) { }

        m(new X<>(), X.default);
    }

}

```

You will see that first argument is typed to be DeferredType and second one incorrectly to be X<T>

jespersm:
Hey @sadayapalam and @mcimadamore, I think this patch is ready for review again.

I've not merged lworld (to keep a clearer history), but it still applies cleanly to the current lworld tip, so it should merge OK.

src/jdk.compiler/share/classes/com/sun/tools/javac/tree/TreeInfo.java
```diff
@@ -276,6 +276,27 @@ public static boolean isDiamond(JCTree tree) {
         }
     }
 
+    /** Return true if a default expression may represent a poly expression (with elided types) */
+    public static boolean isPossiblePolyDefault(JCDefaultValue tree, Type clazztype) {
+        JCTypeApply applyTree = TreeInfo.getTypeApplication(tree.clazz);
+        if (applyTree != null) {
+            return applyTree.arguments.isEmpty();
+        } else {
```
sadayapalam, line 284: The return statement is not reachable and so the block is not useful. This is because, Foo<>.default is a syntax error. Given a primitive class Foo<T> we will only support Foo.default involving implicit diamond like inference and will not support Foo<>.default which is what this code is handling.

src/jdk.compiler/share/classes/com/sun/tools/javac/tree/TreeInfo.java
```diff
@@ -276,6 +276,27 @@ public static boolean isDiamond(JCTree tree) {
...
+            // No type arguments before .default - Consider if the type is generic or not
+            return clazztype == null || clazztype.tsym.type.isParameterized();
+        }
```
sadayapalam, line 287: This method is problematic - This will always treat Foo.default as a poly expression in an invocation context even when the concerned primitive type is not a generic class.  For instance the Foo.default expression in the following snippet becomes a poly expression when it is a standalone expression in reality:

```
primitive class X {

    static void foo(Object o) {
    }

        foo(X.default);    
    }

}

```

src/jdk.compiler/share/classes/com/sun/tools/javac/tree/TreeInfo.java
```diff
@@ -276,6 +276,27 @@ public static boolean isDiamond(JCTree tree) {
...
+    }
+
+    /** Return true if a tree directly or indirectly represents a type application. */
+    public static JCTypeApply getTypeApplication(JCTree tree) {
+        switch(tree.getTag()) {
+            case TYPEAPPLY: return (JCTypeApply)tree;
+            case NEWCLASS: return getTypeApplication(((JCNewClass)tree).clazz);
+            case ANNOTATED_TYPE: return getTypeApplication(((JCAnnotatedType)tree).underlyingType);
```
sadayapalam, line 295: I think there is a copy + paste problem. We can never see a NEWCLASS here.  The LHS of .default is just a type node and cannot be an rvalue expression ==>  new X().default is illegal.

sadayapalam: I think perhaps what we really need to do here is to attribute the clazz node of JCDefaultValue first - perhaps by first making a copy of the node and then see if it is a generic class but lacks type arguments. Such as this: 

```
@Override
    public void visitDefaultValue(JCDefaultValue that) {
        JCExpression copy = new TreeCopier<Void>(attr.make).copy(that.clazz);
        attr.attribType(copy, env);
        if (copy.type.getTypeArguments().isEmpty() && copy.type.tsym.type.isParameterized()) {
            processArg(that, speculativeTree -> new ResolvedDefaultType(that, env, speculativeTree));
        } else {
            //not a poly expression, just call Attr
            setResult(that, attr.attribTree(that, env, attr.unknownExprInfo));
        }
    }
```

sadayapalam (changes requested):

src/jdk.compiler/share/classes/com/sun/tools/javac/tree/TreeInfo.java
sadayapalam: (I think I meant to say return applyTree.arguments.isEmpty(); will never be true.)

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/ArgumentAttr.java
```diff
@@ -718,6 +729,36 @@ Type methodType() {
         }
     }
 
+    /**
+     * Argument type for default values.
+     */
+    class ResolvedDefaultType extends ResolvedMemberType<JCDefaultValue> {
+
```
sadayapalam, line 736: I am surprised that ResolvedDefaultType extends ResolvedMemberType rather than ArgumentType directly. While the Inference behavior could be indirectly modelled it as if the class has a generic factory method named 'default':
```
<T> static Foo<T> default() { ... }
```
this implementation detail could be pushed deep down and not manifest here and in  elsewhere (setPolyKind())

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java
```diff
@@ -5020,25 +5021,62 @@ public void visitDefaultValue(JCDefaultValue tree) {
...
+    }
+    // where
+    CheckContext defaultPolyContext(JCDefaultValue tree, TypeSymbol tsym, CheckContext checkContext) {
+        return new Check.NestedCheckContext(checkContext) {
+            @Override
+            public void report(DiagnosticPosition _unused, JCDiagnostic details) {
+                enclosingContext.report(tree.clazz,
+                        diags.fragment(Fragments.CantApplyDiamond1(Fragments.Diamond(tsym), details)));
+            }
+        };
```
sadayapalam, line 5079: I don't see a test that triggers this diagnostic.

src/jdk.compiler/share/classes/com/sun/tools/javac/tree/TreeInfo.java
```diff
@@ -294,6 +315,9 @@ public static void setPolyKind(JCTree tree, PolyKind pkind) {
             case NEWCLASS:
                 ((JCNewClass)tree).polyKind = pkind;
                 break;
+            case DEFAULT_VALUE:
+                ((JCDefaultValue)tree).polyKind = pkind;
+                break;
             case REFERENCE:
```
sadayapalam, line 321: This is surprising. I would expected polyKind to be set in a similar fashion to switch and conditional expressions.

sadayapalam:
Hi Jesper, I have left some comments - but I will need to make a couple of more passes to fully understand the implementation as this is a fairly involved and complex task. I will actually have to tinker with the version you have posted to comprehend it - Could you give me some time - I will play around with it and propose a version that addresses the problems I have outlined and also includes further tweaks that could simplify the implementation more ? Then we can study it, finalize it and then ask Maurizio for a final review once both of us are satisfied.

jespersm:
@sadayapalam : I can take another go at this one to adress the review comments above, which may make a second review pass easier.

sadayapalam:
> @sadayapalam : I can take another go at this one to adress the review comments above, which may make a second review pass easier.

Sure, this is recognized to be a tricky area, so it is normal to have iterate over it multiple times incrementally improving it. As long as this is had in mind, by all means it helps to have the comments made so far addressed. Thanks!

PR #370, opened 2021-03-20
8263900: [lworld] Make .default a separate node type in the parser.

jespersm:
To prepare for making .default a poly expression and to implement further logic, .default-expressions should be split off as a separate AST node.

---------
### Progress

### Issue
 * [JDK-8263900](https://bugs.openjdk.java.net/browse/JDK-8263900): [lworld] Make .default a separate node type in the parser.

### Reviewers

### Download
To checkout this PR locally:
`$ git fetch https://git.openjdk.java.net/valhalla pull/370/head:pull/370`
`$ git checkout pull/370`

To update a local copy of the PR:
`$ git checkout pull/370`
`$ git pull https://git.openjdk.java.net/valhalla pull/370/head`

### DIFF
--- a/src/jdk.compiler/share/classes/com/sun/source/tree/DefaultExpressionTree.java
+++ b/src/jdk.compiler/share/classes/com/sun/source/tree/DefaultExpressionTree.java
@@ -1,0 +1,48 @@
+/*
+ * Copyright (c) 2021, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+package com.sun.source.tree;
+
+/**
+ * A tree node for a {@code default} instance initializion expression.
+ *
+ * For example:
+ * <pre>
+ *   Optional<String>.default
+ * </pre>
+ *
+ * @jls todo
+ *
+ * @since valhalla
+ */
+public interface DefaultExpressionTree extends ExpressionTree {
+
+    /**
+     * Returns the name of the class of the instance being initialized.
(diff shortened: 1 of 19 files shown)


sadayapalam:
Thanks Jesper, I will review this one and help you take it forward.

sadayapalam:
I have made one high level pass over the changes, more verifying on the boiler plate parts. Need to delve deeper in Attr, Gen, Parser, TransValues, will do this tomorrow.

Here are some comments so far:

(1) Is DefaultExpression better named as DefaultInstance or DefaultValue ??
(2) Should com.sun.tools.javac.tree.JCTree.JCDefaultExpression#getClazz - for consistency sake return JCExpression ??
(compare similar cases in JCTree)
(3) Do we need a visitor in SourceComputer in CRTable.java ??

sadayapalam:
Here are cumulative comments:

(1) Is DefaultExpression better named as DefaultInstance or even better DefaultValue ??
(2) Should com.sun.tools.javac.tree.JCTree.JCDefaultExpression#getClazz - for consistency sake return JCExpression ??
(compare similar cases in JCTree)
(3) Should JCDefaultExpression#getClazz be named getType() a la JCTypeCast, JCNewArray, JCVariableDecl, JCInstanceof ...
(4) Do we need a visitor in SourceComputer in CRTable.java ??
(5) Attr.java:4464: reference to .default needs to be removed from comment
(6) Attr.visitDefaultExpression: sitesym seems to be unused variable declaration.
(7) This program: 
```
primitive class X {
    Object o2 = java.util.default;
}
```

generates this strange error: 
error: package java does not exist
    Object o2 = java.util.default;
                               ^
1 error
(That it emits the same message without the patch notwithstanding)

I think we should attribute the clazz as a TYP_PCK and for package case emit an error.
(8) com.sun.source.util.TreeScanner#visitDefaultExpression: Eliminate the local and directly return ?
 
Let me know if you have any questions/clarifications.

Thanks for your effort, this looks very promising prework for the poly expr typing task

jespersm:
Regarding (7), I hit that earlier as well - but it is a entirely seperate bug, also available on the released product:

```
% javac -version
javac 16
% cat X.java 
public class X {
    void x() {
        Class c = java.lang.class;
    }
}
% javac X.java    
X.java:3: error: package java does not exist
        Class c = java.lang.class;
                      ^
1 error
```

I get the same confusing error with e.g. "new java.lang()"

This bug goes back to at least Java 11, but Java 8 u232 appears to get it right, explaining that there's no java.lang class. I'm guessing it may have come in with Jigsaw.

While I might be able to fix this locally, I think it should really be addressed in the main repo first. I tried finding the bug for it, but didn't find it -- would you like me to report it?

sadayapalam:
> Regarding (7), I hit that earlier as well - but it is a entirely seperate bug, also available in the released product:

[...]

> While I might be able to fix this locally, I think it should really be addressed in the main repo first. I tried finding the bug for it, but didn't find it -- would you like me to report it?

I agree this should be addressed in mainline first, yes please, could you raise a defect report for that ? Thanks!

sadayapalam:
> > Regarding (7), I hit that earlier as well - but it is a entirely seperate bug, also available in the released product:
> 
> [...]
> 
> > While I might be able to fix this locally, I think it should really be addressed in the main repo first. I tried finding the bug for it, but didn't find it -- would you like me to report it?
> 
> I agree this should be addressed in mainline first, yes please, could you raise a defect report for that ? Thanks!

Actually, I think with your work to split Foo.default away from JCFieldAccess, this should be fixed independently in both mainline (for java.lang.class example you report) and in Valhalla for java.lang.default case we are discussing. With the code split, fixing mainline won't automatically fix Valhalla issue with java.lang.default. 

However, I am fine with that being handled on a separate ticket by itself without blocking or side tracking this PR/RFR and also the work on treating .default as a poly expression.

sadayapalam (approved):
Looks good Jesper, Can you issue an /integrate command and I will follow it up with /sponsor

Thanks!

PR #371, opened 2021-03-25
8262831: [lworld] C2 compilation fails with assert "should be addp but is Phi"

kelthuzadx:
Hi,

Can I have a review of this bug fix?

The following test case is crashed at `(5)`:
```
   void test(MyValue[] array) {
        for (int i = 0; i < 10; ++i) {
            for (int j = 0; j < 10; ++j) {                                (1)
                iField = 6;					                   (2)
            }									  (3)
            for (int j = 0; j < 2; ++j) {
                iField += array[0].b;
            }
            MyValue[] array2 = {new MyValue()};           (4)
            c = array[0];						  (5) // hit the assertion
            array2[0] = t;						  (6)
        }
    }
```
I did some investigations. C2 wants to check whether there are other Mem nodes between `(4)` and `(6)` that read or write the array2, because it hopes to merge `(4)` and `(6)` into an InitializeNode. If it finds that there are any reads or writes, such as LoadI in `(5)`, then its Address input must be an AddPNode. But in fact, it may be a PhiNode, so an assertion is hit.

crash1

Why does PhiNode appear on (5) as the input of LoadINode? Because the loop unrolling (PhaseIdealLoop::do_unroll) occurred in `(1)-(3)`, it produced a cloned node of the parameter array(not as straightforward as I said, actually it's a CastPPNode which produced via extra steps), and then the parameter array and the cloned nodes were merged, thus a PhiNode node appeared.

crash2

Thanks Tobias for pointing out that this scenario is not reproducible in mainline JDK because we can't perform such aggressive scalarization for non-inline types. So we'd better fix it in Valhalla for now.

Thanks!
Yang

---------
### Progress

### Issue
 * [JDK-8262831](https://bugs.openjdk.java.net/browse/JDK-8262831): [lworld] C2 compilation fails with assert "should be addp but is Phi"

### Reviewers
 * [Tobias Hartmann](https://openjdk.java.net/census#thartmann) (@TobiHartmann - Committer) ⚠️ Review applies to 57704eafd9744621f53ea9be7f02f0e9b4cf6709

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/valhalla pull/371/head:pull/371` \
`$ git checkout pull/371`

`$ git checkout pull/371` \
`$ git pull https://git.openjdk.java.net/valhalla pull/371/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 371`

`$ git pr show -t 371`

</details>

<a href="https://git.openjdk.java.net/valhalla/pull/371.diff">https://git.openjdk.java.net/valhalla/pull/371.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/opto/memnode.cpp
+++ b/src/hotspot/share/opto/memnode.cpp
@@ -3922,6 +3922,12 @@ intptr_t InitializeNode::can_capture_store(StoreNode* st, PhaseGVN* phase, bool 
                 // the store control then we cannot capture the store.
                 assert(!n->is_Store(), "2 stores to same slice on same control?");
                 Node* base = other_adr;
+                if (base->is_Phi()) {
+                  // In rare case, base may be a PhiNode and it may reads
+                  // the same memory slice between InitializeNode and store.
+                  failed = true;
+                  break;
+                }
                 assert(base->is_AddP(), "should be addp but is %s", base->Name());
                 base = base->in(AddPNode::Base);
                 if (base != NULL) {
--- a/src/hotspot/share/opto/memnode.cpp
+++ b/src/hotspot/share/opto/memnode.cpp
@@ -3923,7 +3923,7 @@ intptr_t InitializeNode::can_capture_store(StoreNode* st, PhaseGVN* phase, bool 
                 assert(!n->is_Store(), "2 stores to same slice on same control?");
                 Node* base = other_adr;
                 if (base->is_Phi()) {
-                  // In rare case, base may be a PhiNode and it may reads
+                  // In rare case, base may be a PhiNode and it may read
                   // the same memory slice between InitializeNode and store.
                   failed = true;
                   break;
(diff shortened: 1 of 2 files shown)


src/hotspot/share/opto/memnode.cpp
```diff
@@ -3922,6 +3922,12 @@ intptr_t InitializeNode::can_capture_store(StoreNode* st, PhaseGVN* phase, bool
                 // the store control then we cannot capture the store.
                 assert(!n->is_Store(), "2 stores to same slice on same control?");
                 Node* base = other_adr;
+                if (base->is_Phi()) {
+                  // In rare case, base may be a PhiNode and it may reads
```
TobiHartmann, line 3926: `it may reads` -> `it may read`

TobiHartmann (approved):
Looks good to me. Thanks for fixing.

PR #375, opened 2021-03-31
8244227: [lworld] Explore an implementation where the reference projection and value projection types are backed by a single class symbol

sadayapalam:
This PR overhauls the existing implementation/representation of reference projection types  and the inline types so that both share the same underlying symbol, much like how an infinite family of parameterized types that originate from the same generic type are modelled using the same underlying symbol.

---------
### Progress

### Issue
 * [JDK-8244227](https://bugs.openjdk.java.net/browse/JDK-8244227): [lworld] Explore an implementation where the reference projection and value projection types are backed by a single class symbol

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/valhalla pull/375/head:pull/375` \
`$ git checkout pull/375`

`$ git checkout pull/375` \
`$ git pull https://git.openjdk.java.net/valhalla pull/375/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 375`

`$ git pr show -t 375`

</details>

<a href="https://git.openjdk.java.net/valhalla/pull/375.diff">https://git.openjdk.java.net/valhalla/pull/375.diff</a>

</details>


src/jdk.compiler/share/classes/com/sun/tools/javac/comp/LambdaToMethod.java
```diff
@@ -370,7 +370,7 @@ public void visitLambda(JCLambda tree) {
         //create the method declaration hoisting the lambda body
         JCMethodDecl lambdaDecl = make.MethodDef(make.Modifiers(sym.flags_field),
                 sym.name,
-                make.QualIdent(lambdaType.getReturnType().tsym),
+                make.QualIdent(lambdaType.getReturnType().tsym).setType(lambdaType.getReturnType()),
```
sadayapalam, line 373: Discussion item:This spot indicates one of the pain points in the new approach, There are lots of places where we construct fresh AST nodes from a symbol and in the process decorate the AST's type from what we see in the symbol. This is not a problem for parameterized types because the type arguments are erased by TransTypes anyway. But with Foo.ref and Foo.val sharing the same symbol, we will loose reference "projection'ness". I would like to discuss if there are mechanical ways to  detect and prevent such loss of type information. ATM, I am dependent on tests breaking to expose such problems and this is dependent heavily on coverage and coverage is still in infancy

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Type.java
```diff
@@ -1036,23 +1036,33 @@ public String toString() {
...
         public ClassType(Type outer, List<Type> typarams, TypeSymbol tsym) {
-            this(outer, typarams, tsym, TypeMetadata.EMPTY);
+            this(outer, typarams, tsym, TypeMetadata.EMPTY, false);
         }
 
         public ClassType(Type outer, List<Type> typarams, TypeSymbol tsym,
                          TypeMetadata metadata) {
+            this(outer, typarams, tsym, metadata, false);
+        }
+
```
sadayapalam, line 1056: I have rewired most of the constructor invocations to the new one which handles a reference projection flag parameter. A few call sites are still left in for the old constructors, these seem harmless.

sadayapalam:
Here are some notes that should help in the review process. I have broken down the changes involved into six logical units each forming a separate commit. Of these, the very first ("In diagnostics and debug strings while mentioning projection types, use .ref/.val rather than $ref/$val") and perhaps even the second commit ("Get out of the business of having to maintain the doppleganger symbol in sync with original") can be skipped or cursorily glanced through.

Real action starts at the third commit ("Genesis, construction, preservation and propagation of reference projection type") and proceeds through the next three commits.

sadayapalam:
Overall, I see two major challenges with the new approach (although it is simpler in many ways and has much to like about vis a vis what we have now):

1. Types.asSuper(Type, Symbol) would answer non-null for Foo.ref, Foo.val.tsym being passed in because they both have the same sym and deep down in Types.asSuper we return nonnull if the symbol is equal. So, I have to invent a version of Types.asSuper(Type, Type) instead and reroute the calls to that. However this was tricky in that there are places we do want asSuper to answer non-null when the Type is reference projection and the Symbol is primitive class. For example given Foo<T> { void add(T) {} } a call to Types.memberType(Foo.ref<String>, add(String)) internally results in a call to asSuper(Foo.ref<String>, Foo<T>) -
2.  There are places where we go from Symbol -> AST using the TreeMaker, - these Tree factory methods internally set the AST node type from the symbol. This results in "reference projectionness" being lost. The problem is to mechanically identify the places where such loss may occur (and recover, which is easy to do). ATM, I fixed the places where the tests failed thereby indicating type loss. But this approach is heavily dependent on coverage being good, and our tests are still not super good there being a work in progress (See that 2 is not a problem for parameterized types because after TransType, type arguments are gone AND more importantly the erasure of all parameterized types of the same generic class is the same, whereas we want Foo.ref and Foo.val to share a symbol but erase differently.)

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
```diff
@@ -2191,6 +2192,21 @@ public ArrayType makeArrayType(Type t) {
...
 
     // <editor-fold defaultstate="collapsed" desc="asSuper">
+    /**
+     * Return the (most specific) base type of `t' that starts with the
+     * given type `s'.  If none exists, return null.
+     *
+     * @param t a type
+     * @param s a type
+     */
+    public Type asSuper(Type t, Type s) {
```
mcimadamore, line 2202: A possible approach here (discussed offline) would be to leave asSuper as it is today - e.g. assume that asSuper speaks about "extends/implements" clause, and deal with .ref vs. .val mismatches at the callsite (if possible) - e.g. in Types::isSubtypes.

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
```diff
@@ -2555,15 +2568,26 @@ public Type visitWildcardType(WildcardType t, Boolean recurse) {
...
             public Type visitClassType(ClassType t, Boolean recurse) {
-                Type erased = t.tsym.erasure(Types.this);
-                if (recurse) {
-                    erased = new ErasedClassType(erased.getEnclosingType(),erased.tsym,
-                            t.getMetadata().without(Kind.ANNOTATIONS));
-                    return erased;
-                } else {
-                    return combineMetadata(erased, t);
+                // erasure(projection(primitive)) = projection(erasure(primitive))
+                Type erased = eraseClassType(t, recurse);
```
mcimadamore, line 2572: Random non-code comment: this seems a case of erasure applied to early. The first time that happens in lworld, but in the old branch which featured generic specializations, at some point we decided to postpone type erasure, as we needed sharp generic types during code gen. This smells like more of the same - e.g. having to fight with erasure so that the info that we care about is preserved in some form.

mcimadamore: There is also the issue that, no matter how delayed erasure is, at the moment we need to generate two different class artifacts, so, at some point the symbol duality is going to pop back up; that's understandable - hopefully we'll come up with a strategy which doesn't require javac to generate two artifacts - at which point javac can simply insert conversions when going from L to Q - and mostly be done with it (of course to generate such conversions it will need access to full type info - see above).

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/LambdaToMethod.java
mcimadamore: Again, my test here is this: suppose that here we weren't talking about Point.ref/Point.val - but, instead of List<int> vs. List<float>. Would anything be different? I don't think so. What you are observing is the result of a "wrong" assumption made by the code - or, rather, an assumption that "used to" be true, but is no longer. That is, this code (and code like this spread all over javac) assumes that we can safely ignore type info once we're past the frontend, as code generation is only expressed in terms of erased types (which are also the types attached to the symbol). This assumption is starting to crumble now, and will come crashing down when specialized generics are introduced. So I see this as technical debt (e.g. javac throwing aways types and dealing with symbols too early) as opposed to a specific limitation of the patch proposed in this PR. To get a cleaner fix, we should fix the underlying problem (eager erasure).

src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/ClassWriter.java
```diff
@@ -1511,13 +1531,62 @@ public JavaFileObject writeClass(ClassSymbol c)
...
-            ClassSymbol refProjection = c.referenceProjection();
-            refProjection.flags_field = (refProjection.flags_field & ~FINAL) | ABSTRACT;
-            writeClassInternal(refProjection);
+            writeClassInternal(getReferenceProjection(c));
         }
         return javaFileObject;
     }
 
+        // where
+        private static ClassSymbol getReferenceProjection(ClassSymbol c) {
```
mcimadamore, line 1540: Maybe one day we will be able to remove this...

src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/ClassReader.java
```diff
@@ -561,31 +540,46 @@ Type classSigToType() {
...
         while (true) {
             final byte c = signature[sigp++];
             switch (c) {
 
             case ';': {         // end
-                ClassSymbol t = enterClass(names.fromUtf(signatureBuffer,
-                                                         startSbp,
-                                                         sbp - startSbp));
-
+                name = names.fromUtf(signatureBuffer,
```
mcimadamore, line 552: This is another bit which might get simpler if we end up generating only a single .class.

mcimadamore (commented):
This seems like a good cleanup overall. IMHO the issues can be bucketized as follows:

* some more experiment is required to see if we can avoid changing asSuper too much
* there is a bigger issue (which we'll have to confront with, sooner or later, because of specialization) regarding type erasure kicking in too early, and subsequent pipeline steps only dealing with symbol (not type) info

Since the approach moves things in the way generic types are modeled, I'm not surprised to see some of the issues I had to face when writing a backend for the specializer compiler pop back here. I'd say, for now, let's try to deal with the first bullet (which is localized with these changes), and let's come back (in a separate PR) with some workaround for the second bullet (perhaps borrowing some old valhalla code?).

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/LambdaToMethod.java
sadayapalam: I agree with your observation that "this as technical debt (e.g. javac throwing aways types and dealing with symbols too early) as opposed to a specific limitation of the patch proposed in this PR. To get a cleaner fix, we should fix the underlying problem (eager erasure)." So what call out as the problem inherent in the new approach is but one instance of the larger problem that will occur in other contexts. I agree with your suggestion elsewhere that this could be handled on a PR of its own.

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
sadayapalam: Yes, I'll pursue what you outline in upcoming commits

sadayapalam:
Hello Maurzio, Please take another look. I have reverted all the earlier changes to asSuper per suggestion and transformed calls to asSuper into isSubtype calls where possible (This is not always possible - see the elaborate javadoc comment attached to Types.asSuper - in fact starting the review there would be a good thing)

There are still a few calls to asSuper that are not studied through the lens of the twin hazards - these are calls from Types itself and Infer.java  - outside of these two files I have studied all call sites of asSuper and suitably amended them and added tests to guard against any regressions and prove the need for and correctness of the transformation. I decided to defer the study of these remaining call sites to the pre-existing ticket JDK-8244712 - Javac should switch to reference projection before walking type hierarchy. (This ticket would include analysis of not only the remaining asSuper calls, but also isSubtype calls) 

Of the two commits (ignoring the merge related ones), one of them is simply a revert of earlier change for asSuper and can be ignored.

Let me know if you see any further issues that block integration. Thanks!

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
```diff
@@ -2201,27 +2202,36 @@ public ArrayType makeArrayType(Type t) {
...
-     * given symbol.  If none exists, return null.
+     * Further caveats in Valhalla: There are two "hazards" we need to watch out for when using
+     * this method.
      *
-     * Caveat Emptor: Since javac represents the class of all arrays with a singleton
-     * symbol Symtab.arrayClass, which by being a singleton cannot hold any discriminant,
-     * this method could yield surprising answers when invoked on arrays. For example when
-     * invoked with t being byte [] and sym being t.sym itself, asSuper would answer null.
+     * 1. Since Foo.ref and Foo.val share the same symbol, that of Foo.class, a call to
+     *    asSuper(Foo.ref.type, Foo.val.type.tsym) would return non-null. This is NOT correct
```
mcimadamore, line 2209: I don't think I agree with the spirit of this comment :-)

There is only one symbol, so Foo.val.type.tsym == Foo. So:

asSuper(Foo.ref.type, Foo.val.type.tsym) ==
asSuper(Foo.ref.type, Foo) ==
true

This is not different from when you have:

asSuper(ListOfString.type, ListOfInteger.type.tsym) != null

I think the real issue here, is that we can't rely on "asSuper" for code generation, because, really, this routine isn't sharp enough. In the case of Gen::visitTypeCast, the issue is that we have to decide whether a checkcast should be emitted or not; we typically do that by comparing symbols. That strategy here is simply insufficient - as the symbols are the same - just the types will be different. This will be, again, the case when we will face a cast from List to List<int> - these two types will have same symbol, but different representation in the classfile, so a checkcast will be needed there.

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
```diff
@@ -2201,27 +2202,36 @@ public ArrayType makeArrayType(Type t) {
...
+     *    but use !types.isSubtype(tree.expr.type, tree.clazz.type) which operates in terms of
+     *    types. When we operate in terms of symbols, there is a loss of type information leading
+     *    to a hazard. Whether a call to asSuper should be transformed into a isSubtype call is
+     *    tricky. isSubtype returns just a boolean while asSuper returns richer information which
+     *    may be required at the call site. Also where the concerned symbol corresponds to a
+     *    generic class, an asSuper call cannot be conveniently rewritten as an isSubtype call
+     *    (see that asSuper(ArrayList<String>.type, List<T>.tsym) != null while
+     *    isSubType(ArrayList<String>.type, List<T>.type) is false;) So care needs to be exercised.
+     *
+     * 2. Given a primitive class Foo, a call to asSuper(Foo.type, SuperclassOfFoo.tsym) and/or
```
mcimadamore, line 2223: Question here: is Foo.type refusing to play because the "default" semantics attached to Foo.type is the one of the "value projection" ? If so, then I agree with the fix (e.g. using reference projection or self). Also, given where the model is, I don't think this is an "hack" or a "workaround" - it's basically a requirement: inheritance info lives on the .ref side of the fence.

In fact, so much so that I'll raise: shouldn't this call to "refProjectionOrSelf" be part of asSuper? If asSuper is always use to check for inheritance  relationship, isn't it always the case that you want the reference projection to be used? (and then, if a callsite ends up with too lose a semantics, maybe that's a sign that using asSuper there was the wrong choice?)

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Symbol.java
```diff
@@ -2042,7 +2042,7 @@ public boolean binaryOverrides(Symbol _other, TypeSymbol origin, Types types) {
 
             // check for a direct implementation
             if (other.isOverridableIn((TypeSymbol)owner) &&
-                types.asSuper(owner.type, other.owner) != null &&
+                types.isSubtype(owner.type, other.owner.type) &&
```
mcimadamore, line 2045: This seems inconsistent with what has been done below (see other comment)

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Symbol.java
```diff
@@ -2111,7 +2111,7 @@ public boolean overrides(Symbol _other, TypeSymbol origin, Types types, boolean
 
             // check for a direct implementation
             if (other.isOverridableIn((TypeSymbol)owner) &&
-                types.asSuper(owner.type, other.owner) != null) {
+                types.asSuper(owner.type.referenceProjectionOrSelf(), other.owner) != null) {
```
mcimadamore, line 2114: So, for source ovrride we use `asSuper`, but for binary override we use `isSubtype`. Shouldn't we use asSuper in both cases here? Overriding is typically a business between class declarations, so symbols should be good? Or can you envision a case where you want to check if a method on Foo.val overrides something on a Foo.ref and you want to say "no" there? In other words, I think the check here is meant to say: make sure that the owner of the method symbols are related by subclassing - which seems to call for asSuper.

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java
```diff
@@ -1916,7 +1916,7 @@ public void report(DiagnosticPosition pos, JCDiagnostic details) {
 
     void checkAutoCloseable(DiagnosticPosition pos, Env<AttrContext> env, Type resource) {
         if (!resource.isErroneous() &&
-            types.asSuper(resource, syms.autoCloseableType.tsym, true) != null &&
+            types.isSubtype(resource.referenceProjectionOrSelf(), syms.autoCloseableType) &&
```
mcimadamore, line 1919: I think this wants asSuper? You are projecting anyway.

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Check.java
```diff
@@ -2695,8 +2695,8 @@ void checkCompatibleSupertypes(DiagnosticPosition pos, Type c) {
         }
         checkCompatibleConcretes(pos, c);
 
-        boolean implementsIdentityObject = types.asSuper(c, syms.identityObjectType.tsym, true) != null;
-        boolean implementsPrimitiveObject = types.asSuper(c, syms.primitiveObjectType.tsym, true) != null;
+        boolean implementsIdentityObject = types.isSubtype(c.referenceProjectionOrSelf(), syms.identityObjectType);
```
mcimadamore, line 2698: Not sure you need full blown subtyping here. You are just asking: is this thing derived from "identityObject" - in all these cases asSuper is what you want. Same for AutoCloseable checks. Iterable, ...

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/LambdaToMethod.java
```diff
@@ -1859,7 +1859,7 @@ boolean isSerializable() {
                 if (forceSerializable) {
                     return true;
                 }
-                return types.asSuper(tree.target, syms.serializableType.tsym) != null;
+                return types.isSubtype(tree.target, syms.serializableType);
```
mcimadamore, line 1862: That's another "is this a subclass of XYZ" check - asSuper is fine here (with ref projection).

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Lower.java
```diff
@@ -1722,7 +1722,7 @@ private JCBlock makeTwrBlock(List<JCTree> resources, JCBlock block, int depth) {
 
     private JCStatement makeResourceCloseInvocation(JCExpression resource) {
         // convert to AutoCloseable if needed
-        if (types.asSuper(resource.type, syms.autoCloseableType.tsym, true) == null) {
+        if (!types.isSubtype(resource.type.referenceProjectionOrSelf(), syms.autoCloseableType)) {
```
mcimadamore, line 1725: asSuper?

src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/Gen.java
```diff
@@ -2276,7 +2276,7 @@ public void visitTypeCast(JCTypeCast tree) {
         if (!tree.clazz.type.isPrimitive() &&
            !types.isSameType(tree.expr.type, tree.clazz.type) &&
             (!tree.clazz.type.isReferenceProjection() || !types.isSameType(tree.clazz.type.valueProjection(), tree.expr.type)) &&
-           types.asSuper(tree.expr.type, tree.clazz.type.tsym) == null) {
+           !types.isSubtype(tree.expr.type, tree.clazz.type)) {
```
mcimadamore, line 2279: This is, I think, the main place, where isSubtype is truly necessary.

mcimadamore (commented):
In general, it looks good. I have three main comments:
* the refProjectionOrSelf trick is a good one, I believe
* consider using reference projection always when doing asSuper (e.g. adding that projection inside asSuper, directly)
* I believe the patch is replacing too many use sites with subtyping. There are several checks where we want to assert that T is an instance of C, where C can be AutoCloseable, Iterable, Serializable. It is not clear to me as to why we'd want these replaced with subtyping - in the sense that here we seem to be asking about a propoerty of the primitive class _declaration_ rather than the particular projection we happen to have on hand.

Note that, as we're exploring new territory, it is very possible that I have overlooked some underlying issue here - but thanks for taking the time to guide me through all the problems!

sadayapalam:
> In general, it looks good. I have three main comments:
> 
> * the refProjectionOrSelf trick is a good one, I believe
> * consider using reference projection always when doing asSuper (e.g. adding that projection inside asSuper, directly)

Will experiment with this and report back. This could minimize code changes. 

> * I believe the patch is replacing too many use sites with subtyping. There are several checks where we want to assert that T is an instance of C, where C can be AutoCloseable, Iterable, Serializable. It is not clear to me as to why we'd want these replaced with subtyping - in the sense that here we seem to be asking about a propoerty of the primitive class _declaration_ rather than the particular projection we happen to have on hand.

Possible. The rule of thumb I used for changing calls to asSuper into isSubtype was simply, whether (a) the call site is interested in the return value of asSuper as opoosed to just checking whether it is null or not and (b) whether generic class symbol could be passed in as the second argument to original asSuper call and if so retain it as asSuper itself. 

So it is likely more sites are changed than strictly necessary. The way I see it some of the overeaget changes to isSubtype result in equivalent semantics with no regression. Still it is worth minimizing ripples and for that reason I will attempt to restore these.

> 
> Note that, as we're exploring new territory, it is very possible that I have overlooked some underlying issue here - but thanks for taking the time to guide me through all the problems!

I am grappling with similar challenges. I.e clearly characterizing which calls involve hazards of which kind. This involves reasoning about every site.

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
sadayapalam: > I don't think I agree with the spirit of this comment :-)
> 

Indeed, it is stronger than it ought to be. It should say it may not be always what is desired at call site.

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
sadayapalam: Plausible, will experiment.

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Symbol.java
sadayapalam: Good point.

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Symbol.java
sadayapalam: Agreed.

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Check.java
sadayapalam: OK

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/LambdaToMethod.java
sadayapalam: OK

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Lower.java
sadayapalam: OK

sadayapalam:
I pushed a commit accommodating two review comments:

1. asSuper calls at various places employed for the purpose of instanceof determination can be left alone.
2. Javadoc comment attached to asSuper is stronger than it ought to be and asserts something that is not quite true.

The third suggestion around asSuper internally coercing its argument into reference projection is something that will be taken up in a subsequent PR for JDK-8244712

@Maurizio, I'll proceed with integration if you give a quick glance through over the latest commit and approve. TIA

mcimadamore (approved):
Looks good - thanks for following up!

PR #376, opened 2021-04-01
8264414: [lworld] [AArch64] TestBufferTearing.java fails with C1

nick-arm:
We see failures like this on AArch64 when MyValue.incrementAndCheck() is
compiled with C1:

  java.lang.RuntimeException: Inconsistent field values: expected 0 to equal 675128
        at jdk.test.lib.Asserts.fail(Asserts.java:594)
        at jdk.test.lib.Asserts.assertEquals(Asserts.java:205)
        at jdk.test.lib.Asserts.assertEQ(Asserts.java:178)
        at compiler.valhalla.inlinetypes.MyValue.incrementAndCheck(TestBufferTearing.java:81)
        at compiler.valhalla.inlinetypes.TestBufferTearing$Runner.run(TestBufferTearing.java:124)

The barrier that is usually inserted on return from a method that wrote
final fields should be sufficient to prevent another thread seeing the
zero-initialised intermediate state.  However this barrier isn't
inserted at the moment because method()->is_object_constructor() is
false for primitive class constructors.

C2 has a similar guard around the memory barrier in Parse::do_exits().
I'm not sure if that needs amending as well but I've not seen any
failures due to it.

---------
### Progress

### Issue
 * [JDK-8264414](https://bugs.openjdk.java.net/browse/JDK-8264414): [lworld] [AArch64] TestBufferTearing.java fails with C1

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/valhalla pull/376/head:pull/376` \
`$ git checkout pull/376`

`$ git checkout pull/376` \
`$ git pull https://git.openjdk.java.net/valhalla pull/376/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 376`

`$ git pr show -t 376`

</details>

<a href="https://git.openjdk.java.net/valhalla/pull/376.diff">https://git.openjdk.java.net/valhalla/pull/376.diff</a>

</details>

### DIFF
--- a/src/hotspot/cpu/aarch64/interp_masm_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/interp_masm_aarch64.cpp
@@ -323,6 +323,10 @@ void InterpreterMacroAssembler::read_inlined_field(Register holder_klass,
           obj, field_index, holder_klass);
 
   bind(done);
+
+  // Ensure the stores to copy the inline field contents are visible
+  // before any subsequent store that publishes this reference.
+  membar(Assembler::StoreStore);
 }
 
 // Load object from cpool->resolved_references(index)
--- a/src/hotspot/share/c1/c1_GraphBuilder.cpp
+++ b/src/hotspot/share/c1/c1_GraphBuilder.cpp
@@ -1983,6 +1983,7 @@ void GraphBuilder::access_field(Bytecodes::Code code) {
             ciInlineKlass* inline_klass = field->type()->as_inline_klass();
             scope()->set_wrote_final();
             scope()->set_wrote_fields();
+            bool need_membar = false;
             if (inline_klass->is_empty()) {
               apush(append(new Constant(new InstanceConstant(inline_klass->default_instance()))));
               if (has_pending_field_access()) {
@@ -1999,6 +2000,7 @@ void GraphBuilder::access_field(Bytecodes::Code code) {
               apush(append_split(vt));
               append(pending_load_indexed()->load_instr());
               set_pending_load_indexed(NULL);
+              need_membar = true;
             } else {
               NewInlineTypeInstance* new_instance = new NewInlineTypeInstance(inline_klass, state_before);
               _memory->new_instance(new_instance);
@@ -2012,8 +2014,12 @@ void GraphBuilder::access_field(Bytecodes::Code code) {
               } else {
                 copy_inline_content(inline_klass, obj, field->offset(), new_instance, inline_klass->first_field_offset(), state_before);
               }
-              // Ensure the stores to copy the field contents are visible
-              // before any subsequent store that publishes this reference.
+              need_membar = true;
+            }
+            if (need_membar) {
+              // If we allocated a new instance ensure the stores to copy the
+              // field contents are visible before any subsequent store that
+              // publishes this reference.
               append(new MemBar(lir_membar_storestore));
             }
           }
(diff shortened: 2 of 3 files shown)


TobiHartmann:
Hi Nick,

I think we also need a barrier in `GraphBuilder::access_field` when loading from a flattened field and in `GraphBuilder::load_indexed` (the initialization code is emitted in `LIRGenerator::access_flattened_array`) when loading from a flattened array.

It would be good to have tests that triggers this.

Best regards,
Tobias

nick-arm:
So I've managed to generate the following test which triggers this:

```java
public class BarrierTest {

    public static Point[] points = new Point[1];
    static volatile boolean running = true;

    public static void writePoint(int i) {
        Rect r = new Rect(new Point(i, i), new Point(i + 1, i + 1));
        points[0] = r.a;   // Load from flattened field here
    }

    private static void checkMissingBarrier() {
        while (running) {
            // Should not see zero-initialised object here
            if (points[0].x == 0 || points[0].y == 0)
                throw new IllegalStateException();
        }
    }
    
    public static void main(String[] args) throws InterruptedException {
        points[0] = new Point(1, 1);
        
        Thread[] threads = new Thread[10];
        for (int i = 0; i < 10; i++) {
            threads[i] = new Thread(BarrierTest::checkMissingBarrier);
            threads[i].start();
        }
                
        for (int i = 2; i < 1_000_000; i++)
            writePoint(i);
        
        running = false;
        
        for (int i = 0; i < 10; i++)
            threads[i].join();
    }
}
```

Where `Rect` and `Point` are primitive classes with the obvious definition. Then `checkMissingBarrier()` will throw an exception if run with `-XX:InlineFieldMaxFlatSize=-1  -XX:FlatArrayElementMaxSize=0 -XX:TieredStopAtLevel=1`. The field access `r.a` creates an on-heap copy of the flattened field and then stores that reference into the `points` array (which can't be flattened because `FlatArrayElementMaxSize=0`). Without a memory barrier after copying the flattened field contents in `GraphBuilder::access_field` another thread can read through that reference and see the zero-initialised memory. 

I'm not sure if there's an easier way of triggering this without having mis-matched InlineFieldMaxFlatSize and FlatArrayElementMaxSize? To see this I think we need to read from a flattened field and then store into a non-flattened field of the same type that's accessible by another thread.

nick-arm:
@TobiHartmann I've added the two missing membars and a test.

TobiHartmann (commented):
Very nice that you were able to come up with a regression test!

Regarding the combination of `InlineFieldMaxFlatSize` and `FlatArrayElementMaxSize`, couldn't you achieve the same result by using an Object or Interface array/field to prevent flattening?

The fix looks good to me but I think we also need membars in `GraphBuilder::access_field` (there are two places where we create a buffer via `new NewInlineTypeInstance`). Would be great if you could add a regression test for that as well.

nick-arm:
> 
> Regarding the combination of `InlineFieldMaxFlatSize` and `FlatArrayElementMaxSize`, couldn't you achieve the same result by using an Object or Interface array/field to prevent flattening?

Yes that's much simpler, thanks.

> 
> The fix looks good to me but I think we also need membars in `GraphBuilder::access_field` (there are two places where we create a buffer via `new NewInlineTypeInstance`). Would be great if you could add a regression test for that as well.

Yes there's also the case for the "delayed" load-indexed as in `array[n].field`. I saw that before but for some reason thought it didn't need the membar. I've added a test for that too so all three cases fail without the corresponding membar.

While I was testing that I found it would fail occasionally before while `writeRefs()` was running in the interpreter before C1 compiled it. So I think there's also a missing membar at the end of `InterpreterMacroAssembler::read_inlined_field()` which I've added (not necessary for x86).

TobiHartmann (approved):
Very nice. The compiler changes look good to me!

@fparain could you take a look at the interpreter changes?

fparain (approved):
Looks good to me.
Thank you for finding and fixing it!

Fred

PR #394, opened 2021-04-27
8265423: [lworld] Provide experimental support for generating a single class file per primitive class

sadayapalam:
Use -XDunifiedValRefClass option to javac to get unified class file generation for primitive classes. Note: Javac will require some tweaks to consume such files, this is WIP.

---------
### Progress

### Issue
 * [JDK-8265423](https://bugs.openjdk.java.net/browse/JDK-8265423): [lworld] Provide experimental support for generating a single class file per primitive class

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/valhalla pull/394/head:pull/394` \
`$ git checkout pull/394`

`$ git checkout pull/394` \
`$ git pull https://git.openjdk.java.net/valhalla pull/394/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 394`

`$ git pr show -t 394`

</details>

<a href="https://git.openjdk.java.net/valhalla/pull/394.diff">https://git.openjdk.java.net/valhalla/pull/394.diff</a>

</details>

### DIFF
--- a/src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/ClassReader.java
+++ b/src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/ClassReader.java
@@ -2687,7 +2687,7 @@ public class ClassReader {
             ClassSymbol referenceProjection = syms.getClass(currentModule, flatname);
             if (referenceProjection != null) {
                 if (referenceProjection.name != names.ref && referenceProjection.owner.kind == PCK) {
-                    readClassFileInternal(referenceProjection);
+                    referenceProjection.complete();
                     ClassType classType = (ClassType) c.type;
                     classType.supertype_field = ((ClassType) referenceProjection.type).supertype_field;
                     classType.interfaces_field = ((ClassType) referenceProjection.type).interfaces_field;
--- a/src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/ClassWriter.java
+++ b/src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/ClassWriter.java
@@ -845,7 +845,7 @@ public class ClassWriter extends ClassFile {
                 !inner.name.isEmpty() ? poolWriter.putName(inner.name) : 0);
             databuf.appendChar(flags);
             icCount++;
-            if (inner.isPrimitiveClass()) {
+            if (inner.isPrimitiveClass() && types.splitPrimitiveClass) {
                 databuf.appendChar(poolWriter.putClass(inner.type.referenceProjection()));
                 databuf.appendChar(
                         inner.owner.kind == TYP && !inner.name.isEmpty() ? poolWriter.putClass((ClassSymbol)inner.owner) : 0);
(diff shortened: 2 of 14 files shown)


src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/ClassWriter.java
```diff
@@ -845,7 +845,7 @@ void writeInnerClasses() {
                 !inner.name.isEmpty() ? poolWriter.putName(inner.name) : 0);
             databuf.appendChar(flags);
             icCount++;
-            if (inner.isPrimitiveClass()) {
+            if (inner.isPrimitiveClass() && types.splitPrimitiveClass) {
```
mcimadamore, line 848: the test `sym.isPrimitiveClass() && types.splitPrimitiveClass` is so common which I think would be better refactored into a separate predicate, to make the code clearer.

e.g. `Symbol.isSplitPrimitiveClass(Types)`

src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/ClassReader.java
```diff
@@ -2687,7 +2687,7 @@ public void readClassFile(ClassSymbol c) {
             ClassSymbol referenceProjection = syms.getClass(currentModule, flatname);
             if (referenceProjection != null) {
                 if (referenceProjection.name != names.ref && referenceProjection.owner.kind == PCK) {
-                    readClassFileInternal(referenceProjection);
+                    referenceProjection.complete();
```
mcimadamore, line 2690: This change seems sneaky - what caused it?

sadayapalam: Indeed. I should push this as a separate change set. I discovered this bug while working on the now aborted attempt to evolve the code generation scheme to match SoV4. The right way to internalize the class is by completing complete() on the symbol and not by directly calling readClassFileInternal(). Doing so, could result in com.sun.tools.javac.code.ClassFinder#fillIn(com.sun.tools.javac.code.Symbol.ClassSymbol) from being re-entered resulting in an assertion failure.

This change is not required for the present feature, but is a crucial fix that should be made independently.

src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/ClassWriter.java
sadayapalam: Thanks, I will take a look

src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/ClassReader.java
sadayapalam: (As I was studying the code that merges the projections back to see what changes I should make for the unified class model, I remembered that bug fix from that aborted work)

mcimadamore: I'm fine with keeping the change here if that's easier - I just wanted to make sure I understood where this was coming from as it looked an odd change given what of the rest of the code is doing. Thanks for the explanation.

PR #414, opened 2021-05-18
L/Q again, step two

fparain:
Please review this second set of fixes for the L/Q transition.
A few changes in C1, most changes in CI.
CI always delegates to the runtime to know if a null-free array is flattened or not (to avoid duplication of the flattening decision code).

With those changes, all tests in runtime/valhalla/inlinetypes now pass with -Xcomp -XX:TieredStopAtLevel=1 (at least on Linux x64).
Some tests in compiler/valhalla/inlinetypes still fail but they will require more time to be investigated (some tests force use of C2 or expect code to be C2 compiled, but C2 is not fixed yet).

Thank you,

Fred

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/valhalla pull/414/head:pull/414` \
`$ git checkout pull/414`

`$ git checkout pull/414` \
`$ git pull https://git.openjdk.java.net/valhalla pull/414/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 414`

`$ git pr show -t 414`

</details>

<a href="https://git.openjdk.java.net/valhalla/pull/414.diff">https://git.openjdk.java.net/valhalla/pull/414.diff</a>

</details>

### DIFF
--- a/src/hotspot/cpu/x86/c1_CodeStubs_x86.cpp
+++ b/src/hotspot/cpu/x86/c1_CodeStubs_x86.cpp
@@ -310,12 +310,12 @@ void NewTypeArrayStub::emit_code(LIR_Assembler* ce) {
 // Implementation of NewObjectArrayStub
 
 NewObjectArrayStub::NewObjectArrayStub(LIR_Opr klass_reg, LIR_Opr length, LIR_Opr result,
-                                       CodeEmitInfo* info, bool is_inline_type) {
+                                       CodeEmitInfo* info, bool is_null_free) {
   _klass_reg = klass_reg;
   _result = result;
   _length = length;
   _info = new CodeEmitInfo(info);
-  _is_inline_type = is_inline_type;
+  _is_null_free = is_null_free;
 }
 
 
@@ -324,7 +324,7 @@ void NewObjectArrayStub::emit_code(LIR_Assembler* ce) {
   __ bind(_entry);
   assert(_length->as_register() == rbx, "length must in rbx,");
   assert(_klass_reg->as_register() == rdx, "klass_reg must in rdx");
-  if (_is_inline_type) {
+  if (_is_null_free) {
     __ call(RuntimeAddress(Runtime1::entry_for(Runtime1::new_flat_array_id)));
   } else {
     __ call(RuntimeAddress(Runtime1::entry_for(Runtime1::new_object_array_id)));
--- a/src/hotspot/share/ci/ciArrayKlass.cpp
+++ b/src/hotspot/share/ci/ciArrayKlass.cpp
@@ -30,6 +30,7 @@
 #include "ci/ciTypeArrayKlass.hpp"
 #include "ci/ciUtilities.hpp"
 #include "ci/ciUtilities.inline.hpp"
+#include "oops/inlineKlass.inline.hpp"
 
 // ciArrayKlass
 //
@@ -104,10 +105,36 @@ bool ciArrayKlass::is_leaf_type() {
 ciArrayKlass* ciArrayKlass::make(ciType* element_type) {
   if (element_type->is_primitive_type()) {
     return ciTypeArrayKlass::make(element_type->basic_type());
-  } else if (element_type->flatten_array()) {
-    return ciFlatArrayKlass::make(element_type->as_klass());
   } else {
-    return ciObjArrayKlass::make(element_type->as_klass());
+    return make(element_type->as_klass(), element_type->is_null_free());
+  }
+}
+
+ciArrayKlass* ciArrayKlass::make(ciKlass* klass, bool null_free) {
+  if (null_free) {
+    if (klass->is_loaded()) {
+      bool is_array_flattened = false;
+      GUARDED_VM_ENTRY(
+        EXCEPTION_CONTEXT;
+        Klass* ak = InlineKlass::cast(klass->get_Klass())->null_free_inline_array_klass(THREAD);
+        if (HAS_PENDING_EXCEPTION) {
+          CLEAR_PENDING_EXCEPTION;
+        } else {
+          if (ak != NULL && ak->is_flatArray_klass()) {
+            is_array_flattened = true;
+          }
+        }
+      )
+      if (is_array_flattened) {
+        return ciFlatArrayKlass::make(klass);
+      } else {
+        return ciObjArrayKlass::make(klass, true);
+      }
+    } else {
+      return ciEnv::unloaded_ciobjarrayklass();
+    }
+  } else {
+    return ciObjArrayKlass::make(klass);
   }
 }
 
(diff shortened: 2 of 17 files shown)


test/hotspot/jtreg/runtime/valhalla/inlinetypes/MultiANewArrayTest/MultiANewArrayTest.java
```diff
@@ -49,8 +49,7 @@ public static void main(String[] args) {
         } catch(Error e) {
             ex = e;
         }
-        Asserts.assertNotNull(ex, "An ICCE should have been thrown");
-        Asserts.assertEquals(ex.getClass(), IncompatibleClassChangeError.class, "Error is not an ICCE");
+        Asserts.assertNull(ex, "No ICCE should have been thrown");
```
forax, line 52: The negation of "An ICCE should have been thrown" is "No error should have been thrown",
talking about ICCE it that context can be misleading.

fparain: Error message updated in the last commit.

Thanks,

Fred

src/hotspot/cpu/x86/c1_CodeStubs_x86.cpp
```diff
@@ -310,12 +310,12 @@ void NewTypeArrayStub::emit_code(LIR_Assembler* ce) {
 // Implementation of NewObjectArrayStub
 
 NewObjectArrayStub::NewObjectArrayStub(LIR_Opr klass_reg, LIR_Opr length, LIR_Opr result,
-                                       CodeEmitInfo* info, bool is_inline_type) {
+                                       CodeEmitInfo* info, bool is_null_free) {
```
nick-arm, line 313: Would you mind making the same change in `c1_CodeStubs_aarch64.cpp` to keep them in sync?

fparain: Sure, sorry for the breakage of the aarch64 platform.

Fred

src/hotspot/share/ci/ciEnv.cpp
```diff
@@ -475,7 +475,8 @@ ciKlass* ciEnv::get_klass_by_name_impl(ciKlass* accessing_klass,
                              require_local);
     if (elem_klass != NULL && elem_klass->is_loaded()) {
       // Now make an array for it
-      return ciArrayKlass::make(elem_klass);
+      bool null_free_array = sym->is_Q_array_signature() && sym->char_at(1) == JVM_SIGNATURE_INLINE_TYPE;
```
TobiHartmann, line 478: Why do we need the `sym->char_at(1) == JVM_SIGNATURE_INLINE_TYPE` check? Isn't that part of `is_Q_array_signature()`?

src/hotspot/share/ci/ciArrayKlass.cpp
```diff
@@ -104,10 +105,36 @@ bool ciArrayKlass::is_leaf_type() {
...
+          }
+        }
+      )
+      if (is_array_flattened) {
+        return ciFlatArrayKlass::make(klass);
+      } else {
+        return ciObjArrayKlass::make(klass, true);
+      }
+    } else {
+      return ciEnv::unloaded_ciobjarrayklass();
```
TobiHartmann, line 134: Can't we delegate handling the unloaded case to `ciObjArrayKlass::make(klass)` below? I.e. check for `if (null_free && klass->is_loaded())` above.

src/hotspot/share/ci/ciArrayKlass.cpp
```diff
@@ -104,10 +105,36 @@ bool ciArrayKlass::is_leaf_type() {
...
+    if (klass->is_loaded()) {
+      bool is_array_flattened = false;
+      GUARDED_VM_ENTRY(
+        EXCEPTION_CONTEXT;
+        Klass* ak = InlineKlass::cast(klass->get_Klass())->null_free_inline_array_klass(THREAD);
+        if (HAS_PENDING_EXCEPTION) {
+          CLEAR_PENDING_EXCEPTION;
+        } else {
+          if (ak != NULL && ak->is_flatArray_klass()) {
+            is_array_flattened = true;
```
TobiHartmann, line 124: Why is that boolean needed? Can't we simply return `ciFlatArrayKlass::make(klass)` here?

TobiHartmann (changes requested):
Looks good to me. Added some minor comments.

src/hotspot/share/ci/ciEnv.cpp
fparain: The problem with is_Q_array_signature() is that it returns true for arrays of any dimension with the last dimension element klass being a null-free primitive class. It returns true for [QFoo; but also for [[[QFoo;. But for the [[[QFoo; array, the first two dimensions are regular nullable object arrays, only the last array is a null-free array (possibly flattened).

src/hotspot/share/ci/ciArrayKlass.cpp
fparain: I can change that. This was the way it was implemented initially, then I changed it to have the whole logic in a single method, but I don't have a strong opinion about this one.
Updated in the latest commit.

src/hotspot/share/ci/ciArrayKlass.cpp
fparain: Boolean removed in the latest commit.

src/hotspot/share/ci/ciArrayKlass.cpp
```diff
@@ -104,10 +105,27 @@ bool ciArrayKlass::is_leaf_type() {
...
+
+ciArrayKlass* ciArrayKlass::make(ciKlass* klass, bool null_free) {
+  if (null_free && klass->is_loaded()) {
+    GUARDED_VM_ENTRY(
+      EXCEPTION_CONTEXT;
+      Klass* ak = InlineKlass::cast(klass->get_Klass())->null_free_inline_array_klass(THREAD);
+      if (HAS_PENDING_EXCEPTION) {
+        CLEAR_PENDING_EXCEPTION;
+      } else {
+        if (ak != NULL && ak->is_flatArray_klass()) {
```
TobiHartmann, line 121: `else` and `if` can be merged

src/hotspot/share/ci/ciEnv.cpp
TobiHartmann: Ah, right. Thanks for the explanation!

TobiHartmann (approved):
Thanks for updating, looks good to me (added one minor suggestion).

src/hotspot/share/ci/ciArrayKlass.cpp
fparain: Code updated.

TobiHartmann (approved):

PR #420, opened 2021-05-21
8266890: [lworld] [AArch64] add support for InlineTypePassFieldsAsArgs

nick-arm:
This patch implements InlineTypePassFieldsAsArgs on AArch64 and the
associated stack extension/repair mechanism. It mostly follows the x86
implementation closely except for how the stack increment is stored in
the callee frame. On x86 the sp_inc stack slot stores the total size of
the frame which is the sum of the extension space, return address copy,
and the original method frame size (i.e. the total bytes needed to pop
the frame). On AArch64 we just store the size of the extension space. I
did it this way because it simplifies the stack repair code in
MacroAssembler::remove_frame (I've added some notes there to document
this). I don't think this should cause any problem because only the
MacroAssembler and platform-dependant frame walking code need to be
aware of this.

This patch includes JDK-8266609 which is a small refactoring in mainline
JDK to simplify how the frame size is passed around in the AArch64 C1
backend.

There was some X86 specific code in unpack_inline_args() in
macroAssembler_common.cpp. I've split this out into an arch-dependant
MacroAssembler::extend_stack_for_inline_args().

MacroAssembler::pack_inline_helper() and shuffle_inline_args() now take
a new Register val_array argument which is the register holding the
buffered oop array for packing. Previously it assumed this was already
loaded in RAX (x86) or x20 (AArch64) but IMO passing it in explicitly
makes the code easier to understand.

There is one new test failure: TestNullableInlineTypes.java. The test
seems functionally correct but there is a failure verifying the C2 IR
graph (see below). I haven't investigated this but the test enables
InlineTypePassFieldsAsArgs which we don't support yet on AArch64 so that
might be causing the failure.

```
Exception in thread "main" java.lang.RuntimeException: Graph for 'TestNullableInlineTypes::test26' contains forbidden node:
83  StoreL  ===  176  177  84  14  [[ 89 ]]  @compiler/valhalla/inlinetypes/MyValue3:exact+16 *, name=l, idx=6;  Memory: @rawptr:BotPTR, idx=Raw; !orig=82 !jvms: TestNullableInlineTypes::test26 @ bci:13 (line 704): expected false, was true
```

Tested tier1, runtime/valhalla, and compiler/valhalla on x86 and AArch64. There are some "bad AD file" failures on x86 but don't seem to be related to this patch.

---------
### Progress

### Issue
 * [JDK-8266890](https://bugs.openjdk.java.net/browse/JDK-8266890): [lworld] [AArch64] add support for InlineTypePassFieldsAsArgs

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/valhalla pull/420/head:pull/420` \
`$ git checkout pull/420`

`$ git checkout pull/420` \
`$ git pull https://git.openjdk.java.net/valhalla pull/420/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 420`

`$ git pr show -t 420`

</details>

<a href="https://git.openjdk.java.net/valhalla/pull/420.diff">https://git.openjdk.java.net/valhalla/pull/420.diff</a>

</details>

### DIFF
--- a/src/hotspot/cpu/aarch64/aarch64.ad
+++ b/src/hotspot/cpu/aarch64/aarch64.ad
@@ -1964,7 +1964,7 @@ void MachEpilogNode::emit(CodeBuffer &cbuf, PhaseRegAlloc *ra_) const {
   C2_MacroAssembler _masm(&cbuf);
   int framesize = C->output()->frame_slots() << LogBytesPerInt;
 
-  __ remove_frame(framesize);
+  __ remove_frame(framesize, C->needs_stack_repair(), C->output()->sp_inc_offset());
 
   if (StackReservedPages > 0 && C->has_reserved_stack_access()) {
     __ reserved_stack_check();
--- a/src/hotspot/cpu/aarch64/c1_LIRAssembler_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/c1_LIRAssembler_aarch64.cpp
@@ -376,10 +376,7 @@ void LIR_Assembler::jobject2reg_with_patching(Register reg, CodeEmitInfo *info) 
 int LIR_Assembler::initial_frame_size_in_bytes() const {
   // if rounding, must let FrameMap know!
 
-  // The frame_map records size in slots (32bit word)
-
-  // subtract two words to account for return address and link
-  return (frame_map()->framesize() - (2*VMRegImpl::slots_per_word))  * VMRegImpl::stack_slot_size;
+  return in_bytes(frame_map()->framesize_in_bytes());
 }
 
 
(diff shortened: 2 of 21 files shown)


TobiHartmann (approved):
Sorry for the delay, this looks good to me!

>  I did it this way because it simplifies the stack repair code in MacroAssembler::remove_frame

But that only simplifies the C++ code and not the assembly code that is emitted right? I wonder if we shouldn't aim for optimal assembly because this code will be exercised a lot.

Could you please file bugs for the other issues that you were seeing?

We currently don't have testing on aarch64 enabled for Valhalla in our CI but we will do so as soon as the port is stable.

nick-arm:
> 
> But that only simplifies the C++ code and not the assembly code that is emitted right? I wonder if we shouldn't aim for optimal assembly because this code will be exercised a lot.

Yes. I thought about it a bit more and I've modified it slightly so the sp_inc stack slot stores `total_frame_size - 2*wordSize` which is more like x86 and also saves one instruction when removing the frame (see the last commit 
d9fa572).

> 
> Could you please file bugs for the other issues that you were seeing?
> 

Sure. Do you have any ideas about this one? https://bugs.openjdk.java.net/browse/JDK-8264340

TobiHartmann (approved):
Looks good to me.

> Sure. Do you have any ideas about this one? https://bugs.openjdk.java.net/browse/JDK-8264340

That doesn't ring a bell. Is it reproducible?

nick-arm:
> > Sure. Do you have any ideas about this one? https://bugs.openjdk.java.net/browse/JDK-8264340
> 
> That doesn't ring a bell. Is it reproducible?

Yes, it always fails in that way on AArch64. It seems to be related to scheduling: I just ran it on x86 with `-XX:+UseOptoScheduling` and it also fails.

TobiHartmann:
> Yes, it always fails in that way on AArch64. It seems to be related to scheduling: I just ran it on x86 with -XX:+UseOptoScheduling and it also fails.

Okay, thanks for the details!

PR #421, opened 2021-05-21
8266466: [lworld] Enhance javac to consume unified primitive class files generated under the option -XDunifiedValRefClass

sadayapalam:
Post the integration of https://bugs.openjdk.java.net/browse/JDK-8244227,
the constructor of ClassType, takes a boolean isReferenceProjection as an additional
parameter. Till now, it was feasible to upfront decide at the time of constructing
a class type whether the proposed type is supposed to be a reference projection or not.

This is no longer possible under the unified class file generation scheme we are
moving to. When we see LFoo; in field/method signatures/descriptors, we don't know
whether the type is Foo.ref of a primitive class Foo or a plain old reference type.

Likewise when we see QFoo; we have an ambiguity whether this is the primitive class
type Foo or is Foo.val type of a ref-default primitive class Foo. Till a class is
fully built ("completed"), we don't know whether it is a primitive class and if so,
what its ref-val defaultness is.

Completing class Foo every time we see a type descriptor LFoo; or QFoo; is wasteful
of resources. Nor is it possible as it is, since it would call for the ClassReader's
code to be reentered even as some other class is being read in: So the proposed patch
implements an incremental piece meal augmentation of attributes.

I suggest the review be started with Type.java in ClassType#Flavor enumeration to gather
a high level picture of the abstractions put in place to characterize a class type in
an incremental basis. After Type.java, Attr.java, ClassReader.java and Symbol.java in
that order provide a good order for review.

Please note, while this patch puts together the infrastructure and layes the groundwork
for modelling ref-default classes and thereby enable VBC migration work, such work will
arrive separately at a later time.

As of now javac knows only about primitive class and their reference projections and
plain old reference types. There is no notion of ref-val defaultness yet, Nor a way
to declare them.

---------
### Progress

### Issue
 * [JDK-8266466](https://bugs.openjdk.java.net/browse/JDK-8266466): [lworld] Enhance javac to consume unified primitive class files generated under the option -XDunifiedValRefClass

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/valhalla pull/421/head:pull/421` \
`$ git checkout pull/421`

`$ git checkout pull/421` \
`$ git pull https://git.openjdk.java.net/valhalla pull/421/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 421`

`$ git pr show -t 421`

</details>

<a href="https://git.openjdk.java.net/valhalla/pull/421.diff">https://git.openjdk.java.net/valhalla/pull/421.diff</a>

</details>

### DIFF
--- a/src/jdk.compiler/share/classes/com/sun/tools/javac/code/Type.java
+++ b/src/jdk.compiler/share/classes/com/sun/tools/javac/code/Type.java
@@ -35,6 +35,7 @@ import java.util.function.Predicate;
 import javax.lang.model.type.*;
 
 import com.sun.tools.javac.code.Symbol.*;
+import com.sun.tools.javac.code.Type.ClassType.Flavor;
 import com.sun.tools.javac.code.TypeMetadata.Entry;
 import com.sun.tools.javac.code.Types.TypeMapping;
 import com.sun.tools.javac.code.Types.UniqueType;
@@ -241,6 +242,14 @@ public abstract class Type extends AnnoConstruct implements TypeMirror, PoolCons
     }
 
     /**
+     * Return the `flavor' associated with a ClassType.
+     * @see ClassType.Flavor
+     */
+    public Flavor getFlavor() {
+        throw new AssertionError("Unexpected call to getFlavor() on a Type that is not a ClassType: " + this);
+    }
+
+    /**
      * @return true IFF the receiver is a reference projection of an inline type and false
      * for primitives or plain references
      */
@@ -286,7 +295,7 @@ public abstract class Type extends AnnoConstruct implements TypeMirror, PoolCons
             List<Type> typarams = t.getTypeArguments();
             List<Type> typarams1 = visit(typarams, s);
             if (outer1 == outer && typarams1 == typarams) return t;
-            else return new ClassType(outer1, typarams1, t.tsym, t.metadata, t.isReferenceProjection()) {
+            else return new ClassType(outer1, typarams1, t.tsym, t.metadata, t.getFlavor()) {
                 @Override
                 protected boolean needsStripping() {
                     return true;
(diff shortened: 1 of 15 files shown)


src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/ClassReader.java
```diff
@@ -644,12 +659,14 @@ public void setEnclosingType(Type outer) {
                             sbp - startSbp);
                     if (allowPrimitiveClasses && name.toString().endsWith("$ref")) {
                         name = name.subName(0, name.length() - 4);
-                        requireProjection = true;
+                        Assert.check(prefix == 'L');
+                        flavor = L_TypeOf_Q;
                     } else {
-                        requireProjection = false;
+                        // We are seeing QFoo; or LFoo; The name itself does not shine any light on default val-refness
+                        flavor = prefix == 'L' ? Flavor.L_TypeOf_X : Flavor.L_TypeOf_X;
```
mcimadamore, line 666: Is there a typo here - shouldn't be:

```
                        flavor = prefix == 'L' ? Flavor.L_TypeOf_X : Flavor.Q_TypeOf_X;
```

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Type.java
```diff
@@ -1017,6 +1026,61 @@ public String toString() {
     public static class ClassType extends Type implements DeclaredType, LoadableConstant,
                                                           javax.lang.model.type.ErrorType {
 
+        /**
+         * The 'flavor' of a ClassType indicates its reference/primitive projectionness
+         * viewed against the default nature of the associated class.
+         */
+        public enum Flavor {
```
mcimadamore, line 1033: Perhaps you could add a method which "upgrades" a flavor to a new one, given new info.

e.g.

```
L_TypeOf_X.update(sym.isPrimitiveClass()) ->  L_TypeOf_Q/L_TypeOf_L (depending on isPrimitiveClass)
```

This would allow you to rule out (e.g. throw) bad updates. E.g. X flavors can be updated, all the other "leaf" flavors can only be updated to self. 

I think a method such as this would remove a lot of conditional logic in the clients.

mcimadamore (commented):
Overall, a clever strategy.

I'm a bit doubtful that we have seen the end of the string here. In particular, when it comes to LFoo; ambiguities (is it a regular reference Foo or a projection Foo.ref of some primitive type?) I think the old and trusted signature attribute also has a role to play.

The signature attribute will in fact require updates also to support universal generics.

src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/ClassReader.java
sadayapalam: Oops. Yes it should be as you say. I am puzzled because this was copy pasted code. Don't know how the error got introduced! Good catch!

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Type.java
sadayapalam: I'll take a look ...

sadayapalam:
> Overall, a clever strategy.
> 
> I'm a bit doubtful that we have seen the end of the string here. In particular, when it comes to LFoo; ambiguities (is it a regular reference Foo or a projection Foo.ref of some primitive type?) I think the old and trusted signature attribute also has a role to play.
> 
> The signature attribute will in fact require updates also to support universal generics.

You are right that this is not the end of the story. The implementation deals with just a single bit ACC_PRIMITIVE and weaves that into the type system - so as of this patch, we can have plain old L_TypeOf_L, L_Typeof_Q and Q_TypeOf_Q as fully "evolved forms"  and X_Typeof_X, Q_Typeof_X and L_Typeof_X as the "larval forms".

BUT,  the ambiguity you right about i.e "In particular, when it comes to LFoo;  ... (is it a regular reference Foo or a projection Foo.ref of some primitive type?)" is handled already in this patch. The single bit ACC_PRIMITIVE is enough for this.

See com.sun.tools.javac.code.Type.ClassType#isReferenceProjection which disambiguates L_TypeOf_X into one of L_TypeOf_Q or L_TypeOf_L. 

I plan to take up work on JDK-8244231 right after this,  at which point, a larval form may see other evolutions.

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Type.java
sadayapalam: > Perhaps you could add a method which "upgrades" a flavor to a new one, given new info.
> 
> e.g.
> 
> ```
> L_TypeOf_X.update(sym.isPrimitiveClass()) ->  L_TypeOf_Q/L_TypeOf_L (depending on isPrimitiveClass)
> ```
> 
> This would allow you to rule out (e.g. throw) bad updates. E.g. X flavors can be updated, all the other "leaf" flavors can only be updated to self.
> 
> I think a method such as this would remove a lot of conditional logic in the clients.

Thanks, this turned out to be a nice suggestion and helped simplify the code at 3 or 4 places.

PR #428, opened 2021-05-26
8267763: [lworld][lw3] Change "non-tearable" nomenclature to "access atomic"

shipilev:
Current Valhalla code has the experimental marker interface `java.lang.NonTearable`, which is actually about access atomicity. It makes weird claims about word tearing and out-of-thin air values.

First, this is not word tearing. Word tearing, as defined by JLS 17.6 is: _"This problem is sometimes known as word tearing, and on processors that cannot easily update a single byte in isolation some other approach will be required"._ That is, word tearing is when we cannot update the _narrow_ member without doing a _wider_ access, thus necessarily affecting the adjacent members. In Valhalla case, what we are dealing with is access atomicity: we sometimes cannot access the _wide_ member without doing a set of _narrower_ accesses. This is why JLS 17.7 says "non-atomic treatment of double and longs", not "word-tearing of double and longs".

Second, the docs for `j.l.NonTearable` mention "out-of-thin-air" (OOTA) values, which are not related here at all. OOTA are the beasts from the causality loops: those are values that were never written by normal execution of the program (i.e. speculative values). In Valhalla case, the writes that produce the broken hybrid are known and expected writes from the conflicting writers.

This nomenclature percolates to Valhalla VM code, so some change is needed there as well.

Additional testing:
 - [x] `runtime/valhalla` tests

---------
### Progress

### Integration blocker
&nbsp;⚠️ Whitespace errors (failed with the updated jcheck configuration)

### Issue
 * [JDK-8267763](https://bugs.openjdk.org/browse/JDK-8267763): [lworld][lw3] Change "non-tearable" nomenclature to "access atomic" (**Enhancement** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/428/head:pull/428` \
`$ git checkout pull/428`

`$ git checkout pull/428` \
`$ git pull https://git.openjdk.org/valhalla.git pull/428/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 428`

`$ git pr show -t 428`

</details>

<a href="https://git.openjdk.org/valhalla/pull/428.diff">https://git.openjdk.org/valhalla/pull/428.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/428#issuecomment-848546485)

### DIFF
--- a/src/hotspot/share/runtime/globals.hpp
+++ b/src/hotspot/share/runtime/globals.hpp
@@ -2077,8 +2077,8 @@ const intx ObjectAlignmentInBytes = 8;
   product(bool, UseArrayMarkWordCheck, NOT_LP64(false) LP64_ONLY(true),     \
           "Use bits in the mark word to check for flat/null-free arrays")   \
                                                                             \
-  product(ccstrlist, ForceNonTearable, "", DIAGNOSTIC,                      \
-          "List of inline classes which are forced to be atomic "           \
+  product(ccstrlist, ForceAtomicAccess, "", DIAGNOSTIC,                     \
+          "List of inline classes which are forced to be atomic access "    \
           "(whitespace and commas separate names, "                         \
           "and leading and trailing stars '*' are wildcards)")              \
                                                                             \
--- a/src/java.base/share/classes/java/lang/AtomicAccess.java
+++ b/src/java.base/share/classes/java/lang/AtomicAccess.java
@@ -26,38 +26,52 @@
 package java.lang;
 
 /**
- * A primitive class implements the {@code NonTearable} interface to
- * request that the JVM take extra care to avoid structure tearing
+ * A primitive class implements the {@code AtomicAccess} interface to
+ * request that the JVM take extra care to avoid non-atomic operations
  * when loading or storing any value of the class to a field or array
- * element.  Normally, only fields declared {@code volatile} are
- * protected against structure tearing, but a class that implements
- * this marker interface will never have its values torn, even when
+ * element.  Normally, only naturally atomic fields and fields declared
+ * {@code volatile} are always atomic, but a class that implements
+ * this marker interface will always be accessed atomically, even when
  * they are stored in array elements or in non-{@code volatile}
  * fields, and even when multiple threads perform racing writes.
  *
- * <p> An primitive instance of multiple components is said to be "torn"
- * when two racing threads compete to write those components, and one
- * thread writes some components while another thread writes other
- * components, so a subsequent observer will read a hybrid composed,
- * as if "out of thin air", of field values from both racing writes.
- * Tearing can also occur when the effects of two non-racing writes
- * are observed by a racing read.  In general, structure tearing
- * requires a read and two writes (initialization counting as a write)
- * of a multi-component value, with a race between any two of the
- * accesses.  The effect can also be described as if the Java memory
- * model break up primitive classinstance reads and writes into reads and
+ * <p> A primitive instance of multiple components can experience both
+ * transient and persistent access atomicity failures.
+ *
+ * <p> Transient failures usually arise from write-read conflicts:
+ * when one thread writes the components one-by-one, and another
+ * thread reads the components one-by-one. In doing so, reader
+ * observes the intermediate state of the primitive instance.
+ * This failure is transient, since "after" (in memory model sense)
+ * the writer finishes its writes, the instance is observed in full
+ * by any subsequent observer.
+ *
+ * <p> Permanent failures usually arise from write-write conflicts:
+ * when two threads compete to write the components, and one thread
+ * writes some components while another thread writes other
+ * components. This failure is permanent: as every subsequent observer
+ * will read a hybrid composed of field values from both racing writes.
+ *
+ * Both these effects can be described as if the Java memory model
+ * break up primitive class instance reads and writes into reads and
  * writes of their various fields, as it does with longs and doubles
  * (JLS 17.7).
  *
- * <p> In extreme cases, the hybrid observed after structure tearing
+ * <p> In extreme cases, the hybrid observed under non-atomic access
  * might be a value which is impossible to construct by normal means.
  * If data integrity or security depends on proper construction,
- * the class should be declared as implementing {@code NonTearable}.
+ * the class should be declared as implementing {@code AtomicAccess}.
+ *
+ * <p> Note this atomicity guarantee only relates to the <i>individual
+ * accesses</i>, not the compound operations over the values. The
+ * read-modify-write operation over {@code AtomicAccess} would still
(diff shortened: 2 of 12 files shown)


shipilev:
Renamed other "tearing" tests too now. `compiler/valhalla` passes with Linux x86_64 fastdebug builds.

TobiHartmann:
Looks good to me but since most of this touches runtime code, someone from runtime should review.

rose00:
Thank you for explaining to all of us about word-tearing, but that seems to be moot in the context of this PR.  This feature (`NonTearable`) is about *structure tearing*, which is a distinct concept.  Hence the different name, as you might have noticed.

There is also a reason for not using the term "atomic".  Indeed it was my preference, but Brian (as he can explain himself as well) has at least two good reasons for avoiding that term, and going with the new term `NonTearable` (as well as the underlying exposition about the *new* phenomenon of structure tearing).

(The term non-tearable is in fact closely synonymous with the term atomic, which literally means non-cuttable.)

A couple of reasons *not* to use "atomic" here:  First, the term is already taken, in the package `java.util.concurrent.atomic`, with classes like `AtomicInteger`.  Users would be faced with different `AtomicFoo` types, leading to needless pedagogical burden.  (It doesn't matter that a dozen JMM experts would be made happier by using a common term for a common underlying notion.)  Second, although you nicely propose `AtomicAccess` as a good and correct term, the basic meaning of Atomic, at the user level, is already *stronger* than atomic access:  It implies not only that structures won't tear, as longs and doubles can tear in Old Java, but also the word "atomic" implies some kind of state-to-state consistency, such as is provided by CAS or atomic add.  Yes, I see that "access" in "atomic access" is a clever way of backing away from state-to-state consistency claims, but it's too subtle.  A new term is needed for the weaker concept.

In other words, for API design, let's keep using the term Atomic for the stronger (more useful term) and NonTearable for the weaker idea of a atomic-for-single-acceses-only.  That's what you are seeing with the term "non-tearable", and what I hope you can also see with the development of the not-really-so-weird discussion of "structure tearing".

src/java.base/share/classes/java/lang/AtomicAccess.java
```diff
@@ -26,38 +26,52 @@
...
  * writes of their various fields, as it does with longs and doubles
  * (JLS 17.7).
  *
- * <p> In extreme cases, the hybrid observed after structure tearing
+ * <p> In extreme cases, the hybrid observed under non-atomic access
  * might be a value which is impossible to construct by normal means.
  * If data integrity or security depends on proper construction,
- * the class should be declared as implementing {@code NonTearable}.
+ * the class should be declared as implementing {@code AtomicAccess}.
+ *
```
rose00, line 64: This comment is useful, because it distinguishes non-tearable from the
de-facto standard notion (in the JDK) of "atomic", which pertains to
"compound operations".  Atomic compound operations relate multiple
states of a variable in a consistent sequence sequence.  This is a
strictly stronger notion than access atomicity or non-tearability.

I support adding such a comment here, but also note that its
presence in the PR is evidence that the type name should
*not* be changed to `AtomicAccess`, precisely because
users are likely to confuse *this new* use of atomic with
*existing old* uses, all of which imply the stronger concept.

src/hotspot/share/runtime/globals.hpp
```diff
@@ -2077,8 +2077,8 @@ const intx ObjectAlignmentInBytes = 8;
   product(bool, UseArrayMarkWordCheck, NOT_LP64(false) LP64_ONLY(true),     \
           "Use bits in the mark word to check for flat/null-free arrays")   \
                                                                             \
-  product(ccstrlist, ForceNonTearable, "", DIAGNOSTIC,                      \
-          "List of inline classes which are forced to be atomic "           \
+  product(ccstrlist, ForceAtomicAccess, "", DIAGNOSTIC,                     \
+          "List of inline classes which are forced to be atomic access "    \
           "(whitespace and commas separate names, "                         \
           "and leading and trailing stars '*' are wildcards)")              \
                                                                             \
```
rose00, line 2084: Internally to the JVM, the term "atomic" is acceptable, as long as
in context it is correctly understood as *single-access* atomicity.
(Which it is.)  The product flag here is at the boundary between
the JVM internals and the end-user (who sets the JVM flag on
the command line).  At that point, it is better to use the external
term `NonTearable`, which (for pedagogical reasons, as explained
elsewhere) does not use the word "atomic" but rather a synonym.

rose00 (changes requested):
Some of this PR is based on wrong-headed misunderstanding
of the term "structure tearing", and some is based on a difference
of opinion about what's best for the end-user.  The part I would
prefer to keep makes more clear the connections between
the existing "compound-atomic" APIs, all of which have `Atomic`
in their names, and the necessary new feature for multi-component
access atomicity, which should *not* have `Atomic` in its name.

Please reformulate to be less disruptive of existing decisions,
or change the decisions by persuasion.

It looks to me like the only useful change proposed here
is a reformulation of the trailing comment in NonTearable.java,
relating it more usefully to the Atomic ™  APIs.

A better exposition of "structure tearing" would be welcome too.
The discussion of transient vs. permanent failure is illuminating.

I think "out of thin air" is a useful description of what a programmer
sees on such a failure, since the structure as a whole was never
constructed by any thread.  I realize that it might distract someone
familiar with the JMM literature.  Perhaps a less confusing term
would be "a composite value never created by any thread", but
"out of thin air" (in scare quotes) seems descriptive to me at least.

The material point for the end user is that structure tearing can
lead to the appearance (temporary or permanent) of primitive
values which were never created by any thread.  Giving a way
to control that phenomenon ("out of thin air" structures, created
by races if not speculation) is the motivation for this feature.

Notice that this problem is new to primitive classes.  There is
no corresponding notion for identity classes which are
configured like primitives, and therefore have all final fields.
The JMM guarantees safe publication for those, which means
no rogue states are ever visible in *those* multi-component
structures.  But when multi-component primitives are inlined,
the JMM needs a *new* guarantee in the place of safe publication.
Non-tearability is the term we have chosen for this.

briangoetz:
I am not married to the term `non-tearable`, and happy to consider alternatives, but `atomic` is definitely a move in the wrong direction.  John outlined most of the arguments, but basically: atomicity, in the context of user-visible Java APIs, is generally reserved for describing atomicity of user-level mutative operations, including read-modify-write operations (e.g., CAS, incrementAndGet, etc.)  This is the A from "ACID".  (Our friends in the database community might call these operations _serialized_, but as that also means something else in Java, we won't call it that either.)   

So, happy to consider alternatives to non-tearable (it was just the best we came up with at the time), but I think `Atomic*` will definitely send the wrong message about what operations are atomic relative to each other.

PR #429, opened 2021-05-26
8267791: [lworld][lw3] Support compiler blackholes for inline types

shipilev:
JDK-8259316 added the support for compiler blackholes. We need to check and implement the support for inline type arguments for them. I am sure we would find more wrinkles with nano-benchmarks, but insofar sample benchmark work as they should:

```
public class PrimitiveClasses {
    static primitive class Value {
        long x;
        long y;
        public Value(long x, long y) {
            this.x = x;
            this.y = y;
        }
    }

    [volatile] Value v = Value.default;

    @Benchmark
    public Value test() {
        return v;
    }
}
```

Before: crash on assert, `assert(vt->fast_out(i)->is_InlineTypeBase(), "Unexpected inline type user");`

After, generated code in non-`volatile` case shows blackhole of individual components:

```
$ java -Djmh.blackhole.mode=COMPILER -jar target/benchmarks.jar PrimitiveClasses -prof perfasm
...
....[Hottest Region 1]..............................................................................
c2, level 4, org.openjdk.jmh_generated.PrimitiveClasses_test_jmhTest::test_avgt_jmhStub, version 641 (39 bytes) 
...
         ↗  0x00007f67bcc8eee0:   mov    0x18(%r8),%r10        ; read and blackhole MyValue.x
  0.02%  │  0x00007f67bcc8eee4:   mov    0x10(%r8),%r11        ; read and blackhole MyValue.y
  0.10%  │  0x00007f67bcc8eee8:   movzbl 0x94(%r14),%r11d      ; read isDone
  0.10%  │  0x00007f67bcc8eef0:   mov    0x3b0(%r15),%r10      ; TL handshake
  2.79%  │  0x00007f67bcc8eef7:   add    $0x1,%rbp             ; ops++
         │  0x00007f67bcc8eefb:   test   %eax,(%r10)           ; TL handshake
 86.74%  │  0x00007f67bcc8eefe:   test   %r11d,%r11d           ; isDone?
  2.31%  ╰  0x00007f67bcc8ef01:   je     0x00007f67bcc8eee0    
```

And generated code in `volatile` case shows blackhole of the box (with the null-check):

```
          ↗   0x00007f3948c8db70:   shl    $0x3,%r10            ; unpack and blackhole MyValue
  0.02%   │↗  0x00007f3948c8db74:   movzbl 0x94(%r14),%r8d      ; read isDone
  0.68%   ││  0x00007f3948c8db7c:   mov    0x3b0(%r15),%r9      ; TL handshake
  0.72%   ││  0x00007f3948c8db83:   add    $0x1,%rbp            ; ops++
          ││  0x00007f3948c8db87:   test   %eax,(%r9)           ; TL handshake
 92.30%   ││  0x00007f3948c8db8a:   test   %r8d,%r8d            ; isDone?
         ╭││  0x00007f3948c8db8d:   jne    0x00007f3948c8dba4   
         │││  0x00007f3948c8db8f:   mov    0xc(%r11),%r10d      ; load "v" (MyValue box)
  0.35%  │││  0x00007f3948c8db93:   test   %r10d,%r10d          ; null check it
         │╰│  0x00007f3948c8db96:   jne    0x00007f3948c8db70   
```

---------
### Progress

### Issue
 * [JDK-8267791](https://bugs.openjdk.java.net/browse/JDK-8267791): [lworld][lw3] Support compiler blackholes for inline types

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/valhalla pull/429/head:pull/429` \
`$ git checkout pull/429`

`$ git checkout pull/429` \
`$ git pull https://git.openjdk.java.net/valhalla pull/429/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 429`

`$ git pr show -t 429`

</details>

<a href="https://git.openjdk.java.net/valhalla/pull/429.diff">https://git.openjdk.java.net/valhalla/pull/429.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/opto/compile.cpp
+++ b/src/hotspot/share/opto/compile.cpp
@@ -1927,6 +1927,27 @@ void Compile::process_inline_types(PhaseIterGVN &igvn, bool remove) {
       } else if (vt->is_InlineTypePtr()) {
         igvn.replace_node(vt, vt->get_oop());
       } else {
+        // Check if any users are blackholes. If so, rewrite them to use the individual
+        // components, instead of the inline type node that goes away.
+        for (DUIterator i = vt->outs(); vt->has_out(i); i++) {
+          if (vt->out(i)->is_Blackhole()) {
+            BlackholeNode* bh = vt->out(i)->as_Blackhole();
+
+            // Unlink the old input
+            int idx = bh->find_edge(vt);
+            assert(idx != -1, "The edge should be there");
+            bh->del_req(idx);
+            --i;
+
+            // Add the new inputs to the components
+            for (uint c = 0; c < vt->field_count(); c++) {
+              bh->add_req(vt->field_value(c));
+            }
+
+            // Node modified, record for IGVN
+            igvn.record_for_igvn(bh);
+          }
+        }
 #ifdef ASSERT
         for (DUIterator_Fast imax, i = vt->fast_outs(imax); i < imax; i++) {
           assert(vt->fast_out(i)->is_InlineTypeBase(), "Unexpected inline type user");
--- a/src/hotspot/share/opto/compile.cpp
+++ b/src/hotspot/share/opto/compile.cpp
@@ -1941,7 +1941,7 @@ void Compile::process_inline_types(PhaseIterGVN &igvn, bool remove) {
             --i;
 
             if (vt->is_allocated(&igvn)) {
-              // Already has an allocation, blackhole that
+              // Already has the allocated instance, blackhole that
               bh->add_req(vt->get_oop());
             } else {
               // Not allocated yet, blackhole the components
--- a/src/hotspot/share/opto/compile.cpp
+++ b/src/hotspot/share/opto/compile.cpp
@@ -1928,7 +1928,7 @@ void Compile::process_inline_types(PhaseIterGVN &igvn, bool remove) {
         igvn.replace_node(vt, vt->get_oop());
       } else {
         // Check if any users are blackholes. If so, rewrite them to use either the
-        // allocated box, or individual components, instead of the inline type node
+        // allocated buffer, or individual components, instead of the inline type node
         // that goes away.
         for (DUIterator i = vt->outs(); vt->has_out(i); i++) {
           if (vt->out(i)->is_Blackhole()) {
--- a/src/hotspot/share/opto/compile.cpp
+++ b/src/hotspot/share/opto/compile.cpp
@@ -1927,8 +1927,9 @@ void Compile::process_inline_types(PhaseIterGVN &igvn, bool remove) {
       } else if (vt->is_InlineTypePtr()) {
         igvn.replace_node(vt, vt->get_oop());
       } else {
-        // Check if any users are blackholes. If so, rewrite them to use the individual
-        // components, instead of the inline type node that goes away.
+        // Check if any users are blackholes. If so, rewrite them to use either the
+        // allocated box, or individual components, instead of the inline type node
+        // that goes away.
         for (DUIterator i = vt->outs(); vt->has_out(i); i++) {
           if (vt->out(i)->is_Blackhole()) {
             BlackholeNode* bh = vt->out(i)->as_Blackhole();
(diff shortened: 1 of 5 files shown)


test/hotspot/jtreg/runtime/valhalla/inlinetypes/BlackholeTest.java
```diff
@@ -0,0 +1,80 @@
...
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+package runtime.valhalla.inlinetypes;
```
TobiHartmann, line 24: I think this test should go to compiler.

src/hotspot/share/opto/compile.cpp
```diff
@@ -1927,6 +1927,34 @@ void Compile::process_inline_types(PhaseIterGVN &igvn, bool remove) {
       } else if (vt->is_InlineTypePtr()) {
         igvn.replace_node(vt, vt->get_oop());
       } else {
+        // Check if any users are blackholes. If so, rewrite them to use either the
+        // allocated box, or individual components, instead of the inline type node
```
TobiHartmann, line 1931: Better replace `box` by `buffer`.

src/hotspot/share/opto/compile.cpp
```diff
@@ -1927,6 +1927,34 @@ void Compile::process_inline_types(PhaseIterGVN &igvn, bool remove) {
...
+
+            // Unlink the old input
+            int idx = bh->find_edge(vt);
+            assert(idx != -1, "The edge should be there");
+            bh->del_req(idx);
+            --i;
+
+            if (vt->is_allocated(&igvn)) {
+              // Already has the allocated instance, blackhole that
+              bh->add_req(vt->get_oop());
```
TobiHartmann, line 1945: This might keep an otherwise unused buffer allocation alive. Is that the expected behavior of using backholes on inline types? Why not always blackhole the compontents?

shipilev: The expected behavior is debatable. Current behavior does make sense to me: for non-flattened cases, we blackhole the reference to wrapper (as that is what application code would be exposed to), in flattened cases, we blackhole the components (as that is what application code would be exposed to). 

I probably misunderstand the compiler code here. Can there be an allocated box that goes away later, and only the component accesses survive? If so, how should this code ask for "if the allocated instance is alive, then blackhole it"?

src/hotspot/share/opto/compile.cpp
shipilev: Done.

test/hotspot/jtreg/runtime/valhalla/inlinetypes/BlackholeTest.java
shipilev: Done.

src/hotspot/share/opto/compile.cpp
TobiHartmann: On second thought, I think this is fine.

> Can there be an allocated box that goes away later, and only the component accesses survive?

Yes, in general that can happen and is not unlikely. For example, due to late inlining or other optimizations removing a non-scalarizable use of the buffer allocation. However, in this case all such optimizations are already over and unused allocations would have been removed. I.e., if `vt->is_allocated` is true, there is a user of the buffer allocation that won't go away and therefore linking the allocation to the blackhole should be fine.

TobiHartmann (approved):

src/hotspot/share/opto/compile.cpp
shipilev: Right. So my understanding was correct: this is a final elimination phase, and we can trust `vt->_is_allocated`. I did a few more experiments, and they seem to produce the expected (AFAICS) generated code.

PR #431, opened 2021-05-28
8267846: [lworld] JIT support for the L/Q model (step 1)

TobiHartmann:
After https://github.com/openjdk/valhalla/pull/409 and https://github.com/openjdk/valhalla/pull/414 have been integrated, we need to fix the JIT code as well. This mostly means re-applying the old changes from https://github.com/openjdk/valhalla/commit/22dd0da1 and https://github.com/openjdk/valhalla/commit/e895128f and adjust them to the current implementation.

All compiler tests now pass except for `TestNullableInlineTypes`, `TestNullableArrays` and `TestIntrinsics` which need reflection support. The corresponding C2 intrinsics are still broken. I'll fix them with [JDK-8267932](https://bugs.openjdk.java.net/browse/JDK-8267932) once reflection support is available.

Thanks,
Tobias

---------
### Progress

### Issue
 * [JDK-8267846](https://bugs.openjdk.java.net/browse/JDK-8267846): [lworld] JIT support for the L/Q model (step 1)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/valhalla pull/431/head:pull/431` \
`$ git checkout pull/431`

`$ git checkout pull/431` \
`$ git pull https://git.openjdk.java.net/valhalla pull/431/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 431`

`$ git pr show -t 431`

</details>

<a href="https://git.openjdk.java.net/valhalla/pull/431.diff">https://git.openjdk.java.net/valhalla/pull/431.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/c1/c1_Instruction.cpp
+++ b/src/hotspot/share/c1/c1_Instruction.cpp
@@ -108,7 +108,7 @@ void Instruction::state_values_do(ValueVisitor* f) {
 }
 
 ciType* Instruction::exact_type() const {
-  ciType* t =  declared_type();
+  ciType* t = declared_type();
   if (t != NULL && t->is_klass()) {
     return t->as_klass()->exact_klass();
   }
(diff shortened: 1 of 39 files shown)


src/hotspot/share/c1/c1_Instruction.cpp
```diff
@@ -138,15 +138,13 @@ bool Instruction::maybe_flattened_array() {
   if (UseFlatArray) {
     ciType* type = declared_type();
     if (type != NULL) {
-      if (type->is_obj_array_klass()) {
-        // Due to array covariance, the runtime type might be a flattened array.
+      if (type->is_obj_array_klass() && !type->as_obj_array_klass()->is_elem_null_free()) {
+        // The runtime type of [LMyValue might be [QMyValue due to [QMyValue <: [LMyValue.
```
fparain, line 142: if type->is_obj_array_klass() is true and type->as_obj_array_klass()->is_elem_null_free() is also true, the method returns false, is it the expected behavior?

TobiHartmann: Yes, because in that case we have a null-free but non-flattened array. Right?

fparain: Correct. Got it.

fparain (approved):
I cannot review changes in C2 or CI typeFlow, but changes other changes in CI and C1 look good to me.

Fred

TobiHartmann:
Thanks for the review, Fred!

PR #436, opened 2021-06-03
8267948: [lword] Core reflection and method handles support for L/Q model

mlchung:
This PR implements the primary (`ref`) mirror and secondary (`val`) mirror for primitive classes [1].   The following APIs returning `Class` return the primary mirror: 

`Object::getClass`
`Class::getClass`
`Class::forName`
`java.lang.reflect.Member::getDeclaringClass` (i.e. `Field`, `Method` and `Constructor`)

`Class::getName` and `Class::getSimpleName` return the same name for the `ref` and `val` mirror of a primitive class.  Therefore, Class.forName(type.getName()) will work if the type is either `ref` or `val` mirror.

`Class::getTypeName` returns `Foo.ref` if this `Class` is a primitive reference type.  (we will revisit this for reference-favoring primitive class.)

New proposed APIs: 
`Class::isPrimitiveClass` returns true for the `ref` and `val` mirror of a primitive class. 

`Class::isPrimaryType` returns true if this `Class` object represents the primary mirror
`Class::asPrimaryType` returns the primary type of this class or interface.
- An alternative I considered is `Class::isReferenceType` and `Class::asReferenceType` but they will have to specify the special case for primitive types which are not (yet) primitive classes.   I want to go with this initial patch to make progress for the prototype.  We will follow up the API discussion next.

`Class::isValueType` returns true if this `Class` object represents the primary mirror
`Class::asValueType` returns the primary type of this class or interface.

Several tests fail when running with -Xcomp which may depend on the second phase JIT support work (JDK-8267932) which depends on this work to make progress.

[1] https://github.com/openjdk/valhalla-docs/blob/main/site/design-notes/state-of-valhalla/03-vm-model.md

---------
### Progress

### Issue
 * [JDK-8267948](https://bugs.openjdk.java.net/browse/JDK-8267948): [lword] Core reflection and method handles support for L/Q model

### Reviewers
 * [Roger Riggs](https://openjdk.java.net/census#rriggs) (@RogerRiggs - Committer)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/valhalla pull/436/head:pull/436` \
`$ git checkout pull/436`

`$ git checkout pull/436` \
`$ git pull https://git.openjdk.java.net/valhalla pull/436/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 436`

`$ git pr show -t 436`

</details>

<a href="https://git.openjdk.java.net/valhalla/pull/436.diff">https://git.openjdk.java.net/valhalla/pull/436.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/classfile/javaClasses.cpp
+++ b/src/hotspot/share/classfile/javaClasses.cpp
@@ -816,6 +816,8 @@ int java_lang_Class::_class_loader_offset;
 int java_lang_Class::_module_offset;
 int java_lang_Class::_protection_domain_offset;
 int java_lang_Class::_component_mirror_offset;
+int java_lang_Class::_primary_mirror_offset;
+int java_lang_Class::_secondary_mirror_offset;
 int java_lang_Class::_init_lock_offset;
 int java_lang_Class::_signers_offset;
 int java_lang_Class::_name_offset;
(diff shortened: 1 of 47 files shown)


fparain (approved):
The runtime part looks good to me.

Fred

RogerRiggs (approved):
Generally looks ok.
I think there will be confusion between Class.isPrimitive() and Class.isPrimitiveClass()
but that's the way it is.

mlchung:
> Generally looks ok.

Thanks Roger.

> I think there will be confusion between Class.isPrimitive() and Class.isPrimitiveClass()
> but that's the way it is.

It probably becomes not so much of an issue when JEP 402 is in.

PR #442, opened 2021-06-10
8268553: [lworld] CI must return secondary mirror when accessing CONSTANT_Class with Q-signature

fparain:
Please review this small change in CI to return the secondary mirror when accessing a CONSTANT_Class_info entry containing a Q-signature.
This change fixes all compiler tests failures in the lqagain branch but the ones depending on intrinsics (Class::getSuperclass intrinsic and Class::isAssignableFrom intrinsics need to be fixed in a separate changeset).

Thank you,

Fred

---------
### Progress

### Issue
 * [JDK-8268553](https://bugs.openjdk.java.net/browse/JDK-8268553): [lworld] CI must return secondary mirror when accessing CONSTANT_Class with Q-signature

### Reviewers
 * [Roland Westrelin](https://openjdk.java.net/census#roland) (@rwestrel - Committer)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/valhalla pull/442/head:pull/442` \
`$ git checkout pull/442`

`$ git checkout pull/442` \
`$ git pull https://git.openjdk.java.net/valhalla pull/442/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 442`

`$ git pr show -t 442`

</details>

<a href="https://git.openjdk.java.net/valhalla/pull/442.diff">https://git.openjdk.java.net/valhalla/pull/442.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/ci/ciEnv.cpp
+++ b/src/hotspot/share/ci/ciEnv.cpp
@@ -697,7 +697,15 @@ ciConstant ciEnv::get_constant_by_index_impl(const constantPoolHandle& cpool,
     }
     assert (klass->is_instance_klass() || klass->is_array_klass(),
             "must be an instance or array klass ");
-    return ciConstant(T_OBJECT, klass->java_mirror());
+    if (tag.is_unresolved_klass()) {
+      return ciConstant(T_OBJECT, get_unloaded_klass_mirror(klass));
+    } else {
+      if (tag.is_Qdescriptor_klass()) {
+        return ciConstant(T_OBJECT, klass->as_inline_klass()->val_mirror());
+      } else {
+        return ciConstant(T_OBJECT, klass->java_mirror());
+      }
+    }
   } else if (tag.is_method_type()) {
     // must execute Java code to link this CP entry into cache[i].f1
     ciSymbol* signature = get_symbol(cpool->method_type_signature_at(index));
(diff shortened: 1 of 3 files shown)


mlchung (approved):
It looks right to me.

rwestrel (approved):
Looks good to me.

fparain:
Mandy, Roland,

Thank you for your reviews,

Fred

src/hotspot/share/ci/ciEnv.cpp
```diff
@@ -697,7 +697,15 @@ ciConstant ciEnv::get_constant_by_index_impl(const constantPoolHandle& cpool,
     }
     assert (klass->is_instance_klass() || klass->is_array_klass(),
             "must be an instance or array klass ");
-    return ciConstant(T_OBJECT, klass->java_mirror());
+    if (tag.is_unresolved_klass()) {
```
TobiHartmann, line 700: This is too strict because `klass->is_loaded()` can be true even if `tag.is_unresolved_klass()`. This leads to C2 adding uncommon traps to trigger resolution in the interpreter. However, I've observed cases were the interpreter never resolved the constant and `tag.is_unresolved_klass()` remained true, leading to continuous deoptimizations in C2 compiled code. I'm not sure if that is expected behavior or if there is a bug in the interpreter but the CI code should be fixed in any case. I'll do that with [JDK-8267932](https://bugs.openjdk.java.net/browse/JDK-8267932).

fparain: Do you have an example where the interpreter never resolves the entry?

PR #482, opened 2021-07-16
8244231: [lworld] Add support for ref-default and val-default inline classes.

sadayapalam:
Code changes and tests

---------
### Progress

### Issue
 * [JDK-8244231](https://bugs.openjdk.java.net/browse/JDK-8244231): [lworld] Add support for ref-default and val-default inline classes.

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/valhalla pull/482/head:pull/482` \
`$ git checkout pull/482`

`$ git checkout pull/482` \
`$ git pull https://git.openjdk.java.net/valhalla pull/482/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 482`

`$ git pr show -t 482`

</details>

<a href="https://git.openjdk.java.net/valhalla/pull/482.diff">https://git.openjdk.java.net/valhalla/pull/482.diff</a>

</details>


src/jdk.compiler/share/classes/com/sun/tools/javac/code/Type.java
```diff
@@ -250,18 +250,34 @@ public Flavor getFlavor() {
...
+        return false;
+    }
+
+    /**
+     * Returns the ClassType representing the primitive value type
+     * of this type, if the class of this type is a primitive class
+     * null otherwise
      */
-    public Type valueProjection() {
+    public ClassType asValueType() {
```
sadayapalam, line 280: These various internal APIs are still crystallizing. I have a follow up ticket JDK-8268734 to evolve them to final form.

sadayapalam:
Recommended order of review:

Flags.java
JavacParser.java
Type.java
ClassWriter.java
ClassReader.java
Attr.java

then other files.

This order in review provide a good understanding of the work flow.

sadayapalam:
(Although the title says add support for val-default classes, this is all about ref-defaul classes. The plain primitive classes we have been supporting all the while are all val-default)

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
```diff
@@ -1212,7 +1211,8 @@ public Boolean visitClassType(ClassType t, Type s) {
                 // If t is an intersection, sup might not be a class type
                 if (!sup.hasTag(CLASS)) return isSubtypeNoCapture(sup, s);
                 return sup.tsym == s.tsym
-                    && (t.tsym != s.tsym || t.isReferenceProjection() == s.isReferenceProjection())
+                    && (t.tsym != s.tsym ||
+                        (t.isReferenceProjection() == s.isReferenceProjection() && t.isValueProjection() == s.isValueProjection()))
                      // Check type variable containment
```
sadayapalam, line 1216: This check should really be comparing that t.flavor == s.flavor. But this is causing some issues at the moment and will be followed up in JDK-8270882

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
```diff
@@ -1459,14 +1459,15 @@ public Boolean visitClassType(ClassType t, Type s) {
                 }
                 return t.tsym == s.tsym
                     && t.isReferenceProjection() == s.isReferenceProjection()
+                    && t.isValueProjection() == s.isValueProjection()
                     && visit(getEnclosingType(t), getEnclosingType(s))
```
sadayapalam, line 1463: Likewise this should be checking that t.flavor == s.flavor. Will be followed up in JDK-8270882

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
```diff
@@ -2625,10 +2626,18 @@ public Type visitWildcardType(WildcardType t, Boolean recurse) {
...
                 Type erased = eraseClassType(t, recurse);
-                if (erased.hasTag(CLASS) && t.flavor != erased.getFlavor()) {
+                Flavor wantedFlavor = t.flavor;
+                if (t.isIntersection()) {
+                    IntersectionClassType ict = (IntersectionClassType) t;
+                    Type firstExplicitBound = ict.getExplicitComponents().head;
+                    if (firstExplicitBound.hasTag(CLASS))
+                        wantedFlavor = firstExplicitBound.getFlavor();
+                    // Todo: Handle Type variable case.
+                }
```
sadayapalam, line 2636: This TODO is captured as a follow up task in JDK-8270882

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java
```diff
@@ -984,9 +984,6 @@ public void visitClassDef(JCClassDecl tree) {
                         env.tree.hasTag(NEWCLASS)) {
                     c.flags_field |= NOOUTERTHIS;
                 }
-                if (env.tree.hasTag(NEWCLASS) && types.isPrimitiveClass(c.getSuperclass())) {
-                    c.flags_field |= PRIMITIVE_CLASS; // avoid further secondary errors.
-                }
                 attribClass(tree.pos(), c);
```
sadayapalam, line 987: This chunk purportedly introduced to serve the purpose of "avoiding further secondary errors" had outlived its purpose and is removed now. This change is not "essential" to the overall implementation and is a incidental cleanup

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java
```diff
@@ -322,7 +322,7 @@ void checkAssignable(DiagnosticPosition pos, VarSymbol v, JCTree base, Env<AttrC
                    withfield operator -This does not result in mutation of final fields; the code generator
                    would implement `copy on write' semantics via the opcode `withfield'.
                 */
-                if (env.info.inWithField && v.getKind() == ElementKind.FIELD && (v.flags() & STATIC) == 0 && types.isPrimitiveClass(v.owner.type)) {
+                if (env.info.inWithField && v.getKind() == ElementKind.FIELD && (v.flags() & STATIC) == 0 && v.owner.isPrimitiveClass()) {
                     if (env.enclClass.sym.outermostClass() == v.owner.outermostClass())
```
sadayapalam, line 326: This is an example of a code change you will see recur several times: Earlier when we had only value default primitive classes to check if a field is a field of a primitive class it was alright to ask
```
types.isPrimitiveClass(v.owner.type)
```
With ref-default classes entering the picture if v.owner is a ref-default primitive class then v.owner.type will not
answer true for the above check. (For a ref-default primitive class Optional.val, Optional.type is 'L' type not Q type) and we need to check
```
owner.isPrimitiveClass()
```
which will answer true for both val-default and ref-default primitive classes.

src/jdk.jdeps/share/classes/com/sun/tools/classfile/AccessFlags.java
```diff
@@ -60,7 +61,7 @@
     public static final int ACC_MANDATED      = 0x8000; //                          method parameter
     public static final int ACC_MODULE        = 0x8000; // class
 
-    public static enum Kind { Class, InnerClass, Field, Method}
+    public static enum Kind { Class, InnerClass, Field, Method, JavacExtra}
```
sadayapalam, line 65: This is an half hearted attempt to find a place for ACC_REF_DEFAULT - I don't think it belongs here really since it is not a part of the class flags. But I have stuck this in here for now - In javap code to reference ACC_REF_DEFAULT symbolically this is useful (since the compiler's definition from Flags.java is not visible there in javap)

src/jdk.jdeps/share/classes/com/sun/tools/classfile/Attribute.java
```diff
@@ -48,6 +48,7 @@
     public static final String EnclosingMethod          = "EnclosingMethod";
     public static final String Exceptions               = "Exceptions";
     public static final String InnerClasses             = "InnerClasses";
+    public static final String JavaFlags                = "JavaFlags";
     public static final String LineNumberTable          = "LineNumberTable";
```
sadayapalam, line 52: The name and attribute syntax was suggested by Dan, this may incur/undergo some bikeshedding later.

src/jdk.jdeps/share/classes/com/sun/tools/javap/AttributeWriter.java
```diff
@@ -354,6 +355,16 @@ public Void visitInnerClasses(InnerClasses_attribute attr, Void ignore) {
         return null;
     }
 
+    @Override
+    public Void visitJavaFlags(JavaFlags_attribute attr, Void unused) {
+        println("Extended Flags:");
+        indent(+1);
+        if ((attr.extendedFlags & ACC_REF_DEFAULT) != 0)
+            println("ACC_REF_DEFAULT");
+        indent(-1);
```
sadayapalam, line 364: com.sun.tools.javac.code.Flags#ACC_REF_DEFAULT is not visible here. So com.sun.tools.classfile.AccessFlags has been enhanced to define ACC_REF_DEFAULT - but this is a stop gap

test/langtools/tools/javac/valhalla/lworld-values/CheckFinal.out
```diff
@@ -2,6 +2,4 @@ CheckFinal.java:14:13: compiler.err.cant.assign.val.to.final.var: fi
 CheckFinal.java:15:13: compiler.err.cant.assign.val.to.final.var: fe
 CheckFinal.java:17:13: compiler.err.cant.assign.val.to.final.var: xsf
 CheckFinal.java:19:29: compiler.err.cant.inherit.from.final: CheckFinal
-CheckFinal.java:19:42: compiler.err.concrete.supertype.for.primitive.class: compiler.misc.anonymous.class: CheckFinal$1, CheckFinal
-CheckFinal.java:19:25: compiler.err.prob.found.req: (compiler.misc.inconvertible.types: compiler.misc.anonymous.class: CheckFinal, CheckFinal)
-6 errors
+4 errors
```
sadayapalam, line 5: The (now deleted) attempt to avoid further secondary errors in Attr.java having outlived its purpose was actually causing additional errors to be emitted! Fixed.

sadayapalam:
I have annotated the code changes with hopefully helpful comments that will help the reviewer. 
https://bugs.openjdk.java.net/browse/JDK-8270882 has various tasks deliberately deferred.

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Type.java
```diff
@@ -1320,36 +1338,76 @@ public boolean isReference() {
...
+            return flavor == Flavor.L_TypeOf_Q && tsym.type.getFlavor() == Flavor.Q_TypeOf_Q; // discount reference favoring primitives.
+        }
+
+        @Override
+        public boolean isPrimitiveReferenceType() {
+            // guard against over-eager and/or inopportune completion
             if (tsym != null) {
                 if (flavor == Flavor.L_TypeOf_X || tsym.isCompleted()) {
-                    flavor = flavor.metamorphose(tsym.isPrimitiveClass());
+                    flavor = flavor.metamorphose(tsym.flags());
```
mcimadamore, line 1366: (Not related to this review) I think we should probably change the `metamorphose` name to a more neutral one like `update` ?

mcimadamore (commented):
Overall the changes look solid. My main note is a naming gripe - e.g. `Type::isPrimitiveClass` is now no longer expressive - and it is outright confusing. I've left suggestions for possible alternate names which I think sit well with where the code is going.

I'm also a bit unsure (and left comment) on whether `isReferenceProjection` and `isValueProjection` are implemented correctly. They do not seem to be the dual of each other at the moment (or at least do not seem to do what the javadoc says they should).

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
```diff
@@ -1212,7 +1211,8 @@ public Boolean visitClassType(ClassType t, Type s) {
                 // If t is an intersection, sup might not be a class type
                 if (!sup.hasTag(CLASS)) return isSubtypeNoCapture(sup, s);
                 return sup.tsym == s.tsym
-                    && (t.tsym != s.tsym || t.isReferenceProjection() == s.isReferenceProjection())
+                    && (t.tsym != s.tsym ||
+                        (t.isReferenceProjection() == s.isReferenceProjection() && t.isValueProjection() == s.isValueProjection()))
```
mcimadamore, line 1215: I guess this works because `asSuper` will return either:

* a supertype of `t`, in which case we now it's a reference type, so for `t` to be subtype of `s`, it must be that `t` is a reference type too
* it returns `t` itself, in which case subtyping holds as long as `t` and `s` are the same

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java
```diff
@@ -4662,11 +4667,31 @@ Type checkIdInternal(JCTree tree,
...
 
-                    // (a) If symbol is a primitive class and its reference projection
-                    // is requested via the .ref notation, then adjust the computed type to
+                    // (a) If symbol is a primitive class and its reference/value projection
+                    // is requested via the .ref/.val notation, then adjust the computed type to
                     // reflect this.
-                    if (owntype.isPrimitiveClass() && tree.hasTag(SELECT) && ((JCFieldAccess) tree).name == names.ref) {
-                        owntype = new ClassType(owntype.getEnclosingType(), owntype.getTypeArguments(), (TypeSymbol)sym, owntype.getMetadata(), Flavor.L_TypeOf_Q);
+                    if (sym.isPrimitiveClass()) {
+                        if (sym.isReferenceFavoringPrimitiveClass()) {
```
mcimadamore, line 4674: Shouldn't we simplify this to try and use `asValueType` - probably this suggests there should be an `asReferenceType` too?

mcimadamore (commented):
Added some more comments

src/jdk.compiler/share/classes/com/sun/tools/javac/parser/JavacParser.java
```diff
@@ -3990,6 +3990,16 @@ protected JCClassDecl classDeclaration(JCModifiers mods, Comment dc) {
         accept(CLASS);
         Name name = typeName();
 
+        if ((mods.flags & Flags.PRIMITIVE_CLASS) != 0) {
```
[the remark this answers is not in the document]
sadayapalam: Understand, this is directly implementing the syntax prescribed by the JEP and feedback would be welcomed there,

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java
sadayapalam: Fair points. 

Since we already have com.sun.tools.javac.code.Type#isPrimitive it is useful to retain to "Class" somewhere in the new APIs name although perhaps not in the present form of Type#isPrimitiveClass()

Perhaps the old method should be renamed to isBasicPrimitive() and the new one should be isPrimitiveClassType()

I will carry over these comments to JDK-8268734 (Various internal Type/Symbol APIs need rationalization/adjustment)

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Type.java
```diff
@@ -1320,36 +1338,76 @@ public boolean isReference() {
...
+            // guard against over-eager and/or inopportune completion
+            if (tsym != null) {
+                if (flavor == Flavor.Q_TypeOf_X || tsym.isCompleted()) {
+                    flavor = flavor.metamorphose(tsym.flags());
 
-            if (projection !=  null)
-                return projection;
+                }
+            }
+            return flavor == Flavor.Q_TypeOf_L;
```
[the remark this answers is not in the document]
sadayapalam: W.r.t the observation: "I'm also a bit unsure (and left comment) on whether isReferenceProjection and isValueProjection are implemented correctly. They do not seem to be the dual of each other at the moment (or at least do not seem to do what the javadoc says they should)."

Please take a another look! I believe they are dual of each other and do what the javadoc claims - (although as I call out below, I am open to questioning whether their behavior should be redefined)

ATM, the javadoc of isReferenceProjection() and isValueProjection() read:

```
    /**
     * @return true IFF the receiver is a reference projection type of a *value favoring* primitive class
     * and false otherwise. Or in other words, is this a class type non-redundantly notated with .ref ??
     */
    public boolean isReferenceProjection() {
        return false;
    }

    /**
     * @return true IFF the receiver is a value projection of a *reference favoring* primitive class type
     * and false otherwise. Or in other words, is this a class type non-redundantly notated with .val ??
     */
    public boolean isValueProjection() {
        return false;
    }

```

You ask of com.sun.tools.javac.code.Type.ClassType#isValueProjection
```
Shouldn't we discount primitive-default primitive classes in here? (e.g. should this code be symmetric with isReferenceProjection?). E.g. what about .val in an already primitive-default primitive class?
```

We *are* already discounting primitive-default classes, because in that case the flavor would be Q_TypeOf_Q.

The ambiguity exists only for the flavor L_TypeOf_Q - when we have a type that is an L_TypeOf_Q it could be
either the (a) reference projection of value default primitive class or (b) the reference type of a reference default primitive class.   

Of (a) and (b) only (a) we consider to be the *reference projection* as per the javadoc.

On the other hand:

Between Q_Typeof_Q and Q_Typeof_L, only the latter is considered a value projection as per the javadoc.

(Backing up a bit it is a fair question to ask should the javadoc be defined that way - should both Q_Typeof_Q and Q_Typeof_L be considered value *projections* and should L_TypeOf_Q should *always* be considered a reference *projection* irrespective of whether the associated class is a reference default or value default ? I have chosen not to since that seems "natural" - but I am open to revisiting this if our experience proves otherwise.

Please see that this is the same issue JEP 401 grapples with by asking the question in https://openjdk.java.net/jeps/401
```
(Open question: should it be legal to redundantly apply .val to a standard primitive class name, or .ref to a reference-favoring primitive class name?)
```

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Type.java
```diff
@@ -1320,36 +1338,76 @@ public boolean isReference() {
 
         @Override
         public boolean isPrimitiveClass() {
```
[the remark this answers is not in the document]
sadayapalam: Agreed, per earlier comment, I will take up this in JDK-8268734. 

(I am a bit unsettled on using "value type" at all, since the JEP introduces the name primitive class type and we are moving away value/inline nomenclature. But I see that java.lang.Class defines a method by name asValueType)

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Type.java
sadayapalam: Oh, you didn't like the cute name I coined! I really liked metamorphose() since it captures that the Type starts out in larval form where not much is known about and evolves into a full blown form in stages through metamorphosis!

No matter, I am not wedded to the name,  I will add a note to JDK-8268734 to follow up

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
sadayapalam: Right.

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java
sadayapalam: ATM, we have com.sun.tools.javac.code.Type#referenceProjection()

Again, I will use JDK-8268734 to rationalize the various internal APIs we have - while doing so, also looking at choice of terminology adopted by java.lang.Class  

But I didn't understand the comment about the simplification you are suggesting - can you clarify ?

mcimadamore: I mean - this code does a bunch of checks - and then creates a new class type with the opposite flavor/polarity. Isn't that what methods like `Type::asValueType` are supposed to do? E.g. if I have a reference-favoring primitive and I see `.val` can't I just get the type I want by calling `asValueType()` on it? If so, should we also have the dual?

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Type.java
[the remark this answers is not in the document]
mcimadamore: > when we have a type that is an L_TypeOf_Q it could be
> either the (a) reference projection of value default primitive class or (b) the reference type of a reference default primitive class.

This seems at odds with what the Javadoc for that flavor says:

```
/**
             * Reference projection type of a primitive-favoring aka primitive-default
             * plain vanilla primitive class type,
             */
            L_TypeOf_Q,
```

E.g. javadoc suggests that Q means primitive-default already, so there should be no need to check the tsym?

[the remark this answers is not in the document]
mcimadamore: More generally, I think I'm slightly uncomfortable by the use of the letters `Q` and `L` in the enum constants `X_TypeOf_Y`. Now, when `L` and `Q` are in the `X` position, the semantics is simple: `L` means `.ref`, `Q` means `.val`.

But when `L` and `Q` appear in the `Y` position, then it becomes messy - because the choice is no longer binary:

* the class is a primitive class or not (1 bit)
* if primitive, the class is reference-favoring or not (another bit)

Right now, it seems that you use `Q` to denote whether the class is primitive class or not, which is useful, I assume, to detect distinction between a legacy reference type (`L_TypeOf_L`) and a reference projection of a primitive class (`L_TypeOf_Q`). This distinction is, I believe an important one, as the former has identity, the latter doesn't, so type checking would probably differ.

What I'm less sure is whether you want/need different type checking rules for when `L_TypeOf_Q && reference-favoring` vs. `L_TypeOf_Q && !reference-favoring`. What you have is still an identity-less reference type - so why should it be important whether you get a redundant projection or not (which might even be disabled at some point during type checking) ?

So, back to your question, I guess I don't see why you have defined the current behavior as being "natural". To me, getting an X projection on a X-favoring primitive class is an idempotent operation, which you can repeat even 20 times, and the type-system shouldn't care. Are there places in the code where you need this sharper distinction?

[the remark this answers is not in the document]
sadayapalam: > > when we have a type that is an L_TypeOf_Q it could be
> > either the (a) reference projection of value default primitive class or (b) the reference type of a reference default primitive class.
> 
> This seems at odds with what the Javadoc for that flavor says:
> 
> ```
> /**
>              * Reference projection type of a primitive-favoring aka primitive-default
>              * plain vanilla primitive class type,
>              */
>             L_TypeOf_Q,
> ```
> 
> E.g. javadoc suggests that Q means primitive-default already, so there should be no need to check the tsym?

Good catch! This is certainly a problem in the javadoc of L_TypeOf_Q and needs to be corrected. It was always the intent that L_TypeOf_Q is simply a primitive reference type.

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java
sadayapalam: > I mean - this code does a bunch of checks - and then creates a new class type with the opposite flavor/polarity. Isn't that what methods like `Type::asValueType` are supposed to do? E.g. if I have a reference-favoring primitive and I see `.val` can't I just get the type I want by calling `asValueType()` on it? If so, should we also have the dual?

OK, understand your point and agree that this could avoid some code duplication, Will follow up. Thanks!

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Type.java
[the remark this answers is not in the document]
sadayapalam: > More generally, I think I'm slightly uncomfortable by the use of the letters `Q` and `L` in the enum constants `X_TypeOf_Y`. Now, when `L` and `Q` are in the `X` position, the semantics is simple: `L` means `.ref`, `Q` means `.val`.
> 
> But when `L` and `Q` appear in the `Y` position, then it becomes messy - because the choice is no longer binary:
> 
> * the class is a primitive class or not (1 bit)
> * if primitive, the class is reference-favoring or not (another bit)
> 
> Right now, it seems that you use `Q` to denote whether the class is primitive class or not, which is useful, I assume, to detect distinction between a legacy reference type (`L_TypeOf_L`) and a reference projection of a primitive class (`L_TypeOf_Q`). This distinction is, I believe an important one, as the former has identity, the latter doesn't, so type checking would probably differ.

First of all, I agree that the javadoc of com.sun.tools.javac.code.Type.ClassType.Flavor#L_TypeOf_Q is incorrect and confusing and this has cascading ripple effects into other areas when it comes to comprehension.

So let me try to clarify the intent.

L_Typeof_Q is simply a primitive reference type - whose underlying primitive class type could either be reference favoring or value favoring.

With this clarification in place, if you look at the flavors, they naturally model the details associated with the domain.

There is inherent disparity/asymmetry in the concerned flavors in that - there is always a primitive reference type available for a primitive class (whether it is reference default or value default) but the reverse is not true bringing about the asymmetry. (ie you can have a reference type without there being a primitive class type i.e the L_Typeof_L legacy flavor)

> 
> What I'm less sure is whether you want/need different type checking rules for when `L_TypeOf_Q && reference-favoring` vs. `L_TypeOf_Q && !reference-favoring`. What you have is still an identity-less reference type - so why should it be important whether you get a redundant projection or not (which might even be disabled at some point during type checking) ?
> 
> So, back to your question, I guess I don't see why you have defined the current behavior as being "natural". To me, getting an X projection on a X-favoring primitive class is an idempotent operation, which you can repeat even 20 times, and the type-system shouldn't care. Are there places in the code where you need this sharper distinction?

Let me punt the question by saying - with the admission that javadoc of L_Typeof_Q was wrong and confusing, do we agree that once that is corrected, the flavors capture and model the domain information faithfully ? I don't off the top of the head recall places where this sharper distinction is needed - but if there is consensus that the model is valid even if a bit baroque, then that is reason to preserve it till we have more experience to definitively prune what we think is valid but superfluous information captured in the abstraction.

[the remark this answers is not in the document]
mcimadamore: > Let me punt the question by saying - with the admission that javadoc of L_Typeof_Q was wrong and confusing, do we agree that once that is corrected, the flavors capture and model the domain information faithfully ? I don't off the top of the head recall places where this sharper distinction is needed - but if there is consensus that the model is valid even if a bit baroque, then that is reason to preserve it till we have more experience to definitively prune what we think is valid but superfluous information captured in the abstraction.

While the javadoc is the main issue, the current behavior or "isReferenceProjection" and "isValueProjection" remind me of `2 + 0 + 0 != 2`, so I find it confusing.

I think perhaps the name "projection" is also misleading - as now we have primitive reference type and primitive value type, and what you really want to ask a type is in which of these two bucket it falls in. Giving too much importance as to "how" a type has been derived (e.g. redundant `.val` or `.ref` ?) smells of over-specificity which will bite us back, I believe.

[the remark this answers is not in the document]
sadayapalam: > > Let me punt the question by saying - with the admission that javadoc of L_Typeof_Q was wrong and confusing, do we agree that once that is corrected, the flavors capture and model the domain information faithfully ? I don't off the top of the head recall places where this sharper distinction is needed - but if there is consensus that the model is valid even if a bit baroque, then that is reason to preserve it till we have more experience to definitively prune what we think is valid but superfluous information captured in the abstraction.
> 
> While the javadoc is the main issue, the current behavior or "isReferenceProjection" and "isValueProjection" remind me of `2 + 0 + 0 != 2`, so I find it confusing.
> 
> I think perhaps the name "projection" is also misleading - as now we have primitive reference type and primitive value type, and what you really want to ask a type is in which of these two bucket it falls in. Giving too much importance as to "how" a type has been derived (e.g. redundant `.val` or `.ref` ?) smells of over-specificity which will bite us back, I believe.

Alright, I will fix the javadoc issue here and carry over the crux of the comment to JDK-8268734 to be addressed comprehensively there.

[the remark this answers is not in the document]
sadayapalam: > > Let me punt the question by saying - with the admission that javadoc of L_Typeof_Q was wrong and confusing, do we agree that once that is corrected, the flavors capture and model the domain information faithfully ? I don't off the top of the head recall places where this sharper distinction is needed - but if there is consensus that the model is valid even if a bit baroque, then that is reason to preserve it till we have more experience to definitively prune what we think is valid but superfluous information captured in the abstraction.
> 
> While the javadoc is the main issue, the current behavior or "isReferenceProjection" and "isValueProjection" remind me of `2 + 0 + 0 != 2`, so I find it confusing.
> 
> I think perhaps the name "projection" is also misleading - as now we have primitive reference type and primitive value type, and what you really want to ask a type is in which of these two bucket it falls in. Giving too much importance as to "how" a type has been derived (e.g. redundant `.val` or `.ref` ?) smells of over-specificity which will bite us back, I believe.

In the course of discussing further to understand this point, it has been evident that the javadoc of com.sun.tools.javac.code.Type#isReferenceProjection mentioning
```
Or in other words, is this a class type non-redundantly notated with .ref ??
```

(and likewise for isValueProjection())

is not helpful in one bit and has been a source of confusion. I will clean up that part. That part of the wording tries to be helpful, but backfires perhaps

sadayapalam:
Thanks Maurizio!

The confusing and unhelpful javadoc issues have been fixed.

In JDK-8268734 ("Various internal Type/Symbol APIs need rationalization/adjustment") I will follow up with a comprehensive review of all the Type and Symbol APIs - these APIs were rolled up as and when needed at widely different points in time 
and it is an useful exercise to take a step back and review them in a holistic fashion.

In JDK-8270882 (" Loose ends in reference default primitive class implementation.") I will follow up on the various known
issues that have been deliberately deferred to a later pass so as to make it more easily manageable,

PR #521, opened 2021-08-04
universal type variables: initial prototype

vicente-romero-oracle:
First iteration of the protype for universal type variables

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/valhalla pull/521/head:pull/521` \
`$ git checkout pull/521`

`$ git checkout pull/521` \
`$ git pull https://git.openjdk.java.net/valhalla pull/521/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 521`

`$ git pr show -t 521`

</details>

<a href="https://git.openjdk.java.net/valhalla/pull/521.diff">https://git.openjdk.java.net/valhalla/pull/521.diff</a>

</details>


src/java.base/share/classes/jdk/internal/javac/PreviewFeature.java
```diff
@@ -61,6 +61,7 @@
 
     public enum Feature {
         SWITCH_PATTERN_MATCHING,
+        UNIVERSAL_TVARS,
```
mcimadamore, line 64: As Dan noted yesterday, for other Valhalla features we do not generally use preview mechanisms (this code needs to be pushed in the valhalla repo). We can postpone the preview dance for later. In fact, to ease experiments in JDK code it would be better to just make universal type variable syntax available by default.

There are some other places where this UNIVERSAL_TVARS constant is used which should be cleaned up for now (e.g. Preview.java).

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Type.java
```diff
@@ -1943,12 +1943,34 @@ public TypeKind getKind() {
          */
         public Type lower;
 
+        public boolean universal = false;
```
mcimadamore, line 1946: Can we use a flag on the tsym for this? After all, being "universal" is a declaration (symbol) not a use (type) property?

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Type.java
```diff
@@ -2021,6 +2049,26 @@ public boolean isNullOrReference() {
         public <R, P> R accept(TypeVisitor<R, P> v, P p) {
             return v.visitTypeVariable(this, p);
         }
+
+        public boolean isUniversal() {
+            return universal;
+        }
+
+        @Override
+        public Type withTypeVar(Type t) {
```
mcimadamore, line 2058: Uhm - isn't this method used on wildcard types, and it all other types is supposed to just return `this` ?

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Type.java
```diff
@@ -2040,6 +2088,8 @@ public CapturedType(Name name,
             this.lower = Assert.checkNonNull(lower);
             this.setUpperBound(upper);
             this.wildcard = wildcard;
+            this.universal = upper.hasTag(TYPEVAR) && ((TypeVar)upper).universal ||
```
mcimadamore, line 2091: This probably requires tweaking - I see two issues:
* right now it detects cases where upper/lower are type variables themselves - but doesn't detect cases originating from e.g. `List<? extends Point>` - where you just have a captured variable whose bound is `Point`.
* the logic does not recurse - but maybe that's ok if `universal` is always set on creation

I think another approach could be to look at the wildcard type associated with the captured type. The wildcard type keeps track of the declared type variable it comes from, which means you can then look at the type symbol in there and decide if it's an universal variable or not. E.g.

```
captured.isUniversal() == captured.wildcard.bound.isUniversal()
```

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
```diff
@@ -1051,6 +1086,10 @@ private boolean isSubtypeUncheckedInternal(Type t, Type s, boolean capture, Warn
                 }
             } else if (isSubtype(t, s, capture)) {
                 return true;
+            } else if (allowUniversalTVars && t.hasTag(TYPEVAR) && s.hasTag(TYPEVAR) && t.tsym == s.tsym) {
```
mcimadamore, line 1089: Are you sure the check does what the comment says?

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
```diff
@@ -3555,6 +3612,13 @@ public Type visitTypeVar(TypeVar t, Void ignored) {
                 if (t.equalsIgnoreMetadata(from.head)) {
                     return to.head.withTypeVar(t);
                 }
+                if (allowUniversalTVars &&
```
mcimadamore, line 3615: Not sure about this - the main issue here seems that `t.equalsIgnoreMetadata` is too weak and would return `false` for `ref` vs.`val` mismatches. But again I'm uncertain about the need of overriding `withTypeVar` (see comment above).

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java
```diff
@@ -812,8 +819,12 @@ KindSelector attribArgs(KindSelector initialKind, List<JCExpression> trees, Env<
      *  Check that all the types are references.
      */
     List<Type> attribTypes(List<JCExpression> trees, Env<AttrContext> env) {
+        return attribTypes(trees, env, i -> false);
+    }
+
+    List<Type> attribTypes(List<JCExpression> trees, Env<AttrContext> env, Predicate<JCExpression> valueOK) {
```
mcimadamore, line 825: Do you really need a predicate here? Wouldn't a boolean be ok? Or perhaps a flag in env, or some other means?

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java
```diff
@@ -5104,9 +5144,22 @@ public void visitTypeApply(JCTypeApply tree) {
...
+                if (tv.universal) {
+                    valueOKSet.add(args.head);
+                }
+                args = args.tail;
+            }
+        }
 
         // Attribute type parameters
-        List<Type> actuals = attribTypes(tree.arguments, env);
+        List<Type> actuals = !allowUniversalTVars ?
```
mcimadamore, line 5160: If you are doing this, you might be better off just attributing types one at a time inside the loop above, and avoid using a set to collect the ones which need special treatment?

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Check.java
```diff
@@ -686,7 +704,7 @@ private boolean checkExtends(Type a, Type bound) {
              return true;
          } else if (!a.hasTag(WILDCARD)) {
              a = types.cvarUpperBound(a);
-             return types.isSubtype(a, bound);
+             return types.isBoundedBy(a, bound, (t, s, w) -> types.isSubtype(t, s));
```
mcimadamore, line 707: This is a good change - I'm a bit scared about the other calls in this routine, to `isCastable` and `notSoftSubtype`. These type relations are mostly doing things outside the spec, and we probably need to sprinkle `isBounded` in there as well.

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Resolve.java
```diff
@@ -1053,7 +1053,8 @@ public MethodCheckContext(boolean strict, DeferredAttrContext deferredAttrContex
         public boolean compatible(Type found, Type req, Warner warn) {
             InferenceContext inferenceContext = deferredAttrContext.inferenceContext;
             return strict ?
-                    types.isSubtypeUnchecked(inferenceContext.asUndetVar(found), inferenceContext.asUndetVar(req), warn) :
+                    types.isBoundedBy(inferenceContext.asUndetVar(found), inferenceContext.asUndetVar(req), warn,
```
mcimadamore, line 1056: I'm not 100% sure on this. I think step one should use just plain subtyping, while step 2 is allowed to use the more powerful relationship. My mental model is `int -> Object` which is only allowed on step 2 - so I think we want to do something similar for `Point -> Object` ? Of course `Point.ref -> Object` would be allowed on step 1.

mcimadamore (commented):
Overall looks like a solid first attempt. I think we should spend some cycles making sure we get the TypeVar model right, and that we put info (e.g. universality) where it belongs (e.g. symbol), because that can affect how clients interact with these variables quite a lot.

src/java.base/share/classes/jdk/internal/javac/PreviewFeature.java
vicente-romero-oracle: then I guess we will need to update the text in the JEP as the it is explicitly saying that this one is a preview feature

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java
vicente-romero-oracle: the thing is that one value only doesn't apply to all cases, what if some type variables are universal but others aren't?

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
vicente-romero-oracle: will remove the comment, probably a self comment I guess

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
vicente-romero-oracle: not sure I follow you here, `t.equalsIgnoreMetadata` is already being used in the method, not sure I see in what case it can give an incorrect answer

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Type.java
mcimadamore: yes, but `universal` is a property of the declaration, while `ref-ness` is a property of the use-site (hence type). Note that the underlying Valhalla compiler already has machinery to talk about reference projections, so perhaps we should use that to speak about `T` vs. `T.ref`, rather than inventing a new one. But we still need a flag somewhere to say whether the variable was declared as universal or not (as that affects what's compatible with that var).

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
mcimadamore: What I mean is that the newly added code seems to workaround the fact that in some cases we do not call `withTypeVar`. I'm saying that because you added another test which, if satisfied, ends up calling `to.head.withTypeVar(t);` again, which seems to suggest that you found cases where this call was needed, but was not triggered - probably because of a mismatch between T and T.ref?

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java
vicente-romero-oracle: it wont be that simple, see that the method invoked is doing:
```
    List<Type> types = attribAnyTypes(trees, env);
    return chk.checkRefTypes(trees, types, valueOK);
```

and chk.checkRefTypes is doing:
```
        List<JCExpression> tl = trees;
        for (List<Type> l = types; l.nonEmpty(); l = l.tail) {
            l.head = checkRefType(tl.head.pos(), l.head, valueOK.test(tl.head));
            tl = tl.tail;
        }
        return types;
```
and Check.checkRefType is the one using the flag, so in order to do all this at the level of `Attr.visitTypeApply` I will have to unfold all this code and then the code in `Attr.visitTypeApply` will lose readability. This is why I opted for this solution.

mcimadamore: I see what you mean - it seems like the current code is working against us here. In fact, attribAnyTypes specifically says that clients are expected to check for reference types after the fact...

Here's an idea... you could maybe call attribAnyTypes yourself, to attribute all the types, and then, in the loop you already have, only call checkRefTypes on the types you really need?

vicente-romero-oracle: OK

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java
```diff
@@ -5144,23 +5137,22 @@ public void visitTypeApply(JCTypeApply tree) {
...
+        } else {
+            // we use argTypes2 to keep a pointer to the original list as we will use argTypes to iterate over it
+            List<Type> argTypes, argTypes2;
+            argTypes2 = argTypes = attribAnyTypes(args, env);
             for (Type t : ((ClassType) clazztype.tsym.type).typarams_field) {
-                TypeVar tv = (TypeVar) t;
-                if (tv.universal) {
-                    valueOKSet.add(args.head);
-                }
+                argTypes.head = chk.checkRefType(args.head.pos(), argTypes.head, ((TypeVar) t).isValueProjection());
```
mcimadamore, line 5149: this should be `isUniversal()`, no?

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java
```diff
@@ -4603,15 +4597,14 @@ private Symbol selectSym(JCFieldAccess tree,
             case WILDCARD:
                 throw new AssertionError(tree);
             case TYPEVAR:
-                if (allowUniversalTVars && name == names.ref && ((TypeVar)site).universal) {
+                if (allowUniversalTVars && name == names.ref && ((TypeVar)site).isValueProjection()) {
```
mcimadamore, line 4600: This should be `isUniversal()` ?

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java
```diff
@@ -2666,7 +2660,7 @@ public void visitApply(JCMethodInvocation tree) {
                         && (tree.meth.hasTag(SELECT))
                         && ((JCFieldAccess)tree.meth).selected.hasTag(IDENT)
                         && TreeInfo.name(((JCFieldAccess)tree.meth).selected) == names._super;
-                boolean qualifierIsUniversal = allowUniversalTVars && qualifier.hasTag(TYPEVAR) && ((TypeVar)qualifier).universal;
+                boolean qualifierIsUniversal = allowUniversalTVars && qualifier.hasTag(TYPEVAR) && ((TypeVar)qualifier).isValueProjection();
```
mcimadamore, line 2663: This should be `isUniversal()` I think - `wait` and friends should warn, even on `.ref` stuff.

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
```diff
@@ -1185,7 +1183,7 @@ public Boolean visitType(Type t, Type s) {
                  case TYPEVAR:
                      return isSubtypeNoCapture(t.getUpperBound(), s);
                  case BOT:
-                     if (allowUniversalTVars && s.hasTag(TYPEVAR) && ((TypeVar)s).universal) {
+                     if (allowUniversalTVars && s.hasTag(TYPEVAR) && ((TypeVar)s).isValueProjection()) {
```
mcimadamore, line 1186: I agree that here you need `isValueProjection`

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
```diff
@@ -3613,10 +3611,10 @@ public Type visitTypeVar(TypeVar t, Void ignored) {
                     return to.head.withTypeVar(t);
                 }
                 if (allowUniversalTVars &&
-                        t.createdFromUniversalTypeVar &&
+                        !t.isValueProjection() &&
                         from.head.hasTag(TYPEVAR) &&
-                        ((TypeVar)from.head).referenceProjection != null &&
-                        t.equalsIgnoreMetadata(((TypeVar)from.head).referenceProjection)) {
+                        ((TypeVar)from.head).projection != null &&
```
mcimadamore, line 3616: Shouldn't this test be `isUniversal()` ?

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
```diff
@@ -3613,10 +3611,10 @@ public Type visitTypeVar(TypeVar t, Void ignored) {
...
+                        t.equalsIgnoreMetadata(((TypeVar)from.head).projection)) {
```
mcimadamore, line 3617: and here, we should call `referencceProjection()` ?

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
```diff
@@ -3613,10 +3611,10 @@ public Type visitTypeVar(TypeVar t, Void ignored) {
                     return to.head.withTypeVar(t);
                 }
                 if (allowUniversalTVars &&
-                        t.createdFromUniversalTypeVar &&
+                        !t.isValueProjection() &&
```
mcimadamore, line 3614: Not sure about this here - "not value projection might also mean it's a regular tvar". Doesn't this test want to check for `T.ref` (in which case you should check for `isReferenceProjection`

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Type.java
```diff
@@ -1943,12 +1943,30 @@ public TypeKind getKind() {
...
+        public boolean isReferenceProjection;
+
+        // redundant for now but helpful for debug reasons
+        public boolean isUniversal;
+
         public TypeVar(Name name, Symbol owner, Type lower) {
+            this(name, owner, lower, false, false);
+        }
+
+        public TypeVar(Name name, Symbol owner, Type lower, boolean isUniversal, boolean isReferenceProjection) {
```
mcimadamore, line 1962: This constructor probably should not take the `isReferenceProjection` as this one creates a new tsym and you want the reference projection to share the same symbol - so you probably are always passing `false` here.

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java
vicente-romero-oracle: yep will change it

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java
vicente-romero-oracle: yep u are right

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java
vicente-romero-oracle: not sure to be honest, although there is no reference in the JEP to these warnings so we can do either now

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
vicente-romero-oracle: yep I agree

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
vicente-romero-oracle: doing this change right now breaks the build, I think it is because we generate reference projections for universal type vars in a lazy way, so some universal type variables could have its projection field set to null

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java
mcimadamore: I double checked with Brian yesterday - he confirmed that both `T` and `T.ref` should not support identity operation (well, given warning when attempting to).

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
mcimadamore: ugh - I understand - moving forward it'd be nice not to depend on tricky init semantics.

vicente-romero-oracle: yep, will fix this

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java
vicente-romero-oracle: ok, thanks for double checking, will fix this

forax: T.ref still support == null, but it's a corner case

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Type.java
```diff
@@ -1986,6 +1988,9 @@ public TypeVar(TypeSymbol tsym, Type bound, Type lower,
             this.lower = lower;
             this.isReferenceProjection = isReferenceProjection;
             this.isUniversal = (tsym.flags_field & UNIVERSAL) != 0;
+            if (isUniversal && !isReferenceProjection) {
```
mcimadamore, line 1991: isn't `isReferenceProjection` always `false` here (you have just created the class)

mcimadamore (approved):
Looks good - just one nitpick

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Type.java
vicente-romero-oracle: no, because this constructor is used again to build the reference, so I have to add it to the condition to avoid stackoverflow errors

vicente-romero-oracle:
thanks for the review

PR #525, opened 2021-08-06
8272041: [lworld] CDS heap dump fails with primitive objects

iklam:
Two simple fixes to handle primitive objects when dumping the CDS heap:

- When dumping the objects, we enter them into a hashtable. Do not use `oopDesc::identity_hash()` as the hashcode for this table.
- When resetting the header of archived objects, do not call `oopDesc::identity_hash()` on primitive objects.

---------
### Progress

### Issue
 * [JDK-8272041](https://bugs.openjdk.java.net/browse/JDK-8272041): [lworld] CDS heap dump fails with primitive objects

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/valhalla pull/525/head:pull/525` \
`$ git checkout pull/525`

`$ git checkout pull/525` \
`$ git pull https://git.openjdk.java.net/valhalla pull/525/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 525`

`$ git pr show -t 525`

</details>

<a href="https://git.openjdk.java.net/valhalla/pull/525.diff">https://git.openjdk.java.net/valhalla/pull/525.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/cds/heapShared.cpp
+++ b/src/hotspot/share/cds/heapShared.cpp
@@ -143,10 +143,11 @@ void HeapShared::fixup_mapped_regions() {
 }
 
 unsigned HeapShared::oop_hash(oop const& p) {
-  assert(!EnableValhalla || !p->mark().is_inline_type(),
-         "this object should never have been locked");
-  unsigned hash = (unsigned)p->identity_hash();
-  return hash;
+  // We are at a safepoint, so the object won't move. It's OK to use its
+  // address as the hashcode.
+  // We can't use p->identity_hash() as it's not available for primitive oops.
+  assert_at_safepoint();
+  return (unsigned)(p2i(p) >> LogBytesPerWord);
 }
 
 static void reset_states(oop obj, TRAPS) {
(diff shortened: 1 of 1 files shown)


fparain (approved):
Looks good to me.

Just a question by curiosity, using object addresses instead of the identity hash code is likely to change the distribution of keys for the hashtable (all objects being in the same memory area and being aligned on heap words, some bits in their addresses are the same for all objects). Is it going to have a significant impact of the performance of the hashtable or is it negligeable?

Thank you,

Fred

mlchung (approved):
Looks good.  I verified with the patch that converts some JDK classes to ref-default primitive class.

iklam:
> Looks good to me.
> 
> Just a question by curiosity, using object addresses instead of the identity hash code is likely to change the distribution of keys for the hashtable (all objects being in the same memory area and being aligned on heap words, some bits in their addresses are the same for all objects). Is it going to have a significant impact of the performance of the hashtable or is it negligeable?

Hi Fred,

Thanks for the review. I've added code to calculate the stats of the hashtable. The max number of entries per bucket is no more than 3, so I think the distribution is acceptable.

Also, the hash function already shifts down by `LogBytesPerWord`, so it will get rid of the lower 3 zero bits.

My stats code is in https://github.com/iklam/valhalla/commits/heap-shared-seen-obj-table-stats, but it's not part of this PR. I will clean it up and integrate into jdk/jdk.

iklam:
Thanks @fparain and @mlchung for the review.
/integrate

fparain:
Ioi,

Thank you for the additional effort to measure statistics of the hashtable.

Fred

PR #534, opened 2021-08-18
8272566: [lworld] [AArch64] G1 write barrier slow path needs to save more registers

nick-arm:
We call MacroAssembler::store_heap_oop() when packing inline type
arguments in C1's scalarised entry point and that in turn uses the
interpreter's barrier set assembler for the write barrier.  With G1GC
this can call into the runtime on the slow path, but it doesn't preserve
the call-clobbered registers, which include the Java argument registers.

This was fixed on x86 in JDK-8242210 and JDK-8251398 but I missed that
when I updated the AArch64 port earlier in the year.  I haven't actually
seen any failures caused by this, but from looking at the generated
assembly it's certainly possible.

This patch saves only the argument registers, and only if Valhalla is
enabled (unlike x86 which saves all registers unconditionally).  There
are 32 caller-saved registers on AArch64 so I wanted to avoid generating
many additional store/load instructions.

---------
### Progress

### Issue
 * [JDK-8272566](https://bugs.openjdk.java.net/browse/JDK-8272566): [lworld] [AArch64] G1 write barrier slow path needs to save more registers

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/valhalla pull/534/head:pull/534` \
`$ git checkout pull/534`

`$ git checkout pull/534` \
`$ git pull https://git.openjdk.java.net/valhalla pull/534/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 534`

`$ git pr show -t 534`

</details>

<a href="https://git.openjdk.java.net/valhalla/pull/534.diff">https://git.openjdk.java.net/valhalla/pull/534.diff</a>

</details>


TobiHartmann (approved):
Looks good to me.

PR #541, opened 2021-08-31
8273018: [lworld] Property annotation propagation to <init> lacks in primitive records

jespersm:
This fix copies parameter names and annotations from constructor into primitive class factory.

---------
### Progress

### Issues
 * [JDK-8273018](https://bugs.openjdk.java.net/browse/JDK-8273018): [lworld] Property annotation propagation to <init> lacks in primitive records
 * [JDK-8273202](https://bugs.openjdk.java.net/browse/JDK-8273202): [lworld] MethodParameters are not generated for factories for primitive records

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/valhalla pull/541/head:pull/541` \
`$ git checkout pull/541`

`$ git checkout pull/541` \
`$ git pull https://git.openjdk.java.net/valhalla pull/541/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 541`

`$ git pr show -t 541`

</details>

<a href="https://git.openjdk.java.net/valhalla/pull/541.diff">https://git.openjdk.java.net/valhalla/pull/541.diff</a>

</details>

### DIFF
--- a/src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/TransPrimitiveClass.java
+++ b/src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/TransPrimitiveClass.java
@@ -389,7 +389,7 @@ public class TransPrimitiveClass extends TreeTranslator {
         if (factory != null)
             return factory;
 
-        MethodType factoryType = new MethodType(init.erasure(types).getParameterTypes(),
+        MethodType factoryType = new MethodType(init.type.getParameterTypes(),
                                                 init.owner.type.asValueType(),
                                                 init.type.getThrownTypes(),
                                                 init.owner.type.tsym);
@@ -398,6 +398,11 @@ public class TransPrimitiveClass extends TreeTranslator {
                                         factoryType,
                                         init.owner);
         factory.params = init.params;
+        // Re-patch the return type on the erased method type, or code generation will fail
+        factory.erasure_field = new MethodType(init.erasure(types).getParameterTypes(),
+                init.owner.type.asValueType(),
+                init.type.getThrownTypes(),
+                init.owner.type.tsym);
         factory.setAttributes(init);
         init2factory.put(init, factory);
         return factory;
--- a/src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/TransPrimitiveClass.java
+++ b/src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/TransPrimitiveClass.java
@@ -397,6 +397,7 @@ public class TransPrimitiveClass extends TreeTranslator {
                                         names.init,
                                         factoryType,
                                         init.owner);
+        factory.params = init.params;
         factory.setAttributes(init);
         init2factory.put(init, factory);
         return factory;
--- a/src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/ClassWriter.java
+++ b/src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/ClassWriter.java
@@ -1015,7 +1015,10 @@ public class ClassWriter extends ClassFile {
             endAttr(alenIdx);
             acount++;
         }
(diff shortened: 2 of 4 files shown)


sadayapalam:
The title of the PR seems to be incorrect ??

jespersm:
Fixed name of PR. The name of the issue is slightly off, I guess it should be ‘parameter annotation’, really.

jespersm:
I combined this fix (which was incomplete, since it "de-erased" some types incorrectly) with the fix for the problem in JDK-8273202), and a more useful example of incorrect compilation of primitive records.

src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/TransPrimitiveClass.java
```diff
@@ -389,14 +389,20 @@ private MethodSymbol getPrimitiveObjectFactory(MethodSymbol init) {
         if (factory != null)
             return factory;
 
-        MethodType factoryType = new MethodType(init.erasure(types).getParameterTypes(),
+        MethodType factoryType = new MethodType(init.type.getParameterTypes(),
                                                 init.owner.type.asValueType(),
                                                 init.type.getThrownTypes(),
```
sadayapalam, line 394: I am lost and need a bit of help here. What exactly is the change and why is it needed ?

src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/TransPrimitiveClass.java
```diff
@@ -389,14 +389,20 @@ private MethodSymbol getPrimitiveObjectFactory(MethodSymbol init) {
...
                                         names.init,
                                         factoryType,
                                         init.owner);
+        factory.params = init.params;
+        // Re-patch the return type on the erased method type, or code generation will fail
+        factory.erasure_field = new MethodType(init.erasure(types).getParameterTypes(),
+                init.owner.type.asValueType(),
+                init.type.getThrownTypes(),
+                init.owner.type.tsym);
         factory.setAttributes(init);
```
sadayapalam, line 406: Again, I need some help with this change. I understand the part of parameters being carried over, but the dance around recomputation of erasure field - what exactly is the material change there ?? I can see the RecordReading test failing if I backout the changes here (except for factory.params = init.params) but need some explanation - TIA

sadayapalam (changes requested):

sadayapalam:
If you want to close more than issue with the same PR, I think you need to add the other issue here with /issue 

Also please mention what tests have been run. Thanks

src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/TransPrimitiveClass.java
jespersm: This method makes a primitive object factory from the parsed constructor. In the current version, the parameter types are erased and their annotations, and original method signature, are erased from the original. This line preserves this information, so the generated classfile is correct.

See next comment.

src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/TransPrimitiveClass.java
jespersm: Related to comment above: Since we're now preserving the full type information for `factoryType` and thus `factory`, instead of propagating the erased info, information from the other (preceding) lowerings would get lost. To avoid that, the erasure is explicitly reconstructed, but with a fixed return type (instead of void).

There is a different test which blows up if this isn't reconstructed, it might be `InnerValueNew.java` (I'll check later).

sadayapalam:
OK, Thanks for the clarifications. If the suggested changes are made, we can take this forward.

jespersm:
@sadayapalam: I've now incorporated the requested review changes.

I've run the tier1 tests, all is green here.

sadayapalam (approved):
Looks good, please proceed to integrate, Thanks so much Jesper

PR #542, opened 2021-08-31
8273202: Emit parameter names for primitive record factories

jespersm:
8273202: Ensure that parameter primitive record factories are generated.

Note that this fix requires JDK-8273018 (PR #541).


jespersm:
I'm folding this change into PR #541, closing this one.

PR #551, opened 2021-09-10
8273594: [lworld] JITs need to properly handle static inline type field with unloaded type

TobiHartmann:
Both C1 and C2 do not properly handle loads from static inline type fields with an unloaded type. For C1, the fix is to simply remove two asserts that are too strong. For C2, we need to trap during typeflow analysis.

I've added a corresponding test to `TestUnloadedInlineTypeField.java` and noticed that the new IR Test Framework often triggers class loading while the test was carefully designed to avoid that. As a workaround, I slightly modified the framework and run the test with `-DIgnoreCompilerControls=true`. I filed [JDK-8273591](https://bugs.openjdk.java.net/browse/JDK-8273591) to fix this upstream.

I also fixed a `-XX:-XX:+PatchALot` typo in the test that went unnoticed because `-XX:+IgnoreUnrecognizedVMOptions` is set.

Thanks,
Tobias

### DIFF
--- a/src/hotspot/share/c1/c1_LIRGenerator.cpp
+++ b/src/hotspot/share/c1/c1_LIRGenerator.cpp
@@ -1649,7 +1649,7 @@ void LIRGenerator::do_StoreField(StoreField* x) {
   }
 #endif
 
-  if (!inline_type_field_access_prolog(x, info)) {
+  if (!inline_type_field_access_prolog(x)) {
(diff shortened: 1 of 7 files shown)


TobiHartmann:
I found more issues with uninitialized classes and fixed them. The new tests currently trigger [JDK-8273650](https://bugs.openjdk.java.net/browse/JDK-8273650), so I'll wait with integration until that is fixed.

PR #560, opened 2021-10-13
8275195: [lworld] Revisit use of TypePtr::meet_aryptr() after merge

rwestrel:
I tried to move the logic that I added to TypeAryKlassPtr into
meet_aryptr() where it's shared between TypeAryPtr and
TypeAryKlassPtr. When running tests, I ran into a failure that I fixed
by making the element klass of an array of nullable inline types
to be non constant similar to this logic:

    // Even if MyValue is exact, [LMyValue is not exact due to [QMyValue <: [LMyValue.
    bool xk = etype->klass_is_exact() && (!etype->is_inlinetypeptr() || null_free);

in TypeOopPtr::make_from_klass_common().

Actually, shouldn't the element type be non exact in the TypeAryPtr
case too?

### DIFF
--- a/src/hotspot/share/opto/type.cpp
+++ b/src/hotspot/share/opto/type.cpp
@@ -4839,12 +4839,12 @@ const Type *TypeAryPtr::xmeet_helper(const Type *t) const {
     bool res_xk = false;
     bool res_not_flat = false;
     bool res_not_null_free = false;
-    const Type* elem = tary->_elem;
-    if (meet_aryptr(ptr, elem, this->klass(), tap->klass(),
+    const Type* res_elem = NULL;
+    if (meet_aryptr(ptr, _ary->_elem, tap->_ary->_elem, this->klass(), tap->klass(),
                     this->klass_is_exact(), tap->klass_is_exact(), this->ptr(), tap->ptr(),
                     this->is_not_flat(), tap->is_not_flat(),
                     this->is_not_null_free(), tap->is_not_null_free(),
-                    res_klass, res_xk, res_not_flat, res_not_null_free) == NOT_SUBTYPE) {
+                    res_elem, res_klass, res_xk, res_not_flat, res_not_null_free) == NOT_SUBTYPE) {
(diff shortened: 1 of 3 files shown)


src/hotspot/share/opto/type.cpp
```diff
@@ -5934,6 +5944,11 @@ const TypeAryKlassPtr *TypeAryKlassPtr::make(PTR ptr, ciKlass* klass, Offset off
     // Element is an object array. Recursively call ourself.
     ciKlass* eklass = klass->as_obj_array_klass()->element_klass();
     const TypeKlassPtr *etype = TypeKlassPtr::make(eklass)->cast_to_exactness(false);
+
+    if (etype->klass_is_exact() && etype->isa_instklassptr() && etype->is_instklassptr()->klass()->is_inlinetype() && !null_free) {
```
TobiHartmann, line 5948: Please add a comment similar to the other places where we do this.

TobiHartmann:
> Do you want me to take care of it in this change?

Yes, that would be great.

> is that not correct?

I'm a bit confused by this logic:
https://github.com/openjdk/valhalla/blob/cb080234f25f77f24c67d5f62b89a7b0908bf097/src/hotspot/share/opto/type.cpp#L5954-L5964

 `not_flat` is supposed to be true iff we know that the array can never be flat. However, it's initialized with `!k->is_flat_array_klass()` which does **not** mean that the array can never be flat. It could be an object array which can be flat if the element type is not exact. Also not null-free should imply not flat. In general, we can deduce that an array is not flat if (1) `!UseFlatArray`, (2) the array is not null-free or (3) the element type is an inline type that is not flattened in arrays.

Or am I missing something?

I think this logic should be orthogonal to this:
https://github.com/openjdk/valhalla/blob/cb080234f25f77f24c67d5f62b89a7b0908bf097/src/hotspot/share/opto/type.cpp#L3607-L3611

Also, can we use an enum for `TypeAryKlassPtr::_null_free`? The tri-state via an int is confusing as well.

PR #562, opened 2021-10-14
8274131: [lworld] [aarch64] ClassInitializationFailuresTest.java stop: klass not initialized

nick-arm:
This check was removed on x86 as part of JDK-8273650.


nick-arm:
Thanks @TobiHartmann !

PR #563, opened 2021-10-14
8274972: [lworld] TestLWorld.test151() fails with IR verification errors

rwestrel:
The test failure is caused by this check:
bool LateInlineVirtualCallGenerator::do_late_inline_check(Compile* C, JVMState* jvms) {
  // Method handle linker case is handled in CallDynamicJavaNode::Ideal().
  // Unless inlining is performed, _override_symbolic_info bit will be set in DirectCallGenerator::generate().
  // Implicit receiver null checks introduce problems when exception states are combined.
  Node* receiver = jvms->map()->argument(jvms, 0);
  const Type* recv_type = C->initial_gvn()->type(receiver);
  if (recv_type->maybe_null()) {
    return false;
  }
  
That code came with the recent merge. receiver should be not null
(it's a InlineTypePtrNode) but it's not. The InlineTypePtrNode is
created when pushed down through Phis. The fix I propose is to set the
type of the phi that's the Oop input to InlineTypeBase to non null
when it's created if it's observed that all InlineTypeBase nodes
encountered when following Phi inputs are non null.

---------
### Progress

### Issue
 * [JDK-8274972](https://bugs.openjdk.java.net/browse/JDK-8274972): [lworld] TestLWorld.test151() fails with IR verification errors

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/valhalla pull/563/head:pull/563` \
`$ git checkout pull/563`

`$ git checkout pull/563` \
`$ git pull https://git.openjdk.java.net/valhalla pull/563/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 563`

`$ git pr show -t 563`

</details>

<a href="https://git.openjdk.java.net/valhalla/pull/563.diff">https://git.openjdk.java.net/valhalla/pull/563.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/opto/cfgnode.cpp
+++ b/src/hotspot/share/opto/cfgnode.cpp
@@ -1913,12 +1913,12 @@ bool PhiNode::wait_for_region_igvn(PhaseGVN* phase) {
 }
 
 // Push inline type input nodes (and null) down through the phi recursively (can handle data loops).
-InlineTypeBaseNode* PhiNode::push_inline_types_through(PhaseGVN* phase, bool can_reshape, ciInlineKlass* vk, bool not_null) {
+InlineTypeBaseNode* PhiNode::push_inline_types_through(PhaseGVN* phase, bool can_reshape, ciInlineKlass* vk, bool is_init) {
   InlineTypeBaseNode* vt = NULL;
   if (_type->isa_ptr()) {
-    vt = InlineTypePtrNode::make_null(*phase, vk)->clone_with_phis(phase, in(0), not_null);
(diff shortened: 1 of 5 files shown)


TobiHartmann (approved):
Looks good. Just wondering, why does IGVN not update the type of the Phi accordingly?

Please re-enable the test, I've just disabled it with https://github.com/openjdk/valhalla/commit/68265657bf1901965fafba3c258d39d202e1941a#diff-4d835d6d5e9844534a306405cf02e01b816b80a13a2e911d2f37fd38ecb2ac69R4123

TobiHartmann (approved):
Looks good to me. Please file a bug for enabling the assert.

PR #593, opened 2022-01-06
8279428: [lworld] Revalue ACC_PRIMITIVE to be 0x800 to align with draft JVMS

sadayapalam:
Change ACC_PRIMITIVE to be 0x800 across components sources and tests

---------
### Progress

### Issue
 * [JDK-8279428](https://bugs.openjdk.java.net/browse/JDK-8279428): [lworld] Revalue ACC_PRIMITIVE to be 0x800 to align with draft JVMS

### Reviewers
 * [Tobias Hartmann](https://openjdk.java.net/census#thartmann) (@TobiHartmann - Committer) ⚠️ Review applies to accb6894a811a53eed14ca990395f2ea3c2c2e0c
 * [Frederic Parain](https://openjdk.java.net/census#fparain) (@fparain - Committer) ⚠️ Review applies to accb6894a811a53eed14ca990395f2ea3c2c2e0c
 * [Roger Riggs](https://openjdk.java.net/census#rriggs) (@RogerRiggs - Committer) ⚠️ Review applies to accb6894a811a53eed14ca990395f2ea3c2c2e0c

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/valhalla pull/593/head:pull/593` \
`$ git checkout pull/593`

`$ git checkout pull/593` \
`$ git pull https://git.openjdk.java.net/valhalla pull/593/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 593`

`$ git pr show -t 593`

</details>

<a href="https://git.openjdk.java.net/valhalla/pull/593.diff">https://git.openjdk.java.net/valhalla/pull/593.diff</a>

</details>


TobiHartmann (approved):
Changes to the JIT test look good.

fparain (approved):
HotSpot changes look good to me.

Fred

RogerRiggs (approved):
LGTM for libraries changes and jdeps

src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/ClassWriter.java
```diff
@@ -1546,9 +1545,9 @@ public void writeClassFile(OutputStream out, ClassSymbol c)
         if (c.owner.kind == MDL) {
             flags = ACC_MODULE;
         } else {
-            flags = adjustFlags(c.flags() & ~DEFAULT);
+            flags = adjustFlags(c.flags() & ~(DEFAULT | STRICTFP));
             if ((flags & PROTECTED) != 0) flags |= PUBLIC;
-            flags = flags & (ClassFlags | ACC_PRIMITIVE) & ~STRICTFP;
```
biboudis, line 1551: As far as I can understand, `ACC_PRIMITIVE` was removed from here because it is driven by the `PRIMITIVE_CLASS` in `adjustFlags` which makes the code clearer at this point. Is my assumption correct?

sadayapalam: Good question! Thanks for asking. 

Indeed there is a "problem" here - actually two problems, one old and one new. I put problem in quotes because, the bug is only latent at the moment and does not impact anything. but is nevertheless a bug and should be fixed.

Explanation. First the old bug:

In Flags.java we have the following:

```
LocalClassFlags                   = FINAL | ABSTRACT | STRICTFP | ENUM | SYNTHETIC  | PRIMITIVE_CLASS,
ClassFlags                        = LocalClassFlags | INTERFACE | PUBLIC | ANNOTATION,
``` 

This is problematic because PRIMITIVE_CLASS is defined to be 1 << 16 - while the class file representation of flags is only 16 bits. Which is why there is a mapping from PRIMITIVE_CLASS to ACC_PRIMITIVE in the first place.

So the definition of LocalClassFlags should have had the bit corresponding to 0x100 ie  (1 << 8) set by including a suitable symbolic constant in place of PRIMITIVE_CLASS.

The idea behind doing
```
flags = flags & ClassFlags
```

in the old (and new code) is ensure that no bit other than well defined class flags bits are set in the flags being written out to file. By not having 0x100 (old ACC_PRIMITIVE) set in ClassFlags we are inadvertently masking ACC_PRIMITIVE bit and to avoid running into problems there, we or back ACC_PRIMITIVE into the ClassFlags just being &ing.

New problem:

In the new code, again LocalClassFlags is missing ACC_PRIMITIVE, but things work just fine but by accident - only because ACC_PRIMITIVE shares the same value with ACC_STRICT !

I'll post a revision that will make this clear and address the latent bug (which may become real bug if ACC_PRIMITIVE changes value again to have a value that is not already a known bit pattern in LocalClassFlags) 

The right

biboudis: Thank you for the detailed explanation.

> In the new code, again LocalClassFlags is missing ACC_PRIMITIVE, but things work just fine but by accident - only because ACC_PRIMITIVE shares the same value with ACC_STRICT !

I am still a bit confused because I don't see any uses of ACC_STRICT but only STRICT_FP at that level (which is indeed 1<<11, 0x0800). So in my mind it shares the same value with STRICT_FP not ACC_STRICT. I will fully get it with your upcoming revision 👍 💯 

Thanks again for the explanation!

sadayapalam: But ACC_STRICT and STRICT_FP symbolize/represent the same and carry the same value. (The latter is javac's internal symbol.)

biboudis: Got it, right!

sadayapalam: OK, here is an incremental fix that addresses this concern. Some notes.

- First of all, we cant define PRIMITIVE_CLASS to be 0x800 because this bit pattern is already claimed by STRICT_FP (javac speak for ACC_STRICT which is class file speak)
- ACC_STRICT technically can never be applied to a class in the class file although in javac internal ASTs a class can be marked STRICT_FP to indicate that strictness in FP operations should be propagated to methods of the class. So there is a clash and we choose a different bit encoding.
- A note about naming. I chose PRIMITIVE_CLASS rather than PRIMITIVE to avoid confusion with basic/builtin primitives.
- Now PRIMITIVE_CLASS (javac internal symbol/encoding) needs to be mapped to ACC_PRIMITIVE (JVMS symbol/encoding) at the time of class file writing. This happens in com.sun.tools.javac.jvm.ClassWriter#adjustFlags
- According to JVMS Table 4.1-B. Class access and property modifiers and Table 4.7.6-A. Nested class access and property flags ACC_STRICT is not a valid flag on a class file and so ....
- JDK-7165659 (javac incorrectly sets strictfp access flag on inner-classes) should have arguably cleared STRICT_FP from LocalClassFlags in Flags.java but chose to clear it at the time of class file writing.
So the present update

- Removes STRICT_FP from LocalClassFlags
- Remove PRIMITIVE_CLASS from LocalClassFlags - Class flags are only 16 bits and PRIMITIVE_CLASS is 1 << 16 and so can't really fit in the class file flag and also needs mapping to ACC_PRIMITIVE. It was a bug to include it in the first place.
- Add ACC_PRIMITIVE to LocalClassFlags since that is valid with Valhalla
(Note LocalClassFlags is used as the base to subsequently define many other symbolic constants)

- Update the series of Extended*Flags constants to include PRIMITIVE_CLASS since the latter is really an extended flag (beyond 16 bits)
- Introduce two new Extended*Flags symbols for which no need existed till now.

The one still open minor oddity is why in com.sun.tools.javac.jvm.ClassWriter#writeClassFile we would choose to do
```
flags = flags & ClassFlags;
```
while not doing something similar in com.sun.tools.javac.jvm.ClassWriter#writeInnerClasses

but one can convince himself that this is harmless and really a nop. So I have chosen to leave this alone (needs much more ceremony to fix it - we need to define additional constants)

Please take a look and let me know of any more questions

biboudis (approved):
This looks good from my side too and all the fixes are clear.

PR #602, opened 2022-01-13
8283573: [lworld] Make Object an abstract class

fparain:
---------
### Progress

### Issue
 * [JDK-8283573](https://bugs.openjdk.java.net/browse/JDK-8283573): [lworld] Make Object an abstract class

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/valhalla pull/602/head:pull/602` \
`$ git checkout pull/602`

`$ git checkout pull/602` \
`$ git pull https://git.openjdk.java.net/valhalla pull/602/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 602`

`$ git pr show -t 602`

</details>

<a href="https://git.openjdk.java.net/valhalla/pull/602.diff">https://git.openjdk.java.net/valhalla/pull/602.diff</a>

</details>


src/hotspot/share/c1/c1_GraphBuilder.cpp
```diff
@@ -2537,7 +2537,11 @@ void GraphBuilder::new_instance(int klass_index) {
   bool will_link;
   ciKlass* klass = stream()->get_klass(will_link);
   assert(klass->is_instance_klass(), "must be an instance klass");
-  NewInstance* new_instance = new NewInstance(klass->as_instance_klass(), state_before, stream()->is_unresolved_klass());
+  ciKlass* ik = klass;
+  if (ik == compilation()->env()->Object_klass() && compilation()->env()->Object_klass()->is_abstract()) {
```
MrSimms, line 2541: Does adding if EnableValhalla as the first condition, or a "// Valhalla" comment, make this clearer ? There are a few places in the change set with the same review comment, but I won't repeat.

MrSimms (commented):
The pattern:

`if (ik == vmClasses::Object_klass() && vmClasses::Object_klass()->is_abstract()) {
  ik = vmClasses::Identity_klass();
}`

Keep repeating itself, could this be an inline method somewhere, maybe `valhalla_check_replace_object_identity()` (i.e. self documenting)

fparain:
> The pattern:
> 
> `if (ik == vmClasses::Object_klass() && vmClasses::Object_klass()->is_abstract()) { ik = vmClasses::Identity_klass(); }`
> 
> Keep repeating itself, could this be an inline method somewhere, maybe `valhalla_check_replace_object_identity()` (i.e. self documenting)

A single method would not be possible, because some tests are using Klass* while others are using ciKlass*.

MrSimms:
> A single method would not be possible, because some tests are using Klass* while others are using ciKlass*.

Yeah it looks only 2 with ciKlass, hardly worth it

src/java.base/share/classes/java/lang/Identity.java
```diff
@@ -0,0 +1,38 @@
+/*
+ * Copyright (c) 2022, 2022, Oracle and/or its affiliates. All rights reserved.
```
TobiHartmann, line 2:
```suggestion
 * Copyright (c) 2022, Oracle and/or its affiliates. All rights reserved.
```

TobiHartmann: (This has been discussed in mainline before. If both years are equal, there should only be one)

TobiHartmann (approved):
Compiler changes look good to me.

fparain:
Last update addresses the copyright format issue. It also removes the EnableValhalla condition in the interpreter changes to allow the VM to be fully functional even with -XX:-EnableValhalla (we need this configuration to track regressions with mainline, and the current JDK sources support only one version of java.lang.Object which is now abstract in the lworld branch). Last change is the addition of the aarch64 support.

MrSimms (approved):
Looks good

TobiHartmann (approved):
Looks good.

PR #604, opened 2022-01-17
Implement code generation support for value classes

sadayapalam:
Push code code generation tactics and strategies from B3 to B2 classes.

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/valhalla pull/604/head:pull/604` \
`$ git checkout pull/604`

`$ git checkout pull/604` \
`$ git pull https://git.openjdk.java.net/valhalla pull/604/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 604`

`$ git pr show -t 604`

</details>

<a href="https://git.openjdk.java.net/valhalla/pull/604.diff">https://git.openjdk.java.net/valhalla/pull/604.diff</a>

</details>

### DIFF
--- a/src/jdk.compiler/share/classes/com/sun/tools/javac/code/Type.java
+++ b/src/jdk.compiler/share/classes/com/sun/tools/javac/code/Type.java
@@ -1330,7 +1330,7 @@ public abstract class Type extends AnnoConstruct implements TypeMirror, PoolCons
 
         @Override
         public boolean isValueClass() {
-            return tsym != null && tsym.isValueClass();
+            return !isReferenceProjection() && tsym != null && tsym.isValueClass();
         }
 
         @Override
--- a/src/jdk.compiler/share/classes/com/sun/tools/javac/code/Symbol.java
+++ b/src/jdk.compiler/share/classes/com/sun/tools/javac/code/Symbol.java
@@ -468,9 +468,9 @@ public abstract class Symbol extends AnnoConstruct implements PoolConstant, Elem
         return name == name.table.names.init && (flags() & STATIC) == 0;
     }
 
-    /** Is this symbol a primitive object factory?
+    /** Is this symbol a value object factory?
      */
-    public boolean isPrimitiveObjectFactory() {
+    public boolean isValueObjectFactory() {
         return ((name == name.table.names.init && this.type.getReturnType().tsym == this.owner));
     }
 
--- a/src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/ByteCodes.java
+++ b/src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/ByteCodes.java
@@ -243,8 +243,8 @@ public interface ByteCodes {
         jsr_w           = 201,
         breakpoint      = 202,
 
-        // primitive classes related bytecodes
-        defaultvalue    = 203,
+        // value classes related bytecodes
+        aconst_init     = 203,
         withfield       = 204,
 
         ByteCodeCount   = 205;
--- a/src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/ClassReader.java
+++ b/src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/ClassReader.java
@@ -1007,7 +1007,7 @@ public class ClassReader {
                         //- System.err.println(" # " + sym.type);
                         if (sym.kind == MTH && sym.type.getThrownTypes().isEmpty())
                             sym.type.asMethodType().thrown = thrown;
-                        if (sym.kind == MTH  && sym.name == names.init && sym.owner.isPrimitiveClass()) {
+                        if (sym.kind == MTH  && sym.name == names.init && sym.owner.isValueClass()) {
                             sym.type = new MethodType(sym.type.getParameterTypes(),
                                     syms.voidType,
                                     sym.type.getThrownTypes(),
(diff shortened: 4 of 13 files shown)


sadayapalam:
Notes for the reviewer:


This patch implements the code generation support for value classes.

    - Much of this functionality already exists in the Valhalla code base, but
      were implemented at the level of B3 (primitive) classes. What this change
      set does therefore is to push these code generation tactcs and strategies
      down to B2 (value) classes.

    - Rename defaultvalue opcode to aconst_init
    - Transate value class constructors to static factory methods of the name
      <init> with the return type of the class type.
    - Translate source level value instantiation attempts via new into calls to
      the static value factory method.
    - Translate assignments to fields with withfield rather than putfield.
    - In the class reader recognize value class factory methods and massage them
      and internalize them as normal constrctors - This is to minimize changes to
      the other phases of the compiler. A value class constrctor is treated as
      plain constrctor by javac but during class file emission, we do as mandated
      by JVMS.
      
Deferred

    - <init> to be renamed to <new>
    - Emission of Preload attribute
    - Marking stateless abstract classes as value supers.

fparain (approved):
Changes to the runtime test look good to me.

Fred

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Type.java
```diff
@@ -1330,7 +1330,7 @@ public boolean isPrimitiveClass() {
 
         @Override
         public boolean isValueClass() {
-            return tsym != null && tsym.isValueClass();
+            return !isReferenceProjection() && tsym != null && tsym.isValueClass();
```
mcimadamore, line 1333: The new test for `isReferenceProjections` does not match my understanding of what a "value class" is. E.g. I believe a value class doesn't really have a reference projection (or, rather, doing vclass.ref is the same as doing String.ref - it's the identity). So it's not clear to me the rationale behind the new check.

sadayapalam: You are right, a B2 value class does not have a reference projection, in fact String.ref and ValueClass.ref will ATM result in a compile error. In the implementation .ref can be applied only to a primitive class (B3) - the spec may call for .ref to a total operation - but this is what we do at the moment awaiting spec clarification on the whole handling of .ref syntax. There is a ticket open for it - but at the moment I have trouble accessing JIRA)

Now a B3 class can and does have a reference projection and every B3 class is a B2 class. So
the receiver of the method isValueClass() could be a B3 class. This necessitates the new change
to the method.

Consider this methid from Gen:
```
public void visitDefaultValue(JCDefaultValue tree) {
        if (tree.type.isValueClass()) {
            code.emitop2(aconst_init, checkDimension(tree.pos(), tree.type), PoolWriter::putClass);
        } else if (tree.type.isReference()) {
            code.emitop0(aconst_null);
        } else {
            code.emitop0(zero(Code.typecode(tree.type)));
        }
        result = items.makeStackItem(tree.type);
        return;
    } 
```
Without the proposed change to isValueClass(), we would emit aconst_init instead of aconst_null for
PrimitiveClass.ref.default

On a related note, (again hampered by lack of access to JBS), we have a ticket open for rationalization/cleanup of various internal APIs - I had put in some work into it - but that had to be halted because of large scale changes to landscape - (ref default classes being gone for example). This work needs to be taken up in the context of the new
model to bring upto date the APIs. For example com.sun.tools.javac.code.Type.ClassType.Flavor is not current )

Let me know if this addresses this your question, Thanks for reviewing!

biboudis (approved):
Tested the current PR 💯

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Type.java
sadayapalam: It is a fact that quite a bit of changes are in the pipeline, particularly with respect to internal API rationalization/cleanup, but lest I get too comfortable sweeping the present point of discussion under that carpet, here are some clarifications,

> I'm still confused a bit. I get that this is all transitional, in a way - but I would expect javac to differentiate between B2 >and B3 more. 

Yes, the API points isPrimitiveClass() will answer true only for B3 while isValueClass() will answer true for both B2 and B3. One of the next steps is to go through with a fine toothed comb every use of isPrimitiveClass in the code base
to make sure that it is guarding treatment specifc to B3 classes - everything else should have been pushed down to isValueClass() gate.

> It is true, as you claim that `every B3 class is a B2 class`: some of the identity restrictions apply to both declarations. >But in my opinion, the current javac support (as of before this PR) is much closer to B3 than it is to B2 declarations, at >least from a type system perspective. That is, a `value class` should be, to javac, just a plain reference class, with no >reference projection etc. So, if String.ref is an error, I'd expect vclass.ref to be an error in the same way (in this >transitional world).

Yes, String.ref is an error and given value class Point, Point.ref is an error today.

> From a translation perspective I agree that B2 and B3 are not too different (no constructors, etc.) - but I'm a bit afraid that by blurring distinction between B2 and B3 in the type system we might end up with problems further down the road.
> 
> That said, I understand that a lot of this is temporary, so if you think it would be more adequate I'm fine looking again at some point in the future.

I would say there is no blurring of distinction - the internal APIs that were evolving for a set of needs haven't fully been custom fit for the new world yet.

Yes, within the next couple of weeks, I expect to have futher changes that would have removed certain vestiges, retooled others to fit the current model and I will invite you to review that.

PR #605, opened 2022-01-18
8280164: [lworld] Generate Preload attribute to enumerate value classes encountered

sadayapalam:
Modelled after InnerClasses attribute generation, we may be generating more information than necessary/useful but this is initial cut

---------
### Progress

### Issue
 * [JDK-8280164](https://bugs.openjdk.java.net/browse/JDK-8280164): [lworld] Generate Preload attribute to enumerate value classes encountered

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/valhalla pull/605/head:pull/605` \
`$ git checkout pull/605`

`$ git checkout pull/605` \
`$ git pull https://git.openjdk.java.net/valhalla pull/605/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 605`

`$ git pr show -t 605`

</details>

<a href="https://git.openjdk.java.net/valhalla/pull/605.diff">https://git.openjdk.java.net/valhalla/pull/605.diff</a>

</details>

### DIFF
--- a/src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/PoolWriter.java
+++ b/src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/PoolWriter.java
@@ -51,6 +51,7 @@ import java.io.IOException;
 import java.io.OutputStream;
 import java.util.ArrayDeque;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.Map;
@@ -93,6 +94,8 @@ public class PoolWriter {
 
     /** The inner classes to be written, as an ordered set (enclosing first). */
     LinkedHashSet<ClassSymbol> innerClasses = new LinkedHashSet<>();
+
+    HashSet<ClassSymbol> valueClasses = new HashSet<>();
 
     /** The list of entries in the BootstrapMethods attribute. */
     Map<BsmKey, Integer> bootstrapMethods = new LinkedHashMap<>();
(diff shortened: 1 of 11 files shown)


test/langtools/tools/javac/valhalla/value-objects/PreloadAttributeTest.java
```diff
@@ -0,0 +1,98 @@
...
+        }
+        void foo(int x) {
+            V4 [] v4 = null; // local variable.
+        }
+        void goo() {
+            V5 [] v5 = null;
+            if (v5 == null) {
+               V6 [] v61 = null;  // stack map table.
+            } else {
+               V5 [] v52 = null;
```
sadayapalam, line 54: A note: V6 has no mention in the source file anywhere (local variable table, stack map table ...) and so doesn't feature in Preload either. The test should be updated to reflect that - it gives the impression that V6 is expected in the Preloads.

src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/PoolWriter.java
```diff
@@ -235,17 +238,38 @@ int putModule(ModuleSymbol mod) {
...
 
+    /** Enter a value class into the `valueClasses' set.
+     */
+    void enterValueClass(ClassSymbol c) {
+        if (c.type.isCompound()) {
+            throw new AssertionError("Unexpected intersection type: " + c.type);
+        }
+        c.complete();
+        if (c.isValueClass()) {
+            valueClasses.add(c);
```
sadayapalam, line 260: I suspect this should be if (c.isValueClass() && !c.isPrimitiveClass())

src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/PoolWriter.java
```diff
@@ -94,6 +95,8 @@
     /** The inner classes to be written, as an ordered set (enclosing first). */
     LinkedHashSet<ClassSymbol> innerClasses = new LinkedHashSet<>();
 
+    HashSet<ClassSymbol> valueClasses = new HashSet<>();
```
biboudis, line 98: nitpicking:`Set` instead of `HashSet` for the type definition? (hm, the interface is used below in line 101 but not in the line 96 above)

biboudis (commented):
I would expect all V* classes to be part of the `Preload`'s, `classes` field. Maybe my understanding is wrong. I checked with `javap` and it reports only one:

```
Classes to be preloaded:
  #14;                                    // value class PreloadAttributeTest$X
```

In any case, even if my understanding is wrong, can we use the same style as in `test/langtools/tools/javac/valhalla/value-objects/ValueCreationTest.java` to assert that the contents of the `Preload` attribute are the expected ones? (in this case 1).

What do you think? Otherwise it looks good.

sadayapalam:
> I would expect all V* classes to be part of the `Preload`'s, `classes` field. Maybe my understanding is wrong. I checked with `javap` and it reports only one:
> 
> ```
> Classes to be preloaded:
>   #14;                                    // value class PreloadAttributeTest$X
> ```

This is the javap output of PreloadAttributeTest.class - this class does not reference the V* classes at all.
If you look at the javap output of PreloadAttributeTest$X.class, you will see the V* classes

> 
> In any case, even if my understanding is wrong, can we use the same style as in `test/langtools/tools/javac/valhalla/value-objects/ValueCreationTest.java` to assert that the contents of the `Preload` attribute are the expected ones? (in this case 1).
> 

That test uses textual comparison which in general is turns out to be flaky - constant pool indices, byte code offsets can change when code generation changes in some minor unrelated way causing tests to "fail" needlessly.

The test you cite itself needs to move away to more robust form that avoid including immaterial things (such as indices/offsets etc) - this is a separate task that should be taken up. 
> What do you think? Otherwise it looks good.

biboudis:
I was doing `javap -v /Users/bibou/ora/valhalla/test/langtools/tools/javac/valhalla/value-objects/PreloadAttributeTest\$X` without the `\` so it was just `PreloadAttributeTest` effectively in the javap command 👍 👍 

All good then 💯

PR #606, opened 2022-01-19
8280006: [lworld] Compiler interface uses wrong class loader when looking up field type

TobiHartmann:
In the test, the primitive type `ValueOnBootclasspath` of flattened `Wrapper1.field` and the type `MyClass` of its flattened `field` are loaded by the boot classloader (`loader` is NULL):
```
<ciInlineKlass name=ValueOnBootclasspath loader=0x0000000000000000 loaded=true initialized=true finalized=false subklass=false size=16 flags=public,final,super super=java/lang/Object ident=1288 address=0x00007f5108540288>
<ciInstanceKlass name=MyClass loader=0x0000000000000000 loaded=true initialized=true finalized=false subklass=false size=16 flags=DEFAULT_ACCESS,super ident=1292 address=0x00007f5108541160> 
```

During JIT compilation, when the type of `Wrapper1.field` is computed in `ciField::compute_type_impl()`, `Wrapper1` is used as the holder klass and therefore its class loader (`loader=0x000000062c754858`) is used as context for the lookup:

```
<ciField name=TestBootClassloader$Wrapper1.field signature=LMyClass; offset=12 type=(reference) flags=0010 is_constant=true is_flattened=false is_null_free=false>
<ciInlineKlass name=TestBootClassloader$Wrapper1 loader=0x000000062c754858 loaded=true initialized=true finalized=false subklass=false size=16 flags=DEFAULT_ACCESS,final,super super=java/lang/Object ident=1284 address=0x00007f510853f440>
```

But since `MyClass` was loaded by the boot classloader, the lookup via the application class loader fails and leads to an unloaded klass:
```
<ciInstanceKlass name=MyClass loader=0x000000062c754858 loaded=false ident=1289 address=0x00007f5108540708>
```

C2's type system then gets confused when mixing both `MyClass` types and bails out with `COMPILE SKIPPED: Can't determine return type. (retry at different tier)`.

The fix is to use the "original holder", i.e., the klass that defined the field in the source code and not the one that contains it at runtime due to flattening, for the lookup instead.

While debugging, I also noticed that loading of signature classes before JIT compilation does not apply to fields that came in through flattening. I filed [JDK-8280230](https://bugs.openjdk.java.net/browse/JDK-8280230) to investigate this separately.

Best regards,
Tobias

---------
### Progress

### Issue
 * [JDK-8280006](https://bugs.openjdk.java.net/browse/JDK-8280006): [lworld] Compiler interface uses wrong class loader when looking up field type

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/valhalla pull/606/head:pull/606` \
`$ git checkout pull/606`

`$ git checkout pull/606` \
`$ git pull https://git.openjdk.java.net/valhalla pull/606/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 606`

`$ git pr show -t 606`

</details>

<a href="https://git.openjdk.java.net/valhalla/pull/606.diff">https://git.openjdk.java.net/valhalla/pull/606.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/ci/ciField.cpp
+++ b/src/hotspot/share/ci/ciField.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 1999, 2021, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1999, 2022, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -71,7 +71,7 @@
 // ------------------------------------------------------------------
 // ciField::ciField
 ciField::ciField(ciInstanceKlass* klass, int index) :
-  _is_flattened(false), _known_to_link_with_put(NULL), _known_to_link_with_get(NULL) {
+  _original_holder(NULL), _is_flattened(false), _known_to_link_with_put(NULL), _known_to_link_with_get(NULL) {
   ASSERT_IN_VM;
   CompilerThread *THREAD = CompilerThread::current();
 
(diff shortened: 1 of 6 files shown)


fparain (approved):
Looks good to me.
Nice job writing an unit test for this.

Fred

TobiHartmann:
Thanks for the review, Fred! Yes, took me a while :)

PR #609, opened 2022-01-21
8280381: [lworld] VM should accept static factory methods in value class files.

fparain:
Please review this patch that enables loading of value classes.
Support for value classes is minimal, but this is the first step to implement them.
The patch also includes some renaming to align Hotspot code with the JVMS draft.

Tested with Mach5, tiers 1-3 (so it builds on all platforms)

Thank you,

Fred


MrSimms (approved):
Looks good

fparain:
Thank your for the review.

PR #617, opened 2022-01-25
8280524: [lworld] Interpreter incorrectly handles uninitialized static inline type field

TobiHartmann:
Code in the interpreter that's supposed to check if the type of a static, null-free inline type field is initialized assumes that `rcx` contains the `InlineKlass*`. However, `rcx` is only initialized for non-static fields:
https://github.com/openjdk/valhalla/blob/ba87b167f2fe2327508a06ec8652c6ddaa9402c1/src/hotspot/cpu/x86/templateTable_x86.cpp#L2939-L2943

As a result, the check spuriously fails, leading to an "impossible" NPE because the null value is not replaced by the default value. This patch also fixes an unrelated issue where an `instanceHandle` should be used instead of an `instanceOop`.

I found this with compiler stress testing and initially assumed that it is a compiler bug. The patch was contributed by @fparain.

Thanks,
Tobias

---------
### Progress

### Issue
 * [JDK-8280524](https://bugs.openjdk.java.net/browse/JDK-8280524): [lworld] Interpreter incorrectly handles uninitialized static inline type field

### Reviewers

### Contributors

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/valhalla pull/617/head:pull/617` \
`$ git checkout pull/617`

`$ git checkout pull/617` \
`$ git pull https://git.openjdk.java.net/valhalla pull/617/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 617`

`$ git pr show -t 617`

</details>

<a href="https://git.openjdk.java.net/valhalla/pull/617.diff">https://git.openjdk.java.net/valhalla/pull/617.diff</a>

</details>

### DIFF
--- a/src/hotspot/cpu/x86/templateTable_x86.cpp
+++ b/src/hotspot/cpu/x86/templateTable_x86.cpp
@@ -3008,9 +3008,10 @@ void TemplateTable::getfield_or_static(int byte_no, bool is_static, RewriteContr
             __ andl(flags2, ConstantPoolCacheEntry::field_index_mask);
   #ifdef _LP64
             Label slow_case, finish;
-            __ cmpb(Address(rcx, InstanceKlass::init_state_offset()), InstanceKlass::fully_initialized);
+            __ movptr(rbx, Address(obj, java_lang_Class::klass_offset()));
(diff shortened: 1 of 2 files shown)


fparain (approved):
Looks good to me.

Fred

TobiHartmann:
Thanks, Fred!

PR #621, opened 2022-01-27
8280382: [lworld] JCWithField AST node support should evolve to work with value classes

biboudis:
This PR ports basic `withField` tests from lworld-values to value-objects and extending some areas of code to support `withField` within value classes.

Nitpicking and a note: one of the error messages is now `compiler.err.primitive.or.value.class.instance.field.expected.here`. Would you propose to disambiguate further depending whether we are inside a `primitive` or a `value` class?

---------
### Progress

### Issue
 * [JDK-8280382](https://bugs.openjdk.java.net/browse/JDK-8280382): [lworld] JCWithField AST node support should evolve to work with value classes

### Reviewers
 * [Srikanth Adayapalam](https://openjdk.java.net/census#sadayapalam) (@sadayapalam - Committer) ⚠️ Review applies to 5aa7e600a3876fe75d4d906baf18a974a0a2e0bb

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/valhalla pull/621/head:pull/621` \
`$ git checkout pull/621`

`$ git checkout pull/621` \
`$ git pull https://git.openjdk.java.net/valhalla pull/621/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 621`

`$ git pr show -t 621`

</details>

<a href="https://git.openjdk.java.net/valhalla/pull/621.diff">https://git.openjdk.java.net/valhalla/pull/621.diff</a>

</details>


src/jdk.compiler/share/classes/com/sun/tools/javac/resources/compiler.properties
```diff
@@ -3937,8 +3937,8 @@ compiler.err.primitive.class.does.not.support=\
 compiler.err.value.class.may.not.extend=\
     inappropriate super class declaration for a value class
 
-compiler.err.primitive.class.instance.field.expected.here=\
-    withfield operator requires an instance field of a primitive class here
+compiler.err.primitive.or.value.class.instance.field.expected.here=\
+    withfield operator requires an instance field of a primitive or a value class here
```
sadayapalam, line 3942: This message is not in alignment with rest of diagnostics that ony speak of value classes. I suggest we amend this to 
compiler.err.value.class.instance.field.expected.here. I think it is very visible and open that every primitive class is a value class so it is not necessary to call out primitive or value class

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java
```diff
@@ -329,7 +329,7 @@ void checkAssignable(DiagnosticPosition pos, VarSymbol v, JCTree base, Env<AttrC
                    withfield operator -This does not result in mutation of final fields; the code generator
                    would implement `copy on write' semantics via the opcode `withfield'.
                 */
-                if (env.info.inWithField && v.getKind() == ElementKind.FIELD && (v.flags() & STATIC) == 0 && v.owner.type.isPrimitiveClass()) {
+                if (env.info.inWithField && v.getKind() == ElementKind.FIELD && (v.flags() & STATIC) == 0 && (v.owner.type.isPrimitiveClass() || v.owner.type.isValueClass())) {
                     if (env.enclClass.sym.outermostClass() == v.owner.outermostClass())
```
sadayapalam, line 333: I suggest we fix all places where a check is being made for isPrimitiveClass() || isValueClass() to just check for isValueClass(). Every primitive class is also a value class so this check is redundant and also is not in alignment with other places where only isValueClass() check is being made. Thirdly and most importantly, we want to reach a state in the codebase where every check of isPrimitiveClass() is essential to triggering behavior that is unique to B3 classes. Any code controlled by isPrimitiveClass() must be specific for B3 classes. This will significantly improve code readability.

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
```diff
@@ -605,12 +605,12 @@ public boolean isConvertible(Type t, Type s, Warner warn) {
...
-        boolean sValue = s.isPrimitiveClass();
+        boolean tValue = t.isPrimitiveClass() || t.isValueClass();
+        boolean sValue = s.isPrimitiveClass() || s.isValueClass();
         if (tValue != sValue) {
             return tValue ?
-                    isSubtype(t.referenceProjection(), s) :
-                    !t.hasTag(BOT) && isSubtype(t, s.referenceProjection());
+                    isSubtype(t.referenceProjectionOrSelf(), s) :
+                    !t.hasTag(BOT) && isSubtype(t, s.referenceProjectionOrSelf());
         }
```
sadayapalam, line 615: These changes are not necessary and in fact are incorrect. See that with this change, the following code does not compile anymore which it should.

```
public value class X {
        X x = null;
    }
}
```

This triggers an error: 
X.java:3: error: incompatible types: <null> cannot be converted to X
        X x = null;
              ^

X being a B2 value class its instances are very much nullable.

test/langtools/tools/javac/valhalla/value-objects/WithFieldOperatorTest.java
```diff
@@ -0,0 +1,42 @@
...
+            X x = X.default;
+            x = __WithField(new Y(), null);  // not a variable at all.
+            x = __WithField(wfot.xs, 10); // not an instance field.
+            x = __WithField(wfot.ifld, 10); // not a field of value type
+            x = __WithField(xVal, xVal); // not a field
+            x = __WithField(this, this); // not a field
+            x = __WithField(X.this, this); // not a field
+            x = __WithField(x.x, 12.0); // float cannot be assigned to int
+            x = __WithField(x.v, null); // null cannot be assigned to value
+            return x;
```
sadayapalam, line 33: Null can be assigned to value, so this line should not trigger an error

test/langtools/tools/javac/valhalla/value-objects/WithFieldOperatorTest.out
```diff
@@ -0,0 +1,10 @@
+WithFieldOperatorTest.java:25:29: compiler.err.unexpected.type: kindname.variable, kindname.value
+WithFieldOperatorTest.java:26:33: compiler.err.primitive.or.value.class.instance.field.expected.here
+WithFieldOperatorTest.java:27:33: compiler.err.primitive.or.value.class.instance.field.expected.here
+WithFieldOperatorTest.java:28:29: compiler.err.primitive.or.value.class.instance.field.expected.here
+WithFieldOperatorTest.java:29:29: compiler.err.cant.assign.val.to.this
+WithFieldOperatorTest.java:30:30: compiler.err.cant.assign.val.to.this
+WithFieldOperatorTest.java:31:34: compiler.err.prob.found.req: (compiler.misc.possible.loss.of.precision: double, int)
+WithFieldOperatorTest.java:32:34: compiler.err.prob.found.req: (compiler.misc.inconvertible.types: compiler.misc.type.null, WithFieldOperatorTest.V)
+WithFieldOperatorTest.java:40:26: compiler.err.cant.assign.val.to.final.var: x
```
sadayapalam, line 9: Error 8 should not be emitted

test/langtools/tools/javac/valhalla/value-objects/FlattenableNegativeTest.java
```diff
@@ -0,0 +1,38 @@
...
+            V [] va2 = new V[] { null }; // Illegal array initialization
+            void foo(X x) {
+                x = __WithField(x.v, null); // illegal withfield attempt
+                v = null; // illegal assignment.
+                va[0] = null; // Illegal.
+                va = new V[] { null }; // Illegal
+            }
+        }
+    }
+}
```
sadayapalam, line 38: For value classes, this is not a negative test. The test should compile fine.

test/langtools/tools/javac/valhalla/value-objects/FlattenableNegativeTest.out
```diff
@@ -0,0 +1,10 @@
...
+FlattenableNegativeTest.java:22:34: compiler.err.prob.found.req: (compiler.misc.inconvertible.types: compiler.misc.type.null, FlattenableNegativeTest.V)
+FlattenableNegativeTest.java:28:25: compiler.err.prob.found.req: (compiler.misc.inconvertible.types: compiler.misc.type.null, FlattenableNegativeTest.V)
+FlattenableNegativeTest.java:29:34: compiler.err.prob.found.req: (compiler.misc.inconvertible.types: compiler.misc.type.null, FlattenableNegativeTest.V)
+FlattenableNegativeTest.java:31:38: compiler.err.prob.found.req: (compiler.misc.inconvertible.types: compiler.misc.type.null, FlattenableNegativeTest.V)
+FlattenableNegativeTest.java:32:21: compiler.err.prob.found.req: (compiler.misc.inconvertible.types: compiler.misc.type.null, FlattenableNegativeTest.V)
+FlattenableNegativeTest.java:33:25: compiler.err.prob.found.req: (compiler.misc.inconvertible.types: compiler.misc.type.null, FlattenableNegativeTest.V)
+FlattenableNegativeTest.java:34:32: compiler.err.prob.found.req: (compiler.misc.inconvertible.types: compiler.misc.type.null, FlattenableNegativeTest.V)
+9 errors
```
sadayapalam, line 10: None of these errors should be emitted, null is assignable to B2 instances.

sadayapalam (changes requested):
I have indicated problem spots. Please incorporate the changes and update, Thanks!

test/langtools/tools/javac/valhalla/value-objects/WithFieldOperatorTest.java
sadayapalam: Also this test in lworld-values directory should be amended to say "null cannot be assigned to primitive class instance" - it should not say value. Confusingly, at a distant past point, B3 classes used to be called value classes and hence this test refers to stale terminology

sadayapalam (changes requested):
Change value to primitive class instance in one test in lworld-values directory

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
biboudis: You are right. The idea behind it was that I was getting a null reference exception on the subtype check. `t.referenceProjection()` was returning `null`  and it was blowing up on the `t.equalsIgnoreMetadata(s)`. The triggering code was something like the following (so the reference projection of `o`  was `null`).

```
value static final class V {}

Object o = V.default;
```

What do you think?

sadayapalam: > You are right. The idea behind it was that I was getting a null reference exception on the subtype check. `t.referenceProjection()` was returning `null` and it was blowing up on the `t.equalsIgnoreMetadata(s)`. The triggering code was something like the following (so the reference projection of `o` was `null`).
> 
> ```
> value static final class V {}
> 
> Object o = V.default;
> ```
> 
> What do you think?

If this probem is still persisting, I suggest we raise a separate ticket for it. I don't know what is going on there, but can take a look and share observations on a separate task basis.

biboudis: It was only introduced only after changing the lines 608 and 607 above, so it feels related to this change:

> https://github.com/openjdk/valhalla/pull/621/files/e0f302f20eacdbb796d03240b5b727ce053ed451#diff-244efd7111106b403c989b25a6d0c719dcb15d44892320cfdc6056a418defb21L608-L609

biboudis:
Thanks for the review @sadayapalam! 🎉

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
sadayapalam: But those lines need not change at all. The original code in com.sun.tools.javac.code.Types#isConvertible(com.sun.tools.javac.code.Type, com.sun.tools.javac.code.Type, com.sun.tools.javac.util.Warner) viz:
```
boolean tValue = t.isPrimitiveClass();
        boolean sValue = s.isPrimitiveClass();
        if (tValue != sValue) {
            return tValue ?
                    isSubtype(t.referenceProjection(), s) :
                    !t.hasTag(BOT) && isSubtype(t, s.referenceProjection());
        }
```
is necessitated by the fact that in the language model a primitive class is an island. Even though Point may be declared to be:
```
primitive class Point extends Object {}
```
asSuper(Point, Object) would return null meaning Point is not a subclass of Object. We bend over backwards maintaining this isolated world view of primitive classes to make it analogous to how int is not a subtype of Object, but is convertible to Integer which is a subtype of Object. Likewise Point is not a subtype of Object, but is convertible to Point.ref and Point.ref is a subtype of Object.

So is Point convertible to Object ? Yes, not because it is a subtype of Object, but because it can be converted to Point.ref and Point.ref is a subtype of Object.

This is the dance that is being done in the code above. This code should not and need not change at all for the present topic at hand.

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java
biboudis: good remark, thx!

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
biboudis: You are right. I made it it work by rebuilding and reverting this change! Thanks for the detailed explanation as always 🙏🏻🙏🏻🙏🏻

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java
```diff
@@ -1331,7 +1331,7 @@ public void visitVarDef(JCVariableDecl tree) {
                as these can undergo updates via copy on write.
             */
             if (tree.init != null) {
-                if ((v.flags_field & FINAL) == 0 || ((v.flags_field & STATIC) == 0 && v.owner.isPrimitiveClass()) ||
+                if ((v.flags_field & FINAL) == 0 || ((v.flags_field & STATIC) == 0 && (v.owner.isPrimitiveClass() || v.owner.isValueClass())) ||
                     !memberEnter.needsLazyConstValue(tree.init)) {
```
sadayapalam, line 1335: There is still one instanceof isPrimitiveClass() || isValueClass() left in

sadayapalam (approved):
One change is requested, otherwise looks good. Please make this change and once the tests are known to be green, proceed to integrate, Thanks!

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java
biboudis: Oups. Good catch.

PR #622, opened 2022-01-31
8280194: Abstract classes that allow value subclasses should be marked ACC_PERMITS_VALUE (0x0040)

sadayapalam:
Mark abstract classes that qualify with ACC_PERMITS_VALUE flag.

---------
### Progress

### Issues
 * [JDK-8280194](https://bugs.openjdk.java.net/browse/JDK-8280194): Abstract classes that allow value subclasses should be marked ACC_PERMITS_VALUE (0x0040)
 * [JDK-8242619](https://bugs.openjdk.java.net/browse/JDK-8242619): [lworld] Ascertain that no-arg super constructors of inline classes are empty

### Reviewers
 * @biboudis (no known github.com user name / role) ⚠️ Review applies to 28ff0c54f2da86bcaa2224c9a2a349de3ac1d46f

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/valhalla pull/622/head:pull/622` \
`$ git checkout pull/622`

`$ git checkout pull/622` \
`$ git pull https://git.openjdk.java.net/valhalla pull/622/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 622`

`$ git pr show -t 622`

</details>

<a href="https://git.openjdk.java.net/valhalla/pull/622.diff">https://git.openjdk.java.net/valhalla/pull/622.diff</a>

</details>


src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
```diff
@@ -2376,7 +2376,7 @@ private boolean implicitIdentityType(Type t) {
                             }
                             break;
                         case MTH:
-                            if ((s.flags() & SYNCHRONIZED) != 0) {
+                            if ((s.flags() & (SYNCHRONIZED | STATIC)) == SYNCHRONIZED) {
                                 return true;
                             } else if (s.isConstructor()) {
```
sadayapalam, line 2381: This is a bug fix - static methods may be synchronized or not - immaterial.

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java
```diff
@@ -1252,12 +1252,6 @@ public void visitMethodDef(JCMethodDecl tree) {
                         }
                     }
                 }
-                if (m.isConstructor() && m.type.getParameterTypes().size() == 0) {
-                    if ((owner.type == syms.objectType) ||
-                            (tree.body.stats.size() == 1 && TreeInfo.getConstructorInvocationName(tree.body.stats, names, false) == names._super)) {
-                        m.flags_field |= EMPTYNOARGCONSTR;
-                    }
-                }
```
sadayapalam, line 1255: This is too late here. Now this is moved up to MemberEnter

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java
```diff
@@ -1455,11 +1449,7 @@ public void visitBlock(JCBlock tree) {
             final Env<AttrContext> localEnv =
                 env.dup(tree, env.info.dup(env.info.scope.dupUnshared(fakeOwner)));
 
-            if ((tree.flags & STATIC) != 0)
-                localEnv.info.staticLevel++;
-            else if (tree.stats.size() > 0)
-                env.info.scope.owner.flags_field |= HASINITBLOCK;
-
+            if ((tree.flags & STATIC) != 0) localEnv.info.staticLevel++;
```
sadayapalam, line 1452: This is too late in Attr. This is now moved up.

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/MemberEnter.java
```diff
@@ -229,6 +229,13 @@ public void visitMethodDef(JCMethodDecl tree) {
...
             annotate.annotateDefaultValueLater(tree.defaultValue, localEnv, m, tree.pos());
         }
+
+        if (m.isConstructor() && m.type.getParameterTypes().size() == 0) {
+            if (tree.body.stats.size() == 0 || (m.flags() & GENERATEDCONSTR) != 0) {
+                // generated constructors do have a super() call, but these are ignored for value classes.
+                m.flags_field |= EMPTYNOARGCONSTR;
+            }
+        }
     }
```
sadayapalam, line 239: This used to happen in Attr, but that is too late. Pulled up to here now

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/MemberEnter.java
```diff
@@ -249,6 +256,12 @@ public void visitMethodDef(JCMethodDecl tree) {
         return localEnv;
     }
 
+    @Override
+    public void visitBlock(JCBlock tree) {
+        if ((tree.flags & STATIC) == 0 && tree.stats.size() > 0)
+            env.info.scope.owner.flags_field |= HASINITBLOCK;
+    }
+
     public void visitVarDef(JCVariableDecl tree) {
```
sadayapalam, line 265: This used to happen in Attr, but that is too late. Pulled up to here now

test/langtools/tools/javac/classfiles/attributes/innerclasses/InnerClassesInAnonymousClassTest.java
```diff
@@ -58,7 +58,7 @@ public void setProperties() {
     @Override
     public List<TestCase> generateTestCases() {
         currentClassType = ClassType.CLASS;
-        setPrefix("class Anonymous {} {new Anonymous() {");
+        setPrefix("class Anonymous { int f; } {new Anonymous() {");
         List<TestCase> sources = super.generateTestCases();
```
sadayapalam, line 63: Hack to suppress ACC_PERMITS_VALUE bit from being set thereby causing the test to "fail"

test/langtools/tools/javac/classfiles/attributes/innerclasses/InnerClassesInLocalClassTest.java
```diff
@@ -52,7 +52,7 @@ public InnerClassesInLocalClassTest() {
         for (Modifier outerModifier : LOCAL_CLASS_MODIFIERS) {
             StringBuilder sb = new StringBuilder();
             sb.append(outerModifier.getString()).append(' ');
-            sb.append("class Local {");
+            sb.append("class Local { int f; ");
             Map<String, Set<String>> class2Flags = new HashMap<>();
```
sadayapalam, line 56: Hack to suppress ACC_PERMITS_VALUE bit from being set thereby causing the test to "fail"

test/langtools/tools/javac/classfiles/attributes/innerclasses/InnerClassesTestBase.java
```diff
@@ -290,8 +292,10 @@ private void test(String classToTest, TestCase test, String...skipClasses) {
                     privateConstructor = "private A" + count + "() {}";
                     syntheticClasses.add("new A" + count + "();");
                 }
+                String instField = innerClassType == ClassType.CLASS && innerMod.contains(Modifier.ABSTRACT) ?
+                                                "int f; " : ""; // inhibit PERMITS_VALUE
                 sb.append(toString(innerMod)).append(' ');
-                sb.append(String.format("%s A%d {%s}\n", innerClassType, count, privateConstructor));
+                sb.append(String.format("%s A%d { %s %s}\n", innerClassType, count, instField, privateConstructor));
                 Set<String> flags = getFlags(innerClassType, innerMod);
                 class2Flags.put("A" + count, flags);
```
sadayapalam, line 300: Hacks to suppress ACC_PERMITS_VALUE bit from being set thereby causing the test to "fail"

test/langtools/tools/javac/classfiles/attributes/innerclasses/InnerClassesInLocalClassTest.java
biboudis: I was playing around to understand this generated test:

```
public class OuterClass {
 static void method() {abstract class Local { class A0{}
abstract class A1{}
final class A2{}
};}
}
```

Why do we get the following for an inner class that is abstract,

```
[ASSERT] : inner_class_access_flags 1Local
Expected: [ACC_ABSTRACT],
     Got: [ACC_ABSTRACT, 0x40]
```

while `abstract class A8` in `PermitsValueTest.java` does not emit `ACC_PERMITS_VALUE`?

nitpick (and possibly not addressed yet): `Modifiers:getString()` doesn't return the name for `ACC_PERMITS_VALUE` yet. Do we need to address this in this PR? possibly not

biboudis (approved):
This PR looks good. Some clarifications requested.

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
sadayapalam: > Do you think the comment above the signature of `implicitIdentityType` needs to reflect this change in a parenthesis?

Sure, I don't see a comment above implicitIdentityType, but inside it at the top, I have amended it to say "if it declares a synhronized instance method ..."

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java
sadayapalam: This was in being set in Attribution phase - and is also consulted in the attribution phase. 
so when we have classes like:
```
value class V extends A {}
abstract class A { A() {} } 
```
when we are attributing V, A may not be attributed. While attributing V, we will check if V's superclass (A) meets all the requirements of an abstract class to be a value super but A isn't attributed yet.

What we have done now is to move these analysis to MemberEnter - MemberEnter for V and A are both guaranteed to have finished before Attribution of either starts, so we have eliminated the phase ordering problem

test/langtools/tools/javac/classfiles/attributes/innerclasses/InnerClassesInLocalClassTest.java
sadayapalam: > I was playing around to understand this generated test:
> 
> ```
> public class OuterClass {
>  static void method() {abstract class Local { class A0{}
> abstract class A1{}
> final class A2{}
> };}
> }
> ```
> 
> Why do we get the following for an inner class that is abstract,
> 
> ```
> [ASSERT] : inner_class_access_flags 1Local
> Expected: [ACC_ABSTRACT],
>      Got: [ACC_ABSTRACT, 0x40]
> ```
> 
> while `abstract class A8` in `PermitsValueTest.java` does not emit `ACC_PERMITS_VALUE`?

Class A8 in PermitsValue is an inner class that has an enclosing instance. Enclosing instance means the class will have an instance field to store the outer this and also its constructor will receive an argument - ie it cannot be a no-arg constructor only. (You may want to see the class files generated to understand the synthetic fields and constructor args that represent the outer this for an inner class with an enclosing instance) 

> 
> nitpick (and possibly not addressed yet): `Modifiers:getString()` doesn't return the name for `ACC_PERMITS_VALUE` yet. Do we need to address this in this PR? possibly not

Oops - Yes I will raise a follow up defect.

biboudis: Thank you! It's clear now!

sadayapalam:
I took a look at the Modifier.getString enhancement - at the moment ACC_PERMITS_VALUE is not surfaced at all in these tests. So we can leave it as is for the time being.

PR #629, opened 2022-02-07
8281166: [lworld] javac should generate BSM to invoke the static factory for value class

sadayapalam:
When constructor reference expressions invole value classes, ensure that TransValues will see the constructor call in order for that call to be transformed into factory method invocation.

---------
### Progress

### Issue
 * [JDK-8281166](https://bugs.openjdk.java.net/browse/JDK-8281166): [lworld] javac should generate BSM to invoke the static factory for value class

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/valhalla pull/629/head:pull/629` \
`$ git checkout pull/629`

`$ git checkout pull/629` \
`$ git pull https://git.openjdk.java.net/valhalla pull/629/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 629`

`$ git pr show -t 629`

</details>

<a href="https://git.openjdk.java.net/valhalla/pull/629.diff">https://git.openjdk.java.net/valhalla/pull/629.diff</a>

</details>

### DIFF
--- a/src/jdk.compiler/share/classes/com/sun/tools/javac/comp/LambdaToMethod.java
+++ b/src/jdk.compiler/share/classes/com/sun/tools/javac/comp/LambdaToMethod.java
@@ -2341,7 +2341,7 @@ public class LambdaToMethod extends TreeTranslator {
                         receiverIsReferenceProjection() ||
                         (tree.getMode() == ReferenceMode.NEW &&
                           tree.kind != ReferenceKind.ARRAY_CTOR &&
-                          (tree.sym.owner.isDirectlyOrIndirectlyLocal() || tree.sym.owner.isInner() || tree.sym.owner.isPrimitiveClass()));
+                          (tree.sym.owner.isDirectlyOrIndirectlyLocal() || tree.sym.owner.isInner() || tree.sym.owner.isValueClass()));
             }
 
             Type generatedRefSig() {
--- a/test/langtools/tools/javac/valhalla/value-objects/ConstructorRefTest.java
+++ b/test/langtools/tools/javac/valhalla/value-objects/ConstructorRefTest.java
@@ -56,7 +56,7 @@ public class ConstructorRefTest {
         }
     }
 
-    public static void main(String [] args) {   
+    public static void main(String [] args) {
 
         Supplier<P.ref> sxp = P::new;
         P p = (P) sxp.get();
--- a/test/langtools/tools/javac/valhalla/value-objects/ConstructorRefTest.java
+++ b/test/langtools/tools/javac/valhalla/value-objects/ConstructorRefTest.java
@@ -1,0 +1,71 @@
+/*
+ * Copyright (c) 2022, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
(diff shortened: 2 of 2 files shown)


src/jdk.compiler/share/classes/com/sun/tools/javac/comp/LambdaToMethod.java
```diff
@@ -2341,7 +2341,7 @@ final boolean needsConversionToLambda() {
                         receiverIsReferenceProjection() ||
                         (tree.getMode() == ReferenceMode.NEW &&
                           tree.kind != ReferenceKind.ARRAY_CTOR &&
-                          (tree.sym.owner.isDirectlyOrIndirectlyLocal() || tree.sym.owner.isInner() || tree.sym.owner.isPrimitiveClass()));
+                          (tree.sym.owner.isDirectlyOrIndirectlyLocal() || tree.sym.owner.isInner() || tree.sym.owner.isValueClass()));
             }
```
sadayapalam, line 2345: Notes for reviewer: Basically by forcing the constructor references to be converted into lambdas, we ensure that later pipeline stages will see the constructor invocation. For example, with the code change above, given a value class V, a constructor reference of the form V::new will be transformed into a lambda implementation method of the form 

```
/*synthetic*/ private static X$V lambda$main$1() {
    return new X$V();
}
```
This lambda implementation method has the constructor invocation intact which can then be detected and transformed by *com.sun.tools.javac.jvm.TransValues#visitNewClass*

(It is possible to directly emit BSM entries without converting them to lambda but that would result in code duplication - we would have to look at DynamicMethod symbols, pick them apart and reencode static and dynamic arguments to indy so that value constructors (and primitive class constructors are reencoded to static factory)

forax: I wonder if the transformation from a constructor call to a static method call when it's a value class should not be done before the lambda transformation ?

sadayapalam: > I wonder if the transformation from a constructor call to a static method call when it's a value class should not be done before the lambda transformation ?

This is technically feasible, but has some practical complications. As the top level comment in com.sun.tools.javac.jvm.TransValues calls out:

```
/**
 * This pass translates value class constructors into static factory methods and patches up constructor
 * calls to become invocations of those static factory methods.
 *
 * We get commissioned as a subpass of Gen. Constructor trees undergo plenty of change in Lower
 * (enclosing instance injection, captured locals ...) and in Gen (instance field initialization,
 * see normalizeDefs) and so it is most effective to wait until things reach a quiescent state
 * before undertaking the tinkering that we do.
 *
 * See https://bugs.openjdk.java.net/browse/JDK-8198749 for the kind of transformations we do.
 *
 */
```

Lambda lowering happens well before Lower and Gen pipeline phases. TransValues is a subphase of Gen - we need to wait until Gen because constructors undergo lots of changes up until and well into Gen. So if we want to group all transformations - of constructors into static factories and constructor invocations into static value factory invocations - into one centralized localized pipeline stage (i.e TransValues) the present approach is the best

forax: Okay, make sense

biboudis (approved):
💯💯

PR #631, opened 2022-02-07
8281336: [lworld] Remove workaround in LambdaToMethod to circumvent BootstrapMethodError

sadayapalam:
Remove the workaround in Javac and include the fix from Mandy for runtime in lieu.

---------
### Progress

### Issues
 * [JDK-8281336](https://bugs.openjdk.java.net/browse/JDK-8281336): [lworld] Remove workaround in LambdaToMethod to circumvent BootstrapMethodError
 * [JDK-8274399](https://bugs.openjdk.java.net/browse/JDK-8274399): [lworld] LambdaConversionException thrown when the receiver type is a primitive reference type and the implementation type is the primitive value type of the same class

### Reviewers

### Contributors
 * Mandy Chung `<mchung@openjdk.org>`

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/valhalla pull/631/head:pull/631` \
`$ git checkout pull/631`

`$ git checkout pull/631` \
`$ git pull https://git.openjdk.java.net/valhalla pull/631/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 631`

`$ git pr show -t 631`

</details>

<a href="https://git.openjdk.java.net/valhalla/pull/631.diff">https://git.openjdk.java.net/valhalla/pull/631.diff</a>

</details>

### DIFF
--- a/src/java.base/share/classes/java/lang/invoke/AbstractValidatingLambdaMetafactory.java
+++ b/src/java.base/share/classes/java/lang/invoke/AbstractValidatingLambdaMetafactory.java
@@ -268,10 +268,10 @@ import static sun.invoke.util.Wrapper.isWrapperType;
             }
 
             // check receiver type
-            if (!implClass.isAssignableFrom(receiverClass)) {
+            if (!implClass.asPrimaryType().isAssignableFrom(receiverClass.asPrimaryType())) {
                 throw new LambdaConversionException(
                         String.format("Invalid receiver type %s; not a subtype of implementation type %s",
-                                      receiverClass, implClass));
+                                      receiverClass.descriptorString(), implClass.descriptorString()));
             }
         } else {
             // no receiver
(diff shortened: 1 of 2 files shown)


mlchung:
The runtime change looks fine.  Is there an existing test verifying this?

biboudis:
Maybe we can use the test from the the original PR?

```
import java.util.function.Supplier;

public primitive class CheckRefLambda {
    int theInteger;

    CheckRefLambda(int newValue){
        theInteger=newValue;
    }

    public Supplier<CheckRefLambda.ref> makeCopier(){
        return()->new CheckRefLambda(this.theInteger + 1);
    }

    public static void main(String[]args){
        CheckRefLambda rec=new CheckRefLambda(42);
        CheckRefLambda rec2=rec.makeCopier().get();
        if(rec2!=new CheckRefLambda(43)){
            throw new AssertionError("Classes should be equal");
        }
    }
}
```

sadayapalam:
> The runtime change looks fine. Is there an existing test verifying this?

Yes, indeed. test/langtools/tools/javac/valhalla/lworld-values/StreamsTest.java fails with only javac changes without the runtime changes.

mlchung (approved):

sadayapalam:
> Maybe we can use the test from the the original PR?
> 
> ```
> import java.util.function.Supplier;
> 
> public primitive class CheckRefLambda {
>     int theInteger;
> 
>     CheckRefLambda(int newValue){
>         theInteger=newValue;
>     }
> 
>     public Supplier<CheckRefLambda.ref> makeCopier(){
>         return()->new CheckRefLambda(this.theInteger + 1);
>     }
> 
>     public static void main(String[]args){
>         CheckRefLambda rec=new CheckRefLambda(42);
>         CheckRefLambda rec2=rec.makeCopier().get();
>         if(rec2!=new CheckRefLambda(43)){
>             throw new AssertionError("Classes should be equal");
>         }
>     }
> }
> ```

Much water has flown under the bridge - we have ripped out all support for ref-default classes and so the meanings of unadorned/naked names have changed. (in the original scenario the name CheckRefLambda is a primitive reference but in what you have put together, CheckRefLambda is an instance of a primitive class) This test case does not fail anymore without any changes (javac and/or runtime). Since we already have a javac test covering the PR, we should be good I would say.

PR #632, opened 2022-02-07
8281279: [lworld] Add JVM support for ACC_PERMITS_VALUE

hseigel:
Please review this fix to add JVM support for ACC_PERMITS_VALUE.  This change throws an ICCE exception when loading a value class whose super class is not java.lang.Object or does not have ACC_PERMITS_VALUE set.  The fix was tested with Mach5 tiers 1-2 on Linux, Mac OS, and Windows and Mach5 tiers 3-5 on Linux x64.

Thanks, Harold

### DIFF
--- a/src/hotspot/share/classfile/classFileParser.cpp
+++ b/src/hotspot/share/classfile/classFileParser.cpp
@@ -1521,6 +1521,7 @@ class ClassFileParser::FieldAllocationCount : public ResourceObj {
 void ClassFileParser::parse_fields(const ClassFileStream* const cfs,
                                    bool is_interface,
                                    bool is_inline_type,
+                                   bool is_permits_value_class,
                                    FieldAllocationCount* const fac,
                                    ConstantPool* cp,
                                    const int cp_size,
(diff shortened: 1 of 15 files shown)


src/hotspot/share/classfile/classFileParser.cpp
```diff
@@ -4582,6 +4586,14 @@ void ClassFileParser::check_super_class_access(const InstanceKlass* this_klass,
...
     }
 
+    // A value class's super must be either java.lang.Object or have ACC_PERMITS_VALUE set.
+    if (this_klass->access_flags().is_value_class() &&
+        super_ik->name() != vmSymbols::java_lang_Object() &&
+        !super_ik->is_permits_value_class()) {
+      classfile_icce_error("value class %s cannot inherit from class %s", super_ik, THREAD);
+      return;
+    }
+
```
fparain, line 4596: This is not the behavior written in the JVMS draft for value classes, which requires the direct super class of a value class to have the ACC_PERMITS _VALUE flag set. However, because we haven't changed java.lang.Object into an abstract class, java.lang.Object cannot have the ACC_PERMITS _VALUE flag set. Currently, I see no other solution than the one you have implemented, adding a special case when the direct super class is java.lang.Object. But it would be useful to add a comment in the source code and the CR to remember and track this discrepancy.

hseigel: Thanks for suggesting this.  I added a comment.

PR #635, opened 2022-02-08
8281013: Errors.PrimitiveClassDoesNotSupport diagnostic should apply to value classes also

sadayapalam:
Diagnosevalue classes accessing various jlO methods that don't work with B2/B3 classes


biboudis (approved):
Looks good! Should we reference 8237066 as well in the PR for administrative purposes? (I noticed it is referred in the JTREG comment).

PR #640, opened 2022-02-08
8281463: [lworld] VALUE / PRIMITIVE modifiers should be supported by reflection

RogerRiggs:
Update java.lang.reflection.Modifier to reflect the JVMS specified bits in the modifier.

The re-use of bits between method and class modifiers poses a problem for this API.
Previously, all bits were distinct and the modifiers do not include any indication of whether
the modifier bits are from a class, field, or method. As a result there is ambiguity as two which
of two meanings a bit takes on.  NATIVE vs VALUE and STRICT vs PRIMITIVE.

Please review for discussion purposes.

---------
### Progress

### Issue
 * [JDK-8281463](https://bugs.openjdk.java.net/browse/JDK-8281463): [lworld] VALUE / PRIMITIVE modifiers should be supported by reflection

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/valhalla pull/640/head:pull/640` \
`$ git checkout pull/640`

`$ git checkout pull/640` \
`$ git pull https://git.openjdk.java.net/valhalla pull/640/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 640`

`$ git pr show -t 640`

</details>

<a href="https://git.openjdk.java.net/valhalla/pull/640.diff">https://git.openjdk.java.net/valhalla/pull/640.diff</a>

</details>

### DIFF
--- a/src/java.base/share/classes/java/lang/Class.java
+++ b/src/java.base/share/classes/java/lang/Class.java
@@ -202,9 +202,6 @@ public final class Class<T> implements java.io.Serializable,
     private static final int ANNOTATION = 0x00002000;
     private static final int ENUM       = 0x00004000;
     private static final int SYNTHETIC  = 0x00001000;
-    private static final int VALUE_CLASS     = 0x00000100;
-    private static final int PERMITS_VALUE   = 0x00000040;
-    private static final int PRIMITIVE_CLASS = 0x00000800;
 
     private static native void registerNatives();
     static {
@@ -625,7 +622,7 @@ public final class Class<T> implements java.io.Serializable,
      * @since Valhalla
      */
     public boolean isPrimitiveClass() {
-        return (this.getModifiers() & PRIMITIVE_CLASS) != 0;
+        return (this.getModifiers() & Modifier.PRIMITIVE) != 0;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


mlchung:
`ACC_VALUE`, `ACC_PRIMITIVE`, and `ACC_PERMITS_VALUE` are class modifiers.   Not every class modifiers defined in JVMS  is defined in the `java.lang.reflect.Modifier` API for example ACC_ENUM, ACC_ANNOTATION and ACC_MODULE.   But instead a test method such as `Class::isEnum`, `Class::isAnnotation` is provided in the `java.lang.Class` API.

For value and primitive class, a new `Class::isValue` and `Class::isPrimitiveClass` API (subject to the final terminology for B2 and B3) would suffice.   I'm not sure yet if an API is needed for `ACC_PERMITS_VALUE` since this is mostly used by JVM.

ExE-Boss:
I personally think that extending the upcoming `AccessFlag`[^1] enum is a better option.

[^1]: https://github.com/openjdk/jdk/pull/7445

liach:
> I personally think that extending the upcoming `AccessFlag` enum is a better option.

I recall that pr was created exactly for valhalla: that in valhalla, class modifier start to use bits used by method/parameter-exclusive modifiers, which made that issue more urgent.

RogerRiggs:
Withdrawing this PR;  the new modifier/access flags will be added to the new AccessFlags class.
Look for a replacement PR.

PR #654, opened 2022-02-18
fixing bugs in inference and type system more regression tests

vicente-romero-oracle:
this PR is adapting Types::glb so that it can deal with primitive classes. It also fixes a bug in inference, basically we are now failing during overload resolution if any undetVar has a bound that is a primitive class. But a bug was making inference fail after overload resolution in some cases.

---------
### Progress

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/valhalla pull/654/head:pull/654` \
`$ git checkout pull/654`

`$ git checkout pull/654` \
`$ git pull https://git.openjdk.java.net/valhalla pull/654/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 654`

`$ git pr show -t 654`

</details>

<a href="https://git.openjdk.java.net/valhalla/pull/654.diff">https://git.openjdk.java.net/valhalla/pull/654.diff</a>

</details>

### DIFF
--- a/src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
+++ b/src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
@@ -4390,9 +4390,9 @@ public class Types {
             return t;
         else if (t.isPrimitive() || s.isPrimitive())
             return syms.errType;
-        else if (isSubtypeNoCapture(t, s))
+        else if (isBoundedBy(t, s, (t1, s1, w) -> isSubtypeNoCapture(t1, s1)))
             return t;
-        else if (isSubtypeNoCapture(s, t))
+        else if (isBoundedBy(s, t, (s1, t1, w) -> isSubtypeNoCapture(s1, t1)))
             return s;
 
         List<Type> closure = union(closure(t), closure(s));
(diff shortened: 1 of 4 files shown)


mcimadamore (commented):
Looks good. I think we should add more tests - e.g. you want to check that, given two primitive types P1 and P2 which both extend a common superinterface I, then:
* lub(P1, P2) == lub(P1, I) == lub(P2, I) == I.
* lub(P1, P1.ref) = P1.ref
* glb(P1, P1.ref) = P1
* glb(P1, I) == P1
* glb(P2, I) == P2

vicente-romero-oracle:
> Looks good. I think we should add more tests - e.g. you want to check that, given two primitive types P1 and P2 which both extend a common superinterface I, then:
> 
>     * lub(P1, P2) == lub(P1, I) == lub(P2, I) == I.
> 
>     * lub(P1, P1.ref) = P1.ref
> 
>     * glb(P1, P1.ref) = P1
> 
>     * glb(P1, I) == P1
> 
>     * glb(P1.ref, I) == P1.ref
> 
>     * glb(P2, I) == P2
> 
>     * glb(P2.ref, I) == P2.ref



> :

sure, I will add more tests before pushing, thanks

PR #661, opened 2022-03-02
8282107: Valhalla: javac does not correctly recognize value class factory <init>

sadayapalam:
Cope with the ill effects of javac downgrading the value class file at lower source levels leaving fuller treatment to JDK-8282525

---------
### Progress

### Issue
 * [JDK-8282107](https://bugs.openjdk.java.net/browse/JDK-8282107): Valhalla: javac does not correctly recognize value class factory <init>

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/valhalla pull/661/head:pull/661` \
`$ git checkout pull/661`

`$ git checkout pull/661` \
`$ git pull https://git.openjdk.java.net/valhalla pull/661/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 661`

`$ git pr show -t 661`

</details>

<a href="https://git.openjdk.java.net/valhalla/pull/661.diff">https://git.openjdk.java.net/valhalla/pull/661.diff</a>

</details>

### DIFF
--- a/src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/ClassReader.java
+++ b/src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/ClassReader.java
@@ -1007,7 +1007,8 @@ public class ClassReader {
                         //- System.err.println(" # " + sym.type);
                         if (sym.kind == MTH && sym.type.getThrownTypes().isEmpty())
                             sym.type.asMethodType().thrown = thrown;
-                        if (sym.kind == MTH  && sym.name == names.init && sym.owner.isValueClass()) {
+                        // Map value class factory methods back to constructors for the benefit of earlier pipeline stages
+                        if (sym.kind == MTH  && sym.name == names.init && !sym.type.getReturnType().hasTag(TypeTag.VOID)) {
                             sym.type = new MethodType(sym.type.getParameterTypes(),
                                     syms.voidType,
                                     sym.type.getThrownTypes(),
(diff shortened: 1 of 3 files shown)


src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/ClassReader.java
```diff
@@ -1007,7 +1007,8 @@ protected void read(Symbol sym, int attrLen) {
                         //- System.err.println(" # " + sym.type);
                         if (sym.kind == MTH && sym.type.getThrownTypes().isEmpty())
                             sym.type.asMethodType().thrown = thrown;
-                        if (sym.kind == MTH  && sym.name == names.init && sym.owner.isValueClass()) {
+                        // Map value class factory methods back to constructors for the benefit of earlier pipeline stages
```
sadayapalam, line 1010: At source levels less than the one where the value class feature is supported, javac incorrectly downgrades a value class to an ordinary class. (See the method com.sun.tools.javac.jvm.ClassReader#adjustClassFlags which internalizes the flags from a class file's header). As a result isValueClass returns false and we don't restore the correct return type of the constructor. This leads to the present problem. The present fix works around by not relying on isValueClass - the only instance where a constructor may have a non-Void return is the case of value class factory, so the check is redundant. 

https://bugs.openjdk.java.net/browse/JDK-8282525 has been raised for a fuller fix. 

A nuance for the reviewers: Source level constructors are lowered into static factory methods late in the pipeline stage and are remapped into constructors when read from class files - this is so as to minimize the changes to the compiler. Rather than know and deal with an entity called static value factory methods, most of the compiler simply works with constructors.

biboudis: Thanks for the extra explanation!!!

biboudis (approved):
Looks good!

PR #663, opened 2022-03-02
addressing a NPE in the compiler due to empty warnStack

vicente-romero-oracle:
Fixing a compiler crash plus other review comments from Maurizio

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/valhalla pull/663/head:pull/663` \
`$ git checkout pull/663`

`$ git checkout pull/663` \
`$ git pull https://git.openjdk.java.net/valhalla pull/663/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 663`

`$ git pr show -t 663`

</details>

<a href="https://git.openjdk.java.net/valhalla/pull/663.diff">https://git.openjdk.java.net/valhalla/pull/663.diff</a>

</details>

### DIFF
--- a/src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
+++ b/src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
@@ -1084,9 +1084,22 @@ public class Types {
      * Is t an unchecked subtype of s?
      */
     public boolean isSubtypeUnchecked(Type t, Type s, Warner warn) {
-        boolean result = isSubtypeUncheckedInternal(t, s, true, warn);
-        if (result) {
-            checkUnsafeVarargsConversion(t, s, warn);
+        boolean result = false;
+        if (warn != warnStack.head) {
+            try {
+                warnStack = warnStack.prepend(warn);
+                result = isSubtypeUncheckedInternal(t, s, true, warn);
+                if (result) {
+                    checkUnsafeVarargsConversion(t, s, warn);
+                }
+            } finally {
+                warnStack = warnStack.tail;
+            }
+        } else {
+            result = isSubtypeUncheckedInternal(t, s, true, warn);
+            if (result) {
+                checkUnsafeVarargsConversion(t, s, warn);
+            }
         }
         return result;
     }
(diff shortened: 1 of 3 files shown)


src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
```diff
@@ -1672,8 +1685,18 @@ public boolean containsType(Type t, Type s) {
             public Boolean visitType(Type t, Type s) {
                 if (s.isPartial())
                     return containedBy(s, t);
-                else
-                    return isSameType(t, s);
+                else {
+                    boolean result = isSameType(t, s);
+                    // warnStack.head is != null if we are checking for an assignment, in other cases we should be strict
+                    // the order in the condition below matters
+                    if (warnStack.head != null && allowUniversalTVars && !result) {
```
vicente-romero-oracle, line 1692: using the emptiness of the warnStack as the way to know if we are dealing with an assignment or not seems a bit fragile, another option could be to pass down a boolean all the way down from all invocations, what do you think?

mcimadamore: Do you mean, e.g. inside overload resolution, how do we distinguish between when we can apply the "loose" check, and when we need the strict check?

If so, my feeling here is that as we have isSubtype vs. isSubtypeUnchecked we also need isSameTypes vs. isSameTypesUnchecked and a subtyping routine should call the "correct" isSameTypes visitor/routine.

mcimadamore (commented):
My suggestion is to fix the crash first, and then file another PR for adding support for nested unchecked null warnings, as that is probably a biggie feature.

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
```diff
@@ -1727,8 +1750,12 @@ public Boolean visitUndetVar(UndetVar t, Type s) {
             public Boolean visitTypeVar(TypeVar t, Type s) {
                 if (s.hasTag(TYPEVAR)) {
                     TypeVar other = (TypeVar)s;
-                    if (allowUniversalTVars && t.isValueProjection() != other.isValueProjection() && t.tsym == other.tsym)
+                    if (allowUniversalTVars && t.isValueProjection() != other.isValueProjection() && t.tsym == other.tsym) {
+                        if (warnStack.head != null) {
```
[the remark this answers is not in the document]
vicente-romero-oracle: correct

vicente-romero-oracle:
> My suggestion is to fix the crash first, and then file another PR for adding support for nested unchecked null warnings, as that is probably a biggie feature.

sounds good I will do that

mcimadamore (approved):
The fix makes sense, and seems consistent with what we do with isCastable.

PR #666, opened 2022-03-03
bug: no warning shown for parameterized conversions

vicente-romero-oracle:
no warning is shown for parameterized conversions

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/valhalla pull/666/head:pull/666` \
`$ git checkout pull/666`

`$ git checkout pull/666` \
`$ git pull https://git.openjdk.java.net/valhalla pull/666/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 666`

`$ git pr show -t 666`

</details>

<a href="https://git.openjdk.java.net/valhalla/pull/666.diff">https://git.openjdk.java.net/valhalla/pull/666.diff</a>

</details>

### DIFF
--- a/src/jdk.compiler/share/classes/com/sun/tools/javac/code/Type.java
+++ b/src/jdk.compiler/share/classes/com/sun/tools/javac/code/Type.java
@@ -2029,12 +2029,6 @@ public abstract class Type extends AnnoConstruct implements TypeMirror, PoolCons
         public boolean isValueProjection() {
             return isUniversal() && !isReferenceProjection();
         }
-
-        public Type withTypeVar(Type t) {
-            return t.hasTag(TYPEVAR) && t.isReferenceProjection() && t == projection ?
-                    referenceProjection() :
-                    this;
-        }
     }
 
     /** A captured type variable comes from wildcards which can have
@@ -2100,10 +2094,6 @@ public abstract class Type extends AnnoConstruct implements TypeMirror, PoolCons
             sb.append(" of ");
             sb.append(wildcard);
             return sb.toString();
-        }
-
-        public Type withTypeVar(Type t) {
-            return this;
         }
     }
 
--- a/src/jdk.compiler/share/classes/com/sun/tools/javac/code/Type.java
+++ b/src/jdk.compiler/share/classes/com/sun/tools/javac/code/Type.java
@@ -2029,6 +2029,12 @@ public abstract class Type extends AnnoConstruct implements TypeMirror, PoolCons
         public boolean isValueProjection() {
             return isUniversal() && !isReferenceProjection();
         }
+
+        public Type withTypeVar(Type t) {
+            return t.hasTag(TYPEVAR) && t.isReferenceProjection() && t == projection ?
+                    referenceProjection() :
+                    this;
+        }
     }
 
     /** A captured type variable comes from wildcards which can have
@@ -2094,6 +2100,10 @@ public abstract class Type extends AnnoConstruct implements TypeMirror, PoolCons
             sb.append(" of ");
             sb.append(wildcard);
             return sb.toString();
+        }
+
+        public Type withTypeVar(Type t) {
+            return this;
         }
     }
 
--- a/src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
+++ b/src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
@@ -1121,7 +1121,7 @@ public class Types {
                         return false;
                     return true;
                 }
-            } else if (isSubtype(t, s, capture)) {
+            } else if (isSubtype(t, s, capture, new SubtypingParameters(true, isSubtypeUncheckedRelation, containsTypeUnchecked))) {
                 return true;
             } else if (allowUniversalTVars && t.hasTag(TYPEVAR) && s.hasTag(TYPEVAR) && t.tsym == s.tsym) {
                 warn.warn(LintCategory.UNIVERSAL);
@@ -1175,9 +1175,15 @@ public class Types {
         return isSubtype(t, s, false);
     }
     public boolean isSubtype(Type t, Type s, boolean capture) {
+        return isSubtype(t, s, capture, new SubtypingParameters(false, isSubtype, containsType));
+    }
+
+    record SubtypingParameters(boolean uncheckedAllowed, TypeRelation subtypingRelation, TypeRelation containmentRelation) {}
+
+    public boolean isSubtype(Type t, Type s, boolean capture, SubtypingParameters subtypingParameters) {
         if (t.equalsIgnoreMetadata(s))
             return true;
-        if (t.hasTag(TYPEVAR) && s.hasTag(TYPEVAR) && t.tsym == s.tsym) {
+        if (subtypingParameters.uncheckedAllowed() && t.hasTag(TYPEVAR) && s.hasTag(TYPEVAR) && t.tsym == s.tsym) {
             // warnStack.head is != null if we are checking for an assignment
             if (warnStack.head != null && allowUniversalTVars && t.isReferenceProjection() != s.isReferenceProjection()) {
                 warnStack.head.warn(LintCategory.UNCHECKED);
@@ -1189,7 +1195,7 @@ public class Types {
 
         if (s.isCompound()) {
             for (Type s2 : interfaces(s).prepend(supertype(s))) {
-                if (!isSubtype(t, s2, capture))
+                if (!isSubtype(t, s2, capture, subtypingParameters))
                     return false;
             }
             return true;
@@ -1202,15 +1208,16 @@ public class Types {
             // TODO: JDK-8039198, bounds checking sometimes passes in a wildcard as s
             Type lower = cvarLowerBound(wildLowerBound(s));
             if (s != lower && !lower.hasTag(BOT))
-                return isSubtype(capture ? capture(t) : t, lower, false);
+                return isSubtype(capture ? capture(t) : t, lower, false, subtypingParameters);
         }
 
         t = capture ? capture(t) : t;
-        return isSubtype.visit(t, s);
+        return subtypingParameters.subtypingRelation().visit(t, s);
     }
     // where
-        private TypeRelation isSubtype = new TypeRelation()
-        {
+        IsSubtypeRelation isSubtype = new IsSubtypeRelation();
+
+        class IsSubtypeRelation extends TypeRelation {
             @Override
             public Boolean visitType(Type t, Type s) {
                 switch (t.getTag()) {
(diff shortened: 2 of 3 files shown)


vicente-romero-oracle:
> I'm not too convinced by this fix. Using warnStack seems really a workaround, and, on top of my head, I can't think of other code doing this. IMHO, universal vars introduce a new relation between types: `isBoundedBy`. What you want here is, essentially, a form to do `isBoundedByUnchecked` which should do strict check, followed by a looser one (like isSubtypeUnchecked does). There are many ways to do this, but it feels that having a better separation between these different routines could be, in the long term, a simplifying move (even if, in the short term it will create some code duplication). Because then we can fix the callsites to use the relationship we think is correct, depending on the case.

I'm not sure that `isSubtype` / `isSubtypeUnchecked` is the right analogy here as in most cases we get to the place where the warning is being issue in this PR without going through `isBoundedBy`

mcimadamore:
> > I'm not too convinced by this fix. Using warnStack seems really a workaround, and, on top of my head, I can't think of other code doing this. IMHO, universal vars introduce a new relation between types: `isBoundedBy`. What you want here is, essentially, a form to do `isBoundedByUnchecked` which should do strict check, followed by a looser one (like isSubtypeUnchecked does). There are many ways to do this, but it feels that having a better separation between these different routines could be, in the long term, a simplifying move (even if, in the short term it will create some code duplication). Because then we can fix the callsites to use the relationship we think is correct, depending on the case.
> 
> I'm not sure that `isSubtype` / `isSubtypeUnchecked` is the right analogy here as in most cases we get to the place where the warning is being issue in this PR without going through `isBoundedBy`

I got confused, sorry (too many type relations!). Looking at the code some more, it seems to me that what we want is to enhance `isSubtypeUnchecked`, right? E.g. `isSubtypeUnchecked` is already called in places where the compiler allows unchecked conversions. But we need to add a new kind of unchecked conversion, which requires some recursion (because isSubtypeUnchecked can go to containsType which then can go back to isSubtype - not unchecked!).

So, it seems to me that the subtyping routine has to be parameterized by the kind of containment routine we want to use. In the normal case, we'd just use Types::containsType - but if we want to check for nested unchecked warnings, we need to have a version of Types::containsType that is more flexible.

In other words, I think what we need is a Types::containsTypeUnchecked - which we can use as a parameter to the subtyping routine.

vicente-romero-oracle:
> > > I'm not too convinced by this fix. Using warnStack seems really a workaround, and, on top of my head, I can't think of other code doing this. IMHO, universal vars introduce a new relation between types: `isBoundedBy`. What you want here is, essentially, a form to do `isBoundedByUnchecked` which should do strict check, followed by a looser one (like isSubtypeUnchecked does). There are many ways to do this, but it feels that having a better separation between these different routines could be, in the long term, a simplifying move (even if, in the short term it will create some code duplication). Because then we can fix the callsites to use the relationship we think is correct, depending on the case.
> > 
> > 
> > I'm not sure that `isSubtype` / `isSubtypeUnchecked` is the right analogy here as in most cases we get to the place where the warning is being issue in this PR without going through `isBoundedBy`
> 
> I got confused, sorry (too many type relations!). Looking at the code some more, it seems to me that what we want is to enhance `isSubtypeUnchecked`, right? E.g. `isSubtypeUnchecked` is already called in places where the compiler allows unchecked conversions. But we need to add a new kind of unchecked conversion, which requires some recursion (because isSubtypeUnchecked can go to containsType which then can go back to isSubtype - not unchecked!).
> 
> So, it seems to me that the subtyping routine has to be parameterized by the kind of containment routine we want to use. In the normal case, we'd just use Types::containsType - but if we want to check for nested unchecked warnings, we need to have a version of Types::containsType that is more flexible.
> 
> In other words, I think what we need is a Types::containsTypeUnchecked - which we can use as a parameter to the subtyping routine.

I will try this, I was going through other way, to untangle some of the type routines, the good thing is that I already have a separate containtType relation for the unchecked case so I will build from there

vicente-romero-oracle:
first attempt at the type routines parameterization approach, what do you think?

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
```diff
@@ -4864,7 +4946,7 @@ private boolean giveWarning(Type from, Type to) {
 
     private boolean containsTypeEquivalent(Type t, Type s) {
         return isSameType(t, s) || // shortcut
-            containsType(t, s) && containsType(s, t);
+            containsType(t, s, containsTypeUnchecked) && containsType(s, t, containsTypeUnchecked);
```
mcimadamore, line 4949: Is this the right replacement? It seems to me that the answer is "it depends". This is called both by "hasSameArgs" which is used to determine overriding (in which case a loose semantics could be ok), but also by the isSameType visitor - in which case the loose semantics should only be enabled conditionally.

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Type.java
```diff
@@ -2029,6 +2029,12 @@ public boolean isReferenceProjection() {
         public boolean isValueProjection() {
             return isUniversal() && !isReferenceProjection();
         }
+
+        public Type withTypeVar(Type t) {
+            return t.hasTag(TYPEVAR) && t.isReferenceProjection() && t == projection ?
```
mcimadamore, line 2034: Uhm - this method is typically used to set type-variables on wildcard types. It is a bit of a corner case in the compiler code, and I think it would be better to leave this alone. What is this code trying to do - and can the same be achieved in another way?

vicente-romero-oracle: this code is preserving the reference projection of a type var, if it was originally declared as a reference `T.ref`, instead of overriding it with its dual `T`. If the reference projection is not preserved then it leads to information loss. I haven't found another more reliable way

vicente-romero-oracle: OK I will do some more research here to see what alternatives we have, using this method is convenient as it is invoked whenever we do a mapping like for example from `Resolve::rawInstantiate` when we invoke `Types::memberType`, but I can try to find another way probably using a separate method to add the behavior I would like

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
```diff
@@ -1237,7 +1250,7 @@ public Boolean visitType(Type t, Type s) {
 
             private Set<TypePair> cache = new HashSet<>();
 
-            private boolean containsTypeRecursive(Type t, Type s) {
+            public boolean containsTypeRecursive(Type t, Type s) {
```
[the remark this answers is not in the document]
vicente-romero-oracle: sure makes sense

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
vicente-romero-oracle: yep you are right, fixed

vicente-romero-oracle:
I have uploaded another commit, I think that it addresses all your comments so far, thanks a lot for the feedback!

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
```diff
@@ -1502,6 +1489,11 @@ public boolean isSignaturePolymorphic(MethodSymbol msym) {
      * Is t the same type as s?
      */
     public boolean isSameType(Type t, Type s) {
+        return isSameType(t, s, containsType);
+    }
+
+    public boolean isSameType(Type t, Type s, ContainsType containmentRelation) {
+        isSameTypeVisitor.setContainmentRel(containmentRelation);
```
mcimadamore, line 1496: This should be reset after the call?

mcimadamore (commented):
I think the pieces are all there. When looking at the code, the impression I have is that we're doing a lot of work just to pass some kind of boolean or enum flag which tells us if some extra loose behavior should be allowed or not (in both the subtyping, containment and sameType visitors).

I was about to suggest that in the previous review rounds, but then held back because the existing visitors cannot take external parameters. I think your changes to the `isSameType` visitor demonstrate that problem: the visitor is just a function on types, and there's no way to pass an extra boolean/enum to it - so you have to store it in a field, which then leads to issues because either (i) you create one visitor instance per subtyping check (which would be very expensive, I fear) or you resort to tricks using mutable fields.

Unfortunately fixing this mismatch is hard, as we're already using the visitor parameter to pass the second type of these type relations. So some other kind of workaround is required.

A possible suggestion for trying to get to a simpler form would be to introduce a new "parameterized" relation, like this:

```
public abstract static class ParameterizedTypeRelation<P> extends TypeRelation {
        P param;

        public final Boolean visit(Type t, Type s, P param) {
            P prevParam = this.param;
            try {
                this.param = param;
                return visit(t, s);
            } finally {
                this.param = prevParam;
            }
        }
    }
```

This basically generalizes the trick you did with `isSameType` (but resets the field value correctly after a visit). Then you make isSubtype/contains/isSameType subclasses of this visitor class, and define an enum with two values for the strict/loose behavior.

I believe this should give you enough flexibility to do what you want, and, at the same time, by minimizing set of changes compared to mainline.

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
vicente-romero-oracle: right

vicente-romero-oracle:
> I think the pieces are all there. When looking at the code, the impression I have is that we're doing a lot of work just to pass some kind of boolean or enum flag which tells us if some extra loose behavior should be allowed or not (in both the subtyping, containment and sameType visitors).
> 
> I was about to suggest that in the previous review rounds, but then held back because the existing visitors cannot take external parameters. I think your changes to the `isSameType` visitor demonstrate that problem: the visitor is just a function on types, and there's no way to pass an extra boolean/enum to it - so you have to store it in a field, which then leads to issues because either (i) you create one visitor instance per subtyping check (which would be very expensive, I fear) or you resort to tricks using mutable fields.
> 
> Unfortunately fixing this mismatch is hard, as we're already using the visitor parameter to pass the second type of these type relations. So some other kind of workaround is required.
> 
> A possible suggestion for trying to get to a simpler form would be to introduce a new "parameterized" relation, like this:
> 
> ```
> public abstract static class ParameterizedTypeRelation<P> extends TypeRelation {
>         P param;
> 
>         public final Boolean visit(Type t, Type s, P param) {
>             P prevParam = this.param;
>             try {
>                 this.param = param;
>                 return visit(t, s);
>             } finally {
>                 this.param = prevParam;
>             }
>         }
>     }
> ```
> 
> This basically generalizes the trick you did with `isSameType` (but resets the field value correctly after a visit). Then you make isSubtype/contains/isSameType subclasses of this visitor class, and define an enum with two values for the strict/loose behavior.
> 
> I believe this should give you enough flexibility to do what you want, and, at the same time, by minimizing set of changes compared to mainline.

sure I will try this, thanks

vicente-romero-oracle:
I have updated the PR with the refactoring you proposed, thanks

mcimadamore (approved):
Looks good - thanks for the hard work!

vicente-romero-oracle:
> Looks good - thanks for the hard work!

nice, thanks to you for the feedback!

PR #684, opened 2022-04-15
addressing several issues in the implementation of universal tvars

vicente-romero-oracle:
Last review iteration found several issues in the current implementation, addressing them here

---------
### Progress

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/valhalla pull/684/head:pull/684` \
`$ git checkout pull/684`

`$ git checkout pull/684` \
`$ git pull https://git.openjdk.java.net/valhalla pull/684/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 684`

`$ git pr show -t 684`

</details>

<a href="https://git.openjdk.java.net/valhalla/pull/684.diff">https://git.openjdk.java.net/valhalla/pull/684.diff</a>

</details>


src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
```diff
@@ -600,57 +600,94 @@ public Type visitErrorType(ErrorType t, Symbol sym) {
     // </editor-fold>
 
     // <editor-fold defaultstate="collapsed" desc="isConvertible">
+    enum IsConvertibleResult {
+        CONVERTIBLE,
+        CONVERTIBLE_REF_PROJECTION,
+        NOT_CONVERTIBLE
```
ExE-Boss, line 606:
```suggestion
        NOT_CONVERTIBLE;

        static IsConvertibleResult of(boolean isConvertible) {
            return convertible ?
                CONVERTIBLE :
                NOT_COVERTIBLE;
        }

        static IsConvertibleResult ofRefProjection(boolean isConvertible) {
            return convertible ?
                CONVERTIBLE_REF_PROJECTION :
                NOT_COVERTIBLE;
        }
```

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
```diff
@@ -600,57 +600,94 @@ public Type visitErrorType(ErrorType t, Symbol sym) {
...
+                        t.hasTag(CLASS) && supertype(t).isPrimitiveClass() :
+                        s.hasTag(CLASS) && supertype(s).isPrimitiveClass();
+                if (!anonymousPrimitiveClass) {
+                    warn.warn(LintCategory.UNCHECKED);
+                }
+                if (discriminateConvertibleKind)
+                    return IsConvertibleResult.CONVERTIBLE_REF_PROJECTION;
             }
-            return result;
+            return result ? IsConvertibleResult.CONVERTIBLE : IsConvertibleResult.NOT_CONVERTIBLE;
```
ExE-Boss, line 651:
```suggestion
            return IsConvertibleResult.of(result);
```

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
```diff
@@ -600,57 +600,94 @@ public Type visitErrorType(ErrorType t, Symbol sym) {
...
+            boolean sHasValue = s != sRefProjection;
+
+            if (tHasValue != sHasValue) {
+                result = tHasValue ?
+                        isSubtype(allowUniversalTVars && (tUndet || sUndet) ? t : tRefProjection, s) :
+                        !t.hasTag(BOT) && isSubtype(t, allowUniversalTVars && (tUndet || sUndet) ? s : sRefProjection);
+                if (result && (allowUniversalTVars && !t.hasTag(BOT))) {
+                    warn.warn(LintCategory.UNCHECKED);
+                }
+                return result ? IsConvertibleResult.CONVERTIBLE_REF_PROJECTION : IsConvertibleResult.NOT_CONVERTIBLE;
```
ExE-Boss, line 671:
```suggestion
                return IsConvertibleResult.ofRefProjection(result);
```

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
```diff
@@ -600,57 +600,94 @@ public Type visitErrorType(ErrorType t, Symbol sym) {
...
                     isSubtype(t, boxedTypeOrType(s)) :
                     isSubtype(boxedTypeOrType(t), s);
+            return result ? IsConvertibleResult.CONVERTIBLE : IsConvertibleResult.NOT_CONVERTIBLE;
         }
 
-        return tPrimitive
+        boolean result = tPrimitive
             ? isSubtype(boxedClass(t).type, s)
             : isSubtype(unboxedType(t), s);
+        return result ? IsConvertibleResult.CONVERTIBLE : IsConvertibleResult.NOT_CONVERTIBLE;
```
ExE-Boss, line 686:
```suggestion
        return IsConvertibleResult.of(tPrimitive
            ? isSubtype(boxedClass(t).type, s)
            : isSubtype(unboxedType(t), s));
```

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
```diff
@@ -600,57 +600,94 @@ public Type visitErrorType(ErrorType t, Symbol sym) {
...
+            }
+            return IsConvertibleResult.NOT_CONVERTIBLE;
         }
 
         if (tUndet || sUndet) {
-            return tUndet ?
+            boolean result = tUndet ?
                     isSubtype(t, boxedTypeOrType(s)) :
                     isSubtype(boxedTypeOrType(t), s);
+            return result ? IsConvertibleResult.CONVERTIBLE : IsConvertibleResult.NOT_CONVERTIBLE;
```
ExE-Boss, line 680:
```suggestion
            return IsConvertibleResult.of(tUndet ?
                    isSubtype(t, boxedTypeOrType(s)) :
                    isSubtype(boxedTypeOrType(t), s));
```

ExE-Boss (commented):
Maybe instead of repeated ternaries, this could use a helper method for the `boolean` to `IsConvertibleResult` conversion:

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
vicente-romero-oracle: yep, could be thanks for the suggestion

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
```diff
@@ -1772,10 +1809,31 @@ public Boolean visitWildcardType(WildcardType t, Type s) {
...
+
+            boolean checkIfBoundedBy(WildcardType t, Type s) {
+                if (t.isUnbound()) return true;
+                if (t.isExtendsBound()) {
+                    boolean isBoundedBy = isBoundedBy(wildUpperBound(s), wildUpperBound(t),
+                            (t1, s1, w) -> isSubtype(t1, s1, false, SubtypingRelationKind.REF_VAL_ALLOWED));
+                    if (isBoundedBy) {
+                        return true;
+                    }
+                    return (allowUniversalTVars &&
```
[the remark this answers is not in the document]
mcimadamore: And, this is made messier by the fact that:

```
val = ref
```

In my example does not trigger a subtying test. It triggers an assignment conversion, which in turn can use the new form of parameterized conversions.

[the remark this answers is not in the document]
vicente-romero-oracle: > This extra check feels outside the spirit of the rules discussed in https://openjdk.java.net/jeps/8261529. That is, I'd expect type containment to recurse using `isBoundedBy` instead of `isSubtype` (as the first branch of this code does).

just to provide some context, the reason that supported in my view using this extra check comes from the universal tvars jep in particular where it says:

```
Changing a wildcard bound in a parameterized type from a universal type-variable type (T) or a primitive class type (Point) to its reference type (T.ref, Point.ref), or vice versa (where the conversion is not already allowed by subtyping):
```
this is why on a failure of isBoundedBy which at the end is based on subtyping I tried this more general relation. Although I realize from the additional points you mentioned here that these opens questions in other fronts like type well-formedness.

mcimadamore:
As discussed offline, rather than chasing individual bugs, a better approach might be *not* to use "loose" version of subtyping and type containment checks to issue warnings. Instead, the code for `isSubtypeUnchecked(S, T)` could work as follows:

1. first check if S <: T - if so, return `true`, otherwise
2. lift S to the same base type as T, using `Types.asSuper`, call the lifted type `S'`, then
3. if `S'` is raw, apply standard unchecked conversion, otherwise
4. perform a structural comparison between `S'` and `T` - if the two types are identical up to `.val`/`.ref` differences, issue an unchecked warning, and return `true`, otherwise return `false`.

I believe this has the potential to simplify the code quite a bit, since we would no longer need to duplicate subtyping and containment. Step (4) needs to be implemented in a new structural type visitor, but that's brand new code, and less likely to be problematic since it's only used for that specific purposes. Issues like reporting multiple diagnostic can easily be avoided under this scheme.

Since the type comparison doesn't really care about inference variables (`UndetVar`) I also do not think that there could be issues when it comes to (4) generating spurious inference constraints.

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
```diff
@@ -1167,17 +1140,196 @@ private boolean isSubtypeUncheckedInternal(Type t, Type s, boolean capture, Warn
...
+            public Boolean visitType(Type t, Type s) {
+                switch (t.getTag()) {
+                    case BYTE:
+                    case CHAR:
+                    case SHORT: case INT: case LONG:
+                    case FLOAT: case DOUBLE:
+                    case BOOLEAN: case VOID:
+                    case BOT: case NONE:
+                        return t.hasTag(s.getTag());
+                    case WILDCARD: //we shouldn't be here - avoids crash (see 7034495)
```
mcimadamore, line 1179: comment seems copy and paste error?

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
```diff
@@ -1167,17 +1140,196 @@ private boolean isSubtypeUncheckedInternal(Type t, Type s, boolean capture, Warn
...
             return false;
         }
+        // where
+        StructuralTypeComparator structuralComparator = new StructuralTypeComparator();
+        class StructuralTypeComparator extends TypeRelation {
+            private Set<TypePair> cache = new HashSet<>();
+
+            @Override
+            public Boolean visitType(Type t, Type s) {
+                switch (t.getTag()) {
```
mcimadamore, line 1171: In general - it seems like this routine has been brought in from isSubtype, which I think is wrong. This should borrow the most from isSameType. I believe in the case of this visitor, it would have been ok to simply return true if the two types have same tags? E.g. doesn't seem like adding the `switch` really adds much.

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
mcimadamore, line 1179: I do not see a visitor method for wildcards, but I think you need one, as you might get to wildcards recursively, because of `visitClassType`

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
```diff
@@ -1167,17 +1140,196 @@ private boolean isSubtypeUncheckedInternal(Type t, Type s, boolean capture, Warn
...
+                }
+            }
+
+            @Override
+            public Boolean visitArrayType(ArrayType t, Type s) {
+                if (t == s)
+                    return true;
+
+                if (s.isPartial())
+                    return visit(s, t);
```
mcimadamore, line 1192: Note sure these lines are relevant?

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
```diff
@@ -1167,17 +1140,196 @@ private boolean isSubtypeUncheckedInternal(Type t, Type s, boolean capture, Warn
...
+            @Override
+            public Boolean visitTypeVar(TypeVar t, Type s) {
+                if (s.hasTag(TYPEVAR)) {
+                    TypeVar other = (TypeVar)s;
+                    if (allowUniversalTVars &&
+                            t.isValueProjection() != other.isValueProjection() && t.tsym == other.tsym) {
+                        return true;
+                    }
+                }
+                return isSameType(t, s);
```
mcimadamore, line 1217: Here you delegate to isSameType... which seems odd - as this is the only place where we do this.

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
```diff
@@ -1167,17 +1140,196 @@ private boolean isSubtypeUncheckedInternal(Type t, Type s, boolean capture, Warn
...
+            }
+
+            public boolean compareTypeArgsRecursive(Type t, Type s) {
+                TypePair pair = new TypePair(t, s);
```
mcimadamore, line 1221: This is typically needed in subtyping routines, to prevent against cases where we could expand terms forever. But in the case of a structural comparison, this can never happen?

mcimadamore (commented):
I think the cleanup is very good - there are probably some other bits that can be cleaned up as well (see comments). I've added some comments on the new visitor, which seems more complex than it should be.

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
vicente-romero-oracle: agreed

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
vicente-romero-oracle: we analyze the arguments of every parameterized type inside of `visitClassType`, so I think we should be covered for the wildcard case?

vicente-romero-oracle: I think we should be covered here as visitors: `StructuralTypeComparator` and `TypeArgStructComparator` are interdependent, so I think we should be fine there, unless I'm missing something

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
```diff
@@ -1167,17 +1140,196 @@ private boolean isSubtypeUncheckedInternal(Type t, Type s, boolean capture, Warn
...
+                    return visit(s, t);
+
+                return s.hasTag(ARRAY)
+                        && visit(t.elemtype, elemtype(s));
+            }
+
+            @Override
+            public Boolean visitClassType(ClassType t, Type s) {
+                // If t is an intersection, sup might not be a class type
+                if (!t.hasTag(CLASS)) return visit(t, s);
```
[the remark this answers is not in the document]
vicente-romero-oracle: sure

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
vicente-romero-oracle: right, fixing it

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
vicente-romero-oracle: with wildcards yes

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
```diff
@@ -1523,19 +1642,15 @@ public boolean isSignaturePolymorphic(MethodSymbol msym) {
...
-        return isSameTypeVisitor.visit(t, s, subtypingKind);
+        return isSameTypeVisitor.visit(t, s);
     }
     // where
         /**
          * Type-equality relation - type variables are considered
          * equals if they share the same object identity.
          */
-        SameTypeVisitor isSameTypeVisitor = new SameTypeVisitor();
-        class SameTypeVisitor extends ParameterizedTypeRelation<SubtypingRelationKind> {
```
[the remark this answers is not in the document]
vicente-romero-oracle: I will remove it

vicente-romero-oracle:
I have updated the code after the last review comments, thanks!

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
```diff
@@ -1406,25 +1373,23 @@ enum SubtypingRelationKind {
             REF_VAL_ALLOWED
         }
 
-        public abstract static class ParameterizedTypeRelation<P> extends TypeRelation {
-            P param;
+        SubtypingRelation isSubtypeRelation = new SubtypingRelation();
+        class SubtypingRelation extends TypeRelation {
+            SubtypingRelationKind param;
```
mcimadamore, line 1378: What is this needed for? We should get into a position where subtyping is "just" subtyping, no need for having some kind of mode flag - unless I'm misunderstanding something?

vicente-romero-oracle: yep I will tackle that now, if I recall correctly I tried to removed it but there where other changes in areas a bit distant from the visitors that's why I didn't go for it in a first step

test/langtools/tools/javac/valhalla/lworld-values/universal-type-variables/UniversalTVarsCompilationTests.java
```diff
@@ -671,6 +662,17 @@ primitive class Point {}
                 }
                 """
         );
+        assertFail("compiler.err.prob.found.req",
+                """
+                import java.util.function.*;
+                class Test<__universal T> {
+                    T.ref field;
+                    void foo(T t, Consumer<? super T> action) {
+                        action.accept(field = t);
```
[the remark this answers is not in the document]
vicente-romero-oracle: well action is parameterized with a wildcard so here we are assigning a ref to a wildcard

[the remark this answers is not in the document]
vicente-romero-oracle: it could probably be better to write the test as:
```
field = t;
action.accept(field = t);
```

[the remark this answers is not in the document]
mcimadamore: Ah I see. So there's two things here:
1. assignment from `t` to `field` - that works ok (with warning)
2. passing `field` (of type `T.ref`) to `action`.

On (2), `action` is subject to capture conversion, so its type is `Consumer<#CAP>` where `T <: #CAP `. So, we have to check whether `T.ref` is assignable to `#CAP`. I believe this currently fails because `T.ref` cannot be "lifted" to `#CAP`.

I think this is ok for now, but I note that, when type variables are involved, there are always two ways to prove subtyping:

* T <: S iff upper(T) <: S
* S <: T iff S <: lower(T)

It is true that, in (2) you cannot "lift" T.ref to `'#CAP`, but it is also true that you can "unlift" #CAP down to T, at which point you would have a structural comparison between T.ref and T again (which would be ok).

Whether or not this is what we want to do is a specification question - we can revisit later.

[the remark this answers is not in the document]
vicente-romero-oracle: I see, thanks for the review

PR #694, opened 2022-05-19
Proptotype: Implement enum ValuePolicy for WeakHashMap with value objects as keys

RogerRiggs:
Prototype of j.u.WeakHashMap support for entries whose keys are value objects.

WeakHashMap relies on `java.lang.ref.WeakReference` for the retention of entries.
Value objects do not have identity and cannot be the referent in any `java.lang.ref.Reference` including `WeakReference`.
An attempt to use a value object as a key would throw `IllegalArgumentException`. Existing libraries that use `WeakHashMap` as caches may unexpectedly be rendered inoperative without some accommodation for value objects.
 
Keys that are value objects are handled and retained differently than identity objects. 
The retention of entries with keys that are value objects is selected using `WeakHashmap.ValuePolicy` when the `WeakHashMap` is created. The policy choices are: SOFT, STRONG, DISCARD, and THROW.

---------
### Progress

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/valhalla pull/694/head:pull/694` \
`$ git checkout pull/694`

`$ git checkout pull/694` \
`$ git pull https://git.openjdk.java.net/valhalla pull/694/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 694`

`$ git pr show -t 694`

</details>

<a href="https://git.openjdk.java.net/valhalla/pull/694.diff">https://git.openjdk.java.net/valhalla/pull/694.diff</a>

</details>


src/java.base/share/classes/java/util/WeakHashMap.java
```diff
@@ -36,13 +39,35 @@
...
  * When a key has been discarded its entry is effectively removed from the map,
  * so this class behaves somewhat differently from other {@code Map}
  * implementations.
  *
+ * <p>
+ * Keys that are {@linkplain Class#isValue() Value objects} do not have identity and cannot be
+ * the referent in any {@link java.lang.ref.Reference} including {@link WeakReference}.
+ * The retention of entries with keys that are Value objects is selected
+ * using {@link ValuePolicy} when the {@code WeakHashMap} is created.
+ * The default is {@link ValuePolicy#SOFT ValuePolicy.SOFT}, the entries are retained until
```
ExE-Boss, line 58:
```suggestion
 * The default is {@link ValuePolicy#SOFT}, the entries are retained until
```

src/java.base/share/classes/java/util/WeakHashMap.java
```diff
@@ -238,6 +286,15 @@ public WeakHashMap() {
         this(DEFAULT_INITIAL_CAPACITY, DEFAULT_LOAD_FACTOR);
     }
 
+    /**
+     * Constructs a new, empty {@code WeakHashMap} with the {@link ValuePolicy},
+     * the default initial capacity (16) and load factor (0.75).
+     * @param  valuePolicy     The {@link ValuePolicy} for keys that are Value objects; non-null
```
ExE-Boss, line 292: This should include `@throws NullPointerException`:
```suggestion
     *
     * @param  valuePolicy     The {@link ValuePolicy} for keys that are Value objects
     * @throws NullPointerException if {@code valuePolicy} is null
```

briangoetz:
Reading this code, it makes me think that we might want a common 
exception for "you did something identityful with a non-identity 
object."  All the XxxReference constructors will want this, locking will 
want this, and higher-level operations will want to play in the game 
too.  Currently locking throws IMSE, which is a reasonable thing for 
locking to do, but perhaps some sort of IdentityException would be more 
useful in helping users understand what they did wrong.

RogerRiggs:
I'll create a separate issue:   JDK-8287241
In addition to a new top level RuntimeException, I could also see a few methods in java.util.Objects like:
```
 public static <T> requireIdentity(T) {...}
 public static <T> requireValue(T) {...}
```

RogerRiggs:
The history is disturbed, try again.

PR #696, opened 2022-05-24
8287241: Add IdentityException to report that a value object is not valid

RogerRiggs:
Add class java.lang.IdentityException, when thrown indicates that an identity object was required but a value object was supplied.

Also add methods `java.util.Objects.requireIdentity(obj)` and `Objects.requireIdentity(obj, message)` to check and throw.

Updated java.lang.ref.Reference and related tests to use new exception and method.

---------
### Progress

### Issue
 * [JDK-8287241](https://bugs.openjdk.org/browse/JDK-8287241): Add IdentityException to report that a value object is not valid

### Reviewers
 * [Mandy Chung](https://openjdk.java.net/census#mchung) (@mlchung - Committer) ⚠️ Review applies to [0785a81f](https://git.openjdk.org/valhalla/pull/696/files/0785a81fee0508db005c6151b63636680bc280c7)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla pull/696/head:pull/696` \
`$ git checkout pull/696`

`$ git checkout pull/696` \
`$ git pull https://git.openjdk.org/valhalla pull/696/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 696`

`$ git pr show -t 696`

</details>

<a href="https://git.openjdk.org/valhalla/pull/696.diff">https://git.openjdk.org/valhalla/pull/696.diff</a>

</details>

### DIFF
--- a/src/java.base/share/classes/java/lang/IdentityException.java
+++ b/src/java.base/share/classes/java/lang/IdentityException.java
@@ -29,16 +29,18 @@ package java.lang;
  * Identity objects are required for synchronization and locking.
  * <a href="{@docRoot}/java.base/java/lang/doc-files/ValueBased.html">Value-based</a>
  * objects do not have identity and cannot be used for synchronization or locking.
+ *
+ * @since Valhalla
  */
 public class IdentityException extends RuntimeException {
     /**
-     * Create an IdentityException with no message.
+     * Create an {@code IdentityException} with no message.
      */
     public IdentityException() {
     }
 
     /**
-     * Create an IdentityException with the class name and default message.
+     * Create an {@code IdentityException} with the class name and default message.
      *
      * @param clazz the class of the object
      */
@@ -47,7 +49,7 @@ public class IdentityException extends RuntimeException {
     }
 
     /**
-     * Create an IdentityException with a message.
+     * Create an {@code IdentityException} with a message.
      *
      * @param  message the detail message; can be {@code null}
      */
@@ -56,7 +58,7 @@ public class IdentityException extends RuntimeException {
     }
 
     /**
-     * Create an IdentityException with a message and cause.
+     * Create an {@code IdentityException} with a message and cause.
      *
      * @param  message the detail message; can be {@code null}
      * @param  cause the cause; {@code null} is permitted, and indicates
--- a/src/java.base/share/classes/java/lang/IdentityException.java
+++ b/src/java.base/share/classes/java/lang/IdentityException.java
@@ -1,0 +1,71 @@
+/*
+ * Copyright (c) 2022, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+package java.lang;
+
+
+/**
+ * Thrown when an identity object is required but a value object is supplied.
+ * <p>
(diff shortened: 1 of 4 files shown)


ExE-Boss:
There should probably also be a `requireIdentity(T obj, Supplier<String> messageSupplier)` overload, like there is for `requireNonNull`:
https://github.com/openjdk/valhalla/blob/eb1682287408033bf1c7738a95cc11a5d73ff7aa/src/java.base/share/classes/java/util/Objects.java#L397-L402

src/java.base/share/classes/java/util/Objects.java
```diff
@@ -189,6 +189,66 @@ public static String toIdentityString(Object o) {
...
+     * @return {@code obj} if {@code obj} is an identity object
+     * @throws NullPointerException if {@code obj} is {@code null}
+     * @throws IdentityException if {@code obj} is not an identity object
+     * @since Valhalla
+     */
+    @ForceInline
+    public static <T> T requireIdentity(T obj) {
+        Objects.requireNonNull(obj);
+        var cl = obj.getClass();
+        if (cl.isValue())
```
liach, line 206: Just curious, is it that all instances of `Object` (from `new Object()`) are identity, and all other classes that are neither value nor identity are abstract classes?

RogerRiggs: `obj.getClass()` is always concrete and either identity or value. So abstract doesn't really come into play when talking about instances.
This test really should be `!cl.isIdentity()` but that method doesn't exist (yet, TBD soon)

Aabstract classes and interfaces can be either identity or value, or neither.  (But not both).

src/java.base/share/classes/java/lang/IdentityException.java
```diff
@@ -0,0 +1,71 @@
...
+/**
+ * Thrown when an identity object is required but a value object is supplied.
+ * <p>
+ * Identity objects are required for synchronization and locking.
+ * <a href="{@docRoot}/java.base/java/lang/doc-files/ValueBased.html">Value-based</a>
+ * objects do not have identity and cannot be used for synchronization or locking.
+ */
+public class IdentityException extends RuntimeException {
+    /**
+     * Create an IdentityException with no message.
```
mlchung, line 35:
```suggestion
     * Create an {@code IdentityException} with no message.
```

Same applies to the javadoc of all constructors.

src/java.base/share/classes/java/util/Objects.java
mlchung, line 206:
```suggestion
        if (obj.getClass().isValue())
```

Same as the implementation of other variants of the `requireIdentity` method.

src/java.base/share/classes/java/lang/IdentityException.java
```diff
@@ -0,0 +1,71 @@
...
+ */
+package java.lang;
+
+
+/**
+ * Thrown when an identity object is required but a value object is supplied.
+ * <p>
+ * Identity objects are required for synchronization and locking.
+ * <a href="{@docRoot}/java.base/java/lang/doc-files/ValueBased.html">Value-based</a>
+ * objects do not have identity and cannot be used for synchronization or locking.
```
mlchung, line 31: Need `@since Valhalla`

mlchung (approved):

src/java.base/share/classes/java/util/Objects.java
RogerRiggs: `cl` is used to throw the IdentityException in the following line.

mlchung: okay.  thanks

src/java.base/share/classes/java/lang/IdentityException.java
```diff
@@ -0,0 +1,73 @@
...
+    /**
+     * Create an {@code IdentityException} with a message and cause.
+     *
+     * @param  message the detail message; can be {@code null}
+     * @param  cause the cause; {@code null} is permitted, and indicates
+     *               that the cause is nonexistent or unknown.
+     */
+    public IdentityException(String message, Throwable cause) {
+        super(message, cause);
+    }
```
forax, line 69: There is a missing variant, the one with a Throwable but no message that you use when you want to throw an IdentityException as a result of a previous exception (by example an IllegalMonitorStateException).

RogerRiggs: It not clear where to draw the line on convenience methods (message can be null), but ok, added.

mlchung: Exceptions in the JDK do not always have all four standard constructors as in `Throwable`.  There was a discussion to audit the constructor choices for JDK exceptions.  Some new exceptions do add all 4 constructors (`IllegalCallerException` and `WrongThreadException`).   Since this one has a constructor taking the `cause` for exception chaining, it's reasonable to add the variant setting the default message to `cause.getMessage()`.

mlchung: Exceptions in the JDK do not always have all four standard constructors as in `Throwable`.  There was a discussion to audit the constructor choices for JDK exceptions.  Some new exceptions do add all 4 constructors (`IllegalCallerException` and `WrongThreadException`).   Since this one has a constructor taking the `cause` for exception chaining, it's reasonable to add the variant setting the default message to `cause.getMessage()`.

PR #698, opened 2022-05-26
8281463: [lworld] VALUE / PRIMITIVE modifiers should be supported by reflection

RogerRiggs:
Add support to java.lang.reflect.AccessFlag ACC_IDENTITY, ACC_VALUE, and ACC_PRIMITIVE.
Added corresponding flags to j.l.reflect.Modifier so the "source" of the AccessFlag is the Modifer mask.
Remove PERMITS_VALUE; moving the definition used by HotSpot to HotSpot code.

Add comments to java.l.reflect.Modifier suggesting use of AccessFlag instead and limitations.

---------
### Progress

### Issue
 * [JDK-8281463](https://bugs.openjdk.org/browse/JDK-8281463): [lworld] VALUE / PRIMITIVE modifiers should be supported by reflection

### Reviewers
 * [Mandy Chung](https://openjdk.org/census#mchung) (@mlchung - Committer) ⚠️ Review applies to [28ef42ff](https://git.openjdk.org/valhalla/pull/698/files/28ef42ffa081be0e367f2b398184789989aabf1a)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla pull/698/head:pull/698` \
`$ git checkout pull/698`

`$ git checkout pull/698` \
`$ git pull https://git.openjdk.org/valhalla pull/698/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 698`

`$ git pr show -t 698`

</details>

<a href="https://git.openjdk.org/valhalla/pull/698.diff">https://git.openjdk.org/valhalla/pull/698.diff</a>

</details>


RogerRiggs:
WRT AccessFlags...
The description of the "source" linkage should be expanded or possibly the feature removed.
My  prototyping for Valhalla involving adding identity, value, and primitive to the AccessFlag class.
These are new (in Valhalla) source keywords and ACC_* bits for classes.  
This suggests that the "source" value for them should be 'true', however, the condition and
AccessFlagTest require the matching name to be defined in Modifier.

That puts Modifier in a conflicted state, not able to correctly interpret the flags or
return an accurate result from `toString()` because the new flags overlap with existing
masks that apply to fields (and not classes).  Modifier has no information about whether the
value is from a class, method or field.  So regardless of the new AccessFlag API, Modifier
will need updates to mitigate the ambiguity.

My intention was to break the link with the existing j.u.Modifier and only support the new values in AccessFlag.
The "source" linkage to Modifier makes that difficult.
Perhaps separately, Modifier should be deprecated; however, it has wide usage.

RogerRiggs:
The design problem with Modifier is that it (will be) acting on a concrete representation of ambiguous information after the mask have been redefined. The bits returned from Class.getModifier() do not identify as being from that source, similarly Member.getModifiers().

A couple possible avenues going forward:

- Plan the end-of-life of the class Modifier, and also Class.getModifier() and Member.Modifier(). Initially, by freezing the semantics and deprecating the class and methods. Arrange that Class.getModifier() not to return the new bit definitions, so that when they are used with the Modifier class the interpretation is unchanged.
The AccessFlags implementation will need to change to a new perhaps private API to get the flags.

- Enhance the Modifier class and Class.getModifier() have a mask indicating the alternate use of the low order 16 bits.  A new mask in the upper 16 bits could discriminate between the overloaded masks.
This could avoid the existing Modifier API from misrepresenting the flag bits.

src/java.base/share/classes/java/lang/Class.java
```diff
@@ -1505,6 +1504,14 @@ public Set<AccessFlag> accessFlags() {
                                             AccessFlag.Location.CLASS);
     }
 
+    /**
+     * {@return true if this class has the requested {@link AccessFlag}}
+     * @param flag an {@link AccessFlag}
+     */
+    public boolean is(AccessFlag flag) {
```
mlchung, line 1511: Is this API needed?   We can do `Class.accessFlags().contains(flag)`.   This discussion should belong to PR 7445 (see my comment in the `Modifier`).

test/langtools/tools/javac/valhalla/lworld-values/InnerClassAttributeValuenessTest.java
```diff
@@ -48,7 +49,12 @@ public static Inner create(int v) {
     }
 
     public static void main(String[] args) {
-        if ((Inner.class.getModifiers() & AccessFlags.ACC_PRIMITIVE) == 0)
+        Set<AccessFlag> flags = Inner.class.accessFlags();
+        System.out.println("accessFlags: " + flags);
+
+        if (!Inner.class.is(AccessFlag.VALUE))
```
mlchung, line 55:
```suggestion
        if (!flags.contains(AccessFlag.VALUE))
```

test/langtools/tools/javac/valhalla/lworld-values/InnerClassAttributeValuenessTest.java
```diff
@@ -48,7 +49,12 @@ public static Inner create(int v) {
...
             throw new AssertionError("Value flag missing");
+        if (!Inner.class.is(AccessFlag.PRIMITIVE))
```
mlchung, line 57:
```suggestion
        if (!flags.contains(AccessFlag.PRIMITIVE))
```

src/java.base/share/classes/jdk/internal/org/objectweb/asm/Type.java
```diff
@@ -682,8 +683,7 @@ private static void appendDescriptor(final Class<?> clazz, final StringBuilder s
     }
 
     static boolean isPrimitiveClass(Class<?> clazz) {
-        int mods = clazz.getModifiers();
-        return (mods & 0x00000100) != 0;
+        return clazz.is(AccessFlag.PRIMITIVE);
```
mlchung, line 686:
```suggestion
        return clazz.accessFlags().contains(AccessFlag.PRIMITIVE);
```

mlchung (commented):
The recent commit changes `AccessFlag.IDENTITY`, `AccessFlag.VALUE`, `AccessFlag.PRIMITIVE` not a source modifier but they are.  So I expect `sourceModifier()` returns true for these new constants.    Whether these new modifiers should be defined in `Modifier` API requires discussion.   `ACC_IDENTITY` may be set even if the source does not declare `identity`.   `Modifier::toString` would not return a string that matches the source declaration.    So this version not defining new constants in Modifier API looks reasonable.

src/java.base/share/classes/java/lang/reflect/AccessFlag.java
```diff
@@ -403,7 +425,7 @@ private static class LocationToFlags {
                                        BRIDGE, VARARGS, NATIVE,
                                        ABSTRACT, STRICT, SYNTHETIC)),
                           entry(Location.INNER_CLASS,
-                                Set.of(PUBLIC, PRIVATE, PROTECTED,
+                                Set.of(PUBLIC, PRIVATE, PROTECTED, IDENTITY, VALUE, PRIMITIVE,
```
mlchung, line 428: I also caught this issue in the previous version.  Glad that it's fixed in this version but the locations of these new constant should include INNER_CLASS as well.

src/java.base/share/classes/java/lang/reflect/AccessFlag.java
```diff
@@ -123,8 +123,20 @@ public enum AccessFlag {
...
+     * <p>
+     * NOTE: The Valhalla draft JVMS spec does not define a value for ACC_SUPER.
+     * Historically, the value 0x0020 was used to indicate ACC_SUPER, which affected the
+     * semantics of any invokespecial instructions (6.5.invokespecial) appearing in the class.
+     * In Java SE 8, the ACC_SUPER semantics became mandatory, regardless of the setting of
+     * ACC_SUPER or the class file version number, and the flag no longer had any effect.
+     * Now the flag has been repurposed as ACC_IDENTITY.
      */
-    SUPER(0x0000_0020, false, Set.of(Location.CLASS)),
+//    SUPER(0x0000_0020, false, Set.of(Location.CLASS)),
```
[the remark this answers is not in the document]
RogerRiggs: By the time this is finalized, there will be no ACC_SUPER in the JVMs spec.

src/java.base/share/classes/java/lang/reflect/AccessFlag.java
```diff
@@ -123,8 +123,20 @@ public enum AccessFlag {
...
+
+    /**
+     * The access flag {@code ACC_IDENTITY} with a mask value of {@code 0x0020}.
+     */
+    IDENTITY(0x0000_0020, false, Set.of(Location.CLASS)),
```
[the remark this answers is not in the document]
RogerRiggs: fixed

src/java.base/share/classes/java/lang/reflect/Modifier.java
```diff
@@ -29,13 +29,24 @@
...
  * {@jvms 4.1}, {@jvms 4.4}, {@jvms 4.5}, and {@jvms 4.7} of
  * <cite>The Java Virtual Machine Specification</cite>.
+ * <p>
+ * This class cannot distinguish among modifiers for a class, method, or field.
+ * The modifier masks are not unique; it is up to the caller to use the correct
+ * mask or method with the modifier bits returned from {@link Class#getModifiers()}
+ * or {@link Member#getModifiers()}.
+ * The {@link AccessFlag} class provides a model that distinguishes between access flags
+ * for classes, methods, and field. The {@link Class#accessFlags()} and {@link Member#accessFlags()}
+ * methods provide the access flags for the respective class, method, or field.
```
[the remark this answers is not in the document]
RogerRiggs: We'll come back to it when the PR is finally reviewed.
We can resolve any differences when Valhalla is to be integrated.

src/java.base/share/classes/java/lang/Class.java
RogerRiggs: There are significant performance concerns. The current implementation of getModifers() is an intrinsic and all of the tests for modifiers are boolean operations against static constants.
Doing a Set.contains against a set that is re-built every time getAccessFlags() is called is a non-starter.
Many of the access flags have dedicated methods in j.l.Class, for example, isPublic, isStatic, etc.
It would improve usability and performance to be able to check for an accessFlag in a simple direct operation.
I'll keep this for the time being.

test/langtools/tools/javac/valhalla/lworld-values/InnerClassAttributeValuenessTest.java
RogerRiggs: No, not performant.

test/langtools/tools/javac/valhalla/lworld-values/InnerClassAttributeValuenessTest.java
RogerRiggs: Ditto

src/java.base/share/classes/jdk/internal/org/objectweb/asm/Type.java
RogerRiggs: Not performant.

src/java.base/share/classes/java/lang/reflect/AccessFlag.java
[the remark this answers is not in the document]
mlchung: We don't know what Java SE release ACC_SUPER will be dropped.  If `AccessFlag` API is added in Java 20 and `ACC_SUPER` is dropped in the same release, then you are right and this constant will never be defined.

Anyway, I'm ok either the javadoc and `ACC_SUPER` is removed from this file or just keep it with an api note.

src/java.base/share/classes/java/lang/Class.java
mlchung: This new API requires discussion and also whether it needs to be performant.  Can you bring this discussion to PR 7445.

`Class::isValue`, `Class::isIdentity` and `Class::isPrimitive` is performance-sensitive and the current implementation is performant checking against the modifiers.   Is there a case that `Class::isXXX`  cannot be used?

mlchung (approved):
Thanks for the update.   I agree that the `Modifier` class spec needs update.   PR 7445 should finalize that.

RogerRiggs:
The new identity, value, and primitive modifiers are included in java.lang.reflect.Modifier to 
provide straightforward mapping from the VM definitions of the modifier bits to those provided in the API.
Added clarifications to modifier methods to indicate that the source of the modifier bits is significant,
the `isXXX` methods of Modifier only give valid results when the modifier bits come from the respective class, method or field.

RogerRiggs:
This resolves several test issues with defineHiddenClass/BasicTest and ClassAccessFlagTest.

src/java.base/share/classes/java/lang/Class.java
```diff
@@ -637,6 +637,17 @@ public boolean isPrimitiveClass() {
         return (this.getModifiers() & PRIMITIVE_CLASS) != 0;
     }
 
+    /**
+     * Returns {@code true} if this class is an identity class.
```
mlchung, line 641: Nit: can replace this with the simplified form `{@return ...}`

src/java.base/share/classes/java/lang/Class.java
```diff
@@ -645,7 +656,7 @@ public boolean isPrimitiveClass() {
      * @since Valhalla
      */
     public boolean isValue() {
-        return (this.getModifiers() & VALUE_CLASS) != 0;
+        return (this.getModifiers() & Modifier.VALUE) != 0;
```
mlchung, line 659: the private static `VALUE_CLASS` constant should be removed

mlchung (approved):
Looks good with minor comments.

PR #718, opened 2022-06-30
8286824: [lworld] WeakHashMap for value classes

RogerRiggs:
WeakHashMap uses WeakReference for keys, allowing it to retain only keys and values that are strongly referenced elsewhere.

The behavior of WeakReference (or any Reference class) is not defined for value objects.

Four (4) behaviors are proposed for WeakHashMap to give the developer a choice:
 - SOFT - Keys that are value objects are retained until memory pressure clears soft references; then the keys that are value objects are removed automatically
 - STRONG - Keys are value objects are retained until they are explicitly removed
 - THROW - Keys that are value objects can not be put into the map; UnsupportedOperationException is thrown
 - DISCARD - Keys that are value objects are silently discarded; as if they were removed immediately without being visible

The default is SOFT - key/values are removed when there is memory pressure.
Entries will be removed automatically at some point, though less deterministically than with WeakReferences.

---------
### Progress

### Issue
 * [JDK-8286824](https://bugs.openjdk.org/browse/JDK-8286824): [lworld] WeakHashMap for value classes

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla pull/718/head:pull/718` \
`$ git checkout pull/718`

`$ git checkout pull/718` \
`$ git pull https://git.openjdk.org/valhalla pull/718/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 718`

`$ git pr show -t 718`

</details>

<a href="https://git.openjdk.org/valhalla/pull/718.diff">https://git.openjdk.org/valhalla/pull/718.diff</a>

</details>


src/java.base/share/classes/java/util/WeakHashMap.java
```diff
@@ -197,13 +226,31 @@ private Entry<K,V>[] newTable(int n) {
     /**
      * Constructs a new, empty {@code WeakHashMap} with the given initial
      * capacity and the given load factor.
+     * The default {@code ValuePolicy} is {@link ValuePolicy#SOFT} unless
+     * the system property {@systemProperty java.util.WeakHashMap.valueKeyRetention}
+     * is equal to the name of a {@link ValuePolicy} enum.
```
mlchung, line 231: Is the value of the system property case-sensitive?  should specify.

src/java.base/share/classes/java/util/WeakHashMap.java
```diff
@@ -53,7 +78,8 @@
  * {@link Collections#synchronizedMap Collections.synchronizedMap}
  * method.
  *
- * <p> This class is intended primarily for use with key objects whose
+ * <p> <i>Update needed for Value Objects:
```
mlchung, line 81: can you file a JBS issue to follow up this?

src/java.base/share/classes/java/util/WeakHashMap.java
```diff
@@ -197,13 +226,31 @@ private Entry<K,V>[] newTable(int n) {
...
      * @throws IllegalArgumentException if the initial capacity is negative,
      *         or if the load factor is nonpositive.
      */
     public WeakHashMap(int initialCapacity, float loadFactor) {
+        this(initialCapacity, loadFactor, ValuePolicy.DEFAULT_VALUE_POLICY);
+    }
+
+    /**
+     * Constructs a new, empty {@code WeakHashMap} with the given initial
+     * capacity and the given load factor.
```
mlchung, line 244: and the given value policy.

src/java.base/share/classes/java/util/WeakHashMap.java
```diff
@@ -36,13 +39,35 @@
  * Hash table based implementation of the {@code Map} interface, with
  * <em>weak keys</em>.
  * An entry in a {@code WeakHashMap} will automatically be removed when
- * its key is no longer in ordinary use.  More precisely, the presence of a
+ * its key is no longer in ordinary use.
+ * More precisely, for keys that are identity objects, the presence of a
  * mapping for a given key will not prevent the key from being discarded by the
  * garbage collector, that is, made finalizable, finalized, and then reclaimed.
+ * For keys that are {@linkplain Class#isValue() Value objects}, the retention of the
```
mlchung, line 46: should this be "value objects" rather than "Value objects"?  This comment applies to all occurrences.

src/java.base/share/classes/java/util/WeakHashMap.java
```diff
@@ -447,6 +513,8 @@ Entry<K,V> getEntry(Object key) {
      *         {@code null} if there was no mapping for {@code key}.
      *         (A {@code null} return can also indicate that the map
      *         previously associated {@code null} with {@code key}.)
```
mlchung, line 515: Need to specify: A null return can also indicate the entry is discarded if the value policy is DISCARD and the key is a value object.

src/java.base/share/classes/java/util/WeakHashMap.java
```diff
@@ -758,6 +862,96 @@ public String toString() {
         }
     }
 
+    /**
+     * A SoftEntry is used for value class keys in which the entries are retained
+     * until there is some memory pressure.  A anchor object is used as the referent
```
mlchung, line 867: s/a anchor/an anchor/

src/java.base/share/classes/java/util/WeakHashMap.java
```diff
@@ -758,6 +862,96 @@ public String toString() {
...
+     * of a SoftReference and also as the referent of the WeakReference.
+     * After the SoftReference is cleared, due to GC pressure, the WeakReference is cleared too.
+     *
+     * @param <K> key
+     * @param <V> value
+     */
+    private static class SoftEntry<K, V> extends Entry<K, V> {
+        Object realKey;
```
mlchung, line 875: can make it a final field.

src/java.base/share/classes/java/util/WeakHashMap.java
```diff
@@ -758,6 +862,96 @@ public String toString() {
...
+    /**
+     * A StrongEntry is used for value class keys in which the entries are retained
+     * until removed.  A singleton instance is used as the referent of the WeakReference.
+     * Since the anchor is never reclaimed, the Entry is retained forever.
+     *
+     * @param <K> key
+     * @param <V> value
+     */
+    private static class StrongEntry<K, V> extends Entry<K, V> {
+        Object realKey;
```
mlchung, line 920: can be a final field

src/java.base/share/classes/java/util/WeakHashMap.java
```diff
@@ -447,6 +513,8 @@ Entry<K,V> getEntry(Object key) {
      *         {@code null} if there was no mapping for {@code key}.
      *         (A {@code null} return can also indicate that the map
      *         previously associated {@code null} with {@code key}.)
+     * @throws UnsupportedOperationException if {@code key} is a Value object
+     *         and the {@link #valuePolicy() valuePolicy} is {@link ValuePolicy#THROW}.
```
mlchung, line 517: `putAll` and `putIfAbsent` would need spec update.

src/java.base/share/classes/java/util/WeakHashMap.java
```diff
@@ -1335,4 +1529,56 @@ public int characteristics() {
         }
     }
 
+    /**
+     * Enum for the ValuePolicy; when putting a key and value into a WeakHashMap
+     * determines how keys that are Value objects are retained (or not).
```
mlchung, line 1534: `@since valhalla`

src/java.base/share/classes/java/util/WeakHashMap.java
RogerRiggs: A subtle point...  The WeakHashmap *may* contain an identity class instance that compares as `equals()` with the value object. In that case, the instance would be returned from `getEntry`.
Similarly, `put()` would return the previous entry (that was `equals`) to a value entry that is discarded.
This will need some more spec detail.

PR #750, opened 2022-09-08
8293134: Valhalla: Verifier error in method call with abstract value class arg

hseigel:
Please review this small fix for JDK-8293134.  Since a value class can be abstract and a super class, only return false at line 64 for primitive classes.

The fix is tested by running mach5 tiers 1-2 on Linux, Mac OS, and Windows, and mach5 tiers 3-5 on Linux x64.  (testing is in progress.)

Thanks, Harold

---------
### Progress

### Issue
 * [JDK-8293134](https://bugs.openjdk.org/browse/JDK-8293134): Valhalla: Verifier error in method call with abstract value class arg

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla pull/750/head:pull/750` \
`$ git checkout pull/750`

`$ git checkout pull/750` \
`$ git pull https://git.openjdk.org/valhalla pull/750/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 750`

`$ git pr show -t 750`

</details>

<a href="https://git.openjdk.org/valhalla/pull/750.diff">https://git.openjdk.org/valhalla/pull/750.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/classfile/verificationType.cpp
+++ b/src/hotspot/share/classfile/verificationType.cpp
@@ -64,7 +64,13 @@ bool VerificationType::resolve_and_check_assignability(InstanceKlass* klass, Sym
     }
   }
 
-  if (this_class->access_flags().is_primitive_class()) return false;
+  // Need to do this check when called from CDS.
+  if (this_class->access_flags().is_primitive_class()) {
+    Klass* from_class = SystemDictionary::resolve_or_fail(
+      from_name, Handle(THREAD, klass->class_loader()),
+      Handle(THREAD, klass->protection_domain()), true, CHECK_false);
+    return from_class == this_class;
+  }
   if (this_class->is_interface() && (!from_field_is_protected ||
       from_name != vmSymbols::java_lang_Object())) {
     // If we are not trying to access a protected field or method in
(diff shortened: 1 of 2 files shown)


src/hotspot/share/classfile/verificationType.cpp
```diff
@@ -64,7 +64,7 @@ bool VerificationType::resolve_and_check_assignability(InstanceKlass* klass, Sym
     }
   }
 
-  if (this_class->access_flags().is_value_class()) return false;
```
fparain, line 67: This line looks wrong to me. If the method is called with name and from_name being identical, and the corresponding class being a primitive class, the method returns false.
This particular scenario is handled in the caller when this method is invoked from VerificationType::is_reference_assignable_from(), but it not handled when called from SystemDictionaryShared::check_verification_constraints().
This problem is not due to the proposed change, the previous code testing for a value class suffers from the same issue.

fparain (changes requested):

src/hotspot/share/classfile/verificationType.cpp
hseigel: I agree.  See changes to this code in commit 2.   Thanks for looking at this PR.

fparain (approved):
LGTM
Thank you for updating your changes.

Fred

hseigel:
Thanks Fred for the review!

/integrate

PR #753, opened 2022-09-09
val_factory_new

hseigel:
---------
### Progress

### Error
&nbsp;⚠️ The pull request body must not be empty.

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla pull/753/head:pull/753` \
`$ git checkout pull/753`

`$ git checkout pull/753` \
`$ git pull https://git.openjdk.org/valhalla pull/753/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 753`

`$ git pr show -t 753`

</details>

<a href="https://git.openjdk.org/valhalla/pull/753.diff">https://git.openjdk.org/valhalla/pull/753.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/classfile/classFileParser.cpp
+++ b/src/hotspot/share/classfile/classFileParser.cpp
@@ -743,12 +743,13 @@ void ClassFileParser::parse_constant_pool(const ClassFileStream* const stream,
               throwIllegalSignature("Method", name, signature, CHECK);
             }
           }
-          // 4509014: If a class method name begins with '<', it must be "<init>"
+          // If a class method name begins with '<', it must be "<init>" or "<new>"
           const unsigned int name_len = name->utf8_length();
           if (tag == JVM_CONSTANT_Methodref &&
               name_len != 0 &&
               name->char_at(0) == JVM_SIGNATURE_SPECIAL &&
-              name != vmSymbols::object_initializer_name()) {
+              (name != vmSymbols::inline_factory_name() &&
+               name != vmSymbols::object_initializer_name())) {
             classfile_parse_error(
               "Bad method name at constant pool index %u in class file %s",
               name_ref_index, CHECK);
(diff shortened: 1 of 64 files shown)


src/hotspot/share/classfile/classFileParser.cpp
```diff
@@ -769,13 +770,24 @@ void ClassFileParser::parse_constant_pool(const ClassFileStream* const stream,
...
+              // <new> factory methods must be non-void return and invokeStatic.
+              const int signature_ref_index =
+                cp->signature_ref_index_at(name_and_type_ref_index);
+              const Symbol* const signature = cp->symbol_at(signature_ref_index);
+              if (signature->is_void_method_signature() || ref_kind != JVM_REF_invokeStatic) {
+                classfile_parse_error(
+                  "Bad factory method name at constant pool index %u in class file %s",
+                  name_ref_index, CHECK);
+              }
+           } else if (name != vmSymbols::object_initializer_name()) {
```
fparain, line 784: This block has an incorrect indentation.

src/hotspot/share/classfile/classFileParser.cpp
```diff
@@ -769,13 +770,24 @@ void ClassFileParser::parse_constant_pool(const ClassFileStream* const stream,
             const int name_ref_index =
               cp->name_ref_index_at(name_and_type_ref_index);
             const Symbol* const name = cp->symbol_at(name_ref_index);
-            if (name != vmSymbols::object_initializer_name()) {
+
+            if (supports_inline_types() && name == vmSymbols::inline_factory_name()) { // <new>
```
fparain, line 774: Should we add a case where the method name is a factory method name but the class file doesn't support inline types?

src/hotspot/share/classfile/classFileParser.cpp
```diff
@@ -5916,7 +5931,9 @@ void ClassFileParser::verify_legal_method_name(const Symbol* name, TRAPS) const
 
   if (length > 0) {
     if (bytes[0] == JVM_SIGNATURE_SPECIAL) {
-      if (name == vmSymbols::object_initializer_name() || name == vmSymbols::class_initializer_name()) {
+      if (name == vmSymbols::object_initializer_name() ||
+          name == vmSymbols::class_initializer_name() ||
+          name == vmSymbols::inline_factory_name()) {
```
fparain, line 5936: Shouldn't we check that the class file supports inline types before accepting a static factory name?

src/hotspot/share/classfile/vmSymbols.hpp
```diff
@@ -347,6 +347,7 @@
   /* common method and field names */                                                             \
   template(object_initializer_name,                   "<init>")                                   \
   template(class_initializer_name,                    "<clinit>")                                 \
+  template(inline_factory_name,                       "<new>")                                    \
```
fparain, line 350: In JDK-8247569 there's an uncertainty about the static factory name, but the CR mentions "\<vnew\>", not "\<new\>" (the change would need to be propagated to all comments).

src/hotspot/share/oops/method.cpp
```diff
@@ -822,12 +822,12 @@ bool Method::is_class_initializer() const {
 
 // A method named <init>, if non-static, is a classic object constructor.
 bool Method::is_object_constructor() const {
-   return name() == vmSymbols::object_initializer_name() && !is_static();
+  return name() == vmSymbols::object_initializer_name() && !is_static(); //TBD is the static check needed?
```
fparain, line 825: With the new model, all constructors are non-static and all factories are static so the check doesn't seem needed (applies to the method below too).

fparain (changes requested):
Thank you for tackling this task.
In addition to the comments on your proposed changes, the comments of method ClassFileParser::verify_unqualified_name() should also be updated to cover the static factory name case (the comment has not been modified, so I cannot comment directly on it).

Fred

PR #754, opened 2022-09-12
8292883: [lworld] javac fails to detect class files with invalid access flags

vicente-romero-oracle:
javac is loading a forged class file which declares a primitive class which also has the `identity` flag set. This PR is fixing this bug by issuing an error in such cases.

TIA

---------
### Progress

### Issue
 * [JDK-8292883](https://bugs.openjdk.org/browse/JDK-8292883): [lworld] javac fails to detect class files with invalid access flags

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla pull/754/head:pull/754` \
`$ git checkout pull/754`

`$ git checkout pull/754` \
`$ git pull https://git.openjdk.org/valhalla pull/754/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 754`

`$ git pr show -t 754`

</details>

<a href="https://git.openjdk.org/valhalla/pull/754.diff">https://git.openjdk.org/valhalla/pull/754.diff</a>

</details>


fparain (commented):
I cannot comment on the implementation of the fix, but as the reporter of the issue I can tell that the proposed fix correctly detects the invalid class in the provided use case.

vicente-romero-oracle:
> I cannot comment on the implementation of the fix, but as the reporter of the issue I can tell that the proposed fix correctly detects the invalid class in the provided use case.

@fparain thanks for checking that

PR #761, opened 2022-09-19
8281295: Prototype serialization/deserialization using Unsafe.

RogerRiggs:
Specify serialization and deserialization of value objects. Value objects are immutable, Externalizable relies on mutation, Externalizable is not supported for value objects. Revised tests and added tests for graphs of value objects.

Somewhat similar to Record serialization in that back references to the value object store 'null' until the value object is completely deserialized.
Similar to regular serialization/deserialization of identity objects.  The serialized form is the same.
Deserialization is done by reading fields and storing into a Buffered larval object, converted to a value object when all of the fields have been read and stored.

Externalizable and Value Objects are incompatible, value objects are immutable and Externalizable.readExternal is expected to modify the fields of an allocated object.

---------
### Progress

### Issue
 * [JDK-8281295](https://bugs.openjdk.org/browse/JDK-8281295): Prototype serialization/deserialization using Unsafe.

### Reviewers
 * [Mandy Chung](https://openjdk.org/census#mchung) (@mlchung - Committer) ⚠️ Review applies to [bf1f3a17](https://git.openjdk.org/valhalla/pull/761/files/bf1f3a175e11f3bda56911083ea34b430d166f94)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla pull/761/head:pull/761` \
`$ git checkout pull/761`

`$ git checkout pull/761` \
`$ git pull https://git.openjdk.org/valhalla pull/761/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 761`

`$ git pr show -t 761`

</details>

<a href="https://git.openjdk.org/valhalla/pull/761.diff">https://git.openjdk.org/valhalla/pull/761.diff</a>

</details>


src/java.base/share/classes/java/io/ObjectInputStream.java
```diff
@@ -2275,8 +2283,26 @@ private Object readOrdinaryObject(boolean unshared)
...
+            obj = desc.finishValue(obj);
+            if (!unshared)
+                handles.setObject(passHandle, obj);
         } else {
+            // For all other objects, publish the ref and then read the data
+            assert obj != null;
+            if (!unshared)
+                handles.setObject(passHandle, obj);
             readSerialData(obj, desc);
         }
```
ExE-Boss, line 2307: These `assert obj != null` statements can be hoisted before all the `if`‑statements.

RogerRiggs: Except the first, for Records object must be null.

RogerRiggs:
The draft of the serialization specification update for value objects is:
http://cr.openjdk.java.net/~rriggs/lworld-serial-spec/

mlchung (commented):
This seems a reasonable way to provide a minimal serialization support for value classes to allow proper solution to be explored.

src/java.base/share/classes/java/io/ObjectStreamClass.java
```diff
@@ -2047,8 +2094,12 @@ void getObjFieldValues(Object obj, Object[] vals) {
              * in array should be equal to Unsafe.INVALID_FIELD_OFFSET.
              */
             for (int i = numPrimFields; i < fields.length; i++) {
+                Field f = fields[i].getField();
                 vals[offsets[i]] = switch (typeCodes[i]) {
-                    case 'L', '[' -> unsafe.getReference(obj, readKeys[i]);
+                    case 'L', '[' ->
+                            unsafe.isFlattened(f)
```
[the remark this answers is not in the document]
RogerRiggs: As far as I know, a "value type" (aka primitive class) is a value class and the only difference from the Unsafe API point of view is whether it is flattened or not. There may be some additional work for JEP 401.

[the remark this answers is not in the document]
RogerRiggs: Also added a few tests for serializable primitive classes.

ExE-Boss:
The `java.io.Externalizable` interface should probably be marked as an identity interface.

RogerRiggs:
> The `java.io.Externalizable` interface should probably be marked as an identity interface.

That would head-off some misuses, and so far I'm not aware of a viable use case.

mlchung:
To recap our offline discussion, I would expect `ObjectStreamField::getTypeString` should return `Q-`descriptor if the field type is a primitive class value type (consistent with `Field`).   For example,

```
class Test implements Serializable {
   SerializablePrimitivePoint p;
   SerializableValuePoint v;
}
```

where `p` may be a flattened field of a primitive class value type but `v` is a field of a reference type (value class).   In that case, `Field.getType().descriptorString()` on `p`  should return `QSerializablePrimitivePoint;`.

Maybe add the serialization support for primitive classes as a follow-on issue.   Better to revert the change added for `Q` descriptor and keep this PR for value class only.

src/java.base/share/classes/java/io/ObjectOutputStream.java
```diff
@@ -128,6 +128,12 @@
  * implemented by a class they can write and read their own state using all of
  * the methods of ObjectOutput and ObjectInput.  It is the responsibility of
  * the objects to handle any versioning that occurs.
+ * Value classes implementing {@link Externalizable} cannot be serialized
+ * or deserialized, the value object is immutable and the state cannot be restored.
+ * Use {@link Serializable} {@code writeReplace} to delegate to another serializable
+ * object such as a record.
+ *
+ * Value objects cannot be `java.io.Externalizable`.
```
mlchung, line 136:
```suggestion
 * Value objects cannot be {@code java.io.Externalizable}.
```

mlchung (approved):
This looks good as an initial prototype.

PR #770, opened 2022-09-28
8294312: [lworld] Add java.util.Objects.isIdentityObject

RogerRiggs:
Add `java.util.Objects.isIdentityObject(obj)` and update tests.

Correct j.l.Class.isIdentity() and .isValue() to correctly identify all arrays as identity objects.
(The modifiers for array classes do have not reliable ACC_IDENTITY or ACC_VALUE bits).

Updated ValhallaFeaturesTest to use junit.

---------
### Progress

### Issue
 * [JDK-8294312](https://bugs.openjdk.org/browse/JDK-8294312): [lworld] Add java.util.Objects.isIdentityObject

### Reviewers
 * [Mandy Chung](https://openjdk.org/census#mchung) (@mlchung - Committer)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla pull/770/head:pull/770` \
`$ git checkout pull/770`

`$ git checkout pull/770` \
`$ git pull https://git.openjdk.org/valhalla pull/770/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 770`

`$ git pr show -t 770`

</details>

<a href="https://git.openjdk.org/valhalla/pull/770.diff">https://git.openjdk.org/valhalla/pull/770.diff</a>

</details>

### DIFF
--- a/src/java.base/share/classes/java/lang/Class.java
+++ b/src/java.base/share/classes/java/lang/Class.java
@@ -639,23 +639,29 @@ public final class Class<T> implements java.io.Serializable,
 
     /**
      * {@return {@code true} if this class is an identity class, otherwise {@code false}}
+     * When value classes are enabled, classes with the {@linkplain Modifier#IDENTITY IDENTITY}
+     * {@link #getModifiers() modifier} are identified as identity classes;
+     * when not enabled, all classes are identity classes.
+     * Array classes are always identity classes.
      *
      * @since Valhalla
      */
     public boolean isIdentity() {
         return !ValhallaFeatures.isEnabled() ||  // Before Valhalla all classes are identity classes
-                isArray() ||
-                (this.getModifiers() & Modifier.IDENTITY) != 0;
+                (this.getModifiers() & Modifier.IDENTITY) != 0 ||
+                isArray();
     }
 
     /**
      * {@return {@code true} if this class is a value class, otherwise {@code false}}
+     * When value classes are enabled, classes with the {@linkplain Modifier#VALUE VALUE}
+     * {@link #getModifiers() modifier} are identified as value classes;
+     * when not enabled, no classes are value classes.
      *
      * @since Valhalla
      */
     public boolean isValue() {
         return ValhallaFeatures.isEnabled() &&  // Before Valhalla no classes are value classes
-                !isArray() &&
                 (this.getModifiers() & Modifier.VALUE) != 0;
     }
 
--- a/src/java.base/share/classes/java/lang/Class.java
+++ b/src/java.base/share/classes/java/lang/Class.java
@@ -72,6 +72,7 @@ import java.util.stream.Collectors;
 import jdk.internal.loader.BootLoader;
 import jdk.internal.loader.BuiltinClassLoader;
 import jdk.internal.misc.Unsafe;
+import jdk.internal.misc.ValhallaFeatures;
 import jdk.internal.module.Resources;
 import jdk.internal.reflect.CallerSensitive;
 import jdk.internal.reflect.CallerSensitiveAdapter;
@@ -642,7 +643,9 @@ public final class Class<T> implements java.io.Serializable,
      * @since Valhalla
      */
     public boolean isIdentity() {
-        return isArray() || (this.getModifiers() & Modifier.IDENTITY) != 0;
+        return !ValhallaFeatures.isEnabled() ||  // Before Valhalla all classes are identity classes
+                isArray() ||
+                (this.getModifiers() & Modifier.IDENTITY) != 0;
     }
 
     /**
@@ -651,7 +654,9 @@ public final class Class<T> implements java.io.Serializable,
      * @since Valhalla
      */
     public boolean isValue() {
-        return !isArray() && (this.getModifiers() & Modifier.VALUE) != 0;
+        return ValhallaFeatures.isEnabled() &&  // Before Valhalla no classes are value classes
+                !isArray() &&
+                (this.getModifiers() & Modifier.VALUE) != 0;
     }
 
     /**
--- a/src/java.base/share/classes/java/lang/Class.java
+++ b/src/java.base/share/classes/java/lang/Class.java
@@ -639,10 +639,9 @@ public final class Class<T> implements java.io.Serializable,
 
     /**
      * {@return {@code true} if this class is an identity class, otherwise {@code false}}
-     * When value classes are enabled, classes with the {@linkplain Modifier#IDENTITY IDENTITY}
-     * {@link #getModifiers() modifier} are identified as identity classes;
-     * when not enabled, all classes are identity classes.
-     * Array classes are always identity classes.
+     * If this {@code Class} object represents an array type, then this method returns {@code true}.
+     * If this {@code Class} object represents a primitive type, or {@code void},
+     * then this method returns {@code false}.
      *
      * @since Valhalla
      */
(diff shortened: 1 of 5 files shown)


src/java.base/share/classes/java/lang/Class.java
```diff
@@ -642,7 +642,7 @@ private static Class<?> forName(Module module, String name, Class<?> caller) {
      * @since Valhalla
      */
     public boolean isIdentity() {
-        return (this.getModifiers() & Modifier.IDENTITY) != 0;
+        return isArray() || (this.getModifiers() & Modifier.IDENTITY) != 0;
```
mlchung, line 645: I think Class::getModifiers for an array class should have `IDENTITY` flag set.  I think it's a bug in the VM that should set `IDENTITY`.

```
jint ObjArrayKlass::compute_modifier_flags() const {
  // The modifier for an objectArray is the same as its element
  if (element_klass() == NULL) {
    assert(Universe::is_bootstrapping(), "partial objArray only at startup");
    return JVM_ACC_ABSTRACT | JVM_ACC_FINAL | JVM_ACC_PUBLIC;
  }
  // Return the flags of the bottom element type.
  jint element_flags = bottom_klass()->compute_modifier_flags();

  return (element_flags & (JVM_ACC_PUBLIC | JVM_ACC_PRIVATE | JVM_ACC_PROTECTED))
                        | (JVM_ACC_ABSTRACT | JVM_ACC_FINAL);
}
```

RogerRiggs: The spec of Class.getModifiers() is explicit for arrays that only a few bits are defined.
The other bits are unspecified and may be implementation specific.
@rose00 thought it risky and too late to specify that now.

I'd suggest making the `Class.isIdentity()` and `Class.isValue()` native methods and have the VM implement the semantics.  Then it will not depend on the modifier bits. Mostly likely, these would become intrinsics.

mlchung: Alternatively, we could use `getClassAccessFlagsRaw` to return the VM specific flags to indicate that. 

I think checking `isArray` is a good interim solution.

src/java.base/share/classes/java/lang/Object.java
```diff
@@ -35,6 +35,12 @@
  * Class {@code Object} is the root of the class hierarchy.
  * Every class has {@code Object} as a superclass. All objects,
  * including arrays, implement the methods of this class.
+ * <p>
+ * Subclasses of {@code java.lang.Object} can be either identity classes or value classes.
+ * The class {@code Object} itself is neither an identity class nor a value class.
+ * See {@jls The Java Language Specification  8.1.1.5 identity and value Classes}.
+ * An Instance can be created with {@code new Object()}, those instances are
```
mlchung, line 42: s/Instance/instance

src/java.base/share/classes/java/util/Objects.java
```diff
@@ -189,6 +190,17 @@ public static String toIdentityString(Object o) {
...
 
+   /**
+     * {@return {@code true} if the object is an identity object, otherwise {@code false}}
+     * @param o an object
+     */
+//    @IntrinsicCandidate
+    public static boolean isIdentityObject(Object o) {
+        return o.getClass().isIdentity() ||
+                o.getClass() == Object.class ||
+                !ValhallaFeatures.isEnabled(); // Before Valhalla all objects are identity objects.
```
mlchung, line 201: `Class::isIdentity` should check for `!ValhallaFeatures.isEnabled()` as well.   Just a side note: we will need to go through the current implementation to see where this check is needed as well.

test/jdk/valhalla/valuetypes/ObjectMethods.java
```diff
@@ -69,6 +70,31 @@ public class ObjectMethods {
...
+        return new Object[][]{
+                {new Object(), true},
+                {"String", true},
+                {String.class, true},
+                {Object.class, true},
+                {new ValueType1(1), false},
+                {new ValueType2(2), false},
+                {new PrimitiveRecord(1, "A"), false},
+                {new ValueRecord(1,"B"), false},
+                {new int[0], true},  // arrays of primitives classes are identity objects
```
mlchung, line 85: s/primitives/primitive/

mlchung:
Do you have tests to check `Class::accessFlags` for array types?

src/java.base/share/classes/java/lang/Class.java
```diff
@@ -651,7 +654,9 @@ public boolean isIdentity() {
      * @since Valhalla
      */
     public boolean isValue() {
-        return (this.getModifiers() & Modifier.VALUE) != 0;
+        return ValhallaFeatures.isEnabled() &&  // Before Valhalla no classes are value classes
```
mlchung, line 657: `isValue` does not need to check `-XX:+EnableVahalla` since the new `VALUE` flag can only be set in a value class and if the feature is enabled.

src/java.base/share/classes/java/lang/Class.java
```diff
@@ -642,7 +643,9 @@ private static Class<?> forName(Module module, String name, Class<?> caller) {
      * @since Valhalla
      */
     public boolean isIdentity() {
```
mlchung, line 645: The javadoc for `isIdentity` should specify if this Class is an array class, primitive type or `void`.

src/java.base/share/classes/java/lang/Object.java
```diff
@@ -35,6 +35,12 @@
  * Class {@code Object} is the root of the class hierarchy.
  * Every class has {@code Object} as a superclass. All objects,
  * including arrays, implement the methods of this class.
+ * <p>
+ * Subclasses of {@code java.lang.Object} can be either identity classes or value classes.
```
mlchung, line 39: suggest to add a link to `Class::isIdentity` and `Class::isValue`

RogerRiggs:
> Do you have tests to check `Class::accessFlags` for array types?

Joe provided tests for Array classes (in ClassAccessFlagTest).
It looks like he assumed other bits in the modifiers were zero.
There is no spec for other bits. (they are unspecified)

src/java.base/share/classes/java/util/Objects.java
```diff
@@ -189,6 +190,19 @@ public static String toIdentityString(Object o) {
         return o.getClass().getName() + "@" + Integer.toHexString(System.identityHashCode(o));
     }
 
+   /**
+    * {@return {@code true} if the object is an identity object, otherwise {@code false}}
```
mlchung, line 194: nit: s/the object/the specified object/

mlchung (commented):
Looks okay in general.   A couple minor comments.

mlchung:
> Joe provided tests for Array classes (in ClassAccessFlagTest).
> It looks like he assumed other bits in the modifiers were zero.
> There is no spec for other bits. (they are unspecified)

Can you add a valhalla-specific test to validate that `Class::accessFlags` returns the expected flags for identity, value, array classes, and interface, etc.

src/java.base/share/classes/java/lang/Class.java
RogerRiggs: The primitive classes and void also do not have specified modifier bits.
At present, they are not reported to be identity or value classes. The mainline javadoc says they are "value classes" but in reality they are identity classes. JEP 204 (future) would be the authority on that.

src/java.base/share/classes/java/util/Objects.java
RogerRiggs: The pattern in all of Objects.java is to refer to "the object".

RogerRiggs:
> Can you add a valhalla-specific test to validate that `Class::accessFlags` returns the expected flags for identity, value, array classes, and interface, etc.

ok, As a separate Issue/PR: JDK-8294678.

src/java.base/share/classes/java/util/Objects.java
RogerRiggs: WIll fix to match other casesin Objects.java.

mlchung (approved):
Looks okay to me.  Thanks for the updates and creating the follow-up issues.

PR #771, opened 2022-09-28
8247569: [lworld] Align with spec for <vnew> methods

vicente-romero-oracle:
Value classes constructor should be named <vnew> instead of <init>. This PR can't be pushed without the corresponding support being available on the VM side

---------
### Progress

### Issue
 * [JDK-8247569](https://bugs.openjdk.org/browse/JDK-8247569): [lworld] Align with spec for <vnew> methods

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla pull/771/head:pull/771` \
`$ git checkout pull/771`

`$ git checkout pull/771` \
`$ git pull https://git.openjdk.org/valhalla pull/771/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 771`

`$ git pr show -t 771`

</details>

<a href="https://git.openjdk.org/valhalla/pull/771.diff">https://git.openjdk.org/valhalla/pull/771.diff</a>

</details>


MrSimms:
VM support, JDK-8294224 is in progress (next week I'd say)

src/java.base/share/classes/java/lang/constant/DirectMethodHandleDescImpl.java
```diff
@@ -64,7 +64,7 @@ final class DirectMethodHandleDescImpl implements DirectMethodHandleDesc {
      */
     DirectMethodHandleDescImpl(Kind kind, ClassDesc owner, String name, MethodTypeDesc type) {
         if (kind == CONSTRUCTOR)
-            name = "<init>";
+            name = owner.isPrimitiveValueType() ? "<vnew>" : "<init>";
```
MrSimms, line 67: So here is an issue: `owner.isPrimitiveValueType()` simply says owner is Q-type descriptor. Need this name for Value Objects (and Q-Types), but Value Objects are L-types...

How do we differentiate here ? By using `kind` ?

MrSimms: https://cr.openjdk.java.net/~dlsmith/jep8277163/jep8277163-20220830/specs/value-objects-jvms.html suggests REF_invokeStatic

PR #775, opened 2022-10-04
8294678: [lworld] Update AccessFlags tests to include ACC_IDENTITY and ACC_VALUE

RogerRiggs:
The tests for Class.accessFlags should be updated to check ACC_IDENTITY and ACC_VALUE. 
Update ClassAccessFlagTest and ObjectMethods test to check consistency between
Class.isIdentity() and Class.isValue() methods with the Class accessFlags().

---------
### Progress

### Issue
 * [JDK-8294678](https://bugs.openjdk.org/browse/JDK-8294678): [lworld] Update AccessFlags tests to include ACC_IDENTITY and ACC_VALUE

### Reviewers
 * [Mandy Chung](https://openjdk.org/census#mchung) (@mlchung - Committer)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla pull/775/head:pull/775` \
`$ git checkout pull/775`

`$ git checkout pull/775` \
`$ git pull https://git.openjdk.org/valhalla pull/775/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 775`

`$ git pr show -t 775`

</details>

<a href="https://git.openjdk.org/valhalla/pull/775.diff">https://git.openjdk.org/valhalla/pull/775.diff</a>

</details>

### DIFF
--- a/src/java.base/share/classes/java/lang/Class.java
+++ b/src/java.base/share/classes/java/lang/Class.java
@@ -1512,13 +1512,15 @@ public final class Class<T> implements java.io.Serializable,
         // INNER_CLASS forbids. INNER_CLASS allows PRIVATE, PROTECTED,
         // and STATIC, which are not allowed on Location.CLASS.
         // Use getClassAccessFlagsRaw to expose SUPER status.
+        // Arrays are identity classes but not reflected in modifiers
         var location = (isMemberClass() || isLocalClass() ||
                         isAnonymousClass() || isArray()) ?
             AccessFlag.Location.INNER_CLASS :
             AccessFlag.Location.CLASS;
-        return AccessFlag.maskToAccessFlags((location == AccessFlag.Location.CLASS) ?
-                                            getClassAccessFlagsRaw() :
-                                            getModifiers(),
+        final int mask = ((location == AccessFlag.Location.CLASS) ?
+                getClassAccessFlagsRaw() : getModifiers()) |
+                (isArray() ? Modifier.IDENTITY : 0);
+        return AccessFlag.maskToAccessFlags(mask & (~0x800),  // suppress unspecified bit
                                             location);
     }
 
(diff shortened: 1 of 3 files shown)


src/java.base/share/classes/java/lang/Class.java
```diff
@@ -1512,13 +1512,15 @@ public Set<AccessFlag> accessFlags() {
...
         var location = (isMemberClass() || isLocalClass() ||
                         isAnonymousClass() || isArray()) ?
             AccessFlag.Location.INNER_CLASS :
             AccessFlag.Location.CLASS;
-        return AccessFlag.maskToAccessFlags((location == AccessFlag.Location.CLASS) ?
-                                            getClassAccessFlagsRaw() :
-                                            getModifiers(),
+        final int mask = ((location == AccessFlag.Location.CLASS) ?
+                getClassAccessFlagsRaw() : getModifiers()) |
+                (isArray() ? Modifier.IDENTITY : 0);
```
mlchung, line 1522: For an array class `a`, `a.accessFlags()` and `AccessFlag.maskToAccessFlags(a.getModifiers(), Location.CLASS)` won't return the same `AccessFlag` set then.   The access flags for array type is special anyway.    It would be less confusing if `Class::accessFlags()` is consistent with `Class::getModifiers()`.

RogerRiggs: My goal was to have the Class.accessFlags() be consistent with Class methods `isIdentity()` and `isValue()`.
Unless we prevail in adding ACC_IDENTITY and ACC_VALUE to Class.getModifiers() there will be an inconsistency among the APIs.  `getModifiers()`, though the oldest, has the least flexibility.  The goal of the AccessFlags API was to be more flexible, correct, and extensible.  It is a question of what the end state should be and how to get there.
(The check for isArray() to select the INNER_CLASS location was from the original AccessFlag PR.)

RogerRiggs: I'd like to commit this PR as is on this point and propose a separate PR to add ACC_IDENTITY and ACC_VALUE to the modifiers for arrays to be able to judge the impact and get further review.
https://bugs.openjdk.org/browse/JDK-8294866

mlchung: I agree that we should discuss the modifiers for arrays in a follow-on issue.   I suggest this PR to keep  `Class::getModifiers` and `Class::accessFlags` consistent to match the specification.    JDK-8294866 will do the spec change altogether.

RogerRiggs: done

mlchung (approved):

PR #779, opened 2022-10-07
8270852: [lworld] bytecodes spinned at runtime should set minor_version to 65535 if preview feature enabled

mlchung:
This patch prepares for lworld to switch to a preview feature.  Also fix the generation of lambda classes and proxy classes to have `Preload` attribute for value classes (not only primitive classes).

---------
### Progress

### Issue
 * [JDK-8270852](https://bugs.openjdk.org/browse/JDK-8270852): [lworld] bytecodes spinned at runtime should set minor_version to 65535 if preview feature enabled

### Reviewers
 * [Roger Riggs](https://openjdk.org/census#rriggs) (@RogerRiggs - Committer)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla pull/779/head:pull/779` \
`$ git checkout pull/779`

`$ git checkout pull/779` \
`$ git pull https://git.openjdk.org/valhalla pull/779/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 779`

`$ git pr show -t 779`

</details>

<a href="https://git.openjdk.org/valhalla/pull/779.diff">https://git.openjdk.org/valhalla/pull/779.diff</a>

</details>

### DIFF
--- a/src/java.base/share/classes/java/lang/invoke/InnerClassLambdaMetafactory.java
+++ b/src/java.base/share/classes/java/lang/invoke/InnerClassLambdaMetafactory.java
@@ -26,7 +26,6 @@
 package java.lang.invoke;
 
 import jdk.internal.misc.PreviewFeatures;
-import jdk.internal.value.PrimitiveClass;
 import jdk.internal.misc.CDS;
 import jdk.internal.org.objectweb.asm.*;
 import sun.invoke.util.BytecodeDescriptor;
--- a/src/java.base/share/classes/java/lang/invoke/InnerClassLambdaMetafactory.java
+++ b/src/java.base/share/classes/java/lang/invoke/InnerClassLambdaMetafactory.java
@@ -25,6 +25,7 @@
 
 package java.lang.invoke;
 
+import jdk.internal.misc.PreviewFeatures;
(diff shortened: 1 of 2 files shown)


src/java.base/share/classes/java/lang/invoke/InnerClassLambdaMetafactory.java
```diff
@@ -642,7 +643,7 @@ boolean requiresPreload(Class<?> cls) {
             while (c.isArray()) {
                 c = c.getComponentType();
             }
-            return (c.isValue() && !PrimitiveClass.isPrimitiveClass(c)) || PrimitiveClass.isPrimitiveValueType(c);
```
RogerRiggs, line 645: Perhaps there should be a method that computes this compound expression?

RogerRiggs (approved):

src/java.base/share/classes/java/lang/invoke/InnerClassLambdaMetafactory.java
mlchung: This compound expression should not be needed.  Preload attribute is generated for a value class including primitive class.  It's a leftover when we added support for value objects.

PR #788, opened 2022-10-12
8247569: [lworld] Align with spec for <vnew> methods

MrSimms:
Merged javac and VM work together, includes VM support (JDK-8294224)

---------
### Progress

### Issues
 * [JDK-8247569](https://bugs.openjdk.org/browse/JDK-8247569): [lworld] Align with spec for <vnew> methods
 * [JDK-8294224](https://bugs.openjdk.org/browse/JDK-8294224): [lworld] VM support for <vnew> factory methods

### Reviewers
 * [Frederic Parain](https://openjdk.org/census#fparain) (@fparain - Committer) ⚠️ Review applies to [55e570ee](https://git.openjdk.org/valhalla/pull/788/files/55e570ee01d27e456bdc9540d99c9610db26064a)
 * [Mandy Chung](https://openjdk.org/census#mchung) (@mlchung - Committer) ⚠️ Review applies to [55e570ee](https://git.openjdk.org/valhalla/pull/788/files/55e570ee01d27e456bdc9540d99c9610db26064a)

### Contributors
 * Vicente Romero `<vromero@openjdk.org>`
 * Harold Seigel `<hseigel@openjdk.org>`
 * Mandy Chung `<mchung@openjdk.org>`

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla pull/788/head:pull/788` \
`$ git checkout pull/788`

`$ git checkout pull/788` \
`$ git pull https://git.openjdk.org/valhalla pull/788/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 788`

`$ git pr show -t 788`

</details>

<a href="https://git.openjdk.org/valhalla/pull/788.diff">https://git.openjdk.org/valhalla/pull/788.diff</a>

</details>


src/java.base/share/classes/java/lang/constant/DirectMethodHandleDescImpl.java
```diff
@@ -63,9 +63,10 @@ final class DirectMethodHandleDescImpl implements DirectMethodHandleDesc {
      * @jvms 4.2.2 Unqualified Names
      */
     DirectMethodHandleDescImpl(Kind kind, ClassDesc owner, String name, MethodTypeDesc type) {
-        if (kind == CONSTRUCTOR)
-            name = "<init>";
-
+        if (kind == CONSTRUCTOR) {
+            // TODO: missing "value objects", they don't use Q-type desc
```
MrSimms, line 67: Needs fixing to support Value Objects (currently only correct for primitive value types)

ExE-Boss:
Personally, I prefer calling it just `<new>`, which would better support [JDK‑8284239].

[JDK‑8284239]: https://bugs.openjdk.org/browse/JDK-8284239 "Consider adding `array::<new>(…)` and using it to compile `T[]::new` method references"

src/java.base/share/classes/java/lang/invoke/MemberName.java
```diff
@@ -522,9 +523,13 @@ public boolean isMethod() {
     public boolean isObjectConstructor() {
         return testAllFlags(IS_OBJECT_CONSTRUCTOR);
     }
+    /** Query whether this member is a constructor. */
+    boolean isConstructor(String name) {
```
mlchung, line 527: `findConstructor` only finds `<init>`.  To get `<vnew>` method handle, it just uses `Lookup::findStatic` as it's a static method.   This `isConstructor` method would cause confusion.  I can clean this up as a follow-up.

src/java.base/share/classes/java/lang/invoke/MethodHandles.java
```diff
@@ -2768,7 +2762,7 @@ public MethodHandle findConstructor(Class<?> refc, MethodType type) throws NoSuc
             if (type.returnType() != void.class) {
                 throw new NoSuchMethodException("Constructors must have void return type: " + refc.getName());
             }
-            String name = "<init>";
+            String name = refc.isValue() ? "<vnew>" : "<init>";
```
mlchung, line 2765: `findConstructor` should not find `<vnew>` which is just a static method.

src/java.base/share/classes/java/lang/invoke/MethodHandles.java
```diff
@@ -3704,14 +3698,20 @@ MemberName resolveOrNull(byte refKind, MemberName member) {
             return IMPL_NAMES.resolveOrNull(refKind, member, lookupClassOrNull(), allowedModes);
         }
 
+        boolean isIllegalSpecialName(byte refKind, String name) {
+            return (name.startsWith("<") &&
+            ((name.equals("<init>") && refKind != REF_newInvokeSpecial) ||
+            ((name.equals("<vnew>") || name.equals("<clinit>")) && refKind != REF_invokeStatic)));
```
mlchung, line 3704: This method should only check for `<vnew>`.   `<clinit>` cannot be looked up.

src/java.base/share/classes/jdk/internal/reflect/MethodAccessorGenerator.java
```diff
@@ -92,7 +92,7 @@ public ConstructorAccessor generateConstructor(Class<?> declaringClass,
     {
         boolean isStaticFactory = declaringClass.isValue();
         return (ConstructorAccessor) generate(declaringClass,
-                                              "<init>",
+                                              isStaticFactory ? "<vnew>" : "<init>",
```
mlchung, line 95: Core reflection does not use this method accessor generator by default.   I don't expect this change is needed.  Are you seeing any test failure without this change and other relevant change in jdk.internal.reflect?

test/jdk/valhalla/valuetypes/StaticFactoryMethodHandleTest.java
```diff
@@ -82,7 +82,7 @@ public Value(int x) {
      */
     @Test
     public void testNoArgStaticFactory() throws Throwable {
-        // test default static init factory
+        // test default vnew init factory
```
mlchung, line 85:
```suggestion
        // test default static vnew factory
```

src/java.base/share/classes/java/lang/invoke/MemberName.java
MrSimms: Thank you

src/java.base/share/classes/jdk/internal/reflect/MethodAccessorGenerator.java
MrSimms: It's gone, JDK-8281295 removed its use

src/java.base/share/classes/java/lang/constant/DirectMethodHandleDescImpl.java
mlchung: My patch I sent to you already fixed this issue.

src/hotspot/share/oops/method.cpp
```diff
@@ -942,12 +942,12 @@ bool Method::is_class_initializer() const {
 
 // A method named <init>, if non-static, is a classic object constructor.
 bool Method::is_object_constructor() const {
-   return name() == vmSymbols::object_initializer_name() && !is_static();
+  return name() == vmSymbols::object_initializer_name() && !is_static();
```
fparain, line 945: Do we still need to check that the method is not static?
Now that constructors and static factories have different names, the check of the static flag looks superfluous (applies to the method below as well, and to the comments above the method declaration).

fparain (approved):
VM changes look good to me (with one comment on is_object_constructor() and is_static_vnew_factory() methods).

mlchung (approved):
Looks good.  A couple more adjustment to the spec.

src/hotspot/share/oops/method.cpp
MrSimms: You are correct, any kind of correctness check has already happened during parsing

PR #806, opened 2022-11-10
8296767: Support multi-field based vector classes.

jatin-bhateja:
Initial Java side changes for treating vector as a value class based on suggestions from John Rose[1].

Summary of changes:
1) Re-define existing vector class hierarchy to use generic multi-field[2] based payloads of bit sizes 64, 128, 256 and 512.
2) Concrete vectors are now value classes encapsulating primitive class instances as payload.
3) Update fallback implementation to read/write from multi-field payload using Unsafe APIs.
4) New VM symbol definitions for generic payloads.

All existing JTREG tests are passing with -Xint option.

C2 side change, runtime support for vector object reconstruction, shuffle and masks handling is still based on existing flow
and will be handled in subsequent patch.

MAX species queries target supported max vector size, since multi-field annotation does not accept variable argument hence MAX species based vector class generation has been disabled currently.

[1] http://cr.openjdk.java.net/~jrose/values/multi-field.html
[2] https://github.com/fparain/valhalla/tree/multifield

---------
### Progress

### Issue
 * [JDK-8296767](https://bugs.openjdk.org/browse/JDK-8296767): Support multi-field based vector classes.

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla pull/806/head:pull/806` \
`$ git checkout pull/806`

`$ git checkout pull/806` \
`$ git pull https://git.openjdk.org/valhalla pull/806/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 806`

`$ git pr show -t 806`

</details>

<a href="https://git.openjdk.org/valhalla/pull/806.diff">https://git.openjdk.org/valhalla/pull/806.diff</a>

</details>


src/java.base/share/classes/jdk/internal/vm/vector/VectorSupport.java
```diff
@@ -148,35 +152,182 @@ public class VectorSupport {
...
-            super(payload);
+        public static VectorPayloadMF createVectPayloadInstance(int elemSize, int length) {
+            return newInstanceFactory(elemSize, length);
         }
-    }
-    public static class VectorMask<E> extends VectorPayload {
-        public VectorMask(Object payload) {
-            super(payload);
+
+        public static VectorPayloadMF createVectPayloadInstanceB(int elemSize, int length, byte [] init) {
```
PaulSandoz, line 205:
```suggestion
        public static VectorPayloadMF createVectPayloadInstanceB(int elemSize, int length, byte[] init) {
```
Same for others.

src/java.base/share/classes/jdk/internal/vm/vector/VectorSupport.java
```diff
@@ -148,35 +152,182 @@ public class VectorSupport {
...
+            VectorPayloadMF obj = newInstanceFactory(elemSize, length);
+            long start_offset = obj.multiFieldOffset();
+            obj = Unsafe.getUnsafe().makePrivateBuffer(obj);
+            for (int i = 0; i < length; i++) {
+                Unsafe.getUnsafe().putByte(obj, start_offset + i * Byte.BYTES, init[i]);
+            }
```
PaulSandoz, line 211: It may be possibly to unify all these create methods using `Unsafe.copyMemory`

src/jdk.incubator.vector/share/classes/jdk/incubator/vector/gen-src.sh
```diff
@@ -57,6 +57,7 @@ for type in byte short int long float double
 do
   Type="$(tr '[:lower:]' '[:upper:]' <<< ${type:0:1})${type:1}"
   TYPE="$(tr '[:lower:]' '[:upper:]' <<< ${type})"
+  Boxinitials="$(tr '[:lower:]' '[:upper:]' <<< ${type:0:1})"
```
PaulSandoz, line 60: No need for this. Just use the full box type name.

test/jdk/jdk/incubator/vector/ShortMaxVectorTests.java
```diff
@@ -24,7 +24,7 @@
 /*
  * @test
  * @modules jdk.incubator.vector
- * @run testng/othervm -ea -esa -Xbatch -XX:-TieredCompilation ShortMaxVectorTests
+ * @run testng/othervm -ea -esa -Xbatch ShortMaxVectorTests
```
PaulSandoz, line 27: No need for this change and in the other max vector tests?

test/jdk/jdk/incubator/vector/gen-tests.sh
```diff
@@ -133,7 +133,7 @@ do
   # For each size
   Log true "${Type}:"
 
-  for bits in 64 128 256 512 Max
+  for bits in 64 128 256 512
```
PaulSandoz, line 136: No need for this change?

PaulSandoz (commented):
I would recommend if you can to avoid unnecessary renaming such as uOp -> uOpMF, thereby making it easier to read/review as we go.

Something to do in a further iteration: we should be able to place the Unsafe access methods within static methods of the vector payload classe and avoid the use in the vector implementations.

PR #828, opened 2023-03-02
8303416: [lworld] Fix JVM crash at Unsafe_FinishPrivateBuffer

XiaohongGong:
When calling Unsafe.finishPrivateBuffer(), JVM crashes with following
 assertion failure:

```
 Internal Error (/mnt/local/code/valhalla/src/hotspot/share/prims/unsafe.cpp:388), pid=29517, tid=29518
  assert(v->mark().is_larval_state()) failed: must be a larval value

 JRE version: OpenJDK Runtime Environment (21.0) (fastdebug build 21-internal-git-05e4d0dba)
 Java VM: OpenJDK 64-Bit Server VM (fastdebug 21-internal-git-05e4d0dba, mixed mode, compressed oops, compressed class ptrs, g1 gc, linux-aarch64)
 Problematic frame:
 V  [[libjvm.so](http://libjvm.so/)+0x1a92a64]  Unsafe_FinishPrivateBuffer+0xc0
```
The assertion is used to check whether the input value is in larval
 state. Usually this method is called after `Unsafe.makePrivateBuffer()`,
 which will create a new oop and set it to larval state. If everything
 works fine, the above assertion won't fail. But the issue happens if
 all the fields of the input value object of `Unsafe.makePrivateBuffer()`
 are the relative java default values (i.e. `0`). The root cause is the
 C2 compiler will optimize the new buffered oop to the default oop for
 such value object, and the larval state is missed during the optimization.
 Marking the larval bit of the optimized oop can fix the issue.

---------
### Progress

### Issue
 * [JDK-8303416](https://bugs.openjdk.org/browse/JDK-8303416): [lworld] Fix JVM crash at Unsafe_FinishPrivateBuffer

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla pull/828/head:pull/828` \
`$ git checkout pull/828`

`$ git checkout pull/828` \
`$ git pull https://git.openjdk.org/valhalla pull/828/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 828`

`$ git pr show -t 828`

</details>

<a href="https://git.openjdk.org/valhalla/pull/828.diff">https://git.openjdk.org/valhalla/pull/828.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/opto/inlinetypenode.cpp
+++ b/src/hotspot/share/opto/inlinetypenode.cpp
@@ -912,26 +912,37 @@ InlineTypeNode* InlineTypeNode::make_from_multi(GraphKit* kit, MultiNode* multi,
 
 InlineTypeNode* InlineTypeNode::make_larval(GraphKit* kit, bool allocate) const {
   ciInlineKlass* vk = inline_klass();
-  InlineTypeNode* res = make_uninitialized(kit->gvn(), vk);
+  InlineTypeNode* vt = make_uninitialized(kit->gvn(), vk);
   for (uint i = 1; i < req(); ++i) {
-    res->set_req(i, in(i));
+    vt->set_req(i, in(i));
   }
 
+  Node* alloc_oop = NULL;
   if (allocate) {
     // Re-execute if buffering triggers deoptimization
     PreserveReexecuteState preexecs(kit);
     kit->jvms()->set_should_reexecute(true);
     Node* klass_node = kit->makecon(TypeKlassPtr::make(vk));
-    Node* alloc_oop  = kit->new_instance(klass_node, NULL, NULL, true);
+    alloc_oop = kit->new_instance(klass_node, NULL, NULL, true);
     AllocateNode* alloc = AllocateNode::Ideal_allocation(alloc_oop, &kit->gvn());
     alloc->_larval = true;
 
     store(kit, alloc_oop, alloc_oop, vk);
-    res->set_oop(alloc_oop);
+    vt->set_oop(alloc_oop);
   }
   // TODO 8239003
-  //res->set_type(TypeInlineType::make(vk, true));
-  res = kit->gvn().transform(res)->as_InlineType();
+  //vt->set_type(TypeInlineType::make(vk, true));
+  InlineTypeNode* res = kit->gvn().transform(vt)->as_InlineType();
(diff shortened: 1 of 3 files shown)


TobiHartmann (changes requested):
I don't think this is correct, as it would overwrite the larval state of the default oop/buffer of that type. Right?

XiaohongGong:
> I don't think this is correct, as it would overwrite the larval state of the default oop/buffer of that type. Right?

Yes, it would. And the larval state will be set back after calling "finishLarvalBuffer". I was thinking the whole progress happens in a single thread. But regarding to the default oop, changing larval state of it seems not right in multi-thread progress. So how about disgarding the GVN optimization of the `InlineTypeNode` if a private buffer is allocated and marked larval for it?

TobiHartmann:
> So how about disgarding the GVN optimization of the InlineTypeNode if a private buffer is allocated and marked larval for it?

Yes, I think that would make sense, although it's still an incomplete fix until JDK-8239003 is in.

XiaohongGong:
> > So how about disgarding the GVN optimization of the InlineTypeNode if a private buffer is allocated and marked larval for it?
> 
> Yes, I think that would make sense, although it's still an incomplete fix until JDK-8239003 is in.

Yes, agree! I will update the patch once I finish our internal testing. Thanks!

XiaohongGong:
Hi, 

The new commit reverts the larval state setting to the default oop.  As consider the default oop is immutable, the new private buffered oop with larval state is not right to be optimized to the default oop. Since we know its fields will be changed after then.  So the new commit just discards the optimization of the inline type if its oop is an allocated buffer with larval state. Note that current version is still an incomplete fix until [JDK-8239003](https://bugs.openjdk.java.net/browse/JDK-8239003) in. 

Please feel free to review this PR. Thanks a lot in advance!

Best Regards,
Xiaohong

TobiHartmann (approved):
Looks good to me!

XiaohongGong:
Thanks for the review @TobiHartmann !

TobiHartmann:
One review is sufficient for this and testing passed. Feel free to integrate.

XiaohongGong:
> One review is sufficient for this and testing passed. Feel free to integrate.

Thanks for the remainding! I will integrate it soon.

PR #848, opened 2023-05-22
8308363: Initial compiler support for FP16 scalar operations.

jatin-bhateja:
Starting with 4th Generation Xeon, Intel has made extensive extensions to existing ISA to support 16 bit scalar and vector floating point operations based on IEEE 754 binary16 format.

We plan to support this in multiple stages spanning across Java side definition of Float16 type, scalar operation and finally SLP vectorization support.

This patch adds  minimal Java and Compiler side support for one API Float16.add.

**Summary of changes :-**
- Minimal implementation of Float16 primitive class supporting one operation (Float16.add)
- X86 AVX512-FP16 feature detection at VM startup.
- C2 IR and Inline expander changes for Float16.add API.
- FP16 constant folding handling.
- Backend support : Instruction selection patterns and assembler support.
- New IR framework and functional tests.

**Implementation details:-**

1/ Newly defined Float16 class encapsulate a short value holding IEEE 754 binary16 encoded value.

2/ Float16 is a primitive class which in future will be aligned with other enhanced primitive wrapper classes proposed by [JEP-402.](https://openjdk.org/jeps/402)

3/ Float16 to support all the operations supported by corresponding Float class.

4/ Java implementation of each API will internally perform floating point operation at FP32 granularity.

5/ API which can be directly mapped to an Intel AVX512FP16 instruction will be a candidate for intensification by C2 compiler.

6/ With Valhalla, C2 compiler always creates an InlineType IR node for a value class instance.
Total number of inputs of an InlineType node match the number of non-static fields. In this case node will have one input of short type TypeInt::SHORT.

7/ Since all the scalar AVX512FP16 instructions operate on floating point registers and Float16 backing storage is held in a general-purpose register hence we need to introduce appropriate conversion IR which moves a 16-bit value from GPR to a XMM register and vice versa.
image

8/ Current plan is to introduce a new IR node for each operation which is a subclass of its corresponding single precision IR node. This will allow leveraging idealization routines (Ideal/Identity/Value) of its parent operation.

9/ All the single/double precision IR nodes carry a Type::FLOAT/DOUBLE ideal type. This represents entire FP32/64 value range and is different from integral types which explicitly record lower and upper bounds of value ranges. Value resolution routines operating over integral types calibrates the lattice associated with IR node based on flow function depicting the semantics of the operation.

10/ There are two separate ideal type for floating point constants TypeF and TypeD, these are singleton types and represents a unique single/double precision floating point value 

11/ Ideal types are linked to specific register classes and determine the allocation set for register allocator. 

12/ Since FP16 value range is a proper subset of FP32 value range hence all the newly created IR nodes will still carry Type::FLOAT type. 

13/ To handle constant folding scenarios value routines associated with newly introduced conversion IR nodes convert an FP16 constant to a FP32 constant (TypeF) value using existing runtime helpers. Constant folding is performed by existing idealization routines of parent IR nodes.

14/ Auto-vectorizer support will scan through packs of scalar IR to create a vector IR, we plan to address this after scalar support.

15/ Functional and performance test for creation for each newly supported operation.

Above design is generic and should work for any reduced precision types like FP8 values.

Please let me know your views on this.

Best Regards,
Jatin 

PS: Test instantiating Float16 class must be compiled with _-XDenablePrimitiveClasses_ currently,  we need to relax this constraint for known primitive classes in language compiler, JEP-402 talks about converting existing primitive box classes to value classes should address this limitation.

---------
### Progress

### Issue
 * [JDK-8308363](https://bugs.openjdk.org/browse/JDK-8308363): Initial compiler support for FP16 scalar operations. (**Enhancement** - P4)

### Reviewers
 * [Sandhya Viswanathan](https://openjdk.org/census#sviswanathan) (@sviswa7 - no project role) ⚠️ Review applies to [ed10ca90](https://git.openjdk.org/valhalla/pull/848/files/ed10ca90756e17e19c8dbf7f3fb2d7fa32dd2f8e)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/848/head:pull/848` \
`$ git checkout pull/848`

`$ git checkout pull/848` \
`$ git pull https://git.openjdk.org/valhalla.git pull/848/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 848`

`$ git pr show -t 848`

</details>

<a href="https://git.openjdk.org/valhalla/pull/848.diff">https://git.openjdk.org/valhalla/pull/848.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/848#issuecomment-1557603742)


jatin-bhateja:
> ### Webrevs
> * 00: [Full](https://webrevs.openjdk.org/?repo=valhalla&pr=848&range=00) ([98d8ea1f](https://git.openjdk.org/valhalla/pull/848/files/98d8ea1f207eca16d1f8bf02f55115ecbea16f15))

image

src/hotspot/share/classfile/classFileParser.cpp
```diff
@@ -5166,7 +5173,7 @@ const char* ClassFileParser::skip_over_field_signature(const char* signature,
     case JVM_SIGNATURE_PRIMITIVE_OBJECT:
       // Can't enable this check fully until JDK upgrades the bytecode generators (TODO: JDK-8270852).
       // For now, compare to class file version 51 so old verifier doesn't see Q signatures.
-      if ( (_major_version < 51 /* CONSTANT_CLASS_DESCRIPTORS */ ) || (!EnablePrimitiveClasses)) {
+      if ( (_major_version < 51 /* CONSTANT_CLASS_DESCRIPTORS */ ) || (!EnablePrimitiveClasses && !is_jdk_internal_class_sig(signature))) {
```
XiaohongGong, line 5176: The last condition `(!EnablePrimitiveClasses && !is_jdk_internal_class_sig(signature))` seems not right to me. Assume there is a jdk internal primitive class defined , and `EnablePrimitiveClasses` is disabled, then result of `(!EnablePrimitiveClasses && !is_jdk_internal_class_sig(signature))` is `false` (i.e. `true && false`), then the followed error is not printed if the jdk version matches >= 51. But it should report the error since `EnablePrimitiveClasses` is closed.  So it should be:
```
if ( (_major_version < 51 /* CONSTANT_CLASS_DESCRIPTORS */ ) || !EnablePrimitiveClasses || !is_jdk_internal_class_sig(signature))
```
right?

src/hotspot/share/classfile/vmIntrinsics.hpp
```diff
@@ -195,6 +195,12 @@ class methodHandle;
   do_intrinsic(_dsignum,                  java_lang_Math,         signum_name,        double_double_signature,   F_S)   \
   do_intrinsic(_fsignum,                  java_lang_Math,         signum_name,        float_float_signature,     F_S)   \
                                                                                                                         \
+                                                                                                                        \
+  /* Float16 intrinsics, similar to what we have in Math. */                                                            \
+  do_intrinsic(_add_float16,              java_lang_Float16,      add_name,           floa16_float16_signature,  F_R)   \
+   do_name(add_name,    "add")                                                                                          \
```
XiaohongGong, line 201: style: please remove extra spaces between `add_name` and `"add"`.

src/hotspot/share/opto/addnode.hpp
```diff
@@ -129,6 +129,14 @@ class AddFNode : public AddNode {
   virtual uint ideal_reg() const { return Op_RegF; }
 };
 
+//------------------------------AddHFNode---------------------------------------
+// Add 2 floats
```
XiaohongGong, line 135: Change to: `Add 2 half-precision floats` ?

src/hotspot/share/opto/convertnode.hpp
```diff
@@ -174,6 +174,24 @@ class ConvI2FNode : public Node {
   virtual uint  ideal_reg() const { return Op_RegF; }
 };
 
+class ReinterpretS2HFNode : public Node {
+  public:
+  ReinterpretS2HFNode( Node *in1 ) : Node(0,in1) {}
```
XiaohongGong, line 178: Suggest changes:
```
ReinterpretS2HFNode(Node* in1) : Node(0, in1) {}
```

src/hotspot/share/opto/convertnode.hpp
```diff
@@ -174,6 +174,24 @@ class ConvI2FNode : public Node {
...
+  virtual int Opcode() const;
+  virtual const Type *bottom_type() const { return Type::FLOAT; }
```
XiaohongGong, line 180: Suggest to:
```
virtual const Type* bottom_type() const { return Type::FLOAT; }
```

src/hotspot/share/opto/convertnode.hpp
```diff
@@ -174,6 +174,24 @@ class ConvI2FNode : public Node {
...
+  virtual const Type* Value(PhaseGVN* phase) const;
+  virtual uint  ideal_reg() const { return Op_RegF; }
+};
+
+class ReinterpretHF2SNode : public Node {
+  public:
+  ReinterpretHF2SNode( Node *in1 ) : Node(0,in1) {}
+  virtual int Opcode() const;
+  virtual const Type* Value(PhaseGVN* phase) const;
+  virtual const Type *bottom_type() const { return TypeInt::SHORT; }
```
XiaohongGong, line 191: Same as above.

src/hotspot/share/opto/library_call.cpp
```diff
@@ -4788,6 +4790,29 @@ bool LibraryCallKit::inline_native_Reflection_getCallerClass() {
   return false;  // bail-out; let JVM_GetCallerClass do the work
 }
 
+bool LibraryCallKit::inline_fp16_operations(vmIntrinsics::ID id) {
+  Node* result = NULL;
```
XiaohongGong, line 4900: Please use `nullptr` instead of `NULL`.

src/hotspot/share/classfile/classFileParser.cpp
jatin-bhateja: Idea here is to relax the need to use an explicit JVM flag -XX:+EnablePrimtiiveClasses for primitive classes known to VM.

jatin-bhateja:
Performance data for Float16 vector ADD on Sapphire Rapids with different configuration.
image

sviswa7 (commented):
Very good work in general. I only have couple of comments in the code, please take a look. Also there might be some further optimization opportunities for the path that comes from Op_ConvF2HF ->  ReinterpretS2HF.  ConvF2HF is doing the conversion from xmm to xmm register and then moves the xmm to gpr. ReinterpretS2HF then moves from gpr back to xmm. This unnecessary movement from xmm->gpr and from gpr->xmm could be optimized out.

src/hotspot/cpu/x86/x86.ad
```diff
@@ -10149,4 +10161,44 @@ instruct DoubleClassCheck_reg_reg_vfpclass(rRegI dst, regD src, kReg ktmp, rFlag
...
+  ins_pipe(pipe_slow);
+%}
 
+instruct reinterpretH2S (rRegI dst, regF src)
+%{
+  match(Set dst (ReinterpretHF2S src));
+  format %{ "vmovw $dst, $src" %}
+  ins_encode %{
+    __ vmovw($dst$$Register, $src$$XMMRegister);
+    __ movswl($dst$$Register, $dst$$Register);
```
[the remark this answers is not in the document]
jatin-bhateja: > Could we do without movswl here?

This was done keeping in mind JVM semantics where byte/short are internally promoted to int type since JVM operands and local variables are 32 bit values.

make/common/JavaCompilation.gmk
```diff
@@ -274,7 +274,7 @@ define SetupJavaCompilationBody
   PARANOIA_FLAGS := -implicit:none -Xprefer:source -XDignore.symbol.file=true -encoding ascii
 
   $1_FLAGS += -g -Xlint:all $$($1_TARGET_RELEASE) $$(PARANOIA_FLAGS) $$(JAVA_WARNINGS_ARE_ERRORS)
-  $1_FLAGS += $$($1_JAVAC_FLAGS)
+  $1_FLAGS += $$($1_JAVAC_FLAGS) -XDenablePrimitiveClasses
```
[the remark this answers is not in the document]
jatin-bhateja: > Do we need this change now that we have special handling in the VM for Float16?

Yes, Float16 being a primitive class needs this flag during build time compilation.

src/hotspot/share/classfile/vmIntrinsics.hpp
```diff
@@ -195,6 +195,12 @@ class methodHandle;
   do_intrinsic(_dsignum,                  java_lang_Math,         signum_name,        double_double_signature,   F_S)   \
   do_intrinsic(_fsignum,                  java_lang_Math,         signum_name,        float_float_signature,     F_S)   \
                                                                                                                         \
+                                                                                                                        \
+  /* Float16 intrinsics, similar to what we have in Math. */                                                            \
+  do_intrinsic(_sum_float16,              java_lang_Float16,      sum_name,           floa16_float16_signature,  F_S)   \
+   do_name(sum_name, "sum")                                                                                             \
```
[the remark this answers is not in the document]
jatin-bhateja: > could this be called add instead of sum?

I made this change after Paul offline comments, to align it with Float.sum API.

jatin-bhateja:
> Very good work in general. I only have couple of comments in the code, please take a look. Also there might be some further optimization opportunities for the path that comes from Op_ConvF2HF -> ReinterpretS2HF. ConvF2HF is doing the conversion from xmm to xmm register and then moves the xmm to gpr. ReinterpretS2HF then moves from gpr back to xmm. This unnecessary movement from xmm->gpr and from gpr->xmm could be optimized out.

Yes, but this can only be done if we factor out xmm -> gpr movement out of ConvF2HF,  its inputs is a IEEE 754 binary32 bit floating point value and output is a binary16 bit value help in a GPR. It will not be proper to remove ConvF2HF + ReinterpretS2HF by a direct ideal transformation currently.

src/hotspot/cpu/x86/x86.ad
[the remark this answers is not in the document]
sviswa7: But this is actually a float16 here (short is only indicating the 16bit storage) and it is not expected to have any integral operation directly on this. The movswl would unnecessarily add to the pathlength.

sviswa7:
> > Very good work in general. I only have couple of comments in the code, please take a look. Also there might be some further optimization opportunities for the path that comes from Op_ConvF2HF -> ReinterpretS2HF. ConvF2HF is doing the conversion from xmm to xmm register and then moves the xmm to gpr. ReinterpretS2HF then moves from gpr back to xmm. This unnecessary movement from xmm->gpr and from gpr->xmm could be optimized out.
> 
> Yes, but this can only be done if we factor out xmm -> gpr movement out of ConvF2HF, its inputs is a IEEE 754 binary32 bit floating point value and output is a binary16 bit value help in a GPR. It will not be proper to remove ConvF2HF + ReinterpretS2HF by a direct ideal transformation currently.

One path is having an instruct with match rule for this, something like below I think:
match(Set dst (ReinterprestS2HF(ConvF2HF src)));

sviswa7 (approved):

jatin-bhateja:
> > > Very good work in general. I only have couple of comments in the code, please take a look. Also there might be some further optimization opportunities for the path that comes from Op_ConvF2HF -> ReinterpretS2HF. ConvF2HF is doing the conversion from xmm to xmm register and then moves the xmm to gpr. ReinterpretS2HF then moves from gpr back to xmm. This unnecessary movement from xmm->gpr and from gpr->xmm could be optimized out.
> > 
> > 
> > Yes, but this can only be done if we factor out xmm -> gpr movement out of ConvF2HF, its inputs is a IEEE 754 binary32 bit floating point value and output is a binary16 bit value help in a GPR. It will not be proper to remove ConvF2HF + ReinterpretS2HF by a direct ideal transformation currently.
> 
> One path is having an instruct with match rule for this, something like below I think: match(Set dst (ReinterprestS2HF(ConvF2HF src)));

Agree, patten match will break if ConF2HF is shared across multiple nodes, but its still better to have than no optimization.

jatin-bhateja:
Hi @sviswa7 , @XiaohongGong , All your comments have been addressed. I am committing this patch

Best Regards

PR #860, opened 2023-06-09
8309650: [lworld] Fix mismatch inline type issue during method calls

XiaohongGong:
JVM crashes with following assertion failure when running the new added
 test case (i.e. `compiler/valhalla/inlinetypes/TestCastMismatch.java`):

```

 Internal Error (/mnt/local/code/valhalla/src/hotspot/share/opto/graphKit.cpp:3414), pid=1533674, tid=1533690
 assert(stopped() || !toop->is_inlinetypeptr() || obj->is_InlineType()) failed: should have been scalarized

 Problematic frame:
 V [[libjvm.so](http://libjvm.so/)+0xd2c070] GraphKit::gen_checkcast(Node*, Node*, Node**, bool)+0x46c
```

The failure happens when compiler knows the type of the `checkcast`
 input is a value/primitive class, while the input node is not an
 `InlineTypeNode`. Note that `InlineTypeNode` is use to represent each
 value/primitive class instance.

In the test case, the input of cast is a method call. And from the
 method declaration, the return type of this method is an abstract class
 whose concrete subclass is a value class. From the method signature,
 the compiler knows the return type is not a value class. Hence, when
 it handles the return result of this method call, no `InlineTypeNode`
 is generated. But when compiler handles the `checkcast` bytecode, it
 checks the speculative type of the input, which is actually the concrete
 class of the instance. In this case, it is the value class. So compiler
 expects the input node is an `InlineTypeNode`.

To fix this issue, the compiler has to use the consistent type system
 for different routines. This patch changes to use the speculative
 type of the method call as well when handling the call's return result.
 With this change, an `InlineTypeNode` is generated for this case after
 the call, and the assertion pass.

Verified with `hotspot::hotspot_all, jdk::tier1,tier2,tier3, langtool::tier1`

---------
### Progress

### Issue
 * [JDK-8309650](https://bugs.openjdk.org/browse/JDK-8309650): [lworld] Fix mismatch inline type issue during method calls (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/860/head:pull/860` \
`$ git checkout pull/860`

`$ git checkout pull/860` \
`$ git pull https://git.openjdk.org/valhalla.git pull/860/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 860`

`$ git pr show -t 860`

</details>

<a href="https://git.openjdk.org/valhalla/pull/860.diff">https://git.openjdk.org/valhalla/pull/860.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/860#issuecomment-1584190509)

### DIFF
--- a/src/hotspot/share/opto/callGenerator.cpp
+++ b/src/hotspot/share/opto/callGenerator.cpp
@@ -1167,7 +1167,7 @@ static void cast_argument(int nargs, int arg_nb, ciType* t, GraphKit& kit, bool 
     kit.set_argument(arg_nb, arg);
   }
   if (sig_type->is_inlinetypeptr()) {
-    arg = InlineTypeNode::make_from_oop(&kit, arg, t->as_inline_klass(), !kit.gvn().type(arg)->maybe_null());
+    arg = InlineTypeNode::make_from_oop(&kit, arg, sig_type->inline_klass(), !kit.gvn().type(arg)->maybe_null());
     kit.set_argument(arg_nb, arg);
   }
 }
--- a/src/hotspot/share/opto/graphKit.cpp
+++ b/src/hotspot/share/opto/graphKit.cpp
@@ -1950,8 +1950,12 @@ Node* GraphKit::set_results_for_java_call(CallJavaNode* call, bool separate_io_p
     ret = InlineTypeNode::make_from_multi(this, call, vk, base_input, false, call->method()->signature()->returns_null_free_inline_type());
   } else {
     ret = _gvn.transform(new ProjNode(call, TypeFunc::Parms));
-    if (call->method()->return_type()->is_inlinetype()) {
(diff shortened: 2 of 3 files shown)


XiaohongGong:
Hi @TobiHartmann , could you please kindly take a look at this PR? Thanks a lot in advance!

TobiHartmann (approved):
Looks good to me, thanks for fixing! Let me run this through our testing and report back once it passed.

TobiHartmann:
All testing passed.

XiaohongGong:
Do I need another review or can integrate it now? Thanks!

TobiHartmann:
One review is sufficient. Please integrate.

PR #863, opened 2023-06-20
8309777: [lworld+vector] Fix re-materialize crash for vector objects with un-vectorized multi-fields during deoptimization

XiaohongGong:
Some vector API jtreg tests crash when running on hardwares that do
 not support the relative vector species.

Here is the log:

```

  Internal Error (valhalla/src/hotspot/share/prims/vectorSupport.cpp:317), pid=2529357, tid=2529377
  assert(loc_type == Location::oop || loc_type == Location::narrowoop) failed: expected 'oop'(2) or 'narrowoop'(9) types location but got: 3

  JRE version: OpenJDK Runtime Environment (21.0) (fastdebug build 21-internal-git-62ba7fc27)
  Java VM: OpenJDK 64-Bit Server VM (fastdebug 21-internal-git-62ba7fc27, mixed mode, sharing, compressed oops, compressed class ptrs, g1 gc, linux-aarch64)
  Problematic frame:
  V [[libjvm.so](http://libjvm.so/)+0x1aed828] VectorSupport::allocate_vector_payload(InstanceKlass*, int, BasicType, frame*, RegisterMap*, ObjectValue*, JavaThread*)+0x41
```

The root cause is `VectorSupport::allocate_vector_payload` assumes the
 `ScopeValue` which denotes the field value generated by C2 compiler
 is a vector or an oop. But after the vector's payload is changed to be
 the `MultiField` annotated field, the non-static fields of the vector
 class is changed to be a series of fields with the same primitive type.
 In C2 compiler, operations for such multi-fields can be vectorized only
 if the running hardware supports the vector operations for the specified
 vector size. For such cases, the field value generated by C2 is a vector.
 Otherwise, the fields are the scalarized primitive type values. Hence
 the re-materizalization for such vector objects should be handled just
 like other normal objects.

To make things right when re-materializing vector objects whose fields
 are not vectorized, two parts need to be modified:

1. In c2 compiler, pass the un-vectorized multi-fields on safepoint when
 scalaring an `InlineTypeNode`. To try to vectorize the multi-fields in
 compiler, only the multi-field base is added to the klass's non-static
 field list. And the other multi-fields are saved in a secondary field list
 of their multi-field base. So if the multi-fields are not vectorized, all
 the field members should be found out and passed to the safepoint.

2. During deoptimization, re-materialize the vector objects with different
 routines based on the `ScopeValue` types. That is, go to the `VectorSupport`
 routine to re-allocate the objects from a vector value, while go to the
 normal objects re-allocation path if the field value is not a vector.

---------
### Progress

### Issue
 * [JDK-8309777](https://bugs.openjdk.org/browse/JDK-8309777): [lworld+vector] Fix re-materialize crash for vector objects with un-vectorized multi-fields during deoptimization (**Enhancement** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/863/head:pull/863` \
`$ git checkout pull/863`

`$ git checkout pull/863` \
`$ git pull https://git.openjdk.org/valhalla.git pull/863/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 863`

`$ git pr show -t 863`

</details>

<a href="https://git.openjdk.org/valhalla/pull/863.diff">https://git.openjdk.org/valhalla/pull/863.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/863#issuecomment-1598190157)

### DIFF
--- a/src/hotspot/share/opto/callnode.cpp
+++ b/src/hotspot/share/opto/callnode.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 1997, 2022, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1997, 2023, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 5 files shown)


XiaohongGong:
Since this issue is fixed by https://github.com/openjdk/valhalla/pull/866 which has been merged, close this PR now.

PR #872, opened 2023-06-23
8310685: [lw5] adding support for null restricted types

vicente-romero-oracle:
initial support for null restricted types

---------
### Progress

### Issue
 * [JDK-8310685](https://bugs.openjdk.org/browse/JDK-8310685): [lw5] adding support for null restricted types (**Bug** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/872/head:pull/872` \
`$ git checkout pull/872`

`$ git checkout pull/872` \
`$ git pull https://git.openjdk.org/valhalla.git pull/872/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 872`

`$ git pr show -t 872`

</details>

<a href="https://git.openjdk.org/valhalla/pull/872.diff">https://git.openjdk.org/valhalla/pull/872.diff</a>

</details>

### DIFF
--- a/src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
+++ b/src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
@@ -50,6 +50,8 @@ import com.sun.tools.javac.comp.Check;
 import com.sun.tools.javac.comp.Enter;
 import com.sun.tools.javac.comp.Env;
 import com.sun.tools.javac.jvm.ClassFile;
+import com.sun.tools.javac.resources.CompilerProperties.Errors;
+import com.sun.tools.javac.resources.CompilerProperties.Warnings;
 import com.sun.tools.javac.util.*;
 
 import static com.sun.tools.javac.code.BoundKind.*;
@@ -118,7 +120,12 @@ public class Types {
         capturedName = names.fromString("<captured wildcard>");
         messages = JavacMessages.instance(context);
         diags = JCDiagnostic.Factory.instance(context);
-        noWarnings = new Warner(null);
+        noWarnings = new Warner(null) {
+            @Override
+            public String toString() {
+                return "NO_WARNINGS";
+            }
+        };
     }
     // </editor-fold>
 
@@ -1043,7 +1050,7 @@ public class Types {
                         return false;
                     return true;
                 }
-            } else if (isSubtype(t, s, capture)) {
+            } else if (isSubtype(t, s, capture, warn)) {
                 return true;
             } else if (t.hasTag(TYPEVAR)) {
                 return isSubtypeUncheckedInternal(t.getUpperBound(), s, false, warn);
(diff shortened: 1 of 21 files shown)


src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
```diff
@@ -1136,10 +1163,17 @@ public Boolean visitType(Type t, Type s) {
                      return t.hasTag(s.getTag());
                  case TYPEVAR:
                      return isSubtypeNoCapture(t.getUpperBound(), s);
-                 case BOT:
+                 case BOT: {
+                     /* this method can be invoked even from the backend, and warnings can be printed again, so
+                      * make sure that the caller really wants to warn
+                      */
+                     if (s.isNonNullable() && warnStack.head.pos() != null) {
```
mcimadamore, line 1170: This is a bit odd. We are defining a type relationship - if "BOT" is not a subtype of T! that is fine, we should just return false. Giving an error seems a bit strange here (I don't think it is normal for subtyping to generate errors).

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
```diff
@@ -1204,11 +1238,15 @@ public Boolean visitClassType(ClassType t, Type s) {
                 if (sup == null) return false;
                 // If t is an intersection, sup might not be a class type
                 if (!sup.hasTag(CLASS)) return isSubtypeNoCapture(sup, s);
-                return sup.tsym == s.tsym
+                boolean result = sup.tsym == s.tsym
                      // Check type variable containment
                     && (!s.isParameterized() || containsTypeRecursive(s, sup))
                     && isSubtypeNoCapture(sup.getEnclosingType(),
                                           s.getEnclosingType());
+                if (result && warnStack.head.pos() != null && s.hasNarrowerNullabilityThan(t)) {
```
mcimadamore, line 1246: I have memories of the previous implementation where we discussed maybe keeping subtyping routines as is, but adding some other check to validate the "polarity" of the type arguments. Wouldn't something similar work here too?

src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/Gen.java
```diff
@@ -2115,6 +2123,10 @@ public void visitParens(JCParens tree) {
     public void visitAssign(JCAssign tree) {
         Item l = genExpr(tree.lhs, tree.lhs.type);
         genExpr(tree.rhs, tree.lhs.type).load();
+        if (tree.lhs.type.isNonNullable()) {
+            code.emitop0(dup);
+            genNullCheck(tree.rhs);
```
mcimadamore, line 2128: Should the cast always be generated even if we know that the RHS is non-nullable? (I'm not sure here)

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
vicente-romero-oracle: yep you are right

PR #880, opened 2023-06-28
8310905: [lw5] addressing review comments on null restricted types

vicente-romero-oracle:
addressing some review comments related to PR [1]

[1] https://github.com/openjdk/valhalla/pull/872

---------
### Progress

### Issue
 * [JDK-8310905](https://bugs.openjdk.org/browse/JDK-8310905): [lw5] addressing review comments on null restricted types (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/880/head:pull/880` \
`$ git checkout pull/880`

`$ git checkout pull/880` \
`$ git pull https://git.openjdk.org/valhalla.git pull/880/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 880`

`$ git pr show -t 880`

</details>

<a href="https://git.openjdk.org/valhalla/pull/880.diff">https://git.openjdk.org/valhalla/pull/880.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/880#issuecomment-1610675605)

### DIFF
--- a/src/jdk.compiler/share/classes/com/sun/tools/javac/code/Type.java
+++ b/src/jdk.compiler/share/classes/com/sun/tools/javac/code/Type.java
@@ -396,7 +396,7 @@ public abstract class Type extends AnnoConstruct implements TypeMirror, PoolCons
     @SuppressWarnings("unchecked")
     public <M extends TypeMetadata, Z> Z getMetadata(Class<M> metadataClass, Function<M, Z> metadataFunc, Z defaultValue) {
         for (TypeMetadata m : metadata) {
-            if (m.getClass() == metadataClass) {
+            if (m.getClass().isAssignableFrom(metadataClass)) {
                 return metadataFunc.apply((M)m);
             }
         }
--- a/src/jdk.compiler/share/classes/com/sun/tools/javac/code/Type.java
+++ b/src/jdk.compiler/share/classes/com/sun/tools/javac/code/Type.java
@@ -783,38 +783,6 @@ public abstract class Type extends AnnoConstruct implements TypeMirror, PoolCons
         return getMetadata(TypeMetadata.NullMarker.class) == null;
     }
 
-    public boolean sameNullabilityAs(Type t) {
-        if (t == null) {
-            return isNullUnspecified();
-        }
-        if (isNullUnspecified()) {
-            return t.isNullUnspecified();
-        }
-        if (isNonNullable()) {
-            return t.isNonNullable();
-        }
-        if (isNullable()) {
-            return t.isNullable();
-        }
-        if (isParametric()) {
-            return t.isParametric();
-        }
-        throw new AssertionError("shouldn't get here");
-    }
-
-    public boolean hasNarrowerNullabilityThan(Type t) {
-        if (isNonNullable()) {
-            return !t.isNonNullable();
-        }
-        if (isParametric()) {
-            return t.isNonNullable() || t.isNullUnspecified();
-        }
-        if (isNullable()) {
-            return t.isNullUnspecified();
-        }
-        return false;
-    }
-
     // end of support for null-marked types
 
     @Override @DefinedBy(Api.LANGUAGE_MODEL)
--- a/src/jdk.compiler/share/classes/com/sun/tools/javac/code/Type.java
+++ b/src/jdk.compiler/share/classes/com/sun/tools/javac/code/Type.java
@@ -543,6 +543,13 @@ public abstract class Type extends AnnoConstruct implements TypeMirror, PoolCons
         if (moreInfo && hasTag(TYPEVAR)) {
             sb.append(hashCode());
         }
+        if (isNullable()) {
+            sb.append("?");
+        } else if (isNonNullable()) {
+            sb.append("!");
+        } else if (isParametric()) {
+            sb.append("*");
+        }
         return sb.toString();
     }
 
--- a/src/jdk.compiler/share/classes/com/sun/tools/javac/code/Type.java
+++ b/src/jdk.compiler/share/classes/com/sun/tools/javac/code/Type.java
@@ -784,17 +784,34 @@ public abstract class Type extends AnnoConstruct implements TypeMirror, PoolCons
     }
 
     public boolean sameNullabilityAs(Type t) {
-        if (isNullUnspecified()) return t.isNullUnspecified();
-        if (isNonNullable()) return t.isNonNullable();
-        if (isNullable()) return t.isNullable();
-        if (isParametric()) return t.isParametric();
+        if (t == null) {
+            return isNullUnspecified();
+        }
+        if (isNullUnspecified()) {
+            return t.isNullUnspecified();
+        }
+        if (isNonNullable()) {
+            return t.isNonNullable();
+        }
+        if (isNullable()) {
+            return t.isNullable();
+        }
+        if (isParametric()) {
+            return t.isParametric();
+        }
         throw new AssertionError("shouldn't get here");
     }
 
     public boolean hasNarrowerNullabilityThan(Type t) {
-        if (isNonNullable()) return !t.isNonNullable();
-        if (isParametric()) return t.isNonNullable() || t.isNullUnspecified();
-        if (isNullable()) return t.isNullUnspecified();
+        if (isNonNullable()) {
+            return !t.isNonNullable();
+        }
+        if (isParametric()) {
+            return t.isNonNullable() || t.isNullUnspecified();
+        }
+        if (isNullable()) {
+            return t.isNullUnspecified();
+        }
         return false;
     }
 
--- a/src/jdk.compiler/share/classes/com/sun/tools/javac/code/Type.java
+++ b/src/jdk.compiler/share/classes/com/sun/tools/javac/code/Type.java
@@ -396,7 +396,7 @@ public abstract class Type extends AnnoConstruct implements TypeMirror, PoolCons
     @SuppressWarnings("unchecked")
     public <M extends TypeMetadata, Z> Z getMetadata(Class<M> metadataClass, Function<M, Z> metadataFunc, Z defaultValue) {
         for (TypeMetadata m : metadata) {
-            if (m.getClass().isAssignableFrom(metadataClass)) {
+            if (m.getClass() == metadataClass) {
                 return metadataFunc.apply((M)m);
             }
         }
@@ -1115,7 +1115,9 @@ public abstract class Type extends AnnoConstruct implements TypeMirror, PoolCons
         }
 
         public Type constType(Object constValue) {
-            return addMetadata(new ConstantValue(constValue));
+            return isPrimitive() ?
+                    addMetadata(new ConstantValue(constValue)) :
+                    addMetadata(new ConstantValue(constValue)).addMetadata(new TypeMetadata.NullMarker(NullMarker.NOT_NULL));
         }
 
         /** The Java source which this type represents.
--- a/src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
+++ b/src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
@@ -1301,10 +1301,36 @@ public class Types {
             public Boolean visitType(Type t, Type s) {
                 if (differentNullability.apply(t, s)) {
                     warner.warn(LintCategory.NULL);
-                    return false;
-                } else {
                     return true;
+                } else {
+                    return false;
                 }
+            }
+
+            @Override
+            public Boolean visitClassType(ClassType t, Type s) {
+                if (differentNullability.apply(t, s)) {
+                    warner.warn(LintCategory.NULL);
+                    return true;
+                } else {
+                    return s != null && (!s.isParameterized() || compareTypeArgsRecursive(t, s))
+                            && visit(t.getEnclosingType(), s.getEnclosingType());
+                }
+            }
+            // where
+            boolean compareTypeArgsRecursive(Type t, Type s) {
+                return compareTypeArgs(t.getTypeArguments(), s.getTypeArguments());
+            }
+
+            boolean compareTypeArgs(List<Type> ts, List<Type> ss) {
+                while (ts.nonEmpty() && ss.nonEmpty()) {
+                    if (visit(ts.head, ss.head)) {
+                        return true;
+                    }
+                    ts = ts.tail;
+                    ss = ss.tail;
+                }
+                return false;
             }
         }
 
--- a/src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
+++ b/src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
@@ -1105,7 +1105,7 @@ public class Types {
     }
     public boolean isSubtype(Type t, Type s, boolean capture) {
         if (t.equalsIgnoreMetadata(s)) {
-            nullabilityComparator.reset((t1, t2) -> hasNarrowerNullability(t1, t2)).visit(s, t);
+            new NullabilityComparator((t1, t2) -> hasNarrowerNullability(t1, t2)).visit(s, t);
             return true;
         }
         if (s.isPartial())
@@ -1225,7 +1225,7 @@ public class Types {
                     && isSubtypeNoCapture(sup.getEnclosingType(),
                                           s.getEnclosingType());
                 if (result) {
-                    nullabilityComparator.reset((t1, t2) -> hasNarrowerNullability(t1, t2)).visit(s, t);
+                    new NullabilityComparator((t1, t2) -> hasNarrowerNullability(t1, t2)).visit(s, t);
                 }
                 return result;
             }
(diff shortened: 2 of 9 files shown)


src/jdk.compiler/share/classes/com/sun/tools/javac/code/Type.java
```diff
@@ -543,6 +543,13 @@ public String toString() {
         if (moreInfo && hasTag(TYPEVAR)) {
             sb.append(hashCode());
         }
+        if (isNullable()) {
```
vicente-romero-oracle, line 546: strictly speaking currently we should only have the non-nullable case but I have provided the rest for completeness

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
```diff
@@ -1291,7 +1272,41 @@ public Boolean visitUndetVar(UndetVar t, Type s) {
...
+            }
+
+            public NullabilityComparator clearWarner() {
+                this.warner = null;
+                return this;
+            }
+
+            @Override
+            public Boolean visitType(Type t, Type s) {
+                if (differentNullability.apply(t, s)) {
```
mcimadamore, line 1302: Is this truly a recursive visitor? I'm wondering, because if you don't need the visitor capabilities, then you could just drop the use of lambdas which will make the code a bit cleaner. E.g. instead of:
```
nullabilityComparator.reset((t1, t2) -> !t1.sameNullabilityAs(t2))
                            .visit(s, t);
```
do:
```
if (!s.sameNullabilityAs(t)) {
     nullabilityComparator.warn(...);
}
```
In which case, maybe NullabilityComparator is just NullabilityWarner?

vicente-romero-oracle: yes the visitor is needed, I will be adding another iteration with more test cases, this was an unfinished iteration

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Type.java
```diff
@@ -777,17 +784,34 @@ public boolean isNullUnspecified() {
     }
 
     public boolean sameNullabilityAs(Type t) {
```
mcimadamore, line 786: Should these methods be here or in Types, since they are general relation on types?

mcimadamore: Also, looking at all these tests I wonder it we shouldn't maybe expose the NullMarker enum directly with a single accessor, and then use that (e.g. in a switch) to do the various tests. But this is just minor code style comments, no need to block this PR (mostly thinking aloud).

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
```diff
@@ -1291,7 +1272,67 @@ public Boolean visitUndetVar(UndetVar t, Type s) {
...
+        }
+
+        public NullabilityComparator nullabilityComparator = new NullabilityComparator();
+        public class NullabilityComparator extends TypeRelation {
+            BiFunction<Type, Type, Boolean> differentNullability;
+            Warner warner;
+
+            NullabilityComparator reset(BiFunction<Type, Type, Boolean> differentNullability) {
+                this.differentNullability = differentNullability;
+                if (this.warner == null || this.warner == noWarnings) {
```
mcimadamore, line 1284: I'm not sure I follow what's happening here. We have a warner stack in Types already. But this class seems to somehow override whatever warner is in the stack (if a client called setWarner on the nullability comparator). Shouldn't we try to reuse the warnStack as a way to set warners, and then tweak nullability comparator to use whatever warner is on the stack? E.g. I'm suspicious about the set/clearWarner methods, as they seem to duplicate functionality we already have in other forms.

test/langtools/tools/javac/bang/BangTypesCompilationTests.java
```diff
@@ -131,15 +131,41 @@ class Foo {
...
-                                "compiler.err.non.nullable.cannot.be.assigned.null")
+                                "compiler.err.prob.found.req"),
+                        new DiagAndCode(
+                                """
+                                import java.util.function.*;
+                                class Test<T> {
+                                    void m() {
+                                        Supplier<? extends T> factory = nullFactory();
+                                    }
+                                    Supplier<? extends T!> nullFactory() { return () -> null; }
```
mcimadamore, line 150: Nice!

test/langtools/tools/javac/bang/BangTypesCompilationTests.java
```diff
@@ -278,6 +304,127 @@ void test(List<? extends Object!> ls1, List<? extends String> ls3) {
...
+                                class Test {
+                                    void plot(Function<String!, String> f) {}
+                                    void m(Function<String, String> gradient) {
+                                        plot(gradient);
+                                    }
+                                }
+                                """,
+                                Result.Warning,
+                                "compiler.warn.unchecked.nullness.conversion",
+                                2),  // this needs to be reviewed, the warning is printed twice, should be once only
```
mcimadamore, line 359: Does this have to do with the fact that the subtyping is executed twice (e.g. once in overload, once in the "check" phase) ?

test/langtools/tools/javac/bang/BangTypesCompilationTests.java
```diff
@@ -318,7 +465,114 @@ void test(List<? super String!> ls2) {
...
+                                """,
+                                Result.Clean,
+                                ""),
+                        new DiagAndCode(
+                                """
+                                interface Shape {}
+                                value class Point implements Shape {}
+                                class Box<T> {}
+                                class Test {
+                                    void m(Box<Shape!> lp) {
```
mcimadamore, line 489: This test reminds me of something: capture conversion should probably be tweaked to take nullability into account. Consider this:

```
class Foo<X extends Bar!> { 
    X x() { ... }
}

Foo<? extends Object> foo = ...
Object! restricted = foo.x(); // this should be ok  (e.g. no warnings)
```

I'm basing this on the fact that capture conversion already tries to establish some common facts between declaration-site and use-site. This test is similar to when use-site says `<? extends Object>` but the decl site says `<X extends List>`, in which case `List` is used as the bound for the captured variable. In this case, it seems like the null restriction that is present at the declaration site should be carried over to the captured variable as well.  Of course this is all outside the context of this review, but I thought I'd leave a comment to note the issue.

mcimadamore (commented):
Looks good - I left some suggestions for improvements and consolidation.

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
vicente-romero-oracle: I did this because Check::checkExtends invokes Types::isSubtype, in a previous iteration I think you suggested not to add a Warner argument to isSubtypes, so this is an effort to circumvent that. I will give it another try

test/langtools/tools/javac/bang/BangTypesCompilationTests.java
vicente-romero-oracle: yep I think so

test/langtools/tools/javac/bang/BangTypesCompilationTests.java
vicente-romero-oracle: yes, interesting

test/langtools/tools/javac/bang/BangTypesCompilationTests.java
vicente-romero-oracle: for some cases like:
```
import java.util.function.*;
class Test {
    void plot(Function<String!, String> f) {}
    void m(Function<String, String> gradient) {
        plot(gradient);
    }
}
```
it was warning twice, there is a double visit by NullabilityComparator to the type. In method IsSubtype::visitClassType, there was a warning being issued when the arguments were compared and then at the end of the method this invocation:

```
    if (result) {
        nullabilityComparator.reset((t1, t2) -> hasNarrowerNullability(t1, t2)).visit(s, t);
    }
```
was provoking the same warning being issued in the same position. We need this invocation to compare the types but it also compares the arguments thus the double warning. I'm using now a field to mark if we have already warned or not.

vicente-romero-oracle:
thanks for the review comments, I have uploaded another iteration

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
```diff
@@ -1277,28 +1277,15 @@ public Boolean visitErrorType(ErrorType t, Type s) {
...
-            }
-
-            public NullabilityComparator clearWarner() {
-                this.warner = null;
                 return this;
             }
 
             @Override
             public Boolean visitType(Type t, Type s) {
+                Warner warner = !warnStack.isEmpty() ? warnStack.head : noWarnings;
```
mcimadamore, line 1288: Good cleanup. I assume that now nullability comparator really has no state, so you could in principle create two instances (with the two comparator functions), and just use those instances, right?

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Check.java
```diff
@@ -4491,6 +4487,7 @@ public void warn(LintCategory lint) {
             switch (lint) {
                 case UNCHECKED:
                     Check.this.warnUnchecked(pos(), Warnings.ProbFoundReq(diags.fragment(uncheckedKey), found, expected));
+                    this.warned = true;
```
mcimadamore, line 4490: Odd - this field is clearly meant to be set in this method, but it is not in mainline. Maybe better to set the field only once, after the switch?

vicente-romero-oracle:
offline comments by Dan:
```
A constructor call, and all literals, can be considered to have a null-restricted type. We shouldn't report warnings about things like Foo! f = new Foo(...); or String! s = "abc";
```

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
vicente-romero-oracle: not really, in one case we are using this comparator: `(t1, t2) -> !hasSameNullability(t1, t2)` so we would need three different instances

vicente-romero-oracle:
I have pushed two commits, one addressing your last comments and another that addresses Dan's offline comments above, the one related to constructor invocations and literals. If preferred I can push this commit as part of another PR

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
mcimadamore: I see - but still - there is no state in the comparator - right? E.g. in principle clients can allocate a new one and take it from there each time. So perhaps that would be more honest, rather than having a shared instance but then having to call a `reset` method?

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Type.java
```diff
@@ -396,7 +396,7 @@ public <M extends TypeMetadata> M getMetadata(Class<M> metadataClass) {
     @SuppressWarnings("unchecked")
     public <M extends TypeMetadata, Z> Z getMetadata(Class<M> metadataClass, Function<M, Z> metadataFunc, Z defaultValue) {
         for (TypeMetadata m : metadata) {
-            if (m.getClass() == metadataClass) {
+            if (m.getClass().isAssignableFrom(metadataClass)) {
```
mcimadamore, line 399: isn't TypeMetadata `sealed` ? I suppose the code you wrote is correct (and better than `==`), but I don't think we should have inherited metadata? E.g. an `equals` should be good?

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
```diff
@@ -5403,20 +5403,16 @@ public Type constantType(LoadableConstant c) {
      * Do t and s have the same nullability?
      */
     public boolean hasSameNullability(Type t, Type s) {
+        // special case for literals, a literal is always != null
+        boolean isLiteral = s != null && s.getMetadata(TypeMetadata.ConstantValue.class) != null;
```
mcimadamore, line 5407: I believe this should be addressed at the level of the `isNullUnspecified` predicates. E.g. if something is a constant type, then you know it's null-restricted (e.g. because we infer it so). Once that's dealt with, I believe these higher-level methods here can stay unchanged?

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java
```diff
@@ -2895,6 +2895,9 @@ else if (!skipNonDiamondPath) {
 
             if (tree.constructor != null && tree.constructor.kind == MTH)
                 owntype = clazztype;
+                if (owntype.getMetadata(TypeMetadata.NullMarker.class) == null) {
+                    owntype = owntype.addMetadata(new TypeMetadata.NullMarker(NullMarker.NOT_NULL)); // constructor invocations are always null restricted
```
mcimadamore, line 2899: I note an asymmetry here: for constructor calls, we annotate the type to say it's non-null. But for constant types we don't, and we instead look as to whether there's constant metadata. Perhaps an alternative to the suggestion I gave above is to simply set the null-restricted metadata in the visitLiteral method, so that everything is explicit.

src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/Gen.java
```diff
@@ -912,7 +912,7 @@ public boolean isConstantDynamic(Symbol sym) {
     public void genArgs(List<JCExpression> trees, List<Type> pts) {
         for (List<JCExpression> l = trees; l.nonEmpty(); l = l.tail) {
             genExpr(l.head, pts.head).load();
-            if (pts.head.isNonNullable()) {
+            if (pts.head.isNonNullable() && l.head.type.constValue() == null) {
```
mcimadamore, line 915: These changes should become unnecessary if `isNonNullable` does the right thing.

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
vicente-romero-oracle: yes I did that experiment first but then I got to some unexpected issues. For example: `1L` now would be consider to have type `long!` which doesn't make too much sense, we could exclude primitives from the predicates I guess but there are still some other places where we can have surprises. I found this example:

```
import java.lang.invoke.*;
class Cell {
    final void reset() {
        /* we are testing that the compiler won't infer the arguments of
         * VarHandle::setVolatile as (Cell, String!)
         */
        VALUE.setVolatile(this, "");
    }
    final void reset(String identity) {
        /* if that were the case, see comment above, then this invocation would generate
         * a warning, VarHandle::setVolatile is a polymorphic signature method
         */
        VALUE.setVolatile(this, identity);
    }
                                                                
    private static final VarHandle VALUE;
    static {
        try {
            MethodHandles.Lookup l = MethodHandles.lookup();
            VALUE = l.findVarHandle(Cell.class, "value", long.class);
        } catch (ReflectiveOperationException e) {
            throw new ExceptionInInitializerError(e);
        }
    }
}
```
VarHandle::setVolatile is a polymorphic signature method. So from the first invocation the compiler would infer argument types: `(Cell, String!)` is this what we want? For the second invocation we find that we already have a method that matches but the invocation will generate a warning if compiled with `Xlint:null`, not sure we want this. This is why I preferred to stay with the declared type for literals and tread them as a special case. We can still do what you suggest and do something different when we infer the signature of polymorphic sig methods, like not consider nullability.

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Type.java
vicente-romero-oracle: yes I guess it should be good, I will replace it in the next iteration

vicente-romero-oracle:
I have uploaded another iteration following our discussion.  Thanks!

mcimadamore (approved):
Looks good

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Type.java
```diff
@@ -1123,7 +1115,9 @@ public <R,S> R accept(Type.Visitor<R,S> v, S s) {
         }
 
         public Type constType(Object constValue) {
-            return addMetadata(new ConstantValue(constValue));
+            return isPrimitive() ?
+                    addMetadata(new ConstantValue(constValue)) :
+                    addMetadata(new ConstantValue(constValue)).addMetadata(new TypeMetadata.NullMarker(NullMarker.NOT_NULL));
```
mcimadamore, line 1120: One nitpick: I can see why having the call here is handy. On the other hand, in terms of code maintenance it would probably be better if `constType` added just one kind of metadata (and we added the NOT_NULL explicitly where we need it). That said, I understand that, pragmatically, Attr is probably not the only place where we call this method, and chasing all the usages might be a bit hard. So let's just keep a mental note of this, going forward.

vicente-romero-oracle: I see your point but yes I agree that for now it is more practical to have the call here. Thanks for all the comments!

PR #881, opened 2023-07-03
8311080: [lworld+vector] Fix jdk build failures with different options

XiaohongGong:
There are several issues exposed by building jdk image with following different options:
 - `--disable-precompiled-headers`
 - `--with-jvm-variants=client`
 - `--with-jvm-variants=minimal` or `--disable-cds`

1. With `--disable-precompiled-headers`, using `class MultiFieldInfo` in `oops/fieldInfo.hpp` and `oops/fieldInfo.inline.hpp` builds error with "invalid use of incomplete type".

The reason is that class `MultiFieldInfo` is used in `fieldInfo.hpp`, but the header declaring the class is not included in it. A direct
fixing is including its header file `oops/instanceKlass.hpp` before using it. But it cannot work here since `oops/instanceKlass.hpp` has included `oops/fieldInfo.hpp`.

To resolve the circle, we can move the definition of class `MultiFieldInfo` to `oops/fieldInfo.hpp`, which I think is also reasonable since `multifield` is a kind of `field`.

2. With `--with-jvm-variants=client`, calling `Deoptimization::reassign_fields_by_klass()` in `vectorsupport.cpp` builds error since `reassign_fields_by_klass()` is defined when C2 or JVMCI compiler is enabled (See [1]).

Consider the caller method in `vectorSupport.cpp` is used only for C2 compiler (See [2]), adding the same limitation for definition of all the relative methods in `vectorSupport.cpp` is better and can fix this issue.

3. With `--with-jvm-variants=minimal` or `--disable-cds`, calling several CDS specific methods ([3]) in `oops/inlineKlass.cpp` builds error. Those caller methods defined in `inlineKlass.hpp` are all CDS related as well. Hence, adding the same CDS condition for them sounds reasonable and can fix this issue.

This patch also fixed an return type mismatch issue exposed by building the jdk image on aarch64 windows system. The relative method is:

```
jint SharedRuntime::skip_value_scalarization(InlineKlass *)
```
The declaration type is `int` (see [4]).

[1] https://github.com/openjdk/valhalla/blob/lworld%2Bvector/src/hotspot/share/runtime/deoptimization.hpp#L195
[2] https://github.com/openjdk/valhalla/blob/lworld%2Bvector/src/hotspot/share/runtime/deoptimization.cpp#L1278
[3] https://github.com/openjdk/valhalla/blob/lworld%2Bvector/src/hotspot/share/oops/instanceKlass.hpp#L1236
[4] https://github.com/openjdk/valhalla/blob/lworld%2Bvector/src/hotspot/share/runtime/sharedRuntime.hpp#L186

---------
### Progress

### Issue
 * [JDK-8311080](https://bugs.openjdk.org/browse/JDK-8311080): [lworld+vector] Fix jdk build failures with different options (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/881/head:pull/881` \
`$ git checkout pull/881`

`$ git checkout pull/881` \
`$ git pull https://git.openjdk.org/valhalla.git pull/881/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 881`

`$ git pr show -t 881`

</details>

<a href="https://git.openjdk.org/valhalla/pull/881.diff">https://git.openjdk.org/valhalla/pull/881.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/881#issuecomment-1617659992)

### DIFF
--- a/src/hotspot/share/runtime/sharedRuntime.cpp
+++ b/src/hotspot/share/runtime/sharedRuntime.cpp
@@ -551,8 +551,8 @@ JRT_LEAF(address, SharedRuntime::exception_handler_for_return_address(JavaThread
   return raw_exception_handler_for_return_address(current, return_address);
 JRT_END
 
-JRT_LEAF(jint, SharedRuntime::skip_value_scalarization(InlineKlass* klass))
-  return (jint)VectorSupport::skip_value_scalarization(klass);
+JRT_LEAF(int, SharedRuntime::skip_value_scalarization(InlineKlass* klass))
+  return (int) VectorSupport::skip_value_scalarization(klass);
 JRT_END
 
 address SharedRuntime::get_poll_stub(address pc) {
--- a/src/hotspot/share/oops/fieldInfo.hpp
+++ b/src/hotspot/share/oops/fieldInfo.hpp
@@ -31,6 +31,7 @@
 #include "utilities/vmEnums.hpp"
 
 class MultiFieldInfo;
+
 static constexpr u4 flag_mask(int pos) {
   return (u4)1 << pos;
 }
@@ -231,6 +232,21 @@ class FieldInfo {
   void static print_from_growable_array(outputStream* os, GrowableArray<FieldInfo>* array, ConstantPool* cp);
 };
 
+class MultiFieldInfo : public MetaspaceObj {
+ private:
+  Symbol* _name;
+  u2 _base_index;
+  jbyte _multifield_index;
+ public:
+  MultiFieldInfo() : _name(nullptr), _base_index(0), _multifield_index(-1) {}
+  MultiFieldInfo(Symbol* name, u2 base, jbyte index) : _name(name), _base_index(base), _multifield_index(index) {}
+  Symbol* name() const { return _name; }
+  u2 base_index() const { return _base_index; }
+  jbyte multifield_index() const { return _multifield_index; }
+  FieldInfo base_field_info(InstanceKlass* ik);
+  void metaspace_pointers_do(MetaspaceClosure* it);
+  MetaspaceObj::Type type() const { return MultiFieldInfoType; }
+};
 
 class FieldInfoStream;
 
--- a/src/hotspot/share/oops/inlineKlass.cpp
+++ b/src/hotspot/share/oops/inlineKlass.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017, 2020, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2017, 2023, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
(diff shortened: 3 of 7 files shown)


XiaohongGong:
Hi @jatin-bhateja , could you please help to take a look at this PR? Thanks a lot!

src/hotspot/share/runtime/sharedRuntime.cpp
```diff
@@ -551,8 +551,8 @@ JRT_LEAF(address, SharedRuntime::exception_handler_for_return_address(JavaThread
   return raw_exception_handler_for_return_address(current, return_address);
 JRT_END
 
-JRT_LEAF(jint, SharedRuntime::skip_value_scalarization(InlineKlass* klass))
-  return (jint)VectorSupport::skip_value_scalarization(klass);
+JRT_LEAF(int, SharedRuntime::skip_value_scalarization(InlineKlass* klass))
+  return (int) VectorSupport::skip_value_scalarization(klass);
```
jatin-bhateja, line 555: Thanks for addressing these, it will also fix the build issues with default options on Windows,  I was planning to address these with JDK-8311383

jatin-bhateja:
I still see some issues with minimal build with following configuration.

sudo ./configure --with-conf-name=linux--minimal --enable-debug --disable-precompiled-headers --with-jvm-variants=minimal --with-boot-jdk=/home/jatinbha/softwares/jdk-20  --with-default-make-target="hotspot" --with-zlib=system

XiaohongGong:
> I still see some issues with minimal and 32 bit build with following configuration on x86. sudo ./configure --with-conf-name=linux--minimal --enable-debug --disable-precompiled-headers --with-jvm-variants=minimal --with-boot-jdk=/home/jatinbha/softwares/jdk-20 --with-default-make-target="hotspot" --with-zlib=system
> 
> We can either address it with this patch else I can handle them in a follow up RFE

Thanks for the review! I'm not familiar with x86 systems, so a separate patch to fix the x86 failure is better to me. Besides, I also met another issue with `--with-jvm-variants=client` on x86 systems, which also exists on `lworld` branch. You may take a look as well if possible.

jatin-bhateja:
> > I still see some issues with minimal and 32 bit build with following configuration on x86. sudo ./configure --with-conf-name=linux--minimal --enable-debug --disable-precompiled-headers --with-jvm-variants=minimal --with-boot-jdk=/home/jatinbha/softwares/jdk-20 --with-default-make-target="hotspot" --with-zlib=system
> > We can either address it with this patch else I can handle them in a follow up RFE
> 
> Thanks for the review! I'm not familiar with x86 systems, so a separate patch to fix the x86 failure is better to me. Besides, I also met another issue with `--with-jvm-variants=client` on x86 systems, which also exists on `lworld` branch. You may take a look as well if possible.

Ok, I will address them with JDK-8311383

jatin-bhateja (approved):
Thanks for addressing these.

Best Regards,
Jatin

XiaohongGong:
> Thanks for addressing these.
> 
> Best Regards, Jatin

Thanks for the fixing the review!

PR #883, opened 2023-07-07
8309357: [lworld] CDS is broken with InlineKlasses after the merge

fparain:
Fix the restoration of the internal pointer to the inlineklass_fixed_block when an InlineKlass is loaded from a CDS archive.
Fix the setting of the new calling convention flags when the constMethod object is shared through CDS.

---------
### Progress

### Issue
 * [JDK-8309357](https://bugs.openjdk.org/browse/JDK-8309357): [lworld] CDS is broken with InlineKlasses after the merge (**Bug** - P2)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/883/head:pull/883` \
`$ git checkout pull/883`

`$ git checkout pull/883` \
`$ git pull https://git.openjdk.org/valhalla.git pull/883/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 883`

`$ git pr show -t 883`

</details>

<a href="https://git.openjdk.org/valhalla/pull/883.diff">https://git.openjdk.org/valhalla/pull/883.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/883#issuecomment-1625780045)

### DIFF
--- a/src/hotspot/share/runtime/sharedRuntime.cpp
+++ b/src/hotspot/share/runtime/sharedRuntime.cpp
@@ -3173,11 +3173,15 @@ AdapterHandlerEntry* AdapterHandlerLibrary::get_adapter(const methodHandle& meth
   CompiledEntrySignature ces(method());
   ces.compute_calling_conventions();
   if (ces.has_scalarized_args()) {
-    method->set_has_scalarized_args();
+    if (!method->has_scalarized_args()) {
+      assert(!method()->constMethod()->is_shared(), "Cannot update shared const object");
+      method->set_has_scalarized_args();
+    }
     if (ces.c1_needs_stack_repair()) {
       method->set_c1_needs_stack_repair();
     }
-    if (ces.c2_needs_stack_repair()) {
+    if (ces.c2_needs_stack_repair() && !method->c2_needs_stack_repair()) {
+      assert(!method->constMethod()->is_shared(), "Cannot update a shared const object");
       method->set_c2_needs_stack_repair();
     }
   } else if (method->is_abstract()) {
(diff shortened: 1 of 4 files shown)


src/hotspot/share/runtime/sharedRuntime.cpp
```diff
@@ -3173,11 +3173,15 @@ AdapterHandlerEntry* AdapterHandlerLibrary::get_adapter(const methodHandle& meth
   CompiledEntrySignature ces(method());
   ces.compute_calling_conventions();
   if (ces.has_scalarized_args()) {
-    method->set_has_scalarized_args();
+    if (!method->has_scalarized_args()) {
+      assert(!method()->constMethod()->is_shared(), "Cannot update shared const object");
```
jatin-bhateja, line 3177: Is this assertion mandatory ? since shared archives are mapped with MAP_PRIVATE flag [ (on linux)](https://github.com/openjdk/valhalla/blob/lworld/src/hotspot/os/linux/os_linux.cpp#L3418) thus it  follow COW semantics, any update to shared mapped region will clone the physical frame before update.

jatin-bhateja (commented):
Patch looks good to me.

src/hotspot/share/runtime/sharedRuntime.cpp
fparain: Shared archives are split in two regions. The writable part is using the COW semantic, but all the "const*" metadata are in an unique read-only region.

PR #888, opened 2023-07-20
8311219: [lworld] VM option "InlineFieldMaxFlatSize" cannot work well

XiaohongGong:
Currently all the non-static final fields with inline type can be flattened, even if the layout size of the inline type is beyond
the max flat size specified by `InlineFieldMaxFlatSize`. Please refer to the condition check [1] which decides whether a field
can be flattened or not.

Field flattening has two major side effects: atomicity and size. Fields with atomic access limitation or large size that exceeds
the specified threshold value cannot be flattened. And final fields are special that they are immutable after initialized. So the atomic check for them can be ignored. Hence, 1) for the atomicity free type like the primitive class, the final and non-final fields with such type can be flattened. And 2) for the normal value class that has atomic feature, only the final fields with such type can be flattened. And all kinds of the flattened fields should not exceed the specified max flat size. Please see more details from [1] [2].

The original condition [1] matches the atomicity check but not the flat size limitation. Promoting the flat size check before all other checks matches the flattening policy and can make the VM option `InlineFieldMaxFlatSize` work for final fields as well.

This patch also fixed the jtreg crashes involved after the field flattening condition is changed. Those tests fail with setting
`-XX:+InlineFieldMaxFlatSize=0` by default. The main issue is the non-flattened inline type field is not buffered which is expected to be. The root cause is when parsing `withfield`, the compiler checks whether the field is primitive class type while not its flattening status. Changing to check the flattening status instead can fix the crash.

[1] https://github.com/openjdk/valhalla/blob/lworld/src/hotspot/share/classfile/fieldLayoutBuilder.cpp#L759
[2] https://mail.openjdk.org/pipermail/valhalla-dev/2023-June/011262.html
[3] https://mail.openjdk.org/pipermail/valhalla-dev/2023-July/011265.html

---------
### Progress

### Issue
 * [JDK-8311219](https://bugs.openjdk.org/browse/JDK-8311219): [lworld] VM option "InlineFieldMaxFlatSize" cannot work well (**Enhancement** - P4)

### Reviewers
 * [Jatin Bhateja](https://openjdk.org/census#jbhateja) (@jatin-bhateja - Committer) ⚠️ Review applies to [5742e1e8](https://git.openjdk.org/valhalla/pull/888/files/5742e1e842685e5bd917967a13040f3ee3969cae)
 * [Tobias Hartmann](https://openjdk.org/census#thartmann) (@TobiHartmann - Committer) ⚠️ Review applies to [73d47fc1](https://git.openjdk.org/valhalla/pull/888/files/73d47fc10518c4cdd106608796669d543815609b)

### Contributors
 * Tobias Hartmann `<thartmann@openjdk.org>`

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/888/head:pull/888` \
`$ git checkout pull/888`

`$ git checkout pull/888` \
`$ git pull https://git.openjdk.org/valhalla.git pull/888/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 888`

`$ git pr show -t 888`

</details>

<a href="https://git.openjdk.org/valhalla/pull/888.diff">https://git.openjdk.org/valhalla/pull/888.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/888#issuecomment-1643602230)


src/hotspot/share/classfile/fieldLayoutBuilder.cpp
```diff
@@ -657,7 +657,8 @@ void FieldLayoutBuilder::regular_field_sorting() {
           //too_volatile_to_flatten = false; //FIXME
           // volatile fields are currently never inlined, this could change in the future
         }
-        if (!(too_big_to_flatten | too_atomic_to_flatten | too_volatile_to_flatten) || fieldinfo.access_flags().is_final()) {
+        if (!too_big_to_flatten &&
+            (!(too_atomic_to_flatten | too_volatile_to_flatten) || fieldinfo.access_flags().is_final())) {
```
jatin-bhateja, line 661: Hi @XiaohongGong , specifications (https://openjdk.org/projects/valhalla/design-notes/state-of-valhalla/03-vm-model#flattenable-contexts) also talks about flattening final L descriptor fields of value objects.

XiaohongGong: Thanks for pointing out this! I got through of the specifications just now.
```
Final fields containing value objects, under both L and Q descriptors, can be routinely flattened. Mutable fields containing value objects under Q descriptors can be routinely flattened; mutable fields containing value objects under L descriptors must provide atomic loads and stores, which requires either atomic instructions or an indirection.
```
Per my understanding, this means:
1. The final fields can be flattened for all L/Q descriptors
2. The non-final fields can be flattened for Q descriptors
3. The non-final fields cannot be flattened for L descriptor value objects due to the atomic feature.

```
For value objects with many fields, the JVM may decide to fall back to buffering its value in the heap, which is the traditional indirected representation.
```
And the above means size of the flattened fields should not exceed some value. Right? 

Please correct me if any misunderstanding! Thanks!

jatin-bhateja: Yes, this is correct.

fparain: The state of Valhalla document is not a specification, but a document about the general design of the Valhalla features in the VM. Currently, fields with L-descriptors are never flattened even the final ones. There are some non-trivial problems to be solved regarding their status before initialization and until we have a good solution to those problems, the safest option is to not flattened them.

jatin-bhateja: Thanks @fparain

jatin-bhateja: > The state of Valhalla document is not a specification, but a document about the general design of the Valhalla features in the VM. Currently, fields with L-descriptors are never flattened even the final ones. There are some non-trivial problems to be solved regarding their status before initialization and until we have a good solution to those problems, the safest option is to not flattened them.

null initialized L descriptor final fields may be tricky to handle, being pessimist looks appropriate in this case.

XiaohongGong: Sounds reasonable to me! It also reminds me that the value object JEP [1] said a value object can only be scalarized when it is on stack or passed through a method call in C2 compiler. 

So maybe changing to `!(too_big_to_flatten | too_atomic_to_flatten | too_volatile_to_flatten)` is safer for current status. WDYT @fparain  @jatin-bhateja ?
 
[1] https://openjdk.org/jeps/8277163

jatin-bhateja: Looks fine to me, the case where we are making this check is meant for T_PRIMITIVE_OBJECTS (Q descriptor value object), thus as per above understanding (1 and 2), final constraint can be removed.  Value object anyways carry an L descriptor (T_OBJECT) and hence are currently not being handled.

jatin-bhateja (approved):
LGTM

XiaohongGong:
Hi, could anyone please take a look at this PR? Thanks a lot in advance!

jatin-bhateja:
> Hi, could anyone please take a look at this PR? Thanks a lot in advance!

This should also be looked in light of atomicity constraint during updates which does not apply to Q descriptor value field.  But aggressive flattening can adversely affect caching behavior.  Just going slightly off the topic, in context of Vector APIs (lworld+vector) all the payloads are final primitive class fields and will always be flattened in their respective boxes.

XiaohongGong:
> Just going slightly off the topic, in context of Vector APIs (lworld+vector) all the payloads are final primitive class fields and will always be flattened in their respective boxes.

Sorry for my late reply!  And yes, the payloads in Vector APIs are always be flattened for current vector size. What my concern is the larger vector size like in Arm SVE after we support the SPECIES_MAX vectors, and the expected behavior/difference if people set a smaller max flat size with `InlineFieldMaxFlatSize`.

Thanks,
Xiaohong

jatin-bhateja:
> > Just going slightly off the topic, in context of Vector APIs (lworld+vector) all the payloads are final primitive class fields and will always be flattened in their respective boxes.
> 
> Sorry for my late reply! And yes, the payloads in Vector APIs are always be flattened for current vector size. What my concern is the larger vector size like in Arm SVE after we support the SPECIES_MAX vectors, and the expected behavior/difference if people set a smaller max flat size with `InlineFieldMaxFlatSize`.
> 
> Thanks, Xiaohong

That's correct!, my above comment was in context of current behavior where final fields of [null-restricted value types](https://openjdk.org/jeps/401) are always flattened irrespective of InlineFieldMaxFlatSize.  Are you inclined to re-instantiate the old behavior ?

Best Regards,
Jatin

TobiHartmann:
Thanks for investigating this issue. I've run some quick testing and there are new failures:

`compiler/valhalla/inlinetypes/TestC1.java` with `-XX:TieredStopAtLevel=1`:

```
#
#  Internal Error (/workspace/open/src/hotspot/share/c1/c1_LIRGenerator.cpp:1778), pid=336001, tid=336015
#  assert(field->type()->as_inline_klass()->is_initialized()) failed: Must be
#
# JRE version: Java(TM) SE Runtime Environment (22.0) (fastdebug build 22-lworld4ea-2023-08-16-1408411.tobias.hartmann.valhalla2)
# Java VM: Java HotSpot(TM) 64-Bit Server VM (fastdebug 22-lworld4ea-2023-08-16-1408411.tobias.hartmann.valhalla2, compiled mode, emulated-client, sharing, tiered, compressed oops, compressed class ptrs, g1 gc, linux-amd64)
# Problematic frame:
# V  [libjvm.so+0x7dc168]  LIRGenerator::access_sub_element(LIRItem&, LIRItem&, LIR_Opr&, ciField*, int)+0x808

Current CompileTask:
C1:   7349 3404    b  1       compiler.valhalla.inlinetypes.TestC1::test7_verifier (47 bytes)

Stack: [0x00007f16d9b00000,0x00007f16d9c00000],  sp=0x00007f16d9bfd810,  free space=1014k
V  [libjvm.so+0x7dc168]  LIRGenerator::access_sub_element(LIRItem&, LIRItem&, LIR_Opr&, ciField*, int)+0x808  (c1_LIRGenerator.cpp:1778)
V  [libjvm.so+0x7de2ee]  LIRGenerator::do_LoadIndexed(LoadIndexed*)+0x52e  (c1_LIRGenerator.cpp:2365)
V  [libjvm.so+0x7c2433]  LIRGenerator::do_root(Instruction*)+0x63  (c1_LIRGenerator.cpp:378)
V  [libjvm.so+0x7c799e]  non-virtual thunk to LIRGenerator::block_do(BlockBegin*)+0x5e  (c1_LIRGenerator.cpp:375)
V  [libjvm.so+0x78a891]  BlockList::iterate_forward(BlockClosure*)+0x41  (c1_Instruction.cpp:1026)
V  [libjvm.so+0x74a94e]  Compilation::emit_lir()+0x50e  (c1_Compilation.cpp:258)
V  [libjvm.so+0x74d30f]  Compilation::compile_java_method()+0x32f  (c1_Compilation.cpp:404)
V  [libjvm.so+0x74dc29]  Compilation::compile_method()+0x179  (c1_Compilation.cpp:473)
V  [libjvm.so+0x74e41a]  Compilation::Compilation(AbstractCompiler*, ciEnv*, ciMethod*, int, BufferBlob*, bool, DirectiveSet*)+0x35a  (c1_Compilation.cpp:606)
V  [libjvm.so+0x74ff2e]  Compiler::compile_method(ciEnv*, ciMethod*, int, bool, DirectiveSet*)+0xde  (c1_Compiler.cpp:254)
V  [libjvm.so+0xa37780]  CompileBroker::invoke_compiler_on_method(CompileTask*)+0xa00  (compileBroker.cpp:2265)
V  [libjvm.so+0xa385a8]  CompileBroker::compiler_thread_loop()+0x5b8  (compileBroker.cpp:1944)
V  [libjvm.so+0xef4fec]  JavaThread::thread_main_inner()+0xcc  (javaThread.cpp:720)
V  [libjvm.so+0x17f23ca]  Thread::call_run()+0xba  (thread.cpp:217)
V  [libjvm.so+0x14dbdcc]  thread_native_entry(Thread*)+0x11c  (os_linux.cpp:779)
```

`compiler/valhalla/inlinetypes/TestBasicFunctionality.java` with `-DWarmup=0 -DVerifyIR=false`

```
#
#  Internal Error (/workspace/open/src/hotspot/share/opto/inlinetypenode.cpp:884), pid=3588916, tid=3588932
#  assert(!null_free || vt->is_allocated(&gvn)) failed: inline type should be allocated
#
# JRE version: Java(TM) SE Runtime Environment (22.0) (fastdebug build 22-lworld4ea-2023-08-16-1408411.tobias.hartmann.valhalla2)
# Java VM: Java HotSpot(TM) 64-Bit Server VM (fastdebug 22-lworld4ea-2023-08-16-1408411.tobias.hartmann.valhalla2, mixed mode, tiered, compressed oops, compressed class ptrs, g1 gc, linux-amd64)
# Problematic frame:
# V  [libjvm.so+0xe57aa7]  InlineTypeNode::make_from_oop_impl(GraphKit*, Node*, ciInlineKlass*, bool, GrowableArray<ciType*>&)+0x417

Current CompileTask:
C2:  40242  965    b  4       compiler.valhalla.inlinetypes.TestBasicFunctionality::test37 (2 bytes)

Stack: [0x00007f0d778ff000,0x00007f0d779ff000],  sp=0x00007f0d779fb8e0,  free space=1010k
V  [libjvm.so+0xe57aa7]  InlineTypeNode::make_from_oop_impl(GraphKit*, Node*, ciInlineKlass*, bool, GrowableArray<ciType*>&)+0x417  (inlinetypenode.cpp:884)
V  [libjvm.so+0xe58e54]  InlineTypeNode::initialize_fields(GraphKit*, MultiNode*, unsigned int&, bool, bool, Node*, GrowableArray<ciType*>&)+0xa24  (inlinetypenode.cpp:1146)
V  [libjvm.so+0xe59603]  InlineTypeNode::make_from_multi(GraphKit*, MultiNode*, ciInlineKlass*, unsigned int&, bool, bool)+0x103  (inlinetypenode.cpp:920)
V  [libjvm.so+0x15227ee]  Compile::build_start_state(StartNode*, TypeFunc const*)+0x7ee  (parse1.cpp:878)
V  [libjvm.so+0xa2b956]  Compile::Compile(ciEnv*, ciMethod*, int, Options, DirectiveSet*)+0x1706  (compile.cpp:792)
V  [libjvm.so+0x876b4e]  C2Compiler::compile_method(ciEnv*, ciMethod*, int, bool, DirectiveSet*)+0x10e  (c2compiler.cpp:115)
V  [libjvm.so+0xa37780]  CompileBroker::invoke_compiler_on_method(CompileTask*)+0xa00  (compileBroker.cpp:2265)
V  [libjvm.so+0xa385a8]  CompileBroker::compiler_thread_loop()+0x5b8  (compileBroker.cpp:1944)
V  [libjvm.so+0xef4fec]  JavaThread::thread_main_inner()+0xcc  (javaThread.cpp:720)
V  [libjvm.so+0x17f23ca]  Thread::call_run()+0xba  (thread.cpp:217)
V  [libjvm.so+0x14dbdcc]  thread_native_entry(Thread*)+0x11c  (os_linux.cpp:779)
```

XiaohongGong:
Thanks for looking at this PR @TobiHartmann ! I think there maybe some potential issues exposed after this change, since almost all value objects cannot be flattened anymore with `-XX:InlineFieldMaxFlatSize=0`, which may need the buffer allocated. 

> Thanks for investigating this issue. I've run some quick testing and there are new failures:
> 
> `compiler/valhalla/inlinetypes/TestC1.java` with `-XX:TieredStopAtLevel=1`:
> 
> ```
> # A fatal error has been detected by the Java Runtime Environment:
> #
> #  Internal Error (/workspace/open/src/hotspot/share/c1/c1_LIRGenerator.cpp:1778), pid=336001, tid=336015
> #  assert(field->type()->as_inline_klass()->is_initialized()) failed: Must be
> #
> # JRE version: Java(TM) SE Runtime Environment (22.0) (fastdebug build 22-lworld4ea-2023-08-16-1408411.tobias.hartmann.valhalla2)
> # Java VM: Java HotSpot(TM) 64-Bit Server VM (fastdebug 22-lworld4ea-2023-08-16-1408411.tobias.hartmann.valhalla2, compiled mode, emulated-client, sharing, tiered, compressed oops, compressed class ptrs, g1 gc, linux-amd64)
> # Problematic frame:
> # V  [libjvm.so+0x7dc168]  LIRGenerator::access_sub_element(LIRItem&, LIRItem&, LIR_Opr&, ciField*, int)+0x808
> 
> Current CompileTask:
> C1:   7349 3404    b  1       compiler.valhalla.inlinetypes.TestC1::test7_verifier (47 bytes)
> 
> Stack: [0x00007f16d9b00000,0x00007f16d9c00000],  sp=0x00007f16d9bfd810,  free space=1014k
> Native frames: (J=compiled Java code, j=interpreted, Vv=VM code, C=native code)
> V  [libjvm.so+0x7dc168]  LIRGenerator::access_sub_element(LIRItem&, LIRItem&, LIR_Opr&, ciField*, int)+0x808  (c1_LIRGenerator.cpp:1778)
> V  [libjvm.so+0x7de2ee]  LIRGenerator::do_LoadIndexed(LoadIndexed*)+0x52e  (c1_LIRGenerator.cpp:2365)
> V  [libjvm.so+0x7c2433]  LIRGenerator::do_root(Instruction*)+0x63  (c1_LIRGenerator.cpp:378)
> V  [libjvm.so+0x7c799e]  non-virtual thunk to LIRGenerator::block_do(BlockBegin*)+0x5e  (c1_LIRGenerator.cpp:375)
> V  [libjvm.so+0x78a891]  BlockList::iterate_forward(BlockClosure*)+0x41  (c1_Instruction.cpp:1026)
> V  [libjvm.so+0x74a94e]  Compilation::emit_lir()+0x50e  (c1_Compilation.cpp:258)
> V  [libjvm.so+0x74d30f]  Compilation::compile_java_method()+0x32f  (c1_Compilation.cpp:404)
> V  [libjvm.so+0x74dc29]  Compilation::compile_method()+0x179  (c1_Compilation.cpp:473)
> V  [libjvm.so+0x74e41a]  Compilation::Compilation(AbstractCompiler*, ciEnv*, ciMethod*, int, BufferBlob*, bool, DirectiveSet*)+0x35a  (c1_Compilation.cpp:606)
> V  [libjvm.so+0x74ff2e]  Compiler::compile_method(ciEnv*, ciMethod*, int, bool, DirectiveSet*)+0xde  (c1_Compiler.cpp:254)
> V  [libjvm.so+0xa37780]  CompileBroker::invoke_compiler_on_method(CompileTask*)+0xa00  (compileBroker.cpp:2265)
> V  [libjvm.so+0xa385a8]  CompileBroker::compiler_thread_loop()+0x5b8  (compileBroker.cpp:1944)
> V  [libjvm.so+0xef4fec]  JavaThread::thread_main_inner()+0xcc  (javaThread.cpp:720)
> V  [libjvm.so+0x17f23ca]  Thread::call_run()+0xba  (thread.cpp:217)
> V  [libjvm.so+0x14dbdcc]  thread_native_entry(Thread*)+0x11c  (os_linux.cpp:779)
> ```

Could you please show more env/options information on this issue? I cannot reproduce it even with `-XX:InlineFieldMaxFlatSize=0 -XX:TieredStopAtLevel=1` on our internal Arm NEON machine.  Since this PR will change the flat state of value objects, I'm not sure whether there is some affects on the value class loading/initialization policy?

> `compiler/valhalla/inlinetypes/TestBasicFunctionality.java` with `-DWarmup=0 -DVerifyIR=false`
> 
> ```
> # A fatal error has been detected by the Java Runtime Environment:
> #
> #  Internal Error (/workspace/open/src/hotspot/share/opto/inlinetypenode.cpp:884), pid=3588916, tid=3588932> #  assert(!null_free || vt->is_allocated(&gvn)) failed: inline type should be allocated
> #
> # JRE version: Java(TM) SE Runtime Environment (22.0) (fastdebug build 22-lworld4ea-2023-08-16-1408411.tobias.hartmann.valhalla2)
> # Java VM: Java HotSpot(TM) 64-Bit Server VM (fastdebug 22-lworld4ea-2023-08-16-1408411.tobias.hartmann.valhalla2, mixed mode, tiered, compressed oops, compressed class ptrs, g1 gc, linux-amd64)
> # Problematic frame:
> # V  [libjvm.so+0xe57aa7]  InlineTypeNode::make_from_oop_impl(GraphKit*, Node*, ciInlineKlass*, bool, GrowableArray<ciType*>&)+0x417
> 
> Current CompileTask:
> C2:  40242  965    b  4       compiler.valhalla.inlinetypes.TestBasicFunctionality::test37 (2 bytes)
> 
> Stack: [0x00007f0d778ff000,0x00007f0d779ff000],  sp=0x00007f0d779fb8e0,  free space=1010k
> Native frames: (J=compiled Java code, j=interpreted, Vv=VM code, C=native code)
> V  [libjvm.so+0xe57aa7]  InlineTypeNode::make_from_oop_impl(GraphKit*, Node*, ciInlineKlass*, bool, GrowableArray<ciType*>&)+0x417  (inlinetypenode.cpp:884)
> V  [libjvm.so+0xe58e54]  InlineTypeNode::initialize_fields(GraphKit*, MultiNode*, unsigned int&, bool, bool, Node*, GrowableArray<ciType*>&)+0xa24  (inlinetypenode.cpp:1146)
> V  [libjvm.so+0xe59603]  InlineTypeNode::make_from_multi(GraphKit*, MultiNode*, ciInlineKlass*, unsigned int&, bool, bool)+0x103  (inlinetypenode.cpp:920)
> V  [libjvm.so+0x15227ee]  Compile::build_start_state(StartNode*, TypeFunc const*)+0x7ee  (parse1.cpp:878)
> V  [libjvm.so+0xa2b956]  Compile::Compile(ciEnv*, ciMethod*, int, Options, DirectiveSet*)+0x1706  (compile.cpp:792)
> V  [libjvm.so+0x876b4e]  C2Compiler::compile_method(ciEnv*, ciMethod*, int, bool, DirectiveSet*)+0x10e  (c2compiler.cpp:115)
> V  [libjvm.so+0xa37780]  CompileBroker::invoke_compiler_on_method(CompileTask*)+0xa00  (compileBroker.cpp:2265)
> V  [libjvm.so+0xa385a8]  CompileBroker::compiler_thread_loop()+0x5b8  (compileBroker.cpp:1944)
> V  [libjvm.so+0xef4fec]  JavaThread::thread_main_inner()+0xcc  (javaThread.cpp:720)
> V  [libjvm.so+0x17f23ca]  Thread::call_run()+0xba  (thread.cpp:217)
> V  [libjvm.so+0x14dbdcc]  thread_native_entry(Thread*)+0x11c  (os_linux.cpp:779)
> ```

I also met the same crash on `lworld+vector` branch with this change when I was running Vector API tests. Actually I don't quite understand what this assertion mean? Why the `null_free` value type must be allocated in this routine? I need more investigation on it. And it will be helpful if anyone could give some guide/advice on this part. Thanks in advance!

Best Regards,
Xiaohong

jatin-bhateja:
> > Are you inclined to re-instantiate the old behavior ?
> 
> May I ask what the "old behavior" do you mean? Thanks!

Hi @XiaohongGong ,
Updated my previous comment, I am able reproduce an intermittent crash in another test "compiler/valhalla/inlinetypes/TestGetfieldChains.java" with additional JVM options :  "-Xbatch -XX:TieredStopAtLevel=1 -XX:CompileThresholdScaling=0.1 -XX:InlineFieldMaxFlatSize=0"
[hs_err_pid1570004.txt](https://github.com/openjdk/valhalla/files/12385038/hs_err_pid1570004.txt)
[replay_pid1570004.txt](https://github.com/openjdk/valhalla/files/12385039/replay_pid1570004.txt)
[TestGetfieldChains.txt](https://github.com/openjdk/valhalla/files/12385040/TestGetfieldChains.txt)

Please find attached relevant logs and replay file.

Best Regards,
Jatin

XiaohongGong:
> > > Are you inclined to re-instantiate the old behavior ?
> > 
> > 
> > May I ask what the "old behavior" do you mean? Thanks!
> 
> Hi @XiaohongGong , Updated my previous comment, I am able reproduce an intermittent crash in another test "compiler/valhalla/inlinetypes/TestGetfieldChains.java" with additional JVM options : "-Xbatch -XX:TieredStopAtLevel=1 -XX:CompileThresholdScaling=0.1 -XX:InlineFieldMaxFlatSize=0" [hs_err_pid1570004.txt](https://github.com/openjdk/valhalla/files/12385038/hs_err_pid1570004.txt) [replay_pid1570004.txt](https://github.com/openjdk/valhalla/files/12385039/replay_pid1570004.txt) [TestGetfieldChains.txt](https://github.com/openjdk/valhalla/files/12385040/TestGetfieldChains.txt)
> 
> Please find attached relevant logs and replay file.
> 
> Best Regards, Jatin

Thanks for pointing out this! I can reproduce this issue after I rebase my patch to latest `lworld` branch. I will try my best to look at what is changed. 

Best Regards,
Xiaohong

TobiHartmann:
Hi @XiaohongGong,

> I think there maybe some potential issues exposed after this change, since almost all value objects cannot be flattened anymore with -XX:InlineFieldMaxFlatSize=0, which may need the buffer allocated.

Yes, I think it's likely that this change reveals some existing issues. I can help with debugging but only later next month.

> Could you please show more env/options information on this issue? I cannot reproduce it even with -XX:InlineFieldMaxFlatSize=0 -XX:TieredStopAtLevel=1 on our internal Arm NEON machine.

It fails quite reliable in our testing on x86_64 with one of the following flag combinations:
- `-Xcomp -XX:TieredStopAtLevel=1 -DIgnoreCompilerControls=true`
- `-Xcomp -XX:-TieredCompilation -DIgnoreCompilerControls=true`
- `-DWarmup=0 -DVerifyIR=false`
- `-DWarmup=0 -XX:TieredStopAtLevel=1`

But you can reproduce this now, right?

> Actually I don't quite understand what this assertion mean? Why the null_free value type must be allocated in this routine?

The `is_allocated` assertion checks that the InlineTypeNode should always have a valid oop to a heap buffer now that we just loaded it from an oop value. That's not guaranteed if the field/argument can be NULL though, that's why the assert checks for `!null_free`. Does that make sense?

XiaohongGong:
> Hi @XiaohongGong,
> 
> > I think there maybe some potential issues exposed after this change, since almost all value objects cannot be flattened anymore with -XX:InlineFieldMaxFlatSize=0, which may need the buffer allocated.
> 
> Yes, I think it's likely that this change reveals some existing issues. I can help with debugging but only later next month.

That's great! Thanks so much for your help!

> > Could you please show more env/options information on this issue? I cannot reproduce it even with -XX:InlineFieldMaxFlatSize=0 -XX:TieredStopAtLevel=1 on our internal Arm NEON machine.
> 
> It fails quite reliable in our testing on x86_64 with one of the following flag combinations:
> 
> * `-Xcomp -XX:TieredStopAtLevel=1 -DIgnoreCompilerControls=true`
> * `-Xcomp -XX:-TieredCompilation -DIgnoreCompilerControls=true`
> * `-DWarmup=0 -DVerifyIR=false`
> * `-DWarmup=0 -XX:TieredStopAtLevel=1`
> 
> But you can reproduce this now, right?

Yes, I can reproduce now. The failure is randomly that the test can pass sometimes on my ARM system. I'm currently working on looking at this issue, but not too much progress now. 

> > Actually I don't quite understand what this assertion mean? Why the null_free value type must be allocated in this routine?
> 
> The `is_allocated` assertion checks that the InlineTypeNode should always have a valid oop to a heap buffer now that we just loaded it from an oop value. That's not guaranteed if the field/argument can be NULL though, that's why the assert checks for `!null_free`. Does that make sense?

Yes, make sense to me. Thanks a lot for the explanation!

XiaohongGong:
Hi @TobiHartmann,

```
#
#  Internal Error (/workspace/open/src/hotspot/share/c1/c1_LIRGenerator.cpp:1778), pid=336001, tid=336015
#  assert(field->type()->as_inline_klass()->is_initialized()) failed: Must be
#
# JRE version: Java(TM) SE Runtime Environment (22.0) (fastdebug build 22-lworld4ea-2023-08-16-1408411.tobias.hartmann.valhalla2)
# Java VM: Java HotSpot(TM) 64-Bit Server VM (fastdebug 22-lworld4ea-2023-08-16-1408411.tobias.hartmann.valhalla2, compiled mode, emulated-client, sharing, tiered, compressed oops, compressed class ptrs, g1 gc, linux-amd64)
# Problematic frame:
# V  [libjvm.so+0x7dc168]  LIRGenerator::access_sub_element(LIRItem&, LIRItem&, LIR_Opr&, ciField*, int)+0x808

Current CompileTask:
C1:   7349 3404    b  1       compiler.valhalla.inlinetypes.TestC1::test7_verifier (47 bytes)
```

Recently I spent sometime looking at this crash when the field is not flattened. The basic conclusion is that I didn't find out the field's flatten status can influence the klass's initialization. As a further investigation for this case, I found the code path is different in C1 compiler when the field is flattened or not.  

Here is my understanding to the process:

The relative ops are a flattened array element loading and a followed field loading from the array element. The array element and the field of the element are both primitive class type. To optimize the whole process, C1 compiler merges the two access ops into one with a delay field access. This saves the object re-materialization from the flattened style during array access. 

When the field is not flattened, it goes to method `LIRGenerator::access_sub_element()`. And the assertion is added after the primitive field is loaded. C1 compiler will set the value to the primitive class's default value if the loaded field is null. And the default value requires the corresponding primitive klass is initialized. 

And when the field is flattened, it goes to method `LIRGenerator::access_flattened_array()`. Before it, an oop buffer is allocated, and in this method, it directly fills the fields information to the allocated oop buffer. The whole process doesn't need the klass's fully initialized since the default value is not used.  To prove my assumption that field's flattened status will not influence the klass's initialization, I added the same assertion in this method (after https://github.com/openjdk/valhalla/blob/lworld/src/hotspot/share/c1/c1_LIRGenerator.cpp#L1804). And it can fail sometimes as well.

As a summary, I think we'd better check the klass's initialization state before the whole process in C1 if it is necessary, just like:
```
diff --git a/src/hotspot/share/c1/c1_GraphBuilder.cpp b/src/hotspot/share/c1/c1_GraphBuilder.cpp
index 78cfd1796..35e408f34 100644
--- a/src/hotspot/share/c1/c1_GraphBuilder.cpp
+++ b/src/hotspot/share/c1/c1_GraphBuilder.cpp
@@ -1123,7 +1123,7 @@ void GraphBuilder::load_indexed(BasicType type) {
     if (s.cur_bc() == Bytecodes::_getfield) {
       bool will_link;
       ciField* next_field = s.get_field(will_link);
-      bool next_needs_patching = !next_field->holder()->is_loaded() ||
+      bool next_needs_patching = !next_field->holder()->is_initialized() ||
                                  !next_field->will_link(method(), Bytecodes::_getfield) ||
                                  PatchALot;
       can_delay_access = C1UseDelayedFlattenedFieldReads && !next_needs_patching; 
```
Any better idea to fixing this issue? Please correct me if any misunderstanding! Thanks a lot!

Best Regards,
Xiaohong

TobiHartmann:
Great analysis @XiaohongGong! The fix looks good to me.

TobiHartmann:
I also had a look at the `inline type should be allocated` assert failure. The problem is that the type of a null-free, not-flat field of a primitive class that is passed in scalarized form is not marked as null-free.

Here's the fix:
```
diff --git a/src/hotspot/share/opto/type.cpp b/src/hotspot/share/opto/type.cpp
index 2181a12be42..8f312be063a 100644
--- a/src/hotspot/share/opto/type.cpp
+++ b/src/hotspot/share/opto/type.cpp
@@ -2187,6 +2187,9 @@ static void collect_inline_fields(ciInlineKlass* vk, const Type** field_array, u
     ciField* field = vk->nonstatic_field_at(j);
     BasicType bt = field->type()->basic_type();
     const Type* ft = Type::get_const_type(field->type());
+    if (field->is_null_free()) {
+      ft = ft->join_speculative(TypePtr::NOTNULL);
+    }
     field_array[pos++] = ft;
     if (type2size[bt] == 2) {
       field_array[pos++] = Type::HALF;
```

XiaohongGong:
> Sorry, above fix is not correct. I forgot that it's perfectly fine for non-flat but null-free fields to be initialized with null. We then need to replace null with the default value. Let's adjust the assert instead:
> 
> ```
> -  assert(!null_free || vt->is_allocated(&gvn), "inline type should be allocated");
> +  assert(!null_free || gvn.type(oop)->maybe_null() || vt->is_allocated(&gvn), "inline type should be allocated");
> ```
> 
> I'll revisit these asserts with JDK-8284443.

Thanks for looking at this failure! Do you think it's necessary that I adjust the assertion in this PR and revisit it in future?

TobiHartmann:
Yes, please include it in this PR but on second thought, the proper fix should look like this:

```
-  assert(!null_free || vt->is_allocated(&gvn), "inline type should be allocated");
+  assert(vt->is_allocated(&gvn) || (null_free && !vk->is_initialized()), "inline type should be allocated");
```

All InlineTypeNodes returned by this method should be allocated except for null free with uninitialized value class (because we can't load the default oop for those).

Testing found some additional issues triggered by your change. Please include below fixes as well.

We hit the `"Should have been buffered"` assert because the verification code is too strong. InlineType users don't require buffering because they are removed as well. In addition, the `u->Opcode() != Op_Return || !tf()->returns_inline_type_as_fields()` condition is not required.

```
--- a/src/hotspot/share/opto/compile.cpp
+++ b/src/hotspot/share/opto/compile.cpp
@@ -2045,16 +2045,10 @@ void Compile::process_inline_types(PhaseIterGVN &igvn, bool remove) {
 #ifdef ASSERT
         // Verify that inline type is buffered when replacing by oop
         else if (u->is_InlineType()) {
-          InlineTypeNode* vt2 = u->as_InlineType();
-          for (uint i = 0; i < vt2->field_count(); ++i) {
-            if (vt2->field_value(i) == vt && !vt2->field_is_flattened(i)) {
-              // Use in non-flat field
-              must_be_buffered = true;
-            }
-          }
+          // InlineType uses don't need buffering because they are about to be replaced as well
         } else if (u->is_Phi()) {
           // TODO 8302217 Remove this once InlineTypeNodes are reliably pushed through
-        } else if (u->Opcode() != Op_Return || !tf()->returns_inline_type_as_fields()) {
+        } else {
           must_be_buffered = true;
         }
         if (must_be_buffered && !vt->is_allocated(&igvn)) {
```

When flattening is disabled, we sometimes bail out from compiling `test51` due to `COMPILE SKIPPED: unsupported calling sequence`. That's expected because the call to the `test` method already occupies a lot of stack space. I adjusted the test:

```
diff --git a/test/hotspot/jtreg/compiler/valhalla/inlinetypes/TestLWorld.java b/test/hotspot/jtreg/compiler/valhalla/inlinetypes/TestLWorld.java
index e1ecca66bd0..d34feba1c48 100644
--- a/test/hotspot/jtreg/compiler/valhalla/inlinetypes/TestLWorld.java
+++ b/test/hotspot/jtreg/compiler/valhalla/inlinetypes/TestLWorld.java
@@ -1609,10 +1609,15 @@ public class TestLWorld {
         }
     }
 
+    // Pass arguments via fields to avoid exzessive spilling leading to compilation bailouts
+    static Test51Value test51_arg1;
+    static MyValue1 test51_arg2;
+    static Object test51_arg3;
+
     // Same as test2 but with field holder being an inline type
     @Test
-    public long test51(Test51Value holder, MyValue1 vt1, Object vt2) {
-        return holder.test(holder, vt1, vt2);
+    public long test51() {
+        return test51_arg1.test(test51_arg1, test51_arg2, test51_arg3);
     }
 
     @Run(test = "test51")
@@ -1622,7 +1627,10 @@ public class TestLWorld {
         Test51Value holder = new Test51Value();
         Asserts.assertEQ(testValue1.hash(), vt.hash());
         Asserts.assertEQ(holder.valueField1.hash(), vt.hash());
-        long result = test51(holder, vt, vt);
+        test51_arg1 = holder;
+        test51_arg2 = vt;
+        test51_arg3 = vt;
+        long result = test51();
         Asserts.assertEQ(result, 9*vt.hash() + def.hashPrimitive());
     }

```

XiaohongGong:
> Yes, please include it in this PR but on second thought, the proper fix should look like this:
> 
> ```
> -  assert(!null_free || vt->is_allocated(&gvn), "inline type should be allocated");
> +  assert(vt->is_allocated(&gvn) || (null_free && !vk->is_initialized()), "inline type should be allocated");
> ```
> 
> All InlineTypeNodes returned by this method should be allocated except for null free with uninitialized value class (because we can't load the default oop for those).

New assertion makes sense to me. I will update the change in this PR. Thanks so much for the fixing! 

> Testing found some additional issues triggered by your change. Please include below fixes as well.
> 
> We hit the `"Should have been buffered"` assert because the verification code is too strong. InlineType users don't require buffering because they are removed as well. In addition, the `u->Opcode() != Op_Return || !tf()->returns_inline_type_as_fields()` condition is not required.
> 
> ```
> --- a/src/hotspot/share/opto/compile.cpp
> +++ b/src/hotspot/share/opto/compile.cpp
> @@ -2045,16 +2045,10 @@ void Compile::process_inline_types(PhaseIterGVN &igvn, bool remove) {
>  #ifdef ASSERT
>          // Verify that inline type is buffered when replacing by oop
>          else if (u->is_InlineType()) {
> -          InlineTypeNode* vt2 = u->as_InlineType();
> -          for (uint i = 0; i < vt2->field_count(); ++i) {
> -            if (vt2->field_value(i) == vt && !vt2->field_is_flattened(i)) {
> -              // Use in non-flat field
> -              must_be_buffered = true;
> -            }
> -          }
> +          // InlineType uses don't need buffering because they are about to be replaced as well
>          } else if (u->is_Phi()) {
>            // TODO 8302217 Remove this once InlineTypeNodes are reliably pushed through
> -        } else if (u->Opcode() != Op_Return || !tf()->returns_inline_type_as_fields()) {
> +        } else {
>            must_be_buffered = true;
>          }
>          if (must_be_buffered && !vt->is_allocated(&igvn)) {
> ```

Also thanks for this fixing! Applying such patch can clean new involved tests failures with `-XX:InlineFieldMaxFlatSize=0`. But I don't quite understand it well. Following is my questions:
1. This change seems extended the `buffer check` for users besides `InlineType`, didn't it? So if the current `InlineTypeNode` is used in a method call which expects the `InlineType` is scalarized as arguments/return, is the buffer for the `InlineType` still needed?
2. Conversely, if an `InlineType` is a non-flattened field of another `InlineType`, why is the buffer not needed although the user `InlineType` will be replaced with its oop? Per my understanding, the oop is the class instance of the `InlineType`, which also contains the fields.

With another thinking, I assumed maybe all the `InlineType` have been scalarized to their users if it can be before this final replacement (i.e. replacement with its oop). So for almost all cases, the remaining `InlineTypeNode` must be buffered at this point, right?

src/hotspot/share/opto/inlinetypenode.cpp
```diff
@@ -143,6 +143,9 @@ InlineTypeNode* InlineTypeNode::merge_with(PhaseGVN* gvn, const InlineTypeNode*
     Node* val1 =        field_value(i);
     Node* val2 = other->field_value(i);
     if (val1->is_InlineType()) {
+      if (val2->is_Phi()) {
+        val2 = gvn->transform(val2);
```
jatin-bhateja, line 147: Can you also add an assertion check here to ensure val2 is always an InlineTypeNode which was pushed forward through PhiNode. Please add a test case exercising this control flow, if a test already exist then kindly mention it on this PR.

src/hotspot/share/classfile/fieldLayoutBuilder.cpp
```diff
@@ -658,7 +658,7 @@ void FieldLayoutBuilder::regular_field_sorting() {
             //too_volatile_to_flatten = false; //FIXME
             // volatile fields are currently never flatten, this could change in the future
           }
-          if (!(too_big_to_flatten | too_atomic_to_flatten | too_volatile_to_flatten) || fieldinfo.access_flags().is_final()) {
```
jatin-bhateja, line 661: An IR framework test can be added to check for flattening of Primitive class final/non-final fields.
Given that InlineTypeNode are removed during Optimization check has to done on post Parse IR.

src/hotspot/share/classfile/fieldLayoutBuilder.cpp
```diff
@@ -758,7 +758,7 @@ void FieldLayoutBuilder::inline_class_field_sorting(TRAPS) {
...
+          if (!(too_big_to_flatten | too_atomic_to_flatten | too_volatile_to_flatten)) {
```
jatin-bhateja, line 761: Please also add an IR framework test to check primitive class final/non-final field flattening, given that InlineTypeNode are removed during Optimization hence such a check has to be done on post Parse IR.

src/hotspot/share/opto/inlinetypenode.cpp
XiaohongGong: `val2->as_InlineType()` has already contained the assertion. This original code makes test `compiler.valhalla.inlinetypes.TestNullableArrays` crashes with:
```
#
#
#  Internal Error (/mnt/local/code/valhalla/src/hotspot/share/opto/node.hpp:977), pid=302375, tid=302391
#  assert(is_InlineType()) failed: invalid node class: Phi
#
# JRE version: OpenJDK Runtime Environment (22.0) (fastdebug build 22-internal-git-a477e90b5)
# Java VM: OpenJDK 64-Bit Server VM (fastdebug 22-internal-git-a477e90b5, mixed mode, tiered, compressed oops, compressed class ptrs, g1 gc, linux-aarch64)
# Problematic frame:
# V  [libjvm.so+0xdbeaf8]  Node::as_InlineType() const [clone .part.0]+0x18
#
# Core dump will be written. Default location: Core dumps may be processed with "/usr/share/apport/apport -p%p -s%s -c%c -d%d -P%P -u%u -g%g -- %E" (or dumping to /mnt/local/code/jtreg/jtreg-git/build/images/jtreg/jtwork/hotspot/scratch/8/core.302375)
#
Unsupported internal testing APIs have been used.
```

So I think it can touch this control flow?

TobiHartmann:
> This change seems extended the buffer check for users besides InlineType, didn't it?

My change does two things:
(1) Remove the `u->Opcode() != Op_Return || !tf()->returns_inline_type_as_fields()` condition because if the user would be a ReturnNode and we would return as fields, the InlineTypeNode should be buffered.
(2) For InlineType users, never require buffering because the InlineType user will be replaced by its oop as well (detailed explanation below).

> So if the current InlineTypeNode is used in a method call which expects the InlineType is scalarized as arguments/return, is the buffer for the InlineType still needed?

No, it would not be needed but the InlineTypeNode would not be connected to the call anymore at this point. Scalarization of arguments/returns happens much earlier, we would already have wired the field values directly to the call. But note that the field value itself could be an InlineTypeNode and **that one** would need to be buffered.

> Conversely, if an InlineType is a non-flattened field of another InlineType, why is the buffer not needed although the user InlineType will be replaced with its oop? Per my understanding, the oop is the class instance of the InlineType, which also contains the fields.

It's not needed because we will replace the user separately and then assert that it's buffered (if required). It might well be that the user InlineType does not need to be allocated (it might itself be a flat field value of another InlineType node or something). Asserting that a non-flattened field value is always buffered "itself" is too strong because there can be cases where neither the holder nor the field value need to be buffered.

> With another thinking, I assumed maybe all the InlineType have been scalarized to their users if it can be before this final replacement (i.e. replacement with its oop). So for almost all cases, the remaining InlineTypeNode must be buffered at this point, right?

Ideally yes but there will be chains of InlineTypeNodes and there is no need for the parent nodes to be buffered when their users are going to be replaced anyway. Also, there's an issue with PhiNodes not being properly replaces (see `TODO 8302217`).

I run this through all our testing and the assert never triggered.

XiaohongGong:
> I run this through all our testing and the assert never triggered.

Thanks for all the explanation! Make sense! I didn't find other failures as well and have pushed these changes to this PR.

TobiHartmann (approved):
Thanks for making all these changes. Looks good to me!

src/hotspot/share/opto/inlinetypenode.cpp
jatin-bhateja: Thanks, this should suffice.

src/hotspot/share/classfile/fieldLayoutBuilder.cpp
XiaohongGong: The main problem for me is how to check the flatten status for an `InlineTypeNode`. Maybe we can check the decode operation generated by the field accessing (or loading) to an primitive class field. Any better idea for this?

TobiHartmann: Couldn't you just check that there is no oop load for that particular field, i.e. no indirection to load from a heap buffer?

jatin-bhateja: If a primitive class field is flattened, then compiler will not be creating an InlineTypeNode for field access, sub-field will be part of container, isn't that sufficient to make the check, e.g.

```
primitive class ABC {
   int f1;
   long f2;
   ABC (int f1, long f2) {
      this.f1 = f1;
      this.f2 = f2;
   }
}


public class flatten {
   public final ABC field;

   public flatten(int val) {
     this.field = new ABC(val, (long)val);
   }

   public static long micro(int val) {
      flatten obj =  new flatten(val);
      return obj.field.f1 + obj.field.f2;
   }

      long res = 0;
      for (int i = 0; i < 10000; i++) {
          res += micro(i);
      }
      System.out.println(res);
   }
}

```

We see no InlineTypeNode in post parse IR with following command line:-

-XX:CompileOnly=flatten::micro -XX:CompileCommand=compileonly,flatten::micro -XX:+EnablePrimitiveClasses -cp . flatten

But, field access does generate InlineTypeNode if flatten size explicitly set as 0
-XX:InlineFieldMaxFlatSize=0 -XX:+PauseAtStartup -XX:CompileOnly=flatten::micro -XX:CompileCommand=compileonly,flatten::micro -XX:+EnablePrimitiveClasses -cp . flatten

EA will anyways remove non-escapable allocation, but post parse IR verification should be good enough to test your changes ?

XiaohongGong: Hi @jatin-bhateja @TobiHartmann , I'v added the IR tests which check the oop load number for non-flattened inline type fields. Please help to take a look whether it's fine. Thanks in advance!

jatin-bhateja: Thanks @XiaohongGong, LGTM!

XiaohongGong: I also tried to check `InlineTypeNode`, but unfortunately `InlineTypeNode` is also optimized out after parsing due to no usage in this case, which is the same with the field flattened.

XiaohongGong:
Thanks for the review @jatin-bhateja  @TobiHartmann!

PR #907, opened 2023-08-15
8314190: [lworld] Missing InlineTypeNode re-materialization during type sharpening.

jatin-bhateja:
-  Currently we try to sharpen the object type to a narrower type to optimize object type comparison against a class constant.
 - This is achieved by inserting a CheckCastPP node which casts the object type to a higher speculative type, if cast type is an inlinetype we need to rematerialize InlineTypeNode from newly casted object which was missing and is fixed by the patch.

Kindly review.

Tier1 regressions are clean.

Best Regards,
Jatin Bhateja

### DIFF
--- a/src/hotspot/share/opto/parse2.cpp
+++ b/src/hotspot/share/opto/parse2.cpp
@@ -2437,6 +2437,10 @@ void Parse::sharpen_type_after_if(BoolTest::mask btest,
             // at the control merge.
             _gvn.set_type_bottom(ccast);
             record_for_igvn(ccast);
+            if (tboth->is_inlinetypeptr()) {
+              assert(tboth->exact_klass(true)->is_inlinetype(), "");
+              ccast = InlineTypeNode::make_from_oop(this, ccast, tboth->exact_klass(true)->as_inline_klass());
+            }
             // Here's the payoff.
             replace_in_map(obj, ccast);
           }


src/hotspot/share/opto/parse2.cpp
```diff
@@ -2437,6 +2437,10 @@ void Parse::sharpen_type_after_if(BoolTest::mask btest,
             // at the control merge.
             _gvn.set_type_bottom(ccast);
             record_for_igvn(ccast);
+            if (tboth->is_inlinetypeptr()) {
+              assert(tboth->exact_klass(true)->is_inlinetype(), "");
```
TobiHartmann, line 2441: I don't think that assert is needed since `as_inline_klass` would already assert. And couldn't you use `tboth->inline_klass()`?

jatin-bhateja:
Hi @TobiHartmann , comments addressed.

test/hotspot/jtreg/compiler/valhalla/inlinetypes/TestValueRematDuringTypeSharpening.java
```diff
@@ -0,0 +1,89 @@
...
+       }
+       return getUnsafeFieldValue(obj, val);
+   }
+
+   @Run(test = {"test"}, mode = RunMode.STANDALONE)
+   public static void kernel() {
+       int res = 0;
+       for (int i = 0; i < 10000; i++) {
+           res += test((i % 1000) == 1 ? new dummyValue1(i) : new dummyValue2(i));
+       }
```
TobiHartmann, line 81: I think you should leave warmup to the IR framework. If more iterations are needed, you can control that via the `@Warmup` annotation.

jatin-bhateja: Thanks @TobiHartmann , standalone run mode has no warmups hence there was an explicit warmup loop, changed as per your suggestion.

PR #909, opened 2023-08-21
8314628: [lworld+vector] validation regression fixes and cleanups.

jatin-bhateja:
Patch fixes following problem seen during validation:-

**CDS:-**
- Problem while restoring value object mirrors. These are needed to perform class level comparisons.

**VM:-**
- Remove payload offset computation logic during VM initialization. It was earlier used during object re-materialization.

**ci Model:-**
- Add a generic routine to handle synthetic ciField population for both static and non-static multifield. This enables returning fully populated ciMultiField structure to compilers during parsing.

**C1 compiler:-**
- Incorrect address displacement computation in HIR while updating multifield bundle using "withfield" bytecode in VectorPayalod constructor. This was causing  failures in Double128Mask tests.

**C2 compiler:-**
1.  Allow vector IR creation under strict check for multifield in default_value creation routine. This was incorrectly creating vector IR for non-vector API tests for Double/Long types.
2. Disable inline expansion of Unsafe.finishPrivateBuffer.  This is fixing the incorrectness problems seen with masked operations loop e.g.

>     vt = newPayload();
>     vt = Unsafe.makePrivateBuffer(vt);
>     for (int i = 0; i < Double128.length; i++) {
>          if (mask[i]) {
>               Unsafe.putDouble(vt, apply(src[i]));
>          }
>     }

   Unsafe.putDouble() only modifies the memory state by storing a new value in payload buffer but does not update any local variable in the JVM state corresponding to latch block. Thus no inductive Phi node is inserted in the loop, it also re-materializes InlineTypeNode from modified memory state. This causes incorrectness if second mask value is false as entire payload is overwritten with default value and earlier lane updates are over-written. In order to be safe, disabling intrinsification of finishPrivateBuffer if incoming argument node is of VectorPayload* type, so that it always receive updated buffer argument.

3. With non-incremental in-lining (-XX:-IncrementalInline) flow user defined routines returning a vector object is incorrectly returning uninitialized VBA connected to oop input of InlineType IR. Defer returning VectorBoxAllocation node, they are expanded and initialized during box expansion and replaces all uses of box.

4. Currently we try to sharpen the object type to a narrower type to optimize object type comparison against a class constant. This is achieved by inserting a CheckCastPP node which casts the object type to a higher speculative type, if cast type is an inlinetype we need to re-materialize InlineTypeNode from newly casted object which was missing.

**Validation Status with patch.**
  - All VectorAPI Jtreg test points except following known issues are passing on x86 at various AVX levels and compilation levels.

**Pending/know issues:-**

1. Failures in some of the mask/shuffle related test points with -XX:+DeoptimizeALot, to be resolved after mainline merge to incorporate major shuffle re-structuring changes in in mainline.
2. Test point using Max-Species are failing, to be fixed with max species support in subsequent patch.
3. Assertion failure in cyclic graph check due to unexpected graph shape seen in box expansion.

Best Regards,
Jatin

---------
### Progress

### Issue
 * [JDK-8314628](https://bugs.openjdk.org/browse/JDK-8314628): [lworld+vector] validation regression fixes and cleanups. (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/909/head:pull/909` \
`$ git checkout pull/909`

`$ git checkout pull/909` \
`$ git pull https://git.openjdk.org/valhalla.git pull/909/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 909`

`$ git pr show -t 909`

</details>

<a href="https://git.openjdk.org/valhalla/pull/909.diff">https://git.openjdk.org/valhalla/pull/909.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/909#issuecomment-1685817197)

### DIFF
--- a/src/hotspot/share/c1/c1_GraphBuilder.cpp
+++ b/src/hotspot/share/c1/c1_GraphBuilder.cpp
@@ -1867,7 +1867,7 @@ void GraphBuilder::copy_inline_content(ciInlineKlass* vk, Value src, int src_off
       assert(!inner_field->is_flattened(), "the iteration over nested fields is handled by the loop itself");
       int off = inner_field->offset_in_bytes() + sec_offset - vk->first_field_offset();
       LoadField* load = new LoadField(src, src_off + off, inner_field, false, state_before, false);
-      sec_offset += type2aelembytes(as_BasicType(load->type()));
+      sec_offset += type2aelembytes(inner_field->type()->basic_type());
       Value replacement = append(load);
       StoreField* store = new StoreField(dest, dest_off + off, inner_field, replacement, false, state_before, false);
       store->set_enclosing_field(enclosing_field);
@@ -2197,10 +2197,14 @@ void GraphBuilder::withfield(int field_index) {
             copy_inline_content(vk, obj, offset, new_instance, vk->first_field_offset(), state_before, field);
           }
         } else {
-          LoadField* load = new LoadField(obj, offset, field, false, state_before, false);
-          Value replacement = append(load);
-          StoreField* store = new StoreField(new_instance, offset, field, replacement, false, state_before, false);
-          append(store);
+          for (int i = 0, sec_offset = 0; i < field->secondary_fields_count(); i++) {
+            ciField* temp = i > 0 ? static_cast<ciMultiField*>(field)->secondary_field_at(i-1) : field;
+            LoadField* load = new LoadField(obj, offset + sec_offset, temp, false, state_before, false);
+            sec_offset += type2aelembytes(as_BasicType(load->type()));
+            Value replacement = append(load);
+            StoreField* store = new StoreField(new_instance, offset + sec_offset, temp, replacement, false, state_before, false);
+            append(store);
+          }
         }
       }
     }
@@ -2218,8 +2222,12 @@ void GraphBuilder::withfield(int field_index) {
       copy_inline_content(vk, val, vk->first_field_offset(), new_instance, offset_modify, state_before, field_modify);
     }
   } else {
-    StoreField* store = new StoreField(new_instance, offset_modify, field_modify, val, false, state_before, needs_patching);
-    append(store);
+    int sec_field_size = type2aelembytes(field_type);
+    for (int i = 0; i < field_modify->secondary_fields_count(); i++) {
+      ciField* temp = i > 0 ? static_cast<ciMultiField*>(field_modify)->secondary_field_at(i-1) : field_modify;
+      StoreField* store = new StoreField(new_instance, offset_modify + i * sec_field_size, temp, val, false, state_before, needs_patching);
+      append(store);
+    }
   }
 }
 
--- a/src/hotspot/share/ci/ciInstanceKlass.cpp
+++ b/src/hotspot/share/ci/ciInstanceKlass.cpp
@@ -534,6 +534,43 @@ int ciInstanceKlass::compute_nonstatic_fields() {
   return fields->length();
 }
 
+
+// Generic routine to handle synthetic field population for both static
+// and non-static base multifields.
+ciField* ciInstanceKlass::populate_synthetic_multifields(ciField* field) {
+  ASSERT_IN_VM;
+  if (!field->is_multifield_base()) {
+    return field;
+  }
+  int sec_fields_count = 0;
+  ciMultiField* mfield = static_cast<ciMultiField*>(field);
+  Arena* arena = CURRENT_ENV->arena();
+  InstanceKlass* k = get_instanceKlass();
+
+  for (JavaFieldStream fs(k); !fs.done(); fs.next()) {
+    if (fs.offset() == field->offset_in_bytes()) {
+      assert(mfield == field, "Duplicate multifield for a given offset");
+      fieldDescriptor& fd = fs.field_descriptor();
+      assert(fd.is_multifield_base(), "");
+      mfield = new (arena) ciMultiField(&fd);
+      sec_fields_count = fd.secondary_fields_count(fd.index()) - 1;
+      mfield->add_secondary_fields(new (arena) GrowableArray<ciField*>(arena, sec_fields_count, 0, NULL));
+      if (sec_fields_count == 0) {
+        break;
+      }
+    } else if (sec_fields_count && fs.is_multifield()) {
+      fieldDescriptor& fd = fs.field_descriptor();
+      mfield->secondary_fields()->append(new (arena) ciField(&fd));
+      sec_fields_count--;
+      if (sec_fields_count == 0) {
+        break;
+      }
+    }
+  }
+
+  return mfield;
+}
+
 GrowableArray<ciField*>* ciInstanceKlass::compute_nonstatic_fields_impl(GrowableArray<ciField*>* super_fields, bool flatten) {
   ASSERT_IN_VM;
   Arena* arena = CURRENT_ENV->arena();
@@ -586,7 +623,7 @@ GrowableArray<ciField*>* ciInstanceKlass::compute_nonstatic_fields_impl(Growable
         ciField* field = NULL;
         if (flattened_field->is_multifield_base()) {
           field = new (arena) ciMultiField(flattened_field, this, offset, is_final);
-          static_cast<ciMultiField*>(field)->add_secondary_fields(static_cast<ciMultiField*>(flattened_field)->secondary_fields());
+          static_cast<ciMultiField*>(field)->set_secondary_fields(static_cast<ciMultiField*>(flattened_field)->secondary_fields());
         } else {
           field = new (arena) ciField(flattened_field, this, offset, is_final);
         }
--- a/src/hotspot/share/ci/ciInstanceKlass.cpp
+++ b/src/hotspot/share/ci/ciInstanceKlass.cpp
@@ -595,15 +595,18 @@ GrowableArray<ciField*>* ciInstanceKlass::compute_nonstatic_fields_impl(Growable
     fields->appendAll(super_fields);
   }
 
+  int sec_fields_count = 0;
   for (JavaFieldStream fs(k); !fs.done(); fs.next()) {
     if (fs.access_flags().is_static())  continue;
     if (fs.is_multifield()) {
-      assert(fields->last()->is_multifield_base(), "");
+      assert(sec_fields_count && fields->last()->is_multifield_base(), "");
+      sec_fields_count--;
       ciMultiField* multifield_base = static_cast<ciMultiField*>(fields->last());
       fieldDescriptor& fd = fs.field_descriptor();
       multifield_base->secondary_fields()->append(new (arena) ciField(&fd));
       continue;
     }
+    assert(!sec_fields_count, "");
     fieldDescriptor& fd = fs.field_descriptor();
     if (fd.is_inlined() && flatten) {
       // Inline type fields are embedded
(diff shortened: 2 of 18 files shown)


src/hotspot/share/c1/c1_GraphBuilder.cpp
```diff
@@ -2197,10 +2197,14 @@ void GraphBuilder::withfield(int field_index) {
...
-          StoreField* store = new StoreField(new_instance, offset, field, replacement, false, state_before, false);
-          append(store);
+          for (int i = 0, sec_offset = 0; i < field->secondary_fields_count(); i++) {
+            ciField* temp = i > 0 ? static_cast<ciMultiField*>(field)->secondary_field_at(i-1) : field;
+            LoadField* load = new LoadField(obj, offset + sec_offset, temp, false, state_before, false);
+            sec_offset += type2aelembytes(as_BasicType(load->type()));
+            Value replacement = append(load);
+            StoreField* store = new StoreField(new_instance, offset + sec_offset, temp, replacement, false, state_before, false);
+            append(store);
+          }
```
XiaohongGong, line 2207: Do you have a further plan to refactor the multifields parser in ci stage so that we can remove these similar handling for multifields in many places?

src/hotspot/share/classfile/javaClasses.cpp
```diff
@@ -4963,30 +4968,6 @@ void java_util_concurrent_locks_AbstractOwnableSynchronizer::serialize_offsets(S
...
-#if INCLUDE_CDS
-void vector_VectorPayload::serialize_offsets(SerializeClosure* f) {
-  VECTORPAYLOAD_FIELDS_DO(FIELD_SERIALIZE_OFFSET);
-}
-#endif
-
-void vector_VectorPayload::set_payload(oop o, oop val) {
-  o->obj_field_put(_payload_offset, val);
-}
-
```
XiaohongGong, line 4989: Can we also clean-up the classfile symbols in https://github.com/openjdk/valhalla/blob/lworld%2Bvector/src/hotspot/share/classfile/vmClassMacros.hpp#L197 and https://github.com/openjdk/valhalla/blob/lworld%2Bvector/src/hotspot/share/classfile/vmSymbols.hpp#L97 ?

src/hotspot/share/opto/inlinetypenode.cpp
```diff
@@ -509,11 +509,11 @@ void InlineTypeNode::load(GraphKit* kit, Node* base, Node* ptr, ciInstanceKlass*
       const TypeOopPtr* oop_ptr = kit->gvn().type(base)->isa_oopptr();
       bool is_array = (oop_ptr->isa_aryptr() != NULL);
       bool mismatched = (decorators & C2_MISMATCHED) != 0;
-      if (base->is_Con() && !is_array && !mismatched && ft->bundle_size() == 1) {
+      ciField* field = holder->get_field_by_offset(offset, false);
+      if (base->is_Con() && !is_array && !mismatched && !field->is_multifield_base() && ft->bundle_size() == 1) {
```
XiaohongGong, line 513: Could we remove the last `ft->bundle_size() == 1` check?

src/hotspot/share/opto/inlinetypenode.cpp
```diff
@@ -853,11 +852,13 @@ Node* InlineTypeNode::default_oop(PhaseGVN& gvn, ciInlineKlass* vk) {
...
 }
 
-Node* InlineTypeNode::default_value(PhaseGVN& gvn, ciType* field_type) {
+Node* InlineTypeNode::default_value(PhaseGVN& gvn, ciType* field_type, ciInlineKlass* klass, int index) {
   BasicType bt = field_type->basic_type();
-  Node* value = gvn.zerocon(field_type->basic_type());
   int vec_len = field_type->bundle_size();
-  if (is_java_primitive(bt) &&
+  Node* value = gvn.zerocon(field_type->basic_type());
+  int is_multifield = klass->declared_nonstatic_field_at(index)->is_multifield_base();
```
XiaohongGong, line 859: Rename `is_multifield` to `is_multifield_base` ? 
Or directly change to:
```
if (is_java_primitive(bt) &&
    klass->declared_nonstatic_field_at(index)->is_multifield_base() &&
    Matcher::match_rule_supported_vector(...))
```

src/hotspot/share/opto/library_call.cpp
```diff
@@ -2340,6 +2340,7 @@ bool LibraryCallKit::inline_unsafe_access(bool is_store, const BasicType type, c
   if (base->is_InlineType()) {
     InlineTypeNode* vt = base->as_InlineType();
     if (is_store) {
+      ciInlineKlass* vk = vt->type()->inline_klass();
       if (!vt->is_allocated(&_gvn)) {
```
XiaohongGong, line 2344: `vk` is unused in this block, so it can be removed?

src/hotspot/share/opto/parse1.cpp
```diff
@@ -2393,7 +2396,8 @@ void Parse::return_current(Node* value) {
         inc_sp(1);
         value = value->as_InlineType()->allocate_fields(this);
       }
-    } else if (value->is_InlineType()) {
+    } else if ((is_vector_value && skip_scalarization) ||
+               (value->Opcode() != Op_VectorBox && value->is_InlineType())) {
```
XiaohongGong, line 2400: `is_vector_value` can be removed in the `else-if` condition since `skip_scalarization = is_vector_value && Compile::current()->inlining_incrementally();` , right?

src/hotspot/share/opto/vector.cpp
```diff
@@ -333,7 +333,7 @@ Node* PhaseVector::expand_vbox_alloc_node(VectorBoxAllocateNode* vbox_alloc,
   vector = gvn.transform(vector)->as_InlineType();
 
   Node* klass_node = kit.makecon(TypeKlassPtr::make(vk));
-  Node* alloc_oop  = kit.new_instance(klass_node, NULL, NULL, /* deoptimize_on_exception */ true, vector);
+  Node* alloc_oop  = kit.new_instance(klass_node, NULL, NULL, /* deoptimize_on_exception */ true);
```
XiaohongGong, line 336: Is there any issue if adding `vector` to the alloc node's using edge here?

src/hotspot/share/ci/ciInstanceKlass.cpp
```diff
@@ -534,6 +534,43 @@ int ciInstanceKlass::compute_nonstatic_fields() {
...
+      mfield->secondary_fields()->append(new (arena) ciField(&fd));
+      sec_fields_count--;
+      if (sec_fields_count == 0) {
+        break;
+      }
+    }
+  }
+
+  return mfield;
+}
```
XiaohongGong, line 572: This whole logic is quite similar with the one in `compute_nonstatic_fields_impl` when the field is a multifield. But the code style is not the same. For example, we check the `sec_fields_count` when field is multifield. This is right to me, in case other multifield is appended to a previous multifield_base which is not its expected one, although we do not have such cases now. But this similar check is missing in `compute_nonstatic_fields_impl`, isn't it? 

Another suggestion is: can we directly use `ciMultiField` instead of `ciField` as the method argument type? Besides, may I ask why we have to new another `ciMultiField`, set the secondary_fields to it and return? Can we modify the input field instead and change the method's return type to `void`?

XiaohongGong:
Could you please update the copyright in file `src/hotspot/share/ci/ciStreams.cpp`.

src/hotspot/share/opto/parse1.cpp
XiaohongGong: It seems `value->Opcode() != Op_VectorBox` in the second line is not needed, since the VectorBox's klass is vector which is contained in `skip_scalarization`, right?

src/hotspot/share/opto/library_call.cpp
```diff
@@ -2652,7 +2655,7 @@ bool LibraryCallKit::inline_unsafe_finish_private_buffer() {
     return false;
   }
   InlineTypeNode* vt = buffer->as_InlineType();
-  if (!vt->is_allocated(&_gvn)) {
+  if (!vt->is_allocated(&_gvn) || VectorSupport::is_vector_payload_mf(vt->inline_klass()->get_InlineKlass())) {
```
XiaohongGong, line 2658: This will make the `finishPrivateBuffer` un-inlined in all Vector API cases, won't it? This seems not so efficient to me. But I cannot find a better way to fix the masked cases now.

src/hotspot/share/classfile/javaClasses.cpp
jatin-bhateja: I am planning to use these symbol definitions in Max species support, estimation of multifield bundle size for max species cannot be just done based on MaxVectorSize, e.g., both LongMaxMask and ByteMaxMask will encapsulate VectorPayloadMFMaxZ which will hold @multifield(-1) annotated boolean mfield.

In order to determine the bundle size, we need to resort to VM symbol name comparison during field parsing, payload class names should also be different for Byte and Long max species mask e.g. VectorPayloadMFMaxZL and VectorPayLoadMFMaxZB. Bundle size for Long should be 8 while for Byte should be 64 for 512-bit MaxVectorSize.

src/hotspot/share/opto/vector.cpp
jatin-bhateja: Yes, it causes an [assertion](https://github.com/openjdk/valhalla/blob/lworld%2Bvector/src/hotspot/share/opto/inlinetypenode.cpp#L1312) failure in InlineTypeNode::remove_redundant_allocations pass which expect such an allocation to be remove during InlineTypeNode Idealizations.

src/hotspot/share/opto/library_call.cpp
jatin-bhateja: You are right, but it will not affect the performance given that code is part of fallback implementation.

src/hotspot/share/classfile/javaClasses.cpp
XiaohongGong: I also spent some time looking at this part. It's true that we have to distingunish the max lanecount for different types of VectorMask. As for this part, maybe we could also extend the `MultiField` annotation by adding another variable denotaing the type info like `esize/scale`. And calculate the right lanecount based on the `MaxVectorSize` and `esize/scale`. WDYT?

src/hotspot/share/opto/library_call.cpp
XiaohongGong: OK, it's fine to me now. We can revisit this part and fine tune if possible in future. Thanks!

src/hotspot/share/ci/ciInstanceKlass.cpp
jatin-bhateja: Yes, _compute_nonstatic_fields_impl_ is only called from [ InlineTypeNode constructor](https://github.com/openjdk/valhalla/blob/lworld%2Bvector/src/hotspot/share/opto/inlinetypenode.hpp#L40), but while parsing field access bytecodes "withfield/getfield/putfield" both c1 and c2 compiler uses _ciBytecodeStream::get_field() API_,  since it directly operates over bytecode hence does not take into account the effect of @multifield annotation over it and earlier used to create only one ciField for base multifield instead of creating a hierarchical ciMutifield structure.

In addition, ClassFiledParser creates[ contiguous sequence of FieldInfo ](https://github.com/openjdk/valhalla/blob/lworld%2Bvector/src/hotspot/share/classfile/classFileParser.cpp#L1623) for both base and subsequent synthetic multifield, hense it's not possible to associate synthetic multifields with any other base multifield.

>>> But this similar check is missing in compute_nonstatic_fields_impl, isn't it?

We can add an assertion to check that secondary fields count always match the number of synthetic FieldInfos after base multifield.

jatin-bhateja:
Hi @XiaohongGong , your comments have been addressed.

XiaohongGong (approved):
Generally looks good to me! Thanks!

src/hotspot/share/classfile/javaClasses.cpp
jatin-bhateja: Even if we extend @multifield annotation to pass an explicit lane size, we still need to create separate Java classes for different Max species mask payloads e.g. VectorPayloadMaxLZ, VectorPayloadMaxBZ. In usual scenario field type is known to parser and sufficient to determine its size, only bundle width is passed explicitly. Its only for MaxSpecies where bundle size is dependent on MaxVectorSize we may need to do a symbol comparison, but your suggestion looks good and may save costly symbol lookup.

PR #918, opened 2023-08-25
8314980: [lworld+vector] consider scalarization conditions during ciMultiField creation.

jatin-bhateja:
Patch adds a new API _ciEnv::is_multifield_scalarized_, to scalarize multifield (ciField[s]) in case target vector cannot accommodate multifield bundle size, else it creates a hierarchical structure ciMultiField and expose entire multifield bundle as one field to C2 compiler.

This cleans up special handling done in C2 compiler, ci field query APIs and object reconstruction handling  at SafePoint.

Please review and share your feedback.

Best Regards,
Jatin

---------
### Progress

### Issue
 * [JDK-8314980](https://bugs.openjdk.org/browse/JDK-8314980): [lworld+vector] consider scalarization conditions during ciMultiField creation. (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/918/head:pull/918` \
`$ git checkout pull/918`

`$ git checkout pull/918` \
`$ git pull https://git.openjdk.org/valhalla.git pull/918/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 918`

`$ git pr show -t 918`

</details>

<a href="https://git.openjdk.org/valhalla/pull/918.diff">https://git.openjdk.org/valhalla/pull/918.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/918#issuecomment-1692763185)


src/hotspot/share/ci/ciEnv.cpp
```diff
@@ -1773,3 +1774,11 @@ void ciEnv::dump_inline_data(int compile_id) {
     }
   }
 }
+
+bool ciEnv::is_multifield_scalarized(BasicType bt, int vec_length) {
+#if COMPILER2
+  return InlineTypeNode::is_multifield_scalarized(bt, vec_length);
+#else
+  return false;
```
XiaohongGong, line 1782: 1. I think it's better to implement `is_multifield_scalarized()` here, and reference it in C2 compiler. It's opposite here.
2. By default, it should return `true` instead of `false` if `COMPILER2` is not supported.

src/hotspot/share/ci/ciInstanceKlass.cpp
```diff
@@ -598,9 +587,10 @@ GrowableArray<ciField*>* ciInstanceKlass::compute_nonstatic_fields_impl(Growable
   int sec_fields_count = 0;
   for (JavaFieldStream fs(k); !fs.done(); fs.next()) {
     if (fs.access_flags().is_static())  continue;
-    if (fs.is_multifield()) {
-      assert(sec_fields_count && fields->last()->is_multifield_base(), "");
+    if (fs.is_multifield() && sec_fields_count) {
```
XiaohongGong, line 590: Change to: `(fs.is_mutlfield() && sec_fields_count != 0)` ?

XiaohongGong:
Hi @jatin-bhateja, thanks for this refactoring! Code is much cleaner to me. So does it also need to clean the special handling for multifields in c1/interpreter (e.g. `deoptimize.cpp`) ? Thanks!

src/hotspot/share/ci/ciEnv.cpp
jatin-bhateja: Agree with 2, thanks, will correct it! Idea here is that ci should leverage compiled exposed interfaces to determine scalarizability.

jatin-bhateja:
> Hi @jatin-bhateja, thanks for this refactoring! Code is much cleaner to me. So does it also need to clean the special handling for multifields in c1/interpreter (e.g. `deoptimize.cpp`) ? Thanks!
> 
> Besides, the `secondary_fields_count()` which is broadcasted to the `bundle_size()` of the `ciField` is calculated from `multifield_info` (see: https://github.com/openjdk/valhalla/blob/lworld%2Bvector/src/hotspot/share/runtime/fieldDescriptor.inline.hpp#L81), which I think may not be synced with each other. And it is used widely in C2 and deoptimization. Will it have any issues?
> 
> An alternative is checking the vector size supported when calculating the `secondary_fields_count()` in `fieldDescriptor.inline.hpp`. Return the multifield count if the vector size is supported, and return `1` if not. And then, we can use this information both at the ci stage when calculating the `nonstatic_fields` in `ciInstanceKlass.cpp` and C2 compiler. WDYT?

Thanks for reporting this, I was planning to do in next cleanup patch, but its good to club this with this one. Will update the PR soon.

jatin-bhateja:
> > Hi @jatin-bhateja, thanks for this refactoring! Code is much cleaner to me. So does it also need to clean the special handling for multifields in c1/interpreter (e.g. `deoptimize.cpp`) ? Thanks!
> > Besides, the `secondary_fields_count()` which is broadcasted to the `bundle_size()` of the `ciField` is calculated from `multifield_info` (see: https://github.com/openjdk/valhalla/blob/lworld%2Bvector/src/hotspot/share/runtime/fieldDescriptor.inline.hpp#L81), which I think may not be synced with each other. And it is used widely in C2 and deoptimization. Will it have any issues?


I do not think it should be a problem for de-optimization,  we have separate handling for re-assignment from vector and scalars locations.

> > An alternative is checking the vector size supported when calculating the `secondary_fields_count()` in `fieldDescriptor.inline.hpp`. Return the multifield count if the vector size is supported, and return `1` if not. And then, we can use this information both at the ci stage when calculating the `nonstatic_fields` in `ciInstanceKlass.cpp` and C2 compiler. WDYT?
> 

I am not in favor of making changes in oop population, unless there is a pressing need, all we need is to furnish appropriate data to compilers through ci layer.  

> Hi @XiaohongGong , Thanks for reporting this, I was planning to revisit this in next cleanup patch, but its good to club it with this one. Will update the PR with needed modifications.

I revisited the implementation and captured the flow in following diagram

image

As discussed earlier, ClassFileParser creates separate FieldInfo structures for each synthetic multifield as its needed for field layout computations. Interpreter directly operates over oop model structures,  currently vectors are created by c2 compile only and it accesses oop model through compiler interface (ci).
 
I think to remove any ambiguity, its best to populate valid bundle_size and set is_multifield_base flag only if target vector size is able to accommodate multifield payload.

XiaohongGong:
> I think to remove any ambiguity, its best to populate valid bundle_size and set is_multifield_base flag only if target vector size is able to accommodate multifield payload.

Sounds reasonable to me. Thanks!

src/hotspot/share/ci/ciField.cpp
```diff
@@ -91,6 +91,7 @@ ciField::ciField(ciInstanceKlass* klass, int index, Bytecodes::Code bc) :
   _signature = ciEnv::current(THREAD)->get_symbol(signature);
   _is_multifield = false;
   _is_multifield_base = false;
+  _is_transiant_multifield_base = false;
```
XiaohongGong, line 94: Why do we still need `_is_multifield_base` in `ciField` ?

jatin-bhateja: That is one which is being used by compilers. The transient one is there for ciInstanceKlass::populate_synthetic_multifields. Both c1 and C2 use ciBytecodeStream::get_field API which creates a ciField from field reference in constant pool. It does not take into account annotations over the filed and hence populate synthetic fields operates over originally created field to create a fully populated ciMultifiled. Let me know if have a better way to address this.

XiaohongGong: Yes, I'm wondering whether we can use one like before, by just setting `_is_multifield_base = true` only when the multifields are vectorized.

jatin-bhateja: Fixed the messy nit.

jatin-bhateja:
Hi @XiaohongGong , Let me know if there are any other comments.

src/hotspot/share/ci/ciEnv.cpp
XiaohongGong: What is the expected return value of `is_multifield_scalarized()` for C1 compiler?

XiaohongGong:
> > > Hi @jatin-bhateja, thanks for this refactoring! Code is much cleaner to me. So does it also need to clean the special handling for multifields in c1/interpreter (e.g. `deoptimize.cpp`) ? Thanks!
> > > Besides, the `secondary_fields_count()` which is broadcasted to the `bundle_size()` of the `ciField` is calculated from `multifield_info` (see: https://github.com/openjdk/valhalla/blob/lworld%2Bvector/src/hotspot/share/runtime/fieldDescriptor.inline.hpp#L81), which I think may not be synced with each other. And it is used widely in C2 and deoptimization. Will it have any issues?
> 
> I do not think it should be a problem for de-optimization, we have separate handling for re-assignment from vector and scalars locations.

What I'm worried is whether we can remove the `is_multifield` filter and `secondary_fields_count` check in deoptimization. Since I assumed the `secondary_fields_count` is meaningful only when the multifields are vectorized now, otherwise, the normal multifields have been added into the klass's nonstatic_fields, right? If so, only check whether the field is vector is enough for me.

jatin-bhateja:
> > > > Hi @jatin-bhateja, thanks for this refactoring! Code is much cleaner to me. So does it also need to clean the special handling for multifields in c1/interpreter (e.g. `deoptimize.cpp`) ? Thanks!
> > > > Besides, the `secondary_fields_count()` which is broadcasted to the `bundle_size()` of the `ciField` is calculated from `multifield_info` (see: https://github.com/openjdk/valhalla/blob/lworld%2Bvector/src/hotspot/share/runtime/fieldDescriptor.inline.hpp#L81), which I think may not be synced with each other. And it is used widely in C2 and deoptimization. Will it have any issues?
> > 
> > 
> > I do not think it should be a problem for de-optimization, we have separate handling for re-assignment from vector and scalars locations.
> 
> What I'm worried is whether we can remove the `is_multifield` filter and `secondary_fields_count` check in deoptimization. Since I assumed the `secondary_fields_count` is meaningful only when the multifields are vectorized now, otherwise, the normal multifields have been added into the klass's nonstatic_fields, right? If so, only check whether the field is vector is enough for me.

You assumption still holds good, secondary_fields_count is usable only if multifiled is held in a vector, but we need secondary_fields_count to estimate the vector length unless VectorSupport::klass2length is re-introduced.

src/hotspot/share/ci/ciEnv.cpp
jatin-bhateja: > What is the expected return value of `is_multifield_scalarized()` for C1 compiler?

C2 is the only runtime component which vectorizes multifield bundle and lower tier compiler should return a true value.

XiaohongGong:
> > > > > Hi @jatin-bhateja, thanks for this refactoring! Code is much cleaner to me. So does it also need to clean the special handling for multifields in c1/interpreter (e.g. `deoptimize.cpp`) ? Thanks!
> > > > > Besides, the `secondary_fields_count()` which is broadcasted to the `bundle_size()` of the `ciField` is calculated from `multifield_info` (see: https://github.com/openjdk/valhalla/blob/lworld%2Bvector/src/hotspot/share/runtime/fieldDescriptor.inline.hpp#L81), which I think may not be synced with each other. And it is used widely in C2 and deoptimization. Will it have any issues?
> > > 
> > > 
> > > I do not think it should be a problem for de-optimization, we have separate handling for re-assignment from vector and scalars locations.
> > 
> > 
> > What I'm worried is whether we can remove the `is_multifield` filter and `secondary_fields_count` check in deoptimization. Since I assumed the `secondary_fields_count` is meaningful only when the multifields are vectorized now, otherwise, the normal multifields have been added into the klass's nonstatic_fields, right? If so, only check whether the field is vector is enough for me.
> 
> You assumption still holds good, secondary_fields_count is usable only if multifiled is held in a vector, but we need secondary_fields_count to estimate the vector length unless VectorSupport::klass2length is re-introduced. I agree we can do some more cleanup here.

Yes, I agree it is needed. I mean the filter for multifield (see: https://github.com/openjdk/valhalla/blob/lworld%2Bvector/src/hotspot/share/runtime/deoptimization.cpp#L1567) and the scalar field reassign loop with `secondary_fields_count` as the loop limit. Can these two part be cleaned like before?

src/hotspot/share/ci/ciEnv.cpp
XiaohongGong: Yes, correct. Per my understanding `#if COMPILER2` is a compile flag which is true if C2 is not disabled. Hence this code is just like directly `return InlineTypeNode::is_multifield_scalarized(bt, vec_length);`, right? And then it returns the same value for interpreter/c1/c2.

XiaohongGong: I think what we need is adding an interface here, and implement it in different compilers, or a runtime check like https://github.com/openjdk/valhalla/blob/lworld%2Bvector/src/hotspot/share/compiler/compilerDefinitions.inline.hpp#L49. WDYT?

jatin-bhateja: > Yes, correct. Per my understanding `#if COMPILER2` is a compile flag which is true if C2 is not disabled. Hence this code is just like directly `return InlineTypeNode::is_multifield_scalarized(bt, vec_length);`, right? And then it returns the same value for interpreter/c1/c2.

My bad, I completely overlooked it, you are correct.

XiaohongGong:
> > > > > > > Hi @jatin-bhateja, thanks for this refactoring! Code is much cleaner to me. So does it also need to clean the special handling for multifields in c1/interpreter (e.g. `deoptimize.cpp`) ? Thanks!
> > > > > > > Besides, the `secondary_fields_count()` which is broadcasted to the `bundle_size()` of the `ciField` is calculated from `multifield_info` (see: https://github.com/openjdk/valhalla/blob/lworld%2Bvector/src/hotspot/share/runtime/fieldDescriptor.inline.hpp#L81), which I think may not be synced with each other. And it is used widely in C2 and deoptimization. Will it have any issues?
> > > > > 
> > > > > 
> > > > > I do not think it should be a problem for de-optimization, we have separate handling for re-assignment from vector and scalars locations.
> > > > 
> > > > 
> > > > What I'm worried is whether we can remove the `is_multifield` filter and `secondary_fields_count` check in deoptimization. Since I assumed the `secondary_fields_count` is meaningful only when the multifields are vectorized now, otherwise, the normal multifields have been added into the klass's nonstatic_fields, right? If so, only check whether the field is vector is enough for me.
> > > 
> > > 
> > > You assumption still holds good, secondary_fields_count is usable only if multifiled is held in a vector, but we need secondary_fields_count to estimate the vector length unless VectorSupport::klass2length is re-introduced. I agree we can do some more cleanup here.
> > 
> > 
> > Yes, I agree it is needed. I mean the filter for multifield (see: https://github.com/openjdk/valhalla/blob/lworld%2Bvector/src/hotspot/share/runtime/deoptimization.cpp#L1567) and the scalar field reassign loop with `secondary_fields_count` as the loop limit. Can these two part be cleaned like before?
> 
> Hi @XiaohongGong , I am not in favor of modifying oops structures which are populated during raw bytecode parsing, while we can always prune ci model to present ciField/ciMultifield structures in desired format based on target supported vector size, but a change in oop model will also trigger changes in field layout computations.
> 
> An iteration over raw fields of a klass should see a separate FieldInfo for each synthetic and base multifield. Also Oops model is common across different execution engines (Compilers and Interpreter), while C2 has ability to create vector IR other do not. Hence by construction Oops model should not be influenced by target vector size.
> 
> We can revisit this at a later stage if needed.

OK, make sense to me. I assumed that you'v verified the latest change with vector api jtreg tests, and no new regression is found, right?

XiaohongGong (approved):

jatin-bhateja:
There is one intermittent failure in following test
Double64VectorTests.unsliceBinaryDouble64VectorTestsBinary(double[-i * 5], double[cornerCaseValue(i)]): failure
But, we have also seen similar assertion failures in other shuffle / mask tests. Plan to addressed in subsequent patch.

PR #931, opened 2023-09-28
8311675: [lworld+vector] Max Species support.

jatin-bhateja:
- Patch adds MaxSpecies support for all types of vectors.
- New factory methods and VectorPayload classes for various kinds of vector, shuffle and mask payloads.
- Disable CDS archiving of VectorPayload classes.
- Summary of high level flow :-
  1/ Max species payloads encapsulate @multifield annotated field accepting -1 value as bundle size parameter.
  2/ For Vector payload bundle size is determined using maximum vector size supported by the target. 
  3/ For Shuffles and Masks payloads multifield bundle size is a function of maximum vector size and vector lane size.
  4/ Based on the dynamic bundle size parser creates a separate FieldInfo structure for each base and synthetic multifield and rest of the flow remains the same. 

Best Regards,
Jatin

---------
### Progress

### Issue
 * [JDK-8311675](https://bugs.openjdk.org/browse/JDK-8311675): [lworld+vector] Max Species support. (**Enhancement** - P4)

### Reviewers
 * [Xiaohong Gong](https://openjdk.org/census#xgong) (@XiaohongGong - Committer) ⚠️ Review applies to [945eb0fb](https://git.openjdk.org/valhalla/pull/931/files/945eb0fb9487f1b8b1eb0afb30ff0b209aafd71a)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/931/head:pull/931` \
`$ git checkout pull/931`

`$ git checkout pull/931` \
`$ git pull https://git.openjdk.org/valhalla.git pull/931/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 931`

`$ git pr show -t 931`

</details>

<a href="https://git.openjdk.org/valhalla/pull/931.diff">https://git.openjdk.org/valhalla/pull/931.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/931#issuecomment-1739797968)

### DIFF
--- a/src/hotspot/cpu/aarch64/vm_version_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/vm_version_aarch64.cpp
@@ -659,11 +659,15 @@ void VM_Version::initialize_cpu_information(void) {
 }
 
 int VM_Version::max_vector_size(BasicType bt) {
- // The MaxVectorSize should have been set by detecting SVE max vector register size.
+  // The MaxVectorSize should have been set by detecting SVE max vector register size.
+#ifdef COMPILER2
   int size = MIN2((UseSVE > 0) ? 256 : 16, (int)MaxVectorSize);
   // Minimum 2 values in vector
   if (size < 2*type2aelembytes(bt)) size = 0;
   // But never < 4
   if (size < 4) size = 0;
   return size / type2aelembytes(bt);
+#else
+  return -1;
+#endif
 }
--- a/src/hotspot/cpu/aarch64/vm_version_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/vm_version_aarch64.cpp
@@ -659,5 +659,11 @@ void VM_Version::initialize_cpu_information(void) {
 }
 
 int VM_Version::max_vector_size(BasicType bt) {
-  return MaxVectorSize / type2aelembytes(bt);
+ // The MaxVectorSize should have been set by detecting SVE max vector register size.
+  int size = MIN2((UseSVE > 0) ? 256 : 16, (int)MaxVectorSize);
+  // Minimum 2 values in vector
+  if (size < 2*type2aelembytes(bt)) size = 0;
+  // But never < 4
+  if (size < 4) size = 0;
+  return size / type2aelembytes(bt);
 }
--- a/src/hotspot/cpu/aarch64/vm_version_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/vm_version_aarch64.cpp
@@ -657,3 +657,7 @@ void VM_Version::initialize_cpu_information(void) {
 
   _initialized = true;
 }
+
+int VM_Version::max_vector_size(BasicType bt) {
+  return MaxVectorSize / type2aelembytes(bt);
+}
--- a/src/hotspot/cpu/x86/vm_version_x86.cpp
+++ b/src/hotspot/cpu/x86/vm_version_x86.cpp
@@ -3217,6 +3217,7 @@ intx VM_Version::allocate_prefetch_distance(bool use_watermark_prefetch) {
 
 // Max vector size in bytes. 0 if not supported.
 int VM_Version::max_vector_size(BasicType bt) {
+#ifdef COMPILER2_OR_JVMCI
   assert(is_java_primitive(bt), "only primitive type vectors");
   if (UseSSE < 2) return 0;
   // SSE2 supports 128bit vectors for all types.
@@ -3256,6 +3257,9 @@ int VM_Version::max_vector_size(BasicType bt) {
     assert(false, "Unexpected basic type");
   }
   return size / type2aelembytes(bt);
+#else
+  return -1;
+#endif
 }
 
 
--- a/src/hotspot/cpu/x86/vm_version_x86.cpp
+++ b/src/hotspot/cpu/x86/vm_version_x86.cpp
@@ -3215,6 +3215,50 @@ intx VM_Version::allocate_prefetch_distance(bool use_watermark_prefetch) {
   }
 }
 
+// Max vector size in bytes. 0 if not supported.
+int VM_Version::max_vector_size(BasicType bt) {
+  assert(is_java_primitive(bt), "only primitive type vectors");
+  if (UseSSE < 2) return 0;
+  // SSE2 supports 128bit vectors for all types.
+  // AVX2 supports 256bit vectors for all types.
+  // AVX2/EVEX supports 512bit vectors for all types.
+  int size = (UseAVX > 1) ? (1 << UseAVX) * 8 : 16;
+  // AVX1 supports 256bit vectors only for FLOAT and DOUBLE.
+  if (UseAVX > 0 && (bt == T_FLOAT || bt == T_DOUBLE))
+    size = (UseAVX > 2) ? 64 : 32;
+  if (UseAVX > 2 && (bt == T_BYTE || bt == T_SHORT || bt == T_CHAR))
+    size = (VM_Version::supports_avx512bw()) ? 64 : 32;
+  // Use flag to limit vector size.
+  size = MIN2(size,(int)MaxVectorSize);
+  // Minimum 2 values in vector (or 4 for bytes).
+  switch (bt) {
+  case T_DOUBLE:
+  case T_LONG:
+    if (size < 16) return 0;
+    break;
+  case T_FLOAT:
+  case T_INT:
+    if (size < 8) return 0;
+    break;
+  case T_BOOLEAN:
+    if (size < 4) return 0;
+    break;
+  case T_CHAR:
+    if (size < 4) return 0;
+    break;
+  case T_BYTE:
+    if (size < 4) return 0;
+    break;
+  case T_SHORT:
+    if (size < 4) return 0;
+    break;
+  default:
+    assert(false, "Unexpected basic type");
+  }
+  return size / type2aelembytes(bt);
+}
+
+
 bool VM_Version::is_intrinsic_supported(vmIntrinsicID id) {
   assert(id != vmIntrinsics::_none, "must be a VM intrinsic");
   switch (id) {
--- a/src/hotspot/share/ci/ciEnv.cpp
+++ b/src/hotspot/share/ci/ciEnv.cpp
@@ -1772,7 +1772,7 @@ void ciEnv::dump_inline_data(int compile_id) {
 }
 
 bool ciEnv::is_multifield_scalarized(BasicType bt, int vec_length) {
-#if COMPILER2
+#ifdef COMPILER2
   return InlineTypeNode::is_multifield_scalarized(bt, vec_length);
 #else
   return true;
--- a/src/hotspot/share/ci/ciEnv.cpp
+++ b/src/hotspot/share/ci/ciEnv.cpp
@@ -1773,12 +1773,7 @@ void ciEnv::dump_inline_data(int compile_id) {
 
 bool ciEnv::is_multifield_scalarized(BasicType bt, int vec_length) {
 #if COMPILER2
-  CompilerThread* ct = CompilerThread::current();
-  if (ct && ct->compiler()->is_c2()) {
-    return InlineTypeNode::is_multifield_scalarized(bt, vec_length);
-  } else {
-    return true;
-  }
+  return InlineTypeNode::is_multifield_scalarized(bt, vec_length);
 #else
   return true;
 #endif
--- a/src/hotspot/share/ci/ciObjectFactory.cpp
+++ b/src/hotspot/share/ci/ciObjectFactory.cpp
@@ -52,6 +52,7 @@
 #include "memory/allocation.inline.hpp"
 #include "memory/universe.hpp"
 #include "oops/oop.inline.hpp"
+#include "prims/vectorSupport.hpp"
 #include "runtime/handles.inline.hpp"
 #include "runtime/signature.hpp"
 #include "utilities/macros.hpp"
--- a/src/hotspot/share/classfile/classFileParser.cpp
+++ b/src/hotspot/share/classfile/classFileParser.cpp
@@ -1631,7 +1631,11 @@ void ClassFileParser::parse_fields(const ClassFileStream* const cfs,
     if (parsed_annotations.is_multifield_base() && is_java_primitive(cp->basic_type_for_signature_at(signature_index))) {
       field->field_flags_addr()->update_multifield_base(true);
       char* base_name = cp->symbol_at(name_index)->as_C_string();
-      for (int i = 1; i < parsed_annotations.multifield_arg(); i++) {
+      int bundle_size = parsed_annotations.multifield_arg();
+      if (bundle_size < 0) {
+        bundle_size = VectorSupport::get_max_multifield_count(_class_name);
+      }
+      for (int i = 1; i < bundle_size; i++) {
         field_index++;
         stringStream st;
         st.print("%s", base_name);
--- a/src/hotspot/share/classfile/vmSymbols.cpp
+++ b/src/hotspot/share/classfile/vmSymbols.cpp
@@ -185,6 +185,7 @@ const char* vmSymbols::name_for(vmSymbolID sid) {
 }
 #endif
 
+
 void vmSymbols::symbols_do(SymbolClosure* f) {
   for (auto index : EnumRange<vmSymbolID>{}) {
     f->do_symbol(&Symbol::_vm_symbols[as_int(index)]);
--- a/src/hotspot/share/classfile/vmSymbols.cpp
+++ b/src/hotspot/share/classfile/vmSymbols.cpp
@@ -185,8 +185,6 @@ const char* vmSymbols::name_for(vmSymbolID sid) {
 }
 #endif
 
-
-
 void vmSymbols::symbols_do(SymbolClosure* f) {
   for (auto index : EnumRange<vmSymbolID>{}) {
     f->do_symbol(&Symbol::_vm_symbols[as_int(index)]);
--- a/src/hotspot/share/oops/inlineKlass.cpp
+++ b/src/hotspot/share/oops/inlineKlass.cpp
@@ -153,6 +153,9 @@ bool InlineKlass::flat_array() {
   if (!UseFlatArray) {
     return false;
   }
+  if (VectorSupport::is_vector_payload_mf(this) || VectorSupport::is_vector(this)) {
+    return false;
+  }
   // Too big
   int elem_bytes = get_exact_size_in_bytes();
   if ((FlatArrayElementMaxSize >= 0) && (elem_bytes > FlatArrayElementMaxSize)) {
--- a/src/hotspot/share/prims/vectorSupport.cpp
+++ b/src/hotspot/share/prims/vectorSupport.cpp
@@ -462,7 +462,7 @@ int VectorSupport::vop2ideal(jint id, BasicType bt) {
 int VectorSupport::get_max_multifield_count(const Symbol* payload_name) {
   assert(payload_name, "");
   vmSymbolID sid = vmSymbols::find_sid(payload_name);
-  int size = VM_Version::max_vector_size(T_BYTE);
+  int size = MAX2(VM_Version::max_vector_size(T_BYTE), 8);
   switch(sid) {
     case vmSymbolID::jdk_internal_vm_vector_VectorPayloadMFMaxB_enum:
     case vmSymbolID::jdk_internal_vm_vector_VectorPayloadMFMaxBZ_enum:
--- a/src/hotspot/share/prims/vectorSupport.cpp
+++ b/src/hotspot/share/prims/vectorSupport.cpp
@@ -459,19 +459,50 @@ int VectorSupport::vop2ideal(jint id, BasicType bt) {
 }
 #endif // COMPILER2
 
+int VectorSupport::get_max_multifield_count(const Symbol* payload_name) {
+  assert(payload_name, "");
+  vmSymbolID sid = vmSymbols::find_sid(payload_name);
+  int size = VM_Version::max_vector_size(T_BYTE);
+  switch(sid) {
+    case vmSymbolID::jdk_internal_vm_vector_VectorPayloadMFMaxB_enum:
+    case vmSymbolID::jdk_internal_vm_vector_VectorPayloadMFMaxBZ_enum:
+    case vmSymbolID::jdk_internal_vm_vector_VectorPayloadMFMaxBB_enum:
+       return size;
+    case vmSymbolID::jdk_internal_vm_vector_VectorPayloadMFMaxS_enum:
+    case vmSymbolID::jdk_internal_vm_vector_VectorPayloadMFMaxSZ_enum:
+    case vmSymbolID::jdk_internal_vm_vector_VectorPayloadMFMaxSB_enum:
+       return size / type2aelembytes(T_SHORT);
+    case vmSymbolID::jdk_internal_vm_vector_VectorPayloadMFMaxI_enum:
+    case vmSymbolID::jdk_internal_vm_vector_VectorPayloadMFMaxIZ_enum:
+    case vmSymbolID::jdk_internal_vm_vector_VectorPayloadMFMaxIB_enum:
+       return size / type2aelembytes(T_INT);
+    case vmSymbolID::jdk_internal_vm_vector_VectorPayloadMFMaxL_enum:
+    case vmSymbolID::jdk_internal_vm_vector_VectorPayloadMFMaxLZ_enum:
+    case vmSymbolID::jdk_internal_vm_vector_VectorPayloadMFMaxLB_enum:
+       return size / type2aelembytes(T_LONG);
+    case vmSymbolID::jdk_internal_vm_vector_VectorPayloadMFMaxF_enum:
+       return size / type2aelembytes(T_FLOAT);
+    case vmSymbolID::jdk_internal_vm_vector_VectorPayloadMFMaxD_enum:
+       return size / type2aelembytes(T_DOUBLE);
+    default  : fatal("unknown symbol"); return -1;
+  }
+}
+
+int VectorSupport::max_vector_size(BasicType bt) {
+  if (is_java_primitive(bt)) {
+    return VM_Version::max_vector_size(bt);
+  }
+  return -1;
+}
+
 /**
  * Implementation of the jdk.internal.vm.vector.VectorSupport class
  */
 
 JVM_ENTRY(jint, VectorSupport_GetMaxLaneCount(JNIEnv *env, jclass vsclazz, jobject clazz)) {
-#ifdef COMPILER2
   oop mirror = JNIHandles::resolve_non_null(clazz);
-  if (java_lang_Class::is_primitive(mirror)) {
-    BasicType bt = java_lang_Class::primitive_type(mirror);
-    return Matcher::max_vector_size(bt);
-  }
-#endif // COMPILER2
-  return -1;
+  BasicType bt = java_lang_Class::primitive_type(mirror);
+  return VectorSupport::max_vector_size(bt);
 } JVM_END
 
 // JVM_RegisterVectorSupportMethods
--- a/src/hotspot/share/prims/vectorSupport.hpp
+++ b/src/hotspot/share/prims/vectorSupport.hpp
@@ -140,6 +140,5 @@ class VectorSupport : AllStatic {
   static bool skip_value_scalarization(Klass* klass);
   static int  max_vector_size(BasicType bt);
   static int  get_max_multifield_count(const Symbol* payload_name);
-  static bool is_vector_payload(const Symbol* metadata);
 };
 #endif // SHARE_PRIMS_VECTORSUPPORT_HPP
--- a/src/hotspot/share/prims/vectorSupport.hpp
+++ b/src/hotspot/share/prims/vectorSupport.hpp
@@ -138,5 +138,8 @@ class VectorSupport : AllStatic {
   static bool is_vector(Klass* klass);
   static bool is_vector_payload_mf(Klass* klass);
   static bool skip_value_scalarization(Klass* klass);
+  static int  max_vector_size(BasicType bt);
+  static int  get_max_multifield_count(const Symbol* payload_name);
+  static bool is_vector_payload(const Symbol* metadata);
 };
 #endif // SHARE_PRIMS_VECTORSUPPORT_HPP
--- a/src/java.base/share/classes/jdk/internal/vm/vector/VectorSupport.java
+++ b/src/java.base/share/classes/jdk/internal/vm/vector/VectorSupport.java
@@ -167,7 +167,7 @@ public class VectorSupport {
 
         @ForceInline
         public static VectorPayloadMF newMaskInstanceFactory(Class<?> elemType, int length, boolean max_payload) {
-            if (false == max_payload) {
+            if (!max_payload) {
                 switch(length) {
                     case  1: return new VectorPayloadMF8Z();
                     case  2: return new VectorPayloadMF16Z();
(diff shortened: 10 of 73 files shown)


XiaohongGong:
Great job @jatin-bhateja! Thanks! I will go through the details soon.

src/hotspot/share/prims/vectorSupport.cpp
```diff
@@ -459,19 +459,50 @@ int VectorSupport::vop2ideal(jint id, BasicType bt) {
...
+       return size / type2aelembytes(T_FLOAT);
+    case vmSymbolID::jdk_internal_vm_vector_VectorPayloadMFMaxD_enum:
+       return size / type2aelembytes(T_DOUBLE);
+    default  : fatal("unknown symbol"); return -1;
+  }
+}
+
+int VectorSupport::max_vector_size(BasicType bt) {
+  if (is_java_primitive(bt)) {
+    return VM_Version::max_vector_size(bt);
```
XiaohongGong, line 493: `VM_Version::max_vector_size(bt)` is a platform specific method, and currently only x86 contains it. So this makes the jdk image build failure on other platforms like aarch64. Could we define a public/common method and implement it on each platform like matcher?

src/hotspot/share/classfile/classFileParser.cpp
```diff
@@ -1631,7 +1631,11 @@ void ClassFileParser::parse_fields(const ClassFileStream* const cfs,
     if (parsed_annotations.is_multifield_base() && is_java_primitive(cp->basic_type_for_signature_at(signature_index))) {
       field->field_flags_addr()->update_multifield_base(true);
       char* base_name = cp->symbol_at(name_index)->as_C_string();
-      for (int i = 1; i < parsed_annotations.multifield_arg(); i++) {
+      int bundle_size = parsed_annotations.multifield_arg();
+      if (bundle_size < 0) {
+        bundle_size = VectorSupport::get_max_multifield_count(_class_name);
```
XiaohongGong, line 1636: An alternative is trying to passing the type info to the `MultiField` annotation, so that the `_class_name` is not needed, and so does to the method `VectorSupport::get_max_multifield_count()`. We can use "max_vector_size(bt)" directly.

src/hotspot/share/classfile/vmSymbols.cpp
```diff
@@ -185,8 +185,6 @@ const char* vmSymbols::name_for(vmSymbolID sid) {
 }
 #endif
 
-
-
```
XiaohongGong, line 189: Please revert this change.

src/java.base/share/classes/jdk/internal/vm/vector/VectorSupport.java
```diff
@@ -166,8 +166,8 @@ public abstract static class VectorPayloadMF {
         public abstract long multiFieldOffset();
 
         @ForceInline
-        public static VectorPayloadMF newInstanceFactory(Class<?> elemType, int length) {
-            if (elemType == boolean.class) {
+        public static VectorPayloadMF newMaskInstanceFactory(Class<?> elemType, int length, boolean max_payload) {
+            if (false == max_payload) {
```
XiaohongGong, line 170: Suggested to `if (!max_payload)` ?

src/java.base/share/classes/jdk/internal/vm/vector/VectorSupport.java
```diff
@@ -178,7 +178,28 @@ public static VectorPayloadMF newInstanceFactory(Class<?> elemType, int length)
...
                     default: assert false : "Unhandled vector mask size";
                 }
-            } else if (elemType == byte.class) {
+            } else {
+                if (elemType == byte.class) {
+                   return new VectorPayloadMFMaxBZ();
+                } else if (elemType == short.class) {
+                   return new VectorPayloadMFMaxSZ();
+                } else if (elemType == int.class) {
+                   return new VectorPayloadMFMaxIZ();
```
XiaohongGong, line 187: Suggest:
```
else if (elemType == int.class || elemType == float.class) {
    return new VectorPayloadMFMaxIZ();
} else {
    assert elemType == long.class || elemType == double.class;
    return new VectorPayloadMFMaxLZ();
}
```

src/java.base/share/classes/jdk/internal/vm/vector/VectorSupport.java
```diff
@@ -178,7 +178,28 @@ public static VectorPayloadMF newInstanceFactory(Class<?> elemType, int length)
...
+                   assert elemType == double.class;
+                   return new VectorPayloadMFMaxLZ();
+                }
+            }
+            return null;
+        }
+
+        @ForceInline
+        public static VectorPayloadMF newShuffleInstanceFactory(Class<?> elemType, int length, boolean max_payload) {
+            if (false == max_payload) {
```
XiaohongGong, line 202: ditto

src/java.base/share/classes/jdk/internal/vm/vector/VectorSupport.java
```diff
@@ -187,56 +208,102 @@ public static VectorPayloadMF newInstanceFactory(Class<?> elemType, int length)
...
+                } else if (elemType == int.class) {
+                   return new VectorPayloadMFMaxIB();
+                } else if (elemType == long.class) {
+                   return new VectorPayloadMFMaxLB();
+                } else if (elemType == float.class) {
+                   return new VectorPayloadMFMaxIB();
+                } else {
+                   assert elemType == double.class;
+                   return new VectorPayloadMFMaxLB();
                 }
```
XiaohongGong, line 227: ditto

src/java.base/share/classes/jdk/internal/vm/vector/VectorSupport.java
```diff
@@ -187,56 +208,102 @@ public static VectorPayloadMF newInstanceFactory(Class<?> elemType, int length)
...
+                    return new VectorPayloadMFMaxD();
                 }
             }
             return null;
         }
 
         @ForceInline
-        public static VectorPayloadMF createVectPayloadInstanceB(int length, byte [] init) {
-            VectorPayloadMF obj = newInstanceFactory(byte.class, length);
+        public static VectorPayloadMF createVectPayloadInstanceB(int length, byte [] init, boolean max_payload) {
```
XiaohongGong, line 305: style: `byte [] init` -> `byte[] init`

src/jdk.incubator.vector/share/classes/jdk/incubator/vector/AbstractMask.java
```diff
@@ -41,8 +41,8 @@ abstract class AbstractMask<E> extends VectorMask<E> {
     /*package-private*/
     abstract VectorPayloadMF getBits();
 
-    static VectorPayloadMF prepare(VectorPayloadMF payload, int offset, int length) {
-        VectorPayloadMF res = VectorPayloadMF.newInstanceFactory(boolean.class, length);
+    static VectorPayloadMF prepare(VectorPayloadMF payload, int offset, Class<?> elementType, int length, boolean is_max_species) {
+        VectorPayloadMF res = VectorPayloadMF.newMaskInstanceFactory(elementType, length, is_max_species);
```
XiaohongGong, line 45: Can we just pass the 'VSPECIES' as the argument? From it, we can get the `elementType`, `length` and `is_max_species` values. And we'd better to create a common method, and wrap such code inside it:
For example:
```
  /*package-private*/
    @ForceInline
    static <F> VectorPayloadMF createMaskPayloadInstance(VectorSpecies<F> species) {
        boolean isMaxShape = species.vectorShape() == VectorShape.S_Max_BIT;
        return VectorPayloadMF.newMaskInstanceFactory(species.elementType(), species.length(), isMaxShape);
    }
```

src/jdk.incubator.vector/share/classes/jdk/incubator/vector/AbstractMask.java
```diff
@@ -53,8 +53,8 @@ static VectorPayloadMF prepare(VectorPayloadMF payload, int offset, int length)
         return res;
     }
 
-    static VectorPayloadMF prepare(boolean val, int length) {
-        VectorPayloadMF res = VectorPayloadMF.newInstanceFactory(boolean.class, length);
+    static VectorPayloadMF prepare(boolean val, Class<?> elementType, int length, boolean is_max_species) {
+        VectorPayloadMF res = VectorPayloadMF.newMaskInstanceFactory(elementType, length, is_max_species);
```
XiaohongGong, line 57: Same as above.

src/jdk.incubator.vector/share/classes/jdk/incubator/vector/AbstractShuffle.java
```diff
@@ -39,8 +39,8 @@ abstract class AbstractShuffle<E> extends VectorShuffle<E> {
     /*package-private*/
     abstract VectorPayloadMF indices();
 
-    static VectorPayloadMF prepare(int length, int[] indices, int offset) {
-        VectorPayloadMF payload = VectorPayloadMF.newInstanceFactory(byte.class, length);
+    static VectorPayloadMF prepare(Class<?> elemType, int length, int[] indices, int offset, boolean is_max_species) {
+        VectorPayloadMF payload = VectorPayloadMF.newShuffleInstanceFactory(elemType, length, is_max_species);
```
XiaohongGong, line 43: Same as VectorMask. We can create a common method which returns a shuffe payload from the given species.

src/jdk.incubator.vector/share/classes/jdk/incubator/vector/Byte128Vector.java
```diff
@@ -69,8 +69,8 @@ final VectorPayloadMF vec() {
         return payload;
     }
 
-    static final Byte128Vector ZERO = new Byte128Vector(VectorPayloadMF.newInstanceFactory(byte.class, 16));
-    static final Byte128Vector IOTA = new Byte128Vector(VectorPayloadMF.createVectPayloadInstanceB(16, (byte[])(VSPECIES.iotaArray())));
+    static final Byte128Vector ZERO = new Byte128Vector(VectorPayloadMF.newVectorInstanceFactory(byte.class, 16, false));
+    static final Byte128Vector IOTA = new Byte128Vector(VectorPayloadMF.createVectPayloadInstanceB(VLENGTH, (byte[])(VSPECIES.iotaArray()), false));
```
XiaohongGong, line 73: Same as VectorShuffle/Mask, create a common method which accepts VectorSpecies as the argument.

XiaohongGong:
I met the new JVM crash issues on ARM Neon platform with this PR. Log is:
```
test IntMaxVectorLoadStoreTests.loadArrayMaskIOOBE(int[i * 5], l + speciesl + 1, mask[false]): success
#
#
#  Internal Error (valhalla/src/hotspot/share/opto/type.cpp:4177), pid=1739617, tid=1739655
#  assert(!flat_array || can_be_inline_type()) failed: Only inline types can be flat in array
#
# JRE version: OpenJDK Runtime Environment (22.0) (fastdebug build 22-internal-git-b575ef0fa)
# Java VM: OpenJDK 64-Bit Server VM (fastdebug 22-internal-git-b575ef0fa, mixed mode, compressed oops, compressed class ptrs, g1 gc, linux-aarch64)
# Problematic frame:
# V  [libjvm.so+0x18697d0]  TypeInstPtr::TypeInstPtr(TypePtr::PTR, ciKlass*, TypePtr::InterfaceSet const&, bool, ciObject*, Type::Offset, bool, int, TypePtr const*, int)+0x15c
```

src/hotspot/share/prims/vectorSupport.hpp
```diff
@@ -138,5 +138,8 @@ class VectorSupport : AllStatic {
   static bool is_vector(Klass* klass);
   static bool is_vector_payload_mf(Klass* klass);
   static bool skip_value_scalarization(Klass* klass);
+  static int  max_vector_size(BasicType bt);
+  static int  get_max_multifield_count(const Symbol* payload_name);
+  static bool is_vector_payload(const Symbol* metadata);
```
XiaohongGong, line 143: `is_vector_payload` is not used and can be removed?

src/hotspot/share/ci/ciObjectFactory.cpp
```diff
@@ -52,6 +52,7 @@
 #include "memory/allocation.inline.hpp"
 #include "memory/universe.hpp"
 #include "oops/oop.inline.hpp"
+#include "prims/vectorSupport.hpp"
```
XiaohongGong, line 55: Not used and revert it?

src/hotspot/share/prims/vectorSupport.cpp
jatin-bhateja: Correct, we need handle it for aarch64.  Matcher routines will not be accessible by C1 only builds. That was a limitation in existing [_VectorSupport_GetMaxLaneCount_](https://github.com/openjdk/valhalla/blob/lworld/src/hotspot/share/prims/vectorSupport.cpp#L583) functionality. Max vector size is tightly bound to targets.

> Could we define a public/common method and implement it on each platform like matcher

Yes, x86 already does that currently, we need to have a target specific handling one way or other, We need need max vector size to estimate bundle size during field parsing and wanted to avoid its dependency on opto compiler.

jatin-bhateja:
> I met a new JVM crash issues on ARM Neon platform with this PR. Log is:
> 
> ```
> test IntMaxVectorLoadStoreTests.loadArrayMaskIOOBE(int[i * 5], l + speciesl + 1, mask[false]): success
> #
> # A fatal error has been detected by the Java Runtime Environment:
> #
> #  Internal Error (valhalla/src/hotspot/share/opto/type.cpp:4177), pid=1739617, tid=1739655
> #  assert(!flat_array || can_be_inline_type()) failed: Only inline types can be flat in array
> #
> # JRE version: OpenJDK Runtime Environment (22.0) (fastdebug build 22-internal-git-b575ef0fa)
> # Java VM: OpenJDK 64-Bit Server VM (fastdebug 22-internal-git-b575ef0fa, mixed mode, compressed oops, compressed class ptrs, g1 gc, linux-aarch64)
> # Problematic frame:
> # V  [libjvm.so+0x18697d0]  TypeInstPtr::TypeInstPtr(TypePtr::PTR, ciKlass*, TypePtr::InterfaceSet const&, bool, ciObject*, Type::Offset, bool, int, TypePtr const*, int)+0x15c
> ```
> 
> And a regression in `DoubleMaxVectorTests` such as:
> 
> ```
> test DoubleMaxVectorTests.IS_FINITEMaskedDoubleMaxVectorTests(double[cornerCaseValue(i)], mask[i % 2]): failure
> java.lang.AssertionError: expected [true] but found [false]
>         at org.testng.Assert.fail(Assert.java:99)
>         at org.testng.Assert.failNotEquals(Assert.java:1037)
>         at org.testng.Assert.assertEqualsImpl(Assert.java:140)
>         at org.testng.Assert.assertEquals(Assert.java:122)
>         at org.testng.Assert.assertEquals(Assert.java:819)
>         at org.testng.Assert.assertEquals(Assert.java:829)
>         at DoubleMaxVectorTests.IS_FINITEMaskedDoubleMaxVectorTests(DoubleMaxVectorTests.java:2716)
>         ...
> ```

Correct, we do have some assertions which I also observed, initial root causes shows these are due to incorrect graph shape, was planning to address it in follow up patch. Let me know after more analysis.

image

src/hotspot/share/classfile/classFileParser.cpp
jatin-bhateja: > An alternative is trying to passing the type info to the `MultiField` annotation, so that the `_class_name` is not needed, and so does to the method `VectorSupport::get_max_multifield_count()`. We can use "max_vector_size(bt)" directly.

Thanks @XiaohongGong , you mean pass explicit vector lane type to multifiled annotation like @multifield(-1,  long.class). Please consider following points 
- We will still need explicit VectorPayload class definitions for mask/shuffule max payloads for different vector types.
- Mask/Shuffle backing storage field type is common for all kinds of vector i.e. byte for shuffle and Boolean for mask.
- We can save on complexity of extending the annotation semantics for max species by doing an efficient vmSymbolID lookup. 
- In future if we refactor shuffle lanes to match the lane types of corresponding vector then additional element type in multifield will be redundant as the backing field type will itself be different for max shuffle payloads.

src/hotspot/share/prims/vectorSupport.cpp
XiaohongGong: Yeah, I agree that we need a non-c2 specific method to get the max vector size. But it is added for x86 now. And I can help add the aarch64 part if you need. My concern is for other platforms that do not have such a method, the jdk image will build failure.

src/hotspot/share/classfile/classFileParser.cpp
XiaohongGong: I understand. It's not a problem for mask/shuffle, since we can add the type info for them as well. And I think that's the main reason that we need the type info. Because we can get the type info during class parser for each field of the normal vector classes. Anyway, looking up symbol ID can also work well.

jatin-bhateja:
Hi @XiaohongGong , Can you kindly do a validation run on AARCH64 with latest patch.

src/hotspot/share/oops/inlineKlass.cpp
```diff
@@ -153,6 +153,9 @@ bool InlineKlass::flat_array() {
   if (!UseFlatArray) {
     return false;
   }
+  if (VectorSupport::is_vector_payload_mf(this) || VectorSupport::is_vector(this)) {
+    return false;
+  }
```
XiaohongGong, line 158: Why do you close the array flatten for vectors?

src/jdk.incubator.vector/share/classes/jdk/incubator/vector/AbstractMask.java
```diff
@@ -138,7 +140,7 @@ public <F> VectorMask<F> cast(VectorSpecies<F> dsp) {
                 this.getClass(), vspecies().elementType(), vspecies().laneCount,
                 species.maskType(), species.elementType(), vspecies().laneCount,
                 this, species,
-                (m, s) -> s.maskFactory(m.getBits()).check(s));
+                (m, s) -> VectorMask.fromLong(s, m.toLong()).check(s));
```
XiaohongGong, line 143: Any issues with the original `maskFactory()` ?

src/hotspot/cpu/x86/vm_version_x86.cpp
```diff
@@ -3215,6 +3215,50 @@ intx VM_Version::allocate_prefetch_distance(bool use_watermark_prefetch) {
...
+    break;
+  case T_BYTE:
+    if (size < 4) return 0;
+    break;
+  case T_SHORT:
+    if (size < 4) return 0;
+    break;
+  default:
+    assert(false, "Unexpected basic type");
+  }
```
XiaohongGong, line 3257: If the return value is 0, what is the actual number of the multifields in payload? It seems not reasonable for `0`. WDYT?

src/hotspot/share/prims/vectorSupport.cpp
```diff
@@ -459,19 +459,50 @@ int VectorSupport::vop2ideal(jint id, BasicType bt) {
 }
 #endif // COMPILER2
 
+int VectorSupport::get_max_multifield_count(const Symbol* payload_name) {
+  assert(payload_name, "");
+  vmSymbolID sid = vmSymbols::find_sid(payload_name);
+  int size = VM_Version::max_vector_size(T_BYTE);
```
XiaohongGong, line 465: It seems we'd better check whether the return value `size` is `0`. If so, a max with 64-bit size is needed like what we have done in java code. See https://github.com/openjdk/valhalla/blob/lworld%2Bvector/src/jdk.incubator.vector/share/classes/jdk/incubator/vector/VectorShape.java#L267

src/hotspot/cpu/aarch64/vm_version_aarch64.cpp
```diff
@@ -657,3 +657,7 @@ void VM_Version::initialize_cpu_information(void) {
 
   _initialized = true;
 }
+
+int VM_Version::max_vector_size(BasicType bt) {
+  return MaxVectorSize / type2aelembytes(bt);
+}
```
XiaohongGong, line 663: Please change to use the code in https://github.com/openjdk/valhalla/blob/lworld%2Bvector/src/hotspot/cpu/aarch64/aarch64.ad#L2344 like what you have changed in x86 systems, together with the `Matcher::max_vector_size()`. Thanks a lot!

XiaohongGong:
> Hi @XiaohongGong , Can you kindly do a validation run on AARCH64 with latest patch.

Sure, I will start the test once you changed the aarch64 part. Thanks!

src/hotspot/share/oops/inlineKlass.cpp
jatin-bhateja: Array of vector may be heterogeneous in nature where each element points to a vector of different shape, thus array flatten do not make much sense to vectors. In additional during type resolution flat_array flag is set for imbalanced Phi (having different type of inputs e.g. VectorPayload64 and VectorPayload128) even though Phi's type is lower common ancestor type of participating inputs i.e., VectorPayload which itself is not an inlinetype, this leads to assertion failures in downstream flow.

https://github.com/openjdk/valhalla/blob/lworld/src/hotspot/share/opto/type.cpp#L4636

src/jdk.incubator.vector/share/classes/jdk/incubator/vector/AbstractMask.java
jatin-bhateja: We cannot typecast value object with another value type, thus directly passing payload (m.getBits()) to maskFactory will cause compilation failures.

src/hotspot/cpu/x86/vm_version_x86.cpp
jatin-bhateja: For x86 this leaf level routine is directly imported from Matcher.  _Matcher::max_vector_size_ will query it to get maximum vector size supported by a target. Agree that it needs to be tuned.

BTW I also see a discrepancy b/w lane count and max vector size for MaxSpecies in existing implementation.

```
LPROMPT>cat test_max.java 

import jdk.incubator.vector.*;

public class test_max {
      System.out.println(ByteVector.SPECIES_MAX);
      System.out.println(ByteVector.broadcast(ByteVector.SPECIES_MAX, 1));
   }
} 
LPROMPT>java -XX:MaxVectorSize=0 --add-modules=jdk.incubator.vector -XX:+PrintFlagsFinal -cp . test_max.java  | grep -e "MaxVectorSize"  -e "Species" -e "\["
OpenJDK 64-Bit Server VM warning: MaxVectorSize must be at least 4 on this platform
[Global flags]
     intx MaxVectorSize                            = 4                                      {C2 product} {command line}
WARNING: Using incubator modules: jdk.incubator.vector
warning: using incubating module(s): jdk.incubator.vector
1 warning
Species[byte, 8, S_Max_BIT]
[1, 1, 1, 1, 1, 1, 1, 1]
LPROMPT>
```

XiaohongGong: Yes, it seems werid. I think it is because the min operation with 64-bit shape in java code. Please see my following comments.

src/jdk.incubator.vector/share/classes/jdk/incubator/vector/AbstractMask.java
XiaohongGong: Sounds reasonable! But I'm curious why we didn't meet this issue before?

src/hotspot/share/oops/inlineKlass.cpp
XiaohongGong: Thanks for the explanation! Seems it also causes the jvm crash in `VectorXXXConversionTests`, but I didn't figure out the root cause yet.  Is it possible supporting the flat array for vector objects with some fixes in future? We may lose the performance with this change, right?

jatin-bhateja: Since we never expose concrete vector types to users hence its only possible to create an array of abstract vector references which can be heterogeneous. This is different from an array of value types which are implicitly final classes and may be flattened into a compact layout since such an array will be homogeneous in nature.

src/hotspot/cpu/aarch64/vm_version_aarch64.cpp
```diff
@@ -659,5 +659,11 @@ void VM_Version::initialize_cpu_information(void) {
 }
 
 int VM_Version::max_vector_size(BasicType bt) {
-  return MaxVectorSize / type2aelembytes(bt);
+ // The MaxVectorSize should have been set by detecting SVE max vector register size.
+  int size = MIN2((UseSVE > 0) ? 256 : 16, (int)MaxVectorSize);
```
XiaohongGong, line 663: Style:
```
// The MaxVectorSize ...
int size = MIN2((UseSVE > 0) ? 256 : 16, (int)MaxVectorSize);
```

XiaohongGong (approved):
Generally LGTM! Just some minor code style.

src/hotspot/share/ci/ciObjectFactory.cpp
XiaohongGong: This is not reverted, isn't it?

src/hotspot/share/oops/inlineKlass.cpp
XiaohongGong: Yeah, I understand. Maybe we could find other more general ways to forbit such kind of array flatten like vector objects, instead of hard coding the `vector/vector_payload` classes. We can revisit this in future. WDYT?

src/hotspot/cpu/x86/vm_version_x86.cpp
```diff
@@ -3215,6 +3215,50 @@ intx VM_Version::allocate_prefetch_distance(bool use_watermark_prefetch) {
...
+  // AVX2 supports 256bit vectors for all types.
+  // AVX2/EVEX supports 512bit vectors for all types.
+  int size = (UseAVX > 1) ? (1 << UseAVX) * 8 : 16;
+  // AVX1 supports 256bit vectors only for FLOAT and DOUBLE.
+  if (UseAVX > 0 && (bt == T_FLOAT || bt == T_DOUBLE))
+    size = (UseAVX > 2) ? 64 : 32;
+  if (UseAVX > 2 && (bt == T_BYTE || bt == T_SHORT || bt == T_CHAR))
+    size = (VM_Version::supports_avx512bw()) ? 64 : 32;
+  // Use flag to limit vector size.
+  size = MIN2(size,(int)MaxVectorSize);
```
XiaohongGong, line 3232: Since `MaxVectorSize` is a C2 specific version, what the behavior if c2 is disabled? Do you think we need consider `#if COMPILER2` here?

jatin-bhateja:
> Generally LGTM! Just some minor code style.

Hi @XiaohongGong ,

Just to confirm do you still see the failing regressions on AARCH64 ?

src/hotspot/cpu/x86/vm_version_x86.cpp
jatin-bhateja: We do have a dependency on C2 one way or other,  I think we should preserve the existing behaviour which will return a -1 value.

XiaohongGong:
> > Generally LGTM! Just some minor code style.
> 
> Hi @XiaohongGong ,
> 
> Just to confirm do you still see the failing regressions on AARCH64 ?

Yes, `DoubeMaxVectorTests.java` still has regression like:
```
test DoubleMaxVectorTests.IS_FINITEMaskedDoubleMaxVectorTests(double[cornerCaseValue(i)], mask[i % 2]): failure
java.lang.AssertionError: expected [true] but found [false]
        at org.testng.Assert.fail(Assert.java:99)
        at org.testng.Assert.failNotEquals(Assert.java:1037)
        at org.testng.Assert.assertEqualsImpl(Assert.java:140)
        at org.testng.Assert.assertEquals(Assert.java:122)
        at org.testng.Assert.assertEquals(Assert.java:819)
        at org.testng.Assert.assertEquals(Assert.java:829)
        at DoubleMaxVectorTests.IS_FINITEMaskedDoubleMaxVectorTests(DoubleMaxVectorTests.java:2716)
```
All other tests pass except this one.

jatin-bhateja:
> > > Generally LGTM! Just some minor code style.
> > 
> > 
> > Hi @XiaohongGong ,
> > Just to confirm do you still see the failing regressions on AARCH64 ?
> 
> Yes, `DoubeMaxVectorTests.java` still has regression like:
> 
> ```
> test DoubleMaxVectorTests.IS_FINITEMaskedDoubleMaxVectorTests(double[cornerCaseValue(i)], mask[i % 2]): failure
> java.lang.AssertionError: expected [true] but found [false]
>         at org.testng.Assert.fail(Assert.java:99)
>         at org.testng.Assert.failNotEquals(Assert.java:1037)
>         at org.testng.Assert.assertEqualsImpl(Assert.java:140)
>         at org.testng.Assert.assertEquals(Assert.java:122)
>         at org.testng.Assert.assertEquals(Assert.java:819)
>         at org.testng.Assert.assertEquals(Assert.java:829)
>         at DoubleMaxVectorTests.IS_FINITEMaskedDoubleMaxVectorTests(DoubleMaxVectorTests.java:2716)
> ```
> 
> All other tests pass except this one.

I do not see this on x86, may I request to to root cause and share your observations.

Best Regards

src/hotspot/share/ci/ciEnv.cpp
```diff
@@ -1773,12 +1773,7 @@ void ciEnv::dump_inline_data(int compile_id) {
 
 bool ciEnv::is_multifield_scalarized(BasicType bt, int vec_length) {
 #if COMPILER2
-  CompilerThread* ct = CompilerThread::current();
```
XiaohongGong, line 1776: Could you please change `#if COMPILER2` to `#ifdef COMPILER2` ? This makes the jdk build error with `--configure-args --with-jvm-features=-compiler2`

src/hotspot/cpu/x86/vm_version_x86.cpp
XiaohongGong: This makes jdk build failure with `--configure-args --with-jvm-features=-compiler2 --configure-args --with-jvm-variants=client`. So maybe we have to separate with `#ifdef COMPILER2` and return `-1` for another path.

XiaohongGong:
> > > > Generally LGTM! Just some minor code style.
> > > 
> > > 
> > > Hi @XiaohongGong ,
> > > Just to confirm do you still see the failing regressions on AARCH64 ?
> > 
> > 
> > Yes, `DoubeMaxVectorTests.java` still has regression like:
> > ```
> > test DoubleMaxVectorTests.IS_FINITEMaskedDoubleMaxVectorTests(double[cornerCaseValue(i)], mask[i % 2]): failure
> > java.lang.AssertionError: expected [true] but found [false]
> >         at org.testng.Assert.fail(Assert.java:99)
> >         at org.testng.Assert.failNotEquals(Assert.java:1037)
> >         at org.testng.Assert.assertEqualsImpl(Assert.java:140)
> >         at org.testng.Assert.assertEquals(Assert.java:122)
> >         at org.testng.Assert.assertEquals(Assert.java:819)
> >         at org.testng.Assert.assertEquals(Assert.java:829)
> >         at DoubleMaxVectorTests.IS_FINITEMaskedDoubleMaxVectorTests(DoubleMaxVectorTests.java:2716)
> > ```
> > 
> > 
> > All other tests pass except this one.
> 
> May I request to to root cause and share your observations.
> 
> Best Regards

Not sure whether this is the same reason with the regressions with `-XX:+DeoptimizeALot`. But we can integrate this patch first and revisit this issue together with `-XX:+DeoptimizeALot` issues.

jatin-bhateja:
> > > > > Generally LGTM! Just some minor code style.
> > > > 
> > > > 
> > > > Hi @XiaohongGong ,
> > > > Just to confirm do you still see the failing regressions on AARCH64 ?
> > > 
> > > 
> > > Yes, `DoubeMaxVectorTests.java` still has regression like:
> > > ```
> > > test DoubleMaxVectorTests.IS_FINITEMaskedDoubleMaxVectorTests(double[cornerCaseValue(i)], mask[i % 2]): failure
> > > java.lang.AssertionError: expected [true] but found [false]
> > >         at org.testng.Assert.fail(Assert.java:99)
> > >         at org.testng.Assert.failNotEquals(Assert.java:1037)
> > >         at org.testng.Assert.assertEqualsImpl(Assert.java:140)
> > >         at org.testng.Assert.assertEquals(Assert.java:122)
> > >         at org.testng.Assert.assertEquals(Assert.java:819)
> > >         at org.testng.Assert.assertEquals(Assert.java:829)
> > >         at DoubleMaxVectorTests.IS_FINITEMaskedDoubleMaxVectorTests(DoubleMaxVectorTests.java:2716)
> > > ```
> > > 
> > > 
> > > All other tests pass except this one.
> > 
> > 
> > May I request to to root cause and share your observations.
> > Best Regards
> 
> Not sure whether this is the same reason with the regressions with `-XX:+DeoptimizeALot`. But we can integrate this patch first and revisit this issue together with `-XX:+DeoptimizeALot` issues.

Yes, it looks unrelated to this patch.

PR #947, opened 2023-11-07
Adding internal null-restricted storage API points

fparain:
Adding internal null-restricted storage API points

### DIFF
--- a/src/java.base/share/classes/jdk/internal/vm/VMSupport.java
+++ b/src/java.base/share/classes/jdk/internal/vm/VMSupport.java
@@ -577,4 +577,20 @@ public class VMSupport {
         }
         return length;
     }
+
+    /**
+     * Allocate an array of a value class type with components that behave in
+     * the same way as a {@link jdk.internal.vm.annotation.NullRestricted}
+     * field.
+     * <p>
+     * Because these behaviors are not specified by Java SE, arrays created with
+     * this method should only be used by internal JDK code for experimental
+     * purposes and should not affect user-observable outcomes.
+     *
+     * @throws IllegalArgumentException if {@code componentType} is not a
+     *         value class type or is not annotated with
+     *         {@link jdk.internal.vm.annotation.ImplicitlyConstructible}
+     */
+    public static native Object[] newNullRestrictedArray(Class<?> componentType,
+                                                         int length);
 }
--- a/test/hotspot/jtreg/runtime/valhalla/inlinetypes/AnnotationsTests.java
+++ b/test/hotspot/jtreg/runtime/valhalla/inlinetypes/AnnotationsTests.java
@@ -1,0 +1,302 @@
+/*
+ * Copyright (c) 2023, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+import jdk.test.lib.Asserts;
+import java.lang.reflect.Method;
+import java.lang.reflect.Field;
+import jdk.internal.misc.Unsafe;
+
+
+/*
+ * @test
+ * @summary Test of ImplicitlyConstructible, NullRestricted and LooselyConsistentValue annotations
+ * @modules java.base/jdk.internal.misc
+ * @library /test/lib
+ * @compile -XDenablePrimitiveClasses --add-exports java.base/jdk.internal.vm.annotation=ALL-UNNAMED AnnotationsTests.java
+ * @run main/othervm --add-exports java.base/jdk.internal.vm.annotation=ALL-UNNAMED -XX:+EnableValhalla -XX:+EnablePrimitiveClasses AnnotationsTests
+ */
+
+
+ public class AnnotationsTests {
+
+    private static final Unsafe UNSAFE = Unsafe.getUnsafe();
+
+
+    public static void main(String[] args) {
+        AnnotationsTests tests = new AnnotationsTests();
+        Class c = tests.getClass();
+        for (Method m : c.getDeclaredMethods()) {
+          if (m.getName().startsWith("test_")) {
+            try {
+              System.out.println("Running " + m.getName());
+              m.invoke(tests);
+            } catch (Throwable t) {
+              t.printStackTrace();
+              throw new RuntimeException(t);
+            }
+          }
+        }
+    }
+
+    static class BadClass0 {
+      @jdk.internal.vm.annotation.NullRestricted
+      String s;
+    }
+
+    // Test detection of illegal usage of NullRestricted on an identity field
+    void test_0() {
+      Throwable exception = null;
+      try {
+          BadClass0 bc = new BadClass0();
+      } catch (IncompatibleClassChangeError e) {
+        exception = e;
+        System.out.println("Received " + e);
+      }
+      Asserts.assertNotNull(exception, "Failed to detect illegal use of @NullRestricted");
+    }
+
+    // Test detection of mismatch between a @NullRestricted field and its class that is not @ImplicitlyConstructible
+    static value class ValueClass1 {
+      int i = 0;
+      int j = 0;
+    }
+
+    static class BadClass1 {
+      @jdk.internal.vm.annotation.NullRestricted
+      ValueClass1 vc;
+    }
+
+    void test_1() {
+      Throwable exception = null;
+      try {
+          BadClass1 tc = new BadClass1();
+      } catch (IncompatibleClassChangeError e) {
+        exception = e;
+        System.out.println("Received " + e);
+      }
+      Asserts.assertNotNull(exception, "Failed to detect illegal use of @NullRestricted");
+    }
+
+    // Test a valid @NullRestricted field with a class that is @ImplicitlyConstructible
+    @jdk.internal.vm.annotation.ImplicitlyConstructible
+    static value class ValueClass2 {
+      int i = 0;
+      int j = 0;
+    }
+
+    static class GoodClass2 {
+      @jdk.internal.vm.annotation.NullRestricted
+      ValueClass2 vc;
+    }
+
+    void test_2() {
+      Throwable exception = null;
+      try {
+          GoodClass2 tc = new GoodClass2();
+      } catch (IncompatibleClassChangeError e) {
+        exception = e;
+        System.out.println("Received " + e);
+      }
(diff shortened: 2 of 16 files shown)


src/java.base/share/classes/jdk/internal/vm/VMSupport.java
```diff
@@ -577,4 +577,20 @@ private static int readLength(DataInputStream dis) throws IOException {
...
+     * <p>
+     * Because these behaviors are not specified by Java SE, arrays created with
+     * this method should only be used by internal JDK code for experimental
+     * purposes and should not affect user-observable outcomes.
+     *
+     * @throws IllegalArgumentException if {@code componentType} is not a
+     *         value class type or is not annotated with
+     *         {@link jdk.internal.vm.annotation.ImplicitlyConstructible}
+     */
+    public static native Object[] newNullRestrictedArray(Class<?> componentType,
```
mlchung, line 594: I think this API should be moved to some other runtime class.    VMSupport is the support class used by JVM CI, JVM TI, and attach mechanism.

What about `jdk.internal.value.ValueClassSupport` or `jdk.internal.misc.VM` for now?   We can find a better home later.

src/java.base/share/native/libjava/VMSupport.c
```diff
@@ -41,3 +41,4 @@ Java_jdk_internal_vm_VMSupport_getVMTemporaryDirectory(JNIEnv *env, jclass cls)
 {
     return JVM_GetTemporaryDirectory(env);
 }
+
```
mlchung, line 44: Nit: should have no change in this file.

test/hotspot/jtreg/runtime/valhalla/inlinetypes/ClassInitializationFailuresTest.java
```diff
@@ -103,45 +116,49 @@ static void testArrayInitialization() {
...
+        // Asserts.assertTrue(e.getClass() == ExceptionInInitializerError.class, " Must be an ExceptionInInitializerError");
+        // // Second attempt because it doesn't fail the same way
+        // try {
+        //     BadThree[][][] array = new BadThree[10][30][10];
+        // } catch(Throwable t) {
+        //     e = t;
+        // }
+        // Asserts.assertNotNull(e, "Error should have been thrown");
+        // Asserts.assertTrue(e.getClass() == NoClassDefFoundError.class, "Must be a NoClassDefFoundError");
+        // Asserts.assertTrue(e.getCause().getClass() == ExceptionInInitializerError.class, "Must be an ExceptionInInitializerError");
```
TobiHartmann, line 152: I would suggest to use `/*` `*/` for uncommenting because it does not affect git blame history of the other lines.

test/hotspot/jtreg/runtime/valhalla/inlinetypes/AnnotationsTests.java
```diff
@@ -0,0 +1,363 @@
...
+
+ public class AnnotationsTests {
+
+    private static final Unsafe UNSAFE = Unsafe.getUnsafe();
+
+    public static void main(String[] args) {
+        AnnotationsTests tests = new AnnotationsTests();
+        Class c = tests.getClass();
+        for (Method m : c.getDeclaredMethods()) {
+          if (m.getName().startsWith("test_")) {
```
TobiHartmann, line 52: Test should use 4-whitespace indentation.

test/hotspot/jtreg/runtime/valhalla/inlinetypes/VolatileTest.java
```diff
@@ -28,25 +28,32 @@
  * @summary check effect of volatile keyword on flattenable fields
  * @modules java.base/jdk.internal.misc
  * @library /test/lib
- * @compile -XDenablePrimitiveClasses VolatileTest.java
- * @run main/othervm -XX:+EnableValhalla -XX:+EnablePrimitiveClasses -XX:InlineFieldMaxFlatSize=128 runtime.valhalla.inlinetypes.VolatileTest
+ * @compile -XDenablePrimitiveClasses --add-exports java.base/jdk.internal.vm.annotation=ALL-UNNAMED VolatileTest.java
```
TobiHartmann, line 31: You could also remove the `-XDenablePrimitiveClasses` now.

TobiHartmann (changes requested):
Sorry for all the delayed comments, I'm reviewing this while trying to convert some of the C2 tests. The following test fails when disabling `EnablePrimitiveClasses`:

`java --add-exports java.base/jdk.internal.vm.annotation=ALL-UNNAMED --add-exports java.base/jdk.internal.misc=ALL-UNNAMED -cp . -XX:+EnableValhalla -XX:-EnablePrimitiveClasses -Xint Test`

```
#
#  SIGSEGV (0xb) at pc=0x00007f665bbba4cb, pid=1616701, tid=1616702
#
# JRE version: Java(TM) SE Runtime Environment (22.0) (fastdebug build 22-lworld4ea-2023-11-20-1024510.tobias...)
# Java VM: Java HotSpot(TM) 64-Bit Server VM (fastdebug 22-lworld4ea-2023-11-20-1024510.tobias..., interpreted mode, compressed oops, compressed class ptrs, g1 gc, linux-amd64)
# Problematic frame:
# V  [libjvm.so+0xc174cb]  FieldLayoutBuilder::inline_class_field_sorting(JavaThread*)+0x4cb

Stack: [0x00007f665ad49000,0x00007f665ae4a000],  sp=0x00007f665ae46b90,  free space=1014k
V  [libjvm.so+0xc174cb]  FieldLayoutBuilder::inline_class_field_sorting(JavaThread*)+0x4cb  (array.hpp:126)
V  [libjvm.so+0xc1d3e7]  FieldLayoutBuilder::compute_inline_class_layout(JavaThread*)+0x27  (fieldLayoutBuilder.cpp:871)
V  [libjvm.so+0xc1daaa]  FieldLayoutBuilder::build_layout(JavaThread*)+0x1a  (fieldLayoutBuilder.cpp:1013)
V  [libjvm.so+0x98da69]  ClassFileParser::post_process_parsed_stream(ClassFileStream const*, ConstantPool*, JavaThread*)+0x1799  (classFileParser.cpp:6631)
V  [libjvm.so+0x98ee62]  ClassFileParser::ClassFileParser(ClassFileStream*, Symbol*, ClassLoaderData*, ClassLoadInfo const*, ClassFileParser::Publicity, JavaThread*)+0x392  (classFileParser.cpp:6022)
V  [libjvm.so+0x1215f23]  KlassFactory::create_from_stream(ClassFileStream*, Symbol*, ClassLoaderData*, ClassLoadInfo const&, JavaThread*)+0x133  (klassFactory.cpp:201)
V  [libjvm.so+0x17b19c3]  SystemDictionary::resolve_class_from_stream(ClassFileStream*, Symbol*, Handle, ClassLoadInfo const&, JavaThread*)+0xb3  (systemDictionary.cpp:947)
V  [libjvm.so+0x102dae2]  jvm_define_class_common(char const*, _jobject*, signed char const*, int, _jobject*, char const*, JavaThread*)+0x242  (jvm.cpp:952)
V  [libjvm.so+0x102dd54]  JVM_DefineClassWithSource+0x94  (jvm.cpp:1119)
C  [libjava.so+0xf2d2]  Java_java_lang_ClassLoader_defineClass1+0x182  (ClassLoader.c:139)
```


```
import jdk.internal.misc.VM;
import jdk.internal.vm.annotation.ImplicitlyConstructible;
import jdk.internal.vm.annotation.LooselyConsistentValue;
import jdk.internal.vm.annotation.NullRestricted;

public class Test {

    @ImplicitlyConstructible
    @LooselyConsistentValue
    static value class MyValueEmpty {

    }

    @ImplicitlyConstructible
    @LooselyConsistentValue
    static value class EmptyContainer {
        @NullRestricted
        MyValueEmpty empty = new MyValueEmpty();
    }

        new EmptyContainer();
    }
}
```

TobiHartmann (changes requested):
This test (classes are equivalent to other test) fails only with `-XX:InlineFieldMaxFlatSize=0` but passes with default values:

```
Exception in thread "main" java.lang.RuntimeException: Failed
	at Test.main(Test.java:23)
```

```
        if (new EmptyContainer() != EmptyContainer.default) {
            throw new RuntimeException("Failed");
        }
    }
``` 

I assume that's because it uses `.default` which will go away?

TobiHartmann:
I found another issue. Below test passes with default arguments but fails with `-XX:InlineFieldMaxFlatSize=0 `:

``` 
Exception in thread "main" java.lang.RuntimeException: Failed
	at Test.main(Test.java:26)
```

```
        EmptyContainer empty = new EmptyContainer();
        EmptyContainer[] array = (EmptyContainer[])VM.newNullRestrictedArray(EmptyContainer.class, 1);
        array[0] = new EmptyContainer();
        if (empty != array[0]) {
            throw new RuntimeException("Failed");
        }
    }
```

test/hotspot/jtreg/runtime/valhalla/inlinetypes/ClassInitializationFailuresTest.java
```diff
@@ -103,45 +116,49 @@ static void testArrayInitialization() {
...
         try {
-            BadTwo[] array = new BadTwo[10];
+            BadTwo[] array = (BadTwo[])VM.newNullRestrictedArray(BadTwo.class, 10);
         } catch(Throwable t) {
             e = t;
         }
         Asserts.assertNotNull(e, "Error should have been thrown");
         Asserts.assertTrue(e.getClass() == NoClassDefFoundError.class, "Must be a NoClassDefFoundError");
         Asserts.assertTrue(e.getCause().getClass() == ExceptionInInitializerError.class, "Must be an ExceptionInInitializerError");
+        // Transition model (annotations and array factory) doesn't permit multi-dimentional arrays tests
```
TobiHartmann, line 134: Shouldn't support multi-dimensional arrays as well? Or are there any blockers?

MrSimms (approved):
+1 to Tobias's comments, otherwise looking good

TobiHartmann:
> The problem here is that the substitutability code doesn't recognize the field as a non-free non-flat field, because it isn't aware that the @NullRestricited annotation has the same effect as a Q-descriptor in the field signature. So, the substitutability code performs a simple comparison between the reference from the initialized field and the null reference of the default value, finds them different and returns a value indicating that the two arguments are different.

Right, I hit the same issue when converting some other tests not using `.default`. Do you plan to adjust the substitutability code accordingly?

test/hotspot/jtreg/runtime/valhalla/inlinetypes/ClassInitializationFailuresTest.java
fparain: The new API to create flat arrays only supports single-dimension arrays. But because multi-dimensional arrays are basically a set of reference arrays with only the last-dimension array being an array of the element type, it is possible to create manually multi-dimensional arrays on top of this API.
Those tests were checking the behavior of `multianewarray` but with the removal of Q-descriptors, this bytecode cannot be used to create multi-dimensional arrays with a flat last-dimension array.

fparain:
Last push should fix all cases where crashes were prevented by adding back `-XX:+EnablePrimitiveClasses`.

RogerRiggs:
The substitutably code is using a VarHandle, that bottoms out in Unsafe.getValue(o, offset, valueType).
Does the VM using a  lazy encoding of the default instance as null result in code using unsafe having to know about the VM representation? 
I'd like to think NOT, but I don't think Unsafe has enough information about the field stored at offset to do the right thing.
My second choice is that the VarHandle for the field replaces the literal null of a \@NullRestricted field with the default.

fparain:
Right now, the substitution of the null reference with the default value is performed in Unsafe:

```    public Object getReference(Object o, long offset, Class<?> type) {
        Object ref = getReference(o, offset);
        if (ref == null && PrimitiveClass.isPrimitiveValueType(type)) {
            // If the type of the returned reference is a primitive value type,
            // return an uninitialized default value if null
            ref = uninitializedDefaultValue(type);
        }
        return ref;
    }
```
The problem in this method is that there's no information about the field except the offset, so it would be expensive to retrieve the field information to check the presence of the @NullRestricted annotation.
Moving the logic to VarHandle would make sense to me.

TobiHartmann:
Thanks Fred, I see that fixing the substitutability check is now tracked by [JDK-8321039](https://bugs.openjdk.org/browse/JDK-8321039).

TobiHartmann (approved):

PR #949, opened 2023-11-13
8319971: [lworld+vector] Fallback implementation cleanups.

jatin-bhateja:
- Cleanup Unsafe APIs based fallback implementation.

Thanks,
Jatin


src/jdk.incubator.vector/share/classes/jdk/incubator/vector/Byte128Vector.java
```diff
@@ -55,6 +55,8 @@ value class Byte128Vector extends ByteVector {
 
     static final Class<Byte> ETYPE = byte.class; // used by the JVM
 
+    static final Unsafe U = Unsafe.getUnsafe();
+
```
XiaohongGong, line 59: Can we reuse `U` defined in `ByteVector.java` ?

XiaohongGong (approved):
LGTM!

PR #952, opened 2023-11-14
8319972: [lworld+vector] Enable intrinsification of Unsafe.finishPrivateBuffer.

jatin-bhateja:
Re-enabling intrinsification of Unsafe.finishPrivateBuffer to fix performance degradation seen in Vector API JMH micro benchmarks. It's intrinsifation was disabled to fix incorrectness caused due to missing inductive InlineType Phi node in loops with Unsafe.put* operations. Since Unsafe.put* APIs returns a void hence JVM state is not altered and ciTypeFlow dataflow analysis does not encounter a inductive definition in the loop, due to this C2 parser misses creating an inductive phi node on
encountering [loop header block](https://github.com/openjdk/valhalla/blob/lworld%2Bvector/src/hotspot/share/opto/parse1.cpp#L696). 

In order to maintain IR compliance with ciTypeFlow analysis C2 should prevent scalarizing value object b/w make and finish private buffer calls.

New behavior of unsafe inline expanders
1) makePrivate: Receive InlineTypeNode input and return initialized buffer in larval state.
2) finishPrivateBuffer: Receive value object buffer input and return rematerialize InlineTypeNode in non-larval state.

This may result into creation of unbalanced phi node at control flow merges where one of the phi input may InlineTypeNode and other is a buffer of compatible value type, but the IR still remains valid.

In addition compiler is now preventing elimination of allocation in larval state.

Validation Status:-
  - All the VectorAPI and Valhalla JTREG tests are now passing at various AVX level with / wo -XX:+DoptimizeALot.

Best Regards,
Jatin

---------
### Progress

### Issue
 * [JDK-8319972](https://bugs.openjdk.org/browse/JDK-8319972): [lworld+vector] Enable intrinsification of Unsafe.finishPrivateBuffer. (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/952/head:pull/952` \
`$ git checkout pull/952`

`$ git checkout pull/952` \
`$ git pull https://git.openjdk.org/valhalla.git pull/952/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 952`

`$ git pr show -t 952`

</details>

<a href="https://git.openjdk.org/valhalla/pull/952.diff">https://git.openjdk.org/valhalla/pull/952.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/952#issuecomment-1810416359)

### DIFF
--- a/src/hotspot/share/opto/graphKit.cpp
+++ b/src/hotspot/share/opto/graphKit.cpp
@@ -54,6 +54,7 @@
 #include "utilities/bitMap.inline.hpp"
 #include "utilities/powerOfTwo.hpp"
 #include "utilities/growableArray.hpp"
+#include "classfile/vmSymbols.hpp"
 
 //----------------------------GraphKit-----------------------------------------
 // Main utility constructor.
@@ -1956,6 +1957,15 @@ Node* GraphKit::set_results_for_java_call(CallJavaNode* call, bool separate_io_p
       if (type->is_inlinetypeptr()) {
         ret = InlineTypeNode::make_from_oop(this, ret, type->inline_klass(), type->inline_klass()->is_null_free());
       }
+      Node* receiver = !call->method()->is_static() ? call->in(TypeFunc::Parms) : nullptr;
+      if (ret->is_InlineType() &&
+          receiver && receiver->bottom_type()->isa_instptr() &&
+          receiver->bottom_type()->is_instptr()->instance_klass()->name()->get_symbol() == vmSymbols::jdk_internal_misc_Unsafe() &&
+          call->method()->name()->get_symbol() == vmSymbols::makePrivateBuffer_name()) {
+        // Re-buffer scalarized InlineTypeNodes returned from makePrivateBuffer
+        // and transition the allocation into larval state.
+        ret = ret->as_InlineType()->make_larval(this);
+      }
     }
   }
 
@@ -3416,7 +3426,7 @@ Node* GraphKit::gen_checkcast(Node *obj, Node* superklass, Node* *failure_contro
           assert(safe_for_replace, "must be");
           obj = null_check(obj);
         }
-        assert(stopped() || !toop->is_inlinetypeptr() || obj->is_InlineType(), "should have been scalarized");
+        assert(stopped() || !toop->is_inlinetypeptr() || obj->is_InlineType() || obj->bottom_type()->is_inlinetypeptr(), "should have been scalarized");
         return obj;
       case Compile::SSC_always_false:
         if (null_free) {
--- a/src/hotspot/share/opto/graphKit.cpp
+++ b/src/hotspot/share/opto/graphKit.cpp
@@ -1864,7 +1864,6 @@ void GraphKit::set_arguments_for_java_call(CallJavaNode* call, bool is_late_inli
     if (t->is_inlinetypeptr() && !call->method()->get_Method()->mismatch() && call->method()->is_scalarized_arg(arg_num)) {
       // We don't pass inline type arguments by reference but instead pass each field of the inline type
       if (!arg->is_InlineType()) {
-        assert(_gvn.type(arg)->is_zero_type() && !t->inline_klass()->is_null_free(), "Unexpected argument type");
         arg = InlineTypeNode::make_from_oop(this, arg, t->inline_klass(), t->inline_klass()->is_null_free());
       }
       InlineTypeNode* vt = arg->as_InlineType();
--- a/src/hotspot/share/opto/inlinetypenode.cpp
+++ b/src/hotspot/share/opto/inlinetypenode.cpp
@@ -797,7 +797,7 @@ Node* InlineTypeNode::Ideal(PhaseGVN* phase, bool can_reshape) {
       // but later decide to inline the call after the callee code also triggered allocation.
       for (DUIterator_Fast imax, i = fast_outs(imax); i < imax; i++) {
         AllocateNode* alloc = fast_out(i)->isa_Allocate();
-        if (alloc != nullptr && alloc->in(AllocateNode::InlineType) == this && !alloc->_is_scalar_replaceable) {
+        if (alloc != nullptr && alloc->in(AllocateNode::InlineType) == this && !alloc->_is_scalar_replaceable && !alloc->_larval) {
           // Found a re-allocation
           Node* res = alloc->result_cast();
           if (res != nullptr && res->is_CheckCastPP()) {
@@ -1009,34 +1009,22 @@ InlineTypeNode* InlineTypeNode::make_from_multi(GraphKit* kit, MultiNode* multi,
   return kit->gvn().transform(vt)->as_InlineType();
 }
 
-InlineTypeNode* InlineTypeNode::make_larval(GraphKit* kit, bool allocate) const {
+Node* InlineTypeNode::make_larval(GraphKit* kit) const {
   ciInlineKlass* vk = inline_klass();
-  InlineTypeNode* res = make_uninitialized(kit->gvn(), vk);
-  for (uint i = 1; i < req(); ++i) {
-    res->set_req(i, in(i));
-  }
 
-  if (allocate) {
-    // Re-execute if buffering triggers deoptimization
-    PreserveReexecuteState preexecs(kit);
-    kit->jvms()->set_should_reexecute(true);
-    Node* klass_node = kit->makecon(TypeKlassPtr::make(vk));
-    Node* alloc_oop  = kit->new_instance(klass_node, nullptr, nullptr, true);
-    AllocateNode* alloc = AllocateNode::Ideal_allocation(alloc_oop);
-    alloc->_larval = true;
+  // Re-execute if buffering triggers deoptimization
+  PreserveReexecuteState preexecs(kit);
+  kit->jvms()->set_should_reexecute(true);
+  Node* klass_node = kit->makecon(TypeKlassPtr::make(vk));
+  Node* alloc_oop  = kit->new_instance(klass_node, nullptr, nullptr, true);
+  AllocateNode* alloc = AllocateNode::Ideal_allocation(alloc_oop);
+  alloc->_larval = true;
 
-    store(kit, alloc_oop, alloc_oop, vk);
-    res->set_oop(alloc_oop);
-  }
-  // TODO 8239003
-  //res->set_type(TypeInlineType::make(vk, true));
-  res = kit->gvn().transform(res)->as_InlineType();
-  assert(!allocate || res->is_allocated(&kit->gvn()), "must be allocated");
-  return res;
+  store(kit, alloc_oop, alloc_oop, vk);
+  return alloc_oop;
 }
 
-InlineTypeNode* InlineTypeNode::finish_larval(GraphKit* kit) const {
-  Node* obj = get_oop();
+InlineTypeNode* InlineTypeNode::finish_larval(GraphKit* kit, Node* obj, const TypeInstPtr* vk) {
   Node* mark_addr = kit->basic_plus_adr(obj, oopDesc::mark_offset_in_bytes());
   Node* mark = kit->make_load(nullptr, mark_addr, TypeX_X, TypeX_X->basic_type(), MemNode::unordered);
   mark = kit->gvn().transform(new AndXNode(mark, kit->MakeConX(~markWord::larval_bit_in_place)));
@@ -1048,15 +1036,7 @@ InlineTypeNode* InlineTypeNode::finish_larval(GraphKit* kit) const {
   assert(alloc != nullptr, "must have an allocation node");
   kit->insert_mem_bar(Op_MemBarStoreStore, alloc->proj_out_or_null(AllocateNode::RawAddress));
 
-  ciInlineKlass* vk = inline_klass();
-  InlineTypeNode* res = make_uninitialized(kit->gvn(), vk);
-  for (uint i = 1; i < req(); ++i) {
-    res->set_req(i, in(i));
-  }
-  // TODO 8239003
-  //res->set_type(TypeInlineType::make(vk, false));
-  res = kit->gvn().transform(res)->as_InlineType();
-  return res;
+  return InlineTypeNode::make_from_oop(kit, obj, vk->inline_klass(), !vk->maybe_null());
 }
 
 bool InlineTypeNode::is_larval(PhaseGVN* gvn) const {
@@ -1281,7 +1261,7 @@ void InlineTypeNode::remove_redundant_allocations(PhaseIdealLoop* phase) {
   // will be removed anyway and changing the memory chain will confuse other optimizations.
   for (DUIterator_Fast imax, i = fast_outs(imax); i < imax; i++) {
     AllocateNode* alloc = fast_out(i)->isa_Allocate();
-    if (alloc != nullptr && alloc->in(AllocateNode::InlineType) == this && !alloc->_is_scalar_replaceable) {
+    if (alloc != nullptr && alloc->in(AllocateNode::InlineType) == this && !alloc->_is_scalar_replaceable && !alloc->_larval) {
       Node* res = alloc->result_cast();
       if (res == nullptr || !res->is_CheckCastPP()) {
         break; // No unique CheckCastPP
--- a/src/hotspot/share/opto/library_call.cpp
+++ b/src/hotspot/share/opto/library_call.cpp
@@ -2627,10 +2627,11 @@ bool LibraryCallKit::inline_unsafe_access(bool is_store, const BasicType type, c
     }
   }
 
-  if (argument(1)->is_InlineType() && is_store) {
-    InlineTypeNode* value = InlineTypeNode::make_from_oop(this, base, _gvn.type(argument(1))->inline_klass());
-    value = value->make_larval(this, false);
-    replace_in_map(argument(1), value);
+  if (_gvn.type(base)->is_inlinetypeptr() && is_store) {
+    AllocateNode* alloc = AllocateNode::Ideal_allocation(base);
+    if (alloc != nullptr) {
+      assert(alloc->_larval, "InlineType instance must be in _larval state for unsafe put operation.\n");
+    }
   }
 
   return true;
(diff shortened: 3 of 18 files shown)


src/hotspot/share/opto/library_call.cpp
```diff
@@ -2627,10 +2627,11 @@ bool LibraryCallKit::inline_unsafe_access(bool is_store, const BasicType type, c
...
 
-  if (argument(1)->is_InlineType() && is_store) {
-    InlineTypeNode* value = InlineTypeNode::make_from_oop(this, base, _gvn.type(argument(1))->inline_klass());
-    value = value->make_larval(this, false);
-    replace_in_map(argument(1), value);
+  if (_gvn.type(base)->is_inlinetypeptr() && is_store) {
+    AllocateNode* alloc = AllocateNode::Ideal_allocation(base);
+    if (alloc != nullptr) {
+      assert(alloc->_larval, "InlineType instance must be in _larval state for unsafe put operation.\n");
+    }
```
XiaohongGong, line 2634: Can we move this check after line2397 (https://github.com/openjdk/valhalla/blob/lworld%2Bvector/src/hotspot/share/opto/library_call.cpp#L2397), and remove the same assertion between line2351-2359 (https://github.com/openjdk/valhalla/blob/lworld%2Bvector/src/hotspot/share/opto/library_call.cpp#L2351) ?

XiaohongGong:
Nice job @jatin-bhateja! Verified on NEON with `-XX:+DeoptimizeALot`! 

But I still have some questions that I don't understand very well, together with the reason that you disabled the intrinsify to `finishPrivateBuffer` before.
1. Here is what I pasted from the previous PR:
> Disable inline expansion of Unsafe.finishPrivateBuffer. This is fixing the incorrectness problems seen with masked operations loop e.g.
vt = newPayload();
vt = Unsafe.makePrivateBuffer(vt);
for (int i = 0; i < Double128.length; i++) {
     if (mask[i]) {
          Unsafe.putDouble(vt, apply(src[i]));
     }
}
Unsafe.putDouble() modifies memory state of an InlineTypeNode by storing a new value in payload buffer, it also re-materializes InlineTypeNode from modified memory state. This causes incorrectness if second mask value is false as entire payload is overwritten with default value and earlier lane updates are over-written. In order to be safe, disabling intrinsification of finishPrivateBuffer if incoming argument node is of VectorPayload* type, so that it always receive updated buffer argument.

You mean this causes incorrectness if second mask value is false because the entire payload is overwritten with default value. Do you mean it is optimized to the default value? But if not all its fields are `0`, how this happen? And I cannot see the direct relationship with intrinsifying of `finishPrivateBuffer`. Could you please clarify more?

2. In this PR, I understand the main fixing is avoiding the scalarization of `InlineType` buffer if it is in larval state, right? I guess this is the main cause of the jtreg regression with `-XX:+DeoptimizeALot`? But I still cannot understand the direct relationship with intrinsifying of `finishPrivateBuffer`. Do you mean with the elimination of scalarization for larval state buffer, the above issue in point 1 is gone?

Going through with this PR now. I still need time learning it. 

Thanks,
Xiaohong

src/hotspot/share/opto/inlinetypenode.cpp
```diff
@@ -1009,34 +1009,22 @@ InlineTypeNode* InlineTypeNode::make_from_multi(GraphKit* kit, MultiNode* multi,
...
-    store(kit, alloc_oop, alloc_oop, vk);
-    res->set_oop(alloc_oop);
-  }
-  // TODO 8239003
-  //res->set_type(TypeInlineType::make(vk, true));
-  res = kit->gvn().transform(res)->as_InlineType();
-  assert(!allocate || res->is_allocated(&kit->gvn()), "must be allocated");
-  return res;
+  store(kit, alloc_oop, alloc_oop, vk);
+  return alloc_oop;
```
XiaohongGong, line 1024: Why not setting the oop to allocated buffer and return the `InlineTypeNode` ? We may have optimizations for Unsafe access for `InlineTypeNode` input.

XiaohongGong:
I see the new IR framework check failures in test `compiler/valhalla/inlinetypes/TestIntrinsics.java` with this PR on NEON. 
See the log below:
```
One or more @IR rules failed:

Failed IR Rules (1) of Methods (1)
----------------------------------
1) Method "public void compiler.valhalla.inlinetypes.TestIntrinsics.test22(compiler.valhalla.inlinetypes.MyValue1)" - [Failed IR rules: 1]:
   * @IR rule 1: "@compiler.lib.ir_framework.IR(phase={DEFAULT}, applyIfCPUFeatureAnd={}, applyIf={}, applyIfCPUFeatureOr={}, applyIfCPUFeature={}, counts={}, failOn={"_#CALL_UNSAFE#I_", "_#ALLOC#I_"}, applyIfAnd={}, applyIfOr={}, applyIfNot={})"
     > Phase "PrintOptoAssembly":
       - failOn: Graph contains forbidden nodes:
         * Constraint 2: "(.*precise compiler/valhalla/inlinetypes/MyValue.*\R(.*(?i:mov|xorl|nop|spill).*\R)*.*_new_instance_Java)"
           - Matched forbidden node:
             * 618 +   mov  R1, precise compiler/valhalla/inlinetypes/MyValue1: 0x0000ffff3c5ce718 (compiler/valhalla/inlinetypes/MyInterface):Constant:exact * # ptr
               624     mov R2, #1       # int
               628     call,static 0x0000ffff8bddf240   // ==>  wrapper for: _new_instance_Java

>>> Check stdout for compilation output of the failed methods
```

src/hotspot/share/opto/graphKit.cpp
```diff
@@ -1956,6 +1957,15 @@ Node* GraphKit::set_results_for_java_call(CallJavaNode* call, bool separate_io_p
...
       }
+      Node* receiver = !call->method()->is_static() ? call->in(TypeFunc::Parms) : nullptr;
+      if (ret->is_InlineType() &&
+          receiver && receiver->bottom_type()->isa_instptr() &&
+          receiver->bottom_type()->is_instptr()->instance_klass()->name()->get_symbol() == vmSymbols::jdk_internal_misc_Unsafe() &&
+          call->method()->name()->get_symbol() == vmSymbols::makePrivateBuffer_name()) {
+        // Re-buffer scalarized InlineTypeNodes returned from makePrivateBuffer
+        // and transition the allocation into larval state.
+        ret = ret->as_InlineType()->make_larval(this);
+      }
```
XiaohongGong, line 1967: As my understanding, the expected behavior is compiler should guarantee the returned `InlineTypeNode` has a larval state buffer after calling `makePrivateBuffer()`, right? Why do we still need this here?

src/hotspot/share/opto/library_call.cpp
```diff
@@ -2648,31 +2649,35 @@ bool LibraryCallKit::inline_unsafe_make_private_buffer() {
...
-  InlineTypeNode* vt = buffer->as_InlineType();
-  if (!vt->is_allocated(&_gvn) || VectorSupport::is_vector_payload_mf(vt->inline_klass()->get_InlineKlass())) {
-    return false;
+  // Incoming value should be inline type.
+  if (!buffer->bottom_type()->is_inlinetypeptr()) {
+     return false;
   }
-  // TODO 8239003 Why is this needed?
-  if (AllocateNode::Ideal_allocation(vt->get_oop()) == nullptr) {
+  const TypeInstPtr* ptr = buffer->bottom_type()->isa_instptr();
```
XiaohongGong, line 2667: Please move this before line-2680.

src/hotspot/share/opto/macro.cpp
```diff
@@ -958,7 +958,10 @@ bool PhaseMacroExpand::scalar_replacement(AllocateNode *alloc, GrowableArray <Sa
   if (res != nullptr) { // Could be null when there are no users
     res_type = _igvn.type(res)->isa_oopptr();
   }
-
+  // Bufferes in larval state should not be scalarized.
+  if (alloc->_larval) {
+    return false;
+  }
```
XiaohongGong, line 964: Does it need to also add such checks in `InlineTypeNode::make_scalar_in_safepoints` ?

XiaohongGong:
Do you think it's better to push the main changes (I mean the elimination scalarization of larval state buffer) to `lworld` branch, which may get more reviews from other people?

src/hotspot/share/opto/inlinetypenode.cpp
```diff
@@ -1048,15 +1036,7 @@ InlineTypeNode* InlineTypeNode::finish_larval(GraphKit* kit) const {
...
-  ciInlineKlass* vk = inline_klass();
-  InlineTypeNode* res = make_uninitialized(kit->gvn(), vk);
-  for (uint i = 1; i < req(); ++i) {
-    res->set_req(i, in(i));
-  }
-  // TODO 8239003
-  //res->set_type(TypeInlineType::make(vk, false));
-  res = kit->gvn().transform(res)->as_InlineType();
-  return res;
+  return InlineTypeNode::make_from_oop(kit, obj, vk->inline_klass(), !vk->maybe_null());
```
XiaohongGong, line 1039: I think this is the root cause of incorrectness in https://bugs.openjdk.org/browse/JDK-8317659. The field values of the `InlineTypeNode` are not updated from the changed oop buffer.

So can we just make `Unsafe.finishPrivateBuffer()` accept `InlineTypeNode` like before, checking its oop is in larval state, and return `make_from_oop()` here?

src/hotspot/share/opto/graphKit.cpp
jatin-bhateja: This it to handle larval state propagation if makePrivateBuffer is not inline expanded, in such a case native implementation gets called and at compile time there is not way to detect larval state.

src/hotspot/share/opto/macro.cpp
jatin-bhateja: With new behavior, a value object in larval state is already allocated over heap and should not be scalarized to InlineTypeNode.

src/hotspot/share/opto/inlinetypenode.cpp
jatin-bhateja: Following is the API sequence to modify a vector lane.

obj =  Unsafe.makePrivateBuffer(Value.default);
Unsafe.putDouble(obj, OFFSET, value);
obj = Unsafe.finishPrivateBuffer(obj);

In case putDouble is not in-line expanded, compiler makes a call to Unsafe_MakePrivateBuffer native implimentation. There is no way for compiler to detect any JVM state changes since put* returns a void. Thus finishPrivateBuffer will directly use default value and this will cause incorrectness, preventing scalarization b/w  make and finish private buffer will fix this problem.

src/hotspot/share/opto/macro.cpp
XiaohongGong: Yes, that's true. So can we still generate the `InlineTypeNode` with the larval state buffer as its oops like before? And we can avoid its scalarization by just checking the larval state.

jatin-bhateja:
> Do you think it's better to push the main changes (I mean the elimination scalarization of larval state buffer) to `lworld` branch, which may get more reviews from other people?

I plan to do it separately for JDK-8239003, there have been lots of changes into lworld recently.

src/hotspot/share/opto/macro.cpp
XiaohongGong: I think so. We also need to add the same larval check in https://github.com/openjdk/valhalla/blob/lworld%2Bvector/src/hotspot/share/opto/inlinetypenode.cpp#L365 to avoid the scalarization of the `InlineType`.

src/hotspot/share/opto/graphKit.cpp
XiaohongGong: I see. Thanks! Sounds reasonable. The native implementation of `makePrivateBuffer()` also marks the larval bit of the object. So can we get some help from it?

XiaohongGong:
> > Do you think it's better to push the main changes (I mean the elimination scalarization of larval state buffer) to `lworld` branch, which may get more reviews from other people?
> 
> I plan to do it separately for JDK-8239003, there have been lots of changes into lworld recently.

Sounds good to me!

src/hotspot/share/opto/inlinetypenode.cpp
XiaohongGong: Per my personal understanding, I see that we may miss the memory dependency for above case with the original `fnishPrivateBuffer()` intrinsifaction. But with above changes, it can add the memory dependency to `Unsafe.putDouble`, as `make_from_oop()` can generate the load operations from the private buffer. So the issue is fixing. Right?

I still cannot undersand the relationship with `scalarization` happened between them, as the memory dependence is added (the buffer has def-use) between them, how the scalarization is happened?  Please correct me if any misunderstanding!

src/hotspot/share/opto/graphKit.cpp
jatin-bhateja: Idea here is to prevent scalarization of value objects in larval state, above check ensures that InlineTypeNode rematerialized from return value of non-inlined makePrivateBuffer are rebuffered and transitioned into larval state.

src/hotspot/share/opto/inlinetypenode.cpp
jatin-bhateja: > Per my personal understanding, I see that we may miss the memory dependency for above case with the original `fnishPrivateBuffer()` intrinsifaction. But with above changes, it can add the memory dependency to `Unsafe.putDouble`, as `make_from_oop()` can generate the load operations from the private buffer. So the issue is fixing. Right?
> 
Yes, idea is to rematerialize scalarized representation only though finishPrivateBuffer. 

However, on your observation related to memory dependency binding on finishPrivate buffer is agnostic to intrinsification of putDouble. 
  
> I still cannot understand the relationship with `scalarization` happened between them, as the memory dependence is added (the buffer has def-use) between them, how the scalarization is happened? Please correct me if any misunderstanding!

Scalarization effect is only seen by users of finishPrivateBuffer when InlineTypeNode is rematerialized.

src/hotspot/share/opto/graphKit.cpp
XiaohongGong: It makes sense to me. Thanks!

jatin-bhateja:
Hi @XiaohongGong , Let me know if there are any further clarification here, else I can integrate this.

src/hotspot/share/opto/inlinetypenode.cpp
XiaohongGong: Any feedback on this please?

src/hotspot/share/opto/library_call.cpp
```diff
@@ -2648,31 +2646,34 @@ bool LibraryCallKit::inline_unsafe_make_private_buffer() {
...
   Node* buffer = argument(1);
-  if (!buffer->is_InlineType()) {
+  // Incoming value should be a buffer and not InlineTypeNode.
+  if (buffer->is_InlineType()) {
+     return false;
+  }
+  // Incoming value should be inline type.
+  if (!buffer->bottom_type()->is_inlinetypeptr()) {
+     return false;
+  }
```
XiaohongGong, line 2663:
```suggestion
  // Incoming value should be a buffer with inline type and not InlineTypeNode.
  if (buffer->is_InlineType() || !buffer->bottom_type()->is_inlinetypeptr()) {
     return false;
  }
```

src/hotspot/share/opto/parse1.cpp
```diff
@@ -1777,9 +1777,12 @@ void Parse::merge_common(Parse::Block* target, int pnum) {
           // Allocate inline type in src block to be able to merge it with oop in target block
           map()->set_req(j, n->as_InlineType()->buffer(this));
         } else if (!n->is_InlineType() && t->is_inlinetypeptr()) {
-          // Scalarize null in src block to be able to merge it with inline type in target block
-          assert(gvn().type(n)->is_zero_type(), "Should have been scalarized");
-          map()->set_req(j, InlineTypeNode::make_null(gvn(), t->inline_klass()));
```
XiaohongGong, line 1782: Did you check why this assertion is hit for larval state buffer? I mean is it possible that the inline type buffer in larval state should not go to this path.

src/hotspot/share/opto/inlinetypenode.cpp
jatin-bhateja: The whole idea is to avoid scalarization for value objects in laval state and not construct InlineTypeNode till we encounter a finishPrivateBuffer. A value object is transitioned into larval state to make it mutable, since Unsafe.put* APIs do not modify the JVM state hence to be safe it's better to keep value objects as oops b/w make and finish private buffer, patch also disables re-materialization of InlineTypeNode from updated memory state in Unsafe.put* inline expanders.

src/hotspot/share/opto/parse1.cpp
jatin-bhateja: Since make_larval now generates the oop hence will hit upon this control path, we no longer create InlineTypeNode for value objects in larval state.

src/hotspot/share/opto/inlinetypenode.cpp
XiaohongGong: Yes, I understand your intention. And it makes sense to me. My question is whether we can keep the InlineType with a larval state before, and forbiting the scalarization for it between make/finish private buffer. Changing `InlineType` totally to the oop buffer looks too heavy to me.

jatin-bhateja: Generating InlineTypeNode in make_larval will also necessitate re-materialization of such a node in Unsafe.put* expander which is problematic, given that put only modifies the memory state, totally avoiding creation of InlineTypeNode for value objects in larval state will fix this.

XiaohongGong: OK. Thanks!

XiaohongGong (approved):

jatin-bhateja:
Thanks @XiaohongGong

PR #953, opened 2023-11-15
8319945: [lworld+vector] Fix vector api jtreg crash with "-XX:-EnableVectorSupport"

XiaohongGong:
Several vector api jtreg tests crash with `"-XX:-EnableVectorSupport"`
when running on machines that the operations of vector payload can
be vectorized.

Here is the log:

```
 Internal Error (valhalla/src/hotspot/share/opto/library_call.cpp:2465), pid=467482, tid=467499
 assert(bt == alias_type->basic_type() || bt == T_PRIMITIVE_OBJECT) failed: should match

 JRE version: OpenJDK Runtime Environment (22.0) (fastdebug build 22-internal-git-5b2fa0aab)
 Java VM: OpenJDK 64-Bit Server VM (fastdebug 22-internal-git-5b2fa0aab, mixed mode, sharing, compressed oops, compressed class ptrs, g1 gc, linux-aarch64)
 Problematic frame:
 V [[libjvm.so](http://libjvm.so/)+0x11b1738] LibraryCallKit::inline_unsafe_access(bool, BasicType, LibraryCallKit::AccessKind, bool)+0x13e8
```

The reason is the matched field of the given offset in the vector
payload is not found when using Usafe's access methods. And the root
cause is when the operations of the multifields are vectorizable in
C2, only the multifield base is added into the holder klass's non-static
fields list in ci stage. Hence when accessing other multifields by
Unsafe's get/put methods (which is the fall-back java implementation of
current vector operations), the fields cannot be found with the given
field offset.

This issue is involved by [JDK-8314980](https://bugs.openjdk.java.net/browse/JDK-8314980) (see [1]), which refactors the
multfields parsing in ci stage. The special handling for multifields
in `ciInstanceKlass::get_field_by_offset` is removed. Adding it back
can fix this issue.

This patch also reverted the special consideration for multifields when
printing out the JVM state in a safepoint. It is not needed anymore,
because the fields of an `InlineTypeNode` in C2 is always consistent
with the holder klass's non-static fields after [JDK-8314980](https://bugs.openjdk.java.net/browse/JDK-8314980).

[1] https://github.com/openjdk/valhalla/pull/918

Verified with Vector API jtreg tests with `-XX:-EnableVectorSupport`, and all tests clean.

### DIFF
--- a/src/hotspot/share/ci/ciInstanceKlass.cpp
+++ b/src/hotspot/share/ci/ciInstanceKlass.cpp
@@ -426,6 +426,17 @@ ciField* ciInstanceKlass::get_field_by_offset(int field_offset, bool is_static) 
       if (field_off > field_offset)
         break;
       // could do binary search or check bins, but probably not worth it
+
+      if (field->secondary_fields_count() > 1) {
+        for (int j = 0; j < field->secondary_fields_count() - 1; j++) {
+          ciField* sec_field = static_cast<ciMultiField*>(field)->secondary_fields()->at(j);
+          int sec_field_offset = sec_field->offset_in_bytes();
+          if (sec_field_offset == field_offset)
+            return sec_field;
+          if (sec_field_offset > field_offset)
+            return nullptr;
+        }
+      }
     }
     return nullptr;
   }
--- a/src/hotspot/share/opto/callnode.cpp
+++ b/src/hotspot/share/opto/callnode.cpp
@@ -516,36 +516,25 @@ void JVMState::format(PhaseRegAlloc *regalloc, const Node *n, outputStream* st) 
           format_helper(regalloc, st, larval_node, ":", -1, NULL);
         }
 
+        Node* fld_node = mcall->in(first_ind);
         ciField* cifield;
-        uint sec_fields_count = 0;
-        for (uint j = 0; j < nf; j++) {
-          Node* fld_node = mcall->in(first_ind + j);
-          if (iklass != NULL) {
-            st->print(" [");
-            cifield = iklass->nonstatic_field_at(j - sec_fields_count);
+        if (iklass != nullptr) {
+          st->print(" [");
+          cifield = iklass->nonstatic_field_at(0);
+          cifield->print_name_on(st);
+          format_helper(regalloc, st, fld_node, ":", 0, &scobjs);
+        } else {
+          format_helper(regalloc, st, fld_node, "[", 0, &scobjs);
+        }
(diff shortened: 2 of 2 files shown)


XiaohongGong:
@jatin-bhateja , could you please take a look at this PR? Thanks a lot!

XiaohongGong:
@jatin-bhateja , could you please take a look at this PR? Thanks a lot!

jatin-bhateja:
Hi @XiaohongGong ,
Can you please also add multifield special handling in **_ciInlineKlass::field_index_by_offset(int offset)_**. As of now its only used by C2 and functionalty looks safe.

XiaohongGong:
> Hi @XiaohongGong , Can you please also add multifield special handling in **_ciInlineKlass::field_index_by_offset(int offset)_**. As of now its only used by C2 and functionalty looks safe.

Yes, considering that it is only used by C2 now, it's safe removing the multifields special handing. I suggest keeping it as it is now. One reason is the code may cannot be tested effectively, another is I'm considering whether we can take a further refactory to the multifields vectorization/scalarization in future. We may just keep the scalarized multifields in ci stage, and do the vectorization just in c2 compiler. So that all the special handling in ci could be cleaned-out. I plan to have a try for this in future.

jatin-bhateja:
> > Hi @XiaohongGong , Can you please also add multifield special handling in **_ciInlineKlass::field_index_by_offset(int offset)_**. As of now its only used by C2 and functionalty looks safe.
> 
> Yes, considering that it is only used by C2 now, it's safe removing the multifields special handing. I suggest keeping it as it is now. One reason is the code may cannot be tested effectively, another is I'm considering whether we can take a further refactory to the multifields vectorization/scalarization in future. We may just keep the scalarized multifields in ci stage, and do the vectorization just in c2 compiler. So that all the special handling in ci could be cleaned-out. I plan to have a try for this in future.

Hi @XiaohongGong ,

Sounds reasonable to me, can you kindly add a TODO in field_index_by_offset for multifile handling, we can remove it once our alternate implementation is in place, other change looks ok to me. In future if it's used by C1 then we may need to introduce special handling for multifield since ci model is shared by compilers. 

Best Regards,
Jatin

XiaohongGong:
Hi @jatin-bhateja , your comments have been addressed. Please help take a look again. Thanks a lot!

jatin-bhateja (approved):

XiaohongGong:
Thanks for the review @jatin-bhateja !

PR #964, opened 2023-12-20
8239003: [lworld] C2 should respect larval state when scalarizing

jatin-bhateja:
Hi All,

Patch addresses issues around unsafe updates to value objects within a loop and larval state preservation by suppressing
scalarization of value objects in larval state.

Since Unsafe.put* APIs returns a void, hence it does not alter JVM state. Due to this ciTypeFlow dataflow analysis does
not encounter an inductive definition corresponding to updated value object within the loop, due to this C2 parser misses
creating an inductive phi node on encountering loop header block.

In order to maintain IR compliance with ciTypeFlow analysis C2 should prevent scalarizing value object b/w make and finish
private buffer calls.

New behavior of unsafe inline expanders

* makePrivate: Receive InlineTypeNode input and return initialized buffer in larval state.
* finishPrivateBuffer: Receive value object buffer input and return rematerialize InlineTypeNode in non-larval state.
  This may result into creation of unbalanced phi node at control flow merges where one of the phi input may InlineTypeNode
  and other is a buffer of compatible value type, but the IR still remains valid.

In addition compiler is now preventing elimination of allocation in larval state.

Validation Status:-

All the Valhalla JTREG tests are passing at various AVX level with / wo -XX:+DoptimizeALot.

Best Regards,
Jatin

---------
### Progress

### Issue
 * [JDK-8239003](https://bugs.openjdk.org/browse/JDK-8239003): [lworld] C2 should respect larval state when scalarizing (**Bug** - P2)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/964/head:pull/964` \
`$ git checkout pull/964`

`$ git checkout pull/964` \
`$ git pull https://git.openjdk.org/valhalla.git pull/964/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 964`

`$ git pr show -t 964`

</details>

<a href="https://git.openjdk.org/valhalla/pull/964.diff">https://git.openjdk.org/valhalla/pull/964.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/964#issuecomment-1864490316)

### DIFF
--- a/src/hotspot/share/opto/graphKit.cpp
+++ b/src/hotspot/share/opto/graphKit.cpp
@@ -1862,7 +1862,13 @@ void GraphKit::set_arguments_for_java_call(CallJavaNode* call, bool is_late_inli
     if (t->is_inlinetypeptr() && !call->method()->get_Method()->mismatch() && call->method()->is_scalarized_arg(arg_num)) {
       // We don't pass inline type arguments by reference but instead pass each field of the inline type
       if (!arg->is_InlineType()) {
-        assert(_gvn.type(arg)->is_zero_type() && !t->inline_klass()->is_null_free(), "Unexpected argument type");
+        AllocateNode* alloc = AllocateNode::Ideal_allocation(arg);
+        // Transitioning a value object to larval state disables it scalarization, this will make such
+        // argument non-compliant with pre-computed domain calling convention.
+        if (alloc && alloc->_larval) {
+          env()->record_method_not_compilable("Passing non-scalarized value object in larval state will"
+                                              " break domain calling convention", false);
+        }
         arg = InlineTypeNode::make_from_oop(this, arg, t->inline_klass(), t->inline_klass()->is_null_free());
       }
       InlineTypeNode* vt = arg->as_InlineType();
--- a/src/hotspot/share/opto/graphKit.cpp
+++ b/src/hotspot/share/opto/graphKit.cpp
@@ -27,6 +27,7 @@
 #include "ci/ciInlineKlass.hpp"
 #include "ci/ciUtilities.hpp"
 #include "classfile/javaClasses.hpp"
+#include "classfile/vmSymbols.hpp"
 #include "ci/ciObjArray.hpp"
 #include "asm/register.hpp"
 #include "compiler/compileLog.hpp"
@@ -1954,6 +1955,15 @@ Node* GraphKit::set_results_for_java_call(CallJavaNode* call, bool separate_io_p
       if (type->is_inlinetypeptr()) {
         ret = InlineTypeNode::make_from_oop(this, ret, type->inline_klass(), type->inline_klass()->is_null_free());
       }
+      Node* receiver = !call->method()->is_static() ? call->in(TypeFunc::Parms) : nullptr;
+      if (ret->is_InlineType() &&
+          receiver && receiver->bottom_type()->isa_instptr() &&
+          receiver->bottom_type()->is_instptr()->instance_klass()->name()->get_symbol() == vmSymbols::jdk_internal_misc_Unsafe() &&
+          call->method()->name()->get_symbol() == vmSymbols::makePrivateBuffer_name()) {
+        // Re-buffer scalarized InlineTypeNodes returned from makePrivateBuffer
+        // and transition the allocation into larval state.
+        ret = ret->as_InlineType()->make_larval(this);
+      }
     }
   }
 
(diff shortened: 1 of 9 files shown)


jatin-bhateja:
## RFP: Larval state propagation across methods

Following are C2 compiler scalarization IR intercepts: -
1. On method entry during Initial JVM State construction:- 
    - If incoming parameter is of value type AND JVM flag InlineTypePassFieldsAsArgs is enabled, create a separate Param for each scalarized field.
        - Uses InlineTypeNode::make_from_multi (in=TRUE) to create an InlineTypeNode.
 2. checkcast bytecode parsing:- 
    - User added explicit type casting and profile guided type sharpening.
        - Uses InlineTypeNode::make_from_oop() to create an InlineTypeNode.
 3. Return value of a method.
    - Caller may receive scalarized value object and creates a separate projection for each scalar field and binds them to an InlineTypeNode.
        - Uses InlineTypeNode::make_from_multi(in=F).
    - Profile driven, return type speculation.
        - Uses InlineTypeNode::make_from_oop() to re-materialize InlineTypeNode IR from returned value object.
4. Allocate a new value object,  constructor emits an aconst_init bytecode which creates an InlineTypeNode for default value object followed by creating a new InlineTypeNode for each withfield bytecode.
  
Idea here is to disable creation of a scalarized IR whenever a value object enters larval state through Unsafe.makePrivateBuffer() invocation , which is an unsafe API with native and C2 compiler intrinsic implementation.  
Value object exits the larval state through Unsafe.finishPrivateBuffer(). Between state entry and exit, a checkcast bytecode processing may again try to transform value oop to scalarized IR through intercept B.
 
 Proposal to propagate larval state across method boundaries:-
1. Pass larval transitioned value object argument as buffers.
    - C1 and Interpreter calling methods always pass buffered value objects and jumps to callee’s inline entry point which unpacks the incoming value object fields into GPRs. 
    - C2 calling method to disable scalarization for a larval value object while creating argument list in GraphKit::set_arguments_for_java_call.
    - Extend Inline Entry point to check larval bits in mark word of incoming value object references to bypass scalarization.
    - C2 while construction its initial JVM state to create additional control flow based to larval state check to generate scalarized IR for an incoming parameter or let it pass as it is to downstream graph.
    - Create a new non-scalarized Param node InlineParamNode which represents a value object in larval state.
2. Checkcast to prevent scalarization if value object in larval state.
    - For locally transitioned larval object, allocation is reachable, and its larval state can be checked.
    - For an externally received oop with speculative type promotion to a value object, emit an explicit larval bit check in oop’s mark word to bypass scalarization.
3. Scalarization of return value depends on returns_inline_type_as_fields(), its return value is computed by comparing actual and extended method signatures. 
    - If returning a value object in scalarized form, pass an explicit larval state on the lines of init value.
 4. Caller receiving a value object should emit a check based on larval projections (if scalarized returns) OR larval flag bit in oop’s mark word for non-scalarized return value oop.
5. MakePrivateBuffer C2 intrinsic to buffer contents of the value objects into a new Allocation IR and set associated larval flag. For a non-inlined Unsafe_MakePrivateBuffer bypass scalarization and pass the retuned oop projection to downstream graph.
6. Finish Private Buffer to rematerialize InlineTypeNode from buffers.
7. Above changes may result into imbalanced PhiNodes with one input as buffer and other as a scalarized InlineTypeNode. This may lead to issues while bookkeeping de-optimization state at SafePointNode which will now directly pass buffered narrowOops. To prevent this, we need to buffer all the incoming branches of PhiNode during parsing in Parse::merge_common.

Ideally, we do not need to handle larval state propagation during de-optimization since in mutable state we use value buffers which internally hold the larval information in the mark word.

Test Scenarios
a. Vector API usecase (addressed):-
```
Caller:
      v = makePrivateBuffer();
      for () {
           … Unsafe get / put.
      }
      v = finishPrivateBuffer();
```

b. Caller – Callee, scenario 1
```
Caller:-
           v = makePrivateBuffer();
           invokevirtual callee (v);
                         Callee:-
                                   v = make_from_oop(v);  // during start state construction. 
                                   …

```
c. Caller – Callee, scenario 2
```
Caller:-
           V = Invokevirtual callee (); // make_from_multi (scalarized), make_from_oop (non-scalarized)  
                        Callee
                                  ….
                                  Return makePrivateBuffer (V)

```
d. Caller – Callee , scenario 3
```
Caller:-
           V = makePrivateBuffer(v);
            invokevirtual callee (v);  // Abstract identity class argument descriptor (bi-morphic, poly-morphic)
Callee:-
            ….
             Build_start_state         
              .. oop converted to InlineTypeNode if single sub-class abstract type parameter.  
              ….
                              checkcast param1 , scalarize into InlineTypeNode based on profile. 
```

e. For scalarized arguments (value type descriptor), C2 compiler always assume to receive scalarized objects and method body JIT code is generated based on that assumption.

f. Caller (C1/ Interpreter) -> Callee (C2)
g. C1 use inline entry points to invoke C2 compiled method.
h. Interpreter calls i2c adapter which after argument marshalling invokes inline entry points. 

Performance implications: -
-	Adding checks to test oop’s larval state in inline entry point and checkcast processing will consume additional cycles.

jatin-bhateja:
image

PR #973, opened 2024-01-19
8324114: [lworld] GTestWrapper.java fails with "assert(LockingMode == LM_LEGACY) failed: should only be called with legacy stack locking"

chhagedorn:
This patch fixes the gtests which fail with the lightweight locking mode (`-XX:LockingMode=2`) but also with heavy monitors (`-XX:LockingMode=0`). I've only noticed while working on this bug that in the meantime, we temporary enforce legacy stack-locking (`-XX:LockingMode=1`) since merging jdk-22+17:
https://github.com/openjdk/valhalla/blob/16fa7709f15465e736dd3a83707ffe88d7bc61bd/src/hotspot/share/runtime/arguments.cpp#L1922-L1925

Nevertheless, I think it's still worth to propose these test fixes now since we probably want to support lightweight locking at some point.

The gtest used `markWork::has_locker()` which only works with `LM_LEGACY`. For `LM_LIGHTWEIGHT`, we need to call `markWord::is_fast_locked()`. Additionally, when running with `LM_MONITOR`, the test to match some things in the mark word fails because we only use heavy monitors and do not use the mark word bits.

Thanks,
Christian

### DIFF
--- a/test/hotspot/gtest/oops/test_markWord.cpp
+++ b/test/hotspot/gtest/oops/test_markWord.cpp
@@ -47,6 +47,16 @@ static void assert_test_pattern(Handle object, const char* pattern) {
   ASSERT_THAT(st.base(), testing::HasSubstr(pattern));
 }
 
+static void assert_mark_word_print_pattern(Handle object, const char* pattern) {
+  if (LockingMode == LM_MONITOR) {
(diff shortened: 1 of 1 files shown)


chhagedorn:
Thanks for your reviews!

/integrate

PR #998, opened 2024-02-07
8298659: [lworld] ValueObject isSubstutitable comparison of float/double should use raw bits

RogerRiggs:
Update the substitutability implementation to match the new JLS for JEP 401.
Float and double values should be compared for equality using the raw bits.

### DIFF
--- a/src/java.base/share/classes/java/lang/runtime/ObjectMethods.java
+++ b/src/java.base/share/classes/java/lang/runtime/ObjectMethods.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017, 2022, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2017, 2024, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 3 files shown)


src/java.base/share/classes/java/lang/runtime/ObjectMethods.java
```diff
@@ -179,8 +179,12 @@ private static int hashCombiner(int x, int y) {
     private static boolean eq(char a, char b) { return a == b; }
     private static boolean eq(int a, int b) { return a == b; }
     private static boolean eq(long a, long b) { return a == b; }
-    private static boolean eq(float a, float b) { return Float.compare(a, b) == 0; }
-    private static boolean eq(double a, double b) { return Double.compare(a, b) == 0; }
+    private static boolean eq(float a, float b) {
+        return Float.floatToRawIntBits(a) == Float.floatToRawIntBits(b);
```
mlchung, line 183: `ObjectMethods` are used for records as well and so these methods should not be modified.  Instead I suggest to add new methods in `ValueObjectMethods` class for value object substutitability comparison of float/double.

RogerRiggs: I'm beginning to doubt the wisdom of a difference in field comparisons between value objects and records (and the `==` operator).

liach: I believe the discrepancy in `equals` and `==` is valid; value objects may contain reference pointers too, thus they still should be compared by `equals` instead of `==`. Having double fields act like reference fields is fine.

src/java.base/share/classes/java/lang/runtime/ValueObjectMethods.java
```diff
@@ -662,13 +694,9 @@ private static LinkageError newLinkageError(Throwable e) {
...
      *     this method returns {@code a == b} with the following exception:
      *     <ul>
-     *     <li> If {@code a} and {@code b} both represent {@code NaN},
-     *          this method returns {@code true}, even though {@code NaN == NaN}
-     *          has the value {@code false}.
-     *     <li> If {@code a} is floating point positive zero while {@code b} is
-     *          floating point negative zero, or vice versa, this method
-     *          returns {@code false}, even though {@code +0.0 == -0.0} has
-     *          the value {@code true}.
+     *     <li> For primitive types {@code float} and {@code double} the
```
jensli, line 697: I'm not sure if this is public documentation for user.

If it is then it might be helpful with a description of that the new equality semantics actually implies for NaN and the various zeros and infinities.

Having to look up `toRawBits` and figure it out will be hard for most people.

RogerRiggs: These are implementation notes; the JLS specifies the behavior of value classes, operations of values, and the details of what it means for a value to be substitutable. Its still open as to how/where the semantics would appear in javadoc.

RogerRiggs:
Start fresh after rebase.

PR #1001, opened 2024-02-08
JDK-8325344: [lworld] RedefinePrimitive.java fails javac compilation

alexmenkov:
Fixed the test after changes in library

---------
### Progress

### Issue
 * [JDK-8325344](https://bugs.openjdk.org/browse/JDK-8325344): [lworld] RedefinePrimitive.java fails javac compilation (**Bug** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/1001/head:pull/1001` \
`$ git checkout pull/1001`

`$ git checkout pull/1001` \
`$ git pull https://git.openjdk.org/valhalla.git pull/1001/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1001`

`$ git pr show -t 1001`

</details>

<a href="https://git.openjdk.org/valhalla/pull/1001.diff">https://git.openjdk.org/valhalla/pull/1001.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/1001#issuecomment-1935038412)


mlchung (commented):
This change is fine in fixing the compilation error.

We are migrating tests that use primitive value class to use the internal APIs (you can reference #991).   Can you repurpose this issue to update the test?   The test should be renamed too like `RedefineNullRestrictedValues.java` something like that.

alexmenkov:
> This change is fine in fixing the compilation error.
> 
> We are migrating tests that use primitive value class to use the internal APIs (you can reference #991). Can you repurpose this issue to update the test? The test should be renamed too like `RedefineNullRestrictedValues.java` something like that.

Thank you for the comment.
Tried to update the test to use internal API and realized that testing for NullRestricted value classes are completely different.
Also this test does not actually works with the current code (even for primitive classes).
I think it would be better to drop this test and develop test scenarios for NullRestricted value classes (will create a separate CR for this)

mlchung:
> I think it would be better to drop this test and develop test scenarios for NullRestricted value classes (will create a separate CR for this)

This is fine.   Please create a CR to add a new test `redefineClass` on value classes.

alexmenkov:
> > I think it would be better to drop this test and develop test scenarios for NullRestricted value classes (will create a separate CR for this)
> 
> This is fine. Please create a CR to add a new test `redefineClass` on value classes.

Filed https://bugs.openjdk.org/browse/JDK-8325583

PR #1017, opened 2024-02-22
Bootstrap transition to JEP 401 model

fparain:
Bootstrapping transition to JEP 401 model.
A lot of things are broken, but at least the VM can boot and execute JEP 401 class files in interpreted mode.
Still WIP.


src/hotspot/share/opto/graphKit.cpp
```diff
@@ -3668,7 +3660,8 @@ Node* GraphKit::inline_type_test(Node* obj, bool is_inline) {
 }
 
 Node* GraphKit::is_val_mirror(Node* mirror) {
-  Node* p = basic_plus_adr(mirror, java_lang_Class::secondary_mirror_offset());
+  // JDK-8325660: notion of secondary mirror / val_mirror is gone one JEP 401
+  Node* p = basic_plus_adr(mirror, (int)0 /* java_lang_Class::secondary_mirror_offset() */);
```
TobiHartmann, line 3664: This will break Escape Analysis because such an access with zero offset is unexpected:

```
#
#  Internal Error (/oracle/valhalla/open/src/hotspot/share/opto/escape.cpp:896), pid=600228, tid=600242
#  assert(_igvn->type(adr)->isa_rawptr()) failed: sanity
#
# JRE version: Java(TM) SE Runtime Environment (22.0) (fastdebug build 22-lworld4ea-2024-02-06-1232159.tobias...)
# Java VM: Java HotSpot(TM) 64-Bit Server VM (fastdebug 22-lworld4ea-2024-02-06-1232159.tobias..., compiled mode, tiered, compressed oops, compressed class ptrs, g1 gc, linux-amd64)
# Problematic frame:
# V  [libjvm.so+0xbe2ec4]  ConnectionGraph::add_objload_to_connection_graph(Node*, Unique_Node_List*)+0x284
```

Please add this to disable the check properly:

```
diff --git a/src/hotspot/share/opto/library_call.cpp b/src/hotspot/share/opto/library_call.cpp
index 92425fecf11..bd3ed61744b 100644
--- a/src/hotspot/share/opto/library_call.cpp
+++ b/src/hotspot/share/opto/library_call.cpp
@@ -4310,7 +4310,8 @@ bool LibraryCallKit::inline_native_subtype_check() {
     region->set_req(_both_ref_path, gen_subtype_check(subk, superk));
     // If superc is an inline mirror, we also need to check if superc == subc because LMyValue
     // is not a subtype of QMyValue but due to subk == superk the subtype check will pass.
-    generate_fair_guard(is_val_mirror(args[0]), prim_region);
+    // TODO JDK-8325660
+    // generate_fair_guard(is_val_mirror(args[0]), prim_region);
     // now we have a successful reference subtype check
     region->set_req(_ref_subtype_path, control());
   }
```

src/hotspot/share/opto/library_call.cpp
```diff
@@ -4115,17 +4115,15 @@ bool LibraryCallKit::inline_primitive_Class_conversion(vmIntrinsics::ID id) {
     return false;
   }
 
-  bool is_val_mirror = true;
-  ciType* tm = mirror_con->java_mirror_type(&is_val_mirror);
+  // // JDK-8325660: Code has been modified because secondary mirror are gone in JEP 401
```
TobiHartmann, line 4118:
```suggestion
  // JDK-8325660: Code has been modified because secondary mirror are gone in JEP 401
```

src/hotspot/share/opto/parse2.cpp
```diff
@@ -3503,12 +3503,6 @@ void Parse::do_one_bytecode() {
   case Bytecodes::_new:
     do_new();
     break;
-  case Bytecodes::_aconst_init:
-    do_aconst_init();
-    break;
-  case Bytecodes::_withfield:
-    do_withfield();
```
TobiHartmann, line 3510: Could you add a `// TODO JDK-8325660` to both methods so we don't forget removing them?

TobiHartmann (approved):
The ci, C1 and C2 changes look good to me. Thanks a lot for the first cleanup here, Frederic!

src/hotspot/share/opto/graphKit.cpp
fparain: Thank you for the quick fix.

src/hotspot/share/opto/parse2.cpp
fparain: do_withfield() and do_aconst_init() methods have been removed from C2 parse* code.

src/hotspot/share/opto/library_call.cpp
fparain: Fixed

fparain:
Thanks Tobias for the review and the fix.

PR #1024, opened 2024-02-27
8326707: [lworld] Remove VALUE modifier from Modifier and AccessFlag

RogerRiggs:
Remove the VALUE modifier in Modifier.java and AccessFlag.java.
Update Class.isValue() to be false for primitive, array, and interface classes and otherwise be ~ACC_IDENTITY in the modifiers.

### DIFF
--- a/src/java.base/share/classes/java/lang/reflect/Modifier.java
+++ b/src/java.base/share/classes/java/lang/reflect/Modifier.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 1996, 2022, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1996, 2024, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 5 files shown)


src/java.base/share/classes/java/lang/reflect/Modifier.java
```diff
@@ -156,18 +156,18 @@ public static boolean isVolatile(int mod) {
...
      * @apiNote {@code isValue} should only be called with the modifiers
      * of a {@linkplain Class#getModifiers() class}.
      *
      * @param   mod a set of modifiers
-     * @return {@code true} if {@code mod} includes the
-     * {@code value} modifier; {@code false} otherwise.
+     * @return {@code true} if {@code mod} does not include the
+     * {@code identity} modifier; {@code false} otherwise.
      */
     public static boolean isValue(int mod) {
```
mlchung, line 169: I think we should remove this as there is no longer such a modifier.

test/jdk/java/lang/reflect/AccessFlag/ClassAccessFlagTest.java
```diff
@@ -54,9 +54,12 @@ public static void main(String... args) {
...
             TestInterface.class,
-            TestIdentityInterface.class,
-            TestValueInterface.class,
-            ExpectedClassFlags.class,
+                TestFinalClass.class,
+                TestAbstractClass.class,
+                Foo.class,
+                StaticTestInterface.class,
+                TestMarkerAnnotation.class,
+                ExpectedClassFlags.class,
```
mlchung, line 62: nit: extra 4-space indentation.

mlchung (approved):

PR #1040, opened 2024-03-07
8327636: [lworld] Make primitive wrappers be value class if preview feature is enabled

mlchung:
As specified in JEP 401, when preview features are enabled, the following standard library classes are considered to be value classes:
```
    java.lang.Byte
    java.lang.Short
    java.lang.Integer
    java.lang.Long
    java.lang.Float
    java.lang.Double
    java.lang.Boolean
    java.lang.Character
    java.util.Optional
    java.lang.Number
    java.lang.Record
```

These value classes are loaded only if --enable-preview is set.    All valhalla tests should have preview enabled to ensure testing with the proper set of library classes.

Testing: Run `jdk_valhalla` and `hotspot_valhalla_runtime` with  `-Xint` with and without `--enable-preview`.  The result is pretty good.  [JDK-8327637](https://bugs.openjdk.org/browse/JDK-8327637) and [JDK-8327639](https://bugs.openjdk.org/browse/JDK-8327639) need further investigation.

---------
### Progress

### Issue
 * [JDK-8327636](https://bugs.openjdk.org/browse/JDK-8327636): [lworld] Make primitive wrappers be value class if preview feature is enabled (**Bug** - P3)

### Reviewers
 * [Roger Riggs](https://openjdk.org/census#rriggs) (@RogerRiggs - Committer) ⚠️ Review applies to [ff15f65e](https://git.openjdk.org/valhalla/pull/1040/files/ff15f65eb5c39b40e5c4c92b607b373fd2dbf037)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/1040/head:pull/1040` \
`$ git checkout pull/1040`

`$ git checkout pull/1040` \
`$ git pull https://git.openjdk.org/valhalla.git pull/1040/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1040`

`$ git pr show -t 1040`

</details>

<a href="https://git.openjdk.org/valhalla/pull/1040.diff">https://git.openjdk.org/valhalla/pull/1040.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/1040#issuecomment-1984704910)

### DIFF
--- a/make/modules/java.base/gensrc/GensrcValueClasses.gmk
+++ b/make/modules/java.base/gensrc/GensrcValueClasses.gmk
@@ -27,6 +27,16 @@
 # Generate the value class replacements for selected java.base source files
 
 java.base-VALUE_CLASS-REPLACEMENTS := \
+    java/lang/Byte.java \
+    java/lang/Short.java \
+    java/lang/Integer.java \
+    java/lang/Long.java \
+    java/lang/Float.java \
+    java/lang/Double.java \
+    java/lang/Boolean.java \
+    java/lang/Character.java \
+    java/lang/Number.java \
+    java/lang/Record.java \
     java/util/Optional.java \
     java/util/OptionalInt.java \
     java/util/OptionalLong.java \
--- a/make/modules/java.base/gensrc/GensrcValueClasses.gmk
+++ b/make/modules/java.base/gensrc/GensrcValueClasses.gmk
@@ -53,7 +53,6 @@ java.base-VALUE_CLASS-REPLACEMENTS := \
     java/time/YearMonth.java \
     java/time/Year.java \
     java/time/Period.java \
-    java/lang/Record.java \
     #
 
 java.base-VALUE-CLASS-FILES := \
--- a/src/java.base/share/classes/sun/invoke/util/VerifyAccess.java
+++ b/src/java.base/share/classes/sun/invoke/util/VerifyAccess.java
@@ -208,6 +208,13 @@ public class VerifyAccess {
                 return true;
             }
 
+            // exports are not setup during early VM initialization
+            if (!jdk.internal.misc.VM.isModuleSystemInited()) {
+                //  access java.base classes only
+                assert lookupModule == refModule && refModule == Object.class.getModule();
+                return true;
+            }
+
             // allow access to public types in all unconditionally exported packages
             if ((allowedModes & UNCONDITIONAL_ALLOWED) != 0) {
                 return refModule.isExported(refc.getPackageName());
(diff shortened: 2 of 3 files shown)


make/modules/java.base/gensrc/GensrcValueClasses.gmk
```diff
@@ -27,6 +27,16 @@
...
+    java/lang/Byte.java \
+    java/lang/Short.java \
+    java/lang/Integer.java \
+    java/lang/Long.java \
+    java/lang/Float.java \
+    java/lang/Double.java \
+    java/lang/Boolean.java \
+    java/lang/Character.java \
+    java/lang/Number.java \
+    java/lang/Record.java \
```
liach, line 39: Record is already included in the end of the list.

liach (commented):
https://bugs.openjdk.org/browse/JDK-8327481 can be closed as a duplicate of this.

make/modules/java.base/gensrc/GensrcValueClasses.gmk
mlchung: Good catch.   Fixed.

src/java.base/share/classes/sun/invoke/util/VerifyAccess.java
```diff
@@ -208,6 +208,13 @@ public static boolean isClassAccessible(Class<?> refc,
                 return true;
             }
 
+            // exports are not setup during early VM initialization
+            if (!jdk.internal.misc.VM.isModuleSystemInited()) {
```
liach, line 212: Should we merge this and the `lookupModule == null` early startup cases, and `assert lookupModule == refModule && refModule == Object.class.getModule()` can become `assert lookupModule == refModule && (refModule == null || refModule == Object.class.getModule())`?

mlchung: It can be simpler:

```
            // early VM startup case, java.base not defined or module system not initialized
            if (lookupModule == null || !jdk.internal.misc.VM.isModuleSystemInited()) {
                assert lookupModule == refModule;
                return true;
            }
```

liach: Can `Class::getModule` ever return `null` after module system is initialized? If not we can omit the `lookupModule == null` check.

mlchung: > Can Class::getModule ever return null after module system is initialized?

No.

`lookupModule == null` when module system is being initialized before java.base is defined.   So the null check needs to be there.

RogerRiggs (approved):
LGTM

PR #1046, opened 2024-03-14
8328010: [lworld] C1 assertion failure in LIRAssembler::patching_epilog

fparain:
Fixes in C1 instance creation code.
Compiler tests updates.


TobiHartmann (approved):
That looks good to me. Thanks for fixing!

fparain:
Thanks for the review.

PR #1063, opened 2024-03-26
8304168: [lworld] CDS tests fail with --enable-preview patched value classes

iklam:
[1] Allow `--enable-preview -XX:+EnableValhalla` to be specified when dumping a CDS archive. This allows only the patch modules that are used for the value classes. The user is still forbidden from specifying --patch-module when dumping the CDS archive.

[2] If a CDS archive was created with `--enable-preview -XX:+EnableValhalla`, then it must also be specified at run time, or else the CDS archive will be ignored.

[3] Fixed a few bugs with value types. E.g., some metadata objects need to be placed in R/W region.

[4] Tests with all CDS tests and tiers 1-4.

[5] Modify Image.gmk to be able to generate default CDS archives for `--enable-preview -XX:+EnableValhalla`, but this is not yet enabled as some non-CDS tests will fail. See [JDK-8329108](https://bugs.openjdk.org/browse/JDK-8329108)

---------
### Progress

### Issue
 * [JDK-8304168](https://bugs.openjdk.org/browse/JDK-8304168): [lworld] CDS tests fail with --enable-preview patched value classes (**Bug** - P4)

### Reviewers
 * [Mandy Chung](https://openjdk.org/census#mchung) (@mlchung - Committer) ⚠️ Review applies to [11c6d680](https://git.openjdk.org/valhalla/pull/1063/files/11c6d680bc4d9f95800d91dfed7d6b74cb997ec7)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/1063/head:pull/1063` \
`$ git checkout pull/1063`

`$ git checkout pull/1063` \
`$ git pull https://git.openjdk.org/valhalla.git pull/1063/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1063`

`$ git pr show -t 1063`

</details>

<a href="https://git.openjdk.org/valhalla/pull/1063.diff">https://git.openjdk.org/valhalla/pull/1063.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/1063#issuecomment-2020819916)


joker22008 (changes requested):
- [ ]

joker22008:
.

joker22008:
Ориро

joker22008:
Оооееу

joker22008:
> 
> 
> :wave: Welcome back iklam! A progress list of the required criteria for merging this PR into `lworld` will be added to the body of your pull request. There are additional [pull request commands](https://wiki.openjdk.java.net/display/SKARA/Pull+Request+Commands) available for use with this pull request.

Ь

mlchung (approved):
This approach is reasonable and simplest to add a new CDS archive which adds a small percentage to the image size.

RogerRiggs:
Can we drop `EnableValhalla`; I believe we're trying to remove that and use only --enable-preview.

iklam:
> Can we drop `EnableValhalla`; I believe we're trying to remove that and use only --enable-preview.

There's existing code in arguments.cpp that checks `if (enable_preview() && EnableValhalla)`. I think we can clean that up in a separate PR.

https://github.com/openjdk/valhalla/blob/a226fb4bff53a715acc83ef56187d5e174c10ed0/src/hotspot/share/runtime/arguments.cpp#L2195-L2201

MrSimms:
> Can we drop `EnableValhalla`; I believe we're trying to remove that and use only --enable-preview.

We said within the VM implementation to continue using `EnableValhalla` to easily demarcate Valhalla until we prepare for preview. Outside the VM, i.e. test invocations etc, yeah we are using `--enable-preview`

src/hotspot/share/cds/filemap.cpp
```diff
@@ -2453,6 +2454,24 @@ bool FileMapHeader::validate() {
...
+        err = "created";
+      }
+    }
+    if (err != nullptr) {
+      log_warning(cds)("This archive was %s with --enable-preview -XX:+EnableValhalla. It is "
+                         "incompatible with the current JVM setting", err);
+      return false;
+    }
+  }
+
```
fparain, line 2474: Another VM option, `InlineFieldMaxFlatSize`, can influence the field layout, and by consequence, the validity of the `InstanceKlass` from the archive. It is probably needed to encode the value of `InlineFieldMaxFlatSize` in the archive to be able to check against the current VM options. We can decide to completely reject the archive if `InlineFieldMaxFlatSize` options don't match, or we can have a per class approach, as most classes won't have flat fields, and only reject them with a field layout not compliant with the current flag value.

src/hotspot/share/cds/cdsConfig.cpp
```diff
@@ -38,8 +40,17 @@ bool CDSConfig::_is_dumping_dynamic_archive = false;
...
 
+bool CDSConfig::is_valhalla_preview() {
+  return Arguments::enable_preview() && EnableValhalla;
+}
+
+
 #if INCLUDE_CDS_JAVA_HEAP
 bool CDSConfig::is_dumping_heap() {
+  if (is_valhalla_preview()) {
+    // Not working yet -- e.g., HeapShared::ooo_hash() needs to be implemented for value oops
```
fparain, line 51: HeapShared::ooo_hash() -> HeapShared::oop_hash()

iklam: Fixed.

src/hotspot/share/cds/filemap.cpp
iklam: We have such a check already (see [JDK-8272290](https://bugs.openjdk.org/browse/JDK-8272290)]

fparain (approved):

iklam:
Thanks @mlchung @fparain for the review.
I tested again with tiers 1-4 and no new failures.

PR #1064, opened 2024-03-27
8329196: [lworld] More fixes to value object construction in C2

TobiHartmann:
Follow-up from [JDK-8325660](https://bugs.openjdk.org/browse/JDK-8325660). We can't rely on `GraphKit::replace_in_map` to reliably replace an InlineTypeNode that is initialized in the constructor in the caller map because there can be data flow merges. Instead, we need to explicitly update the node in the exit map. I added corresponding regression tests.

I also noticed that we are extensively cloning InlineTypeNodes during parsing, which often leads to an explosion of redundant nodes that then need to be removed. I added `InlineTypeNode::clone_if_required` to mitigate this, this can probably be further improved (I added corresponding ToDo's).

There are still issues with deoptimization from the constructor (`-XX:+DeoptimizeALot`). I'll address these in another fix.

I also disabled an IR rule in `TestZGCBarrierElision`. The issue will be fixed by [JDK-8329234](https://bugs.openjdk.org/browse/JDK-8329234).

Thanks,
Tobias

---------
### Progress

### Issue
 * [JDK-8329196](https://bugs.openjdk.org/browse/JDK-8329196): [lworld] More fixes to value object construction in C2 (**Bug** - P2)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/1064/head:pull/1064` \
`$ git checkout pull/1064`

`$ git checkout pull/1064` \
`$ git pull https://git.openjdk.org/valhalla.git pull/1064/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1064`

`$ git pr show -t 1064`

</details>

<a href="https://git.openjdk.org/valhalla/pull/1064.diff">https://git.openjdk.org/valhalla/pull/1064.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/1064#issuecomment-2024591650)


TobiHartmann:
Found more issues when there are nested constructors, i.e., calls from one constructor to another. Will continue to work on this after the easter break.

PR #1078, opened 2024-04-18
[lworld draft] Implementation of nullable flat fields

fparain:
First implementation of nullable flat fields for x86_64 and aarch64 with new framework to verify field layouts.


test/hotspot/jtreg/runtime/valhalla/inlinetypes/NullableFlatFieldTest.java
```diff
@@ -0,0 +1,276 @@
...
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+ /*
+ * @test NullableFlatFieldTest
+ * @library /test/lib
+ * @modules java.base/jdk.internal.vm.annotation
+ * @enablePreview
+ * @run main/othervm -XX:+EnableNullableFieldFlattening -XX:+PrintInlineLayout NullableFlatFieldTest
```
MrSimms, line 29: `PrintInlineLayout` isn't a "product flag" is it ? Should the test be debug only, or should we make `PrintInlineLayout` a product flag ? I'd consider the later for EA release.

test/hotspot/jtreg/runtime/valhalla/inlinetypes/field_layout/NullMarkersTest.java
```diff
@@ -0,0 +1,298 @@
...
+  public static void main(String[] args) throws Exception {
+    // Generate test classes
+    NullMarkersTest fat = new NullMarkersTest();
+
+
+    // Execute the test runner in charge of loading all test classes
+    ProcessBuilder pb = exec("NullMarkersTest$TestRunner");
+    OutputAnalyzer out = new OutputAnalyzer(pb.start());
+
+    Asserts.assertEquals(out.getExitValue(), 0, "Something went wrong while running the tests");
```
MrSimms, line 276: Speeds up problem solving if the output and error streams for processes is printed...there is the same issue with `PrintFieldLayout` not being product flag

test/hotspot/jtreg/runtime/valhalla/inlinetypes/NullableFlatFieldTest.java
MrSimms, line 29: Although the tests are short (just now), shouldn't all uses of `EnableNullableFieldFlattening` need `-Xint` just now ?

fparain: Fixed.

test/hotspot/jtreg/runtime/valhalla/inlinetypes/field_layout/NullMarkersTest.java
fparain: The output now contains more information to speed-up investigations. In most cases, the output is too big to be saved by jtreg with a normal configuration, but added information should indicate in which class an error has been detected (last part of the log is saved by jtreg). 
A better solution would be to not print the whole output, but just the layouts of the classes involved in the fail check. But this would require significant work and this patch is already big. I'd prefer to push that in another RFE.

test/hotspot/jtreg/runtime/valhalla/inlinetypes/NullableFlatFieldTest.java
fparain: `PrintFieldLayout` and `PrintInlineLayout` are now product diagnostic flags.

MrSimms (approved):
Good initial runtime implementation to hang the JIT implementation off of, further work for other PRs...good to go !

src/hotspot/share/interpreter/interpreterRuntime.cpp
```diff
@@ -315,6 +315,62 @@ JRT_ENTRY(void, InterpreterRuntime::read_flat_field(JavaThread* current, oopDesc
...
+JRT_ENTRY(void, InterpreterRuntime::write_nullable_flat_field(JavaThread* current, oopDesc* obj, oopDesc* value, ResolvedFieldEntry* entry))
+  assert(oopDesc::is_oop(obj), "Sanity check");
+  Handle obj_h(THREAD, obj);
+  assert(value == nullptr || oopDesc::is_oop(value), "Sanity check");
+  Handle val_h(THREAD, value);
+
+  InstanceKlass* ik = InstanceKlass::cast(obj_h()->klass());
+  int nm_offset = ik->null_marker_offsets_array()->at(entry->field_index());
+  if (val_h() == nullptr) {
+    obj_h()->byte_field_put(nm_offset, (jbyte)0);
```
DanHeidinga, line 355: Do we need a barrier here?  When we write the null marker, it is as if we wrote the null marker and then nulled the fields out (which we don't actually do) which makes me think we always need the barrier after writing the null marker to be consistent with the non-null case.

src/hotspot/share/interpreter/interpreterRuntime.cpp
```diff
@@ -315,6 +315,62 @@ JRT_ENTRY(void, InterpreterRuntime::read_flat_field(JavaThread* current, oopDesc
...
+    return;
+  }
+  InlineKlass* vk = InlineKlass::cast(val_h()->klass());
+  if (entry->has_internal_null_marker()) {
+    // The interpreter copies values with a bulk operation
+    // To avoid accidently setting the null marker to "null" during
+    // the copying, the null marker is set to non zero in the source object
```
DanHeidinga, line 362: Is this saying if the null marker is embedded in the value field layout rather than placed separately, we set it to non-null before writing it into the container?  This tripped me up a little bit reading the code but I think it makes sense

DanHeidinga: How does that work for our memory barriers then?  Do we not need a release here as well given the acquire used to read the null marker?

src/hotspot/share/classfile/fieldLayoutBuilder.cpp
```diff
@@ -338,11 +351,18 @@ bool FieldLayout::reconstruct_layout(const InstanceKlass* ik) {
       if (fs.access_flags().is_static()) continue;
       has_instance_fields = true;
       LayoutRawBlock* block;
-      if (fs.field_flags().is_null_free_inline_type()) {
+      // if (fs.field_flags().is_null_free_inline_type()) {
```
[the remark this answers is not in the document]
fparain: Old code, will be removed in next commit.

src/hotspot/cpu/aarch64/templateTable_aarch64.cpp
```diff
@@ -3340,9 +3352,10 @@ void TemplateTable::fast_storefield(TosState state)
 
   // access field
   switch (bytecode()) {
-  case Bytecodes::_fast_qputfield: //fall through
+  case Bytecodes::_fast_vputfield: //fall through
    {
-      Label is_flat, done;
+      Label is_flat, has_null_marker, done;
+      __ test_field_has_null_marker(r3, noreg /* temp */, has_null_marker);
       __ null_check(r0);
```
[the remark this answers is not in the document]
DanHeidinga: You're right.  Thanks for checking.

src/hotspot/share/interpreter/interpreterRuntime.cpp
fparain: In this case, there is a single write from the current thread, so there's no visibility ordering issue like with the non-null case. A concurrently reading thread will check (or has already read) the null marker, and this would be no different than reading a reference field being set to null. So I don't think we need a barrier here, but if you really think there's a potential issue, I can add one.

src/hotspot/share/interpreter/interpreterRuntime.cpp
fparain: This particular case cannot currently happen with JEP 401 model implementation (nullable flat field are limited to single field values until we have atomic updates supported). But it involves the tearing issue the interpreter has in some cases. So, even adding a barrier would not make the code correct. I'd prefer to address the issue of this particular code when fixing the wider interpreter tearing issue.

src/hotspot/share/classfile/fieldLayoutBuilder.cpp
```diff
@@ -734,7 +793,12 @@ void FieldLayoutBuilder::inline_class_field_sorting(TRAPS) {
       break;
     case T_OBJECT:
     case T_ARRAY:
-      if (!fieldinfo.field_flags().is_null_free_inline_type()) {
+    {
+      bool field_is_known_value_class =  !fieldinfo.field_flags().is_injected() && _inline_type_field_klasses != nullptr && _inline_type_field_klasses->at(fieldinfo.index()) != nullptr;
+      bool value_has_oops = field_is_known_value_class ? _inline_type_field_klasses->at(fieldinfo.index())->nonstatic_oop_count() > 0 : true;
+      bool is_candidate_for_flattening = fieldinfo.field_flags().is_null_free_inline_type() || (EnableNullableFieldFlattening && field_is_known_value_class && !value_has_oops);
+      // if (!fieldinfo.field_flags().is_null_free_inline_type()) {
```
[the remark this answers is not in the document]
fparain: I don't like the duplication either, and having a helper method would make the code easier to read. However, this is a section of code under big transformations, with nullable flat field being added, then next the atomic fields, then the nullable atomic fields. Before refactoring, I'd prefer to wait until all the cases are covered, so we know exactly all the parameters the helper method will require, and if the two cases are still identical or if some divergences appeared.

src/hotspot/cpu/aarch64/templateTable_aarch64.cpp
```diff
@@ -3340,9 +3352,10 @@ void TemplateTable::fast_storefield(TosState state)
 
   // access field
   switch (bytecode()) {
-  case Bytecodes::_fast_qputfield: //fall through
+  case Bytecodes::_fast_vputfield: //fall through
```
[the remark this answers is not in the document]
fparain: Fixed in next commit.

src/hotspot/share/interpreter/interpreterRuntime.cpp
DanHeidinga: I think you're right here and we don't need the `release_byte_field_put` here.  Just as we're happy for reads/writes of null references to slowly make their way to other threads, a slow discovery of the null bit is fine given the other value fields are stable before the bit flips.

The barrier you have in place is sufficient for that constraint.

DanHeidinga (approved):
Thanks for the clarifications.  This looks good to me.

fparain:
Mr Simms, Dan,
Thank you for your reviews.
Fred

PR #1079, opened 2024-04-19
8321734: [lworld] C2: flat_in_array type is not handled correctly leading to crashes

chhagedorn:
There are multiple issues around types being `flat_in_array`:

#### 1. `SubTypeCheckNode` is not folded while `CheckCastPP` is correctly replaced by top due to an impossible type:
Test: `testSubTypeCheckNotFoldedParsingAbstractClass():` 

We have the following:
- `PUnique` is the unique concrete value sub class of the abstract value `AUnique` class
- `Object[] arr` -> elements could be an inline type that is flat

We apply loop unswitching such that we have a loop where the access `arr[i]` is with a flat inline type (i.e. marked as flat-in-array) and one version without. In the flat-in-array loop, we have the following sub type check for the `checkcast` for `(AUnique)arr[i]`:
```
AUnique [maybe-flat-in-array] <: Object [flat-in-array]
```
Since we do not know if `AUnique` is flat-in-array, we cannot fold this check. However, the corresponding `CheckCastPP` for this sub type check uses the unique concrete sub class `PUnique` which is known to be *not*-flat-in-array with `FlatArrayElementMaxSize=4`. Since a not-flat-in-array type cannot be a sub type of a flat-in-array type (determined during class loading), the `CheckCastPP` is replaced by top. But the `SubTypeCheck` is not folded. 

This was found in Valhalla and fixed in mainline with JDK-8328480. I've added the Valhalla specific test and merged the fix from mainline into this patch.

#### 2. Wrong `not_flat_in_array` default value
Test: `testSubTypeCheck()` and `testCmpP()`:
Arrays are always normal objects (i.e. not inline types). They are therefore never flat-in-array and always not-flat-in-array. While they are correctly never marked as flat-in-array, they are currently not marked as *not*-flat-in-array and thus have the implicit state *maybe*-flat-in-array:
https://github.com/openjdk/valhalla/blob/e01ec832189453cc302c7ca8915e69bb63a3d4b1/src/hotspot/share/opto/type.hpp#L1096

This becomes a problem for `SubTypeCheck` and `CmpP` nodes which cannot determine that a flat-in-array klass and an array klass are unrelated. But the type system itself correctly treat arrays as *not*-flat-in-array when doing a meet operation. As a result, data is folded while control is not.

This is fixed by changing the flat-in-array tri-state: As a default, a type is always not-flat-in-array. Inline types then override this property accordingly.

#### 3. Type system wrongly treats "maybe-be-flat-in-array <: flat-in-array" as top
Test: `testUnswitchingAbstractClass()`
 "maybe-be-flat-in-array <: flat-in-array" should not be treated as top. We could have cases where we know more about the super type while the sub type does not. In this case, the sub type should take over the flat-in-array property. We therefore have the following decision matrix for sub und super types:
https://github.com/openjdk/valhalla/blob/beb128981325b71c6891048d98282df08dddf329/src/hotspot/share/opto/type.cpp#L4585-L4592

Thanks,
Christian

---------
### Progress

### Issue
 * [JDK-8321734](https://bugs.openjdk.org/browse/JDK-8321734): [lworld] C2: flat_in_array type is not handled correctly leading to crashes (**Bug** - P3)

### Reviewers
 * [Tobias Hartmann](https://openjdk.org/census#thartmann) (@TobiHartmann - Committer) ⚠️ Review applies to [8501d19f](https://git.openjdk.org/valhalla/pull/1079/files/8501d19fe3dc244ac41ada228b417f171c50ac32)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/1079/head:pull/1079` \
`$ git checkout pull/1079`

`$ git checkout pull/1079` \
`$ git pull https://git.openjdk.org/valhalla.git pull/1079/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1079`

`$ git pr show -t 1079`

</details>

<a href="https://git.openjdk.org/valhalla/pull/1079.diff">https://git.openjdk.org/valhalla/pull/1079.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/1079#issuecomment-2066386286)

### DIFF
--- a/src/hotspot/share/opto/graphKit.cpp
+++ b/src/hotspot/share/opto/graphKit.cpp
@@ -3446,8 +3446,9 @@ Node* GraphKit::gen_instanceof(Node* obj, Node* superklass, bool safe_for_replac
 // uncommon trap or exception is thrown.
 Node* GraphKit::gen_checkcast(Node *obj, Node* superklass, Node* *failure_control, bool null_free) {
   kill_dead_locals();           // Benefit all the uncommon traps
-  const TypeKlassPtr *tk = _gvn.type(superklass)->is_klassptr()->try_improve();
-  const TypeOopPtr *toop = tk->cast_to_exactness(false)->as_instance_type();
+  const TypeKlassPtr* klass_ptr_type = _gvn.type(superklass)->is_klassptr();
+  const TypeKlassPtr* improved_klass_ptr_type = klass_ptr_type->try_improve();
+  const TypeOopPtr* toop = improved_klass_ptr_type->cast_to_exactness(false)->as_instance_type();
   bool safe_for_replace = (failure_control == nullptr);
   assert(!null_free || toop->is_inlinetypeptr(), "must be an inline type pointer");
 
@@ -3457,7 +3458,7 @@ Node* GraphKit::gen_checkcast(Node *obj, Node* superklass, Node* *failure_contro
   // because if the test is going to turn into zero code, we don't
   // want a residual null check left around.  (Causes a slowdown,
   // for example, in some objArray manipulations, such as a[i]=a[j].)
-  if (tk->singleton()) {
+  if (improved_klass_ptr_type->singleton()) {
     const TypeKlassPtr* kptr = nullptr;
     const Type* t = _gvn.type(obj);
     if (t->isa_oop_ptr()) {
@@ -3467,7 +3468,7 @@ Node* GraphKit::gen_checkcast(Node *obj, Node* superklass, Node* *failure_contro
       kptr = TypeInstKlassPtr::make(TypePtr::NotNull, vk, Type::Offset(0));
     }
     if (kptr != nullptr) {
-      switch (C->static_subtype_check(tk, kptr)) {
+      switch (C->static_subtype_check(improved_klass_ptr_type, kptr)) {
       case Compile::SSC_always_true:
         // If we know the type check always succeed then we don't use
         // the profiling data at this bytecode. Don't lose it, feed it
@@ -3561,7 +3562,7 @@ Node* GraphKit::gen_checkcast(Node *obj, Node* superklass, Node* *failure_contro
   }
 
   Node* cast_obj = nullptr;
-  if (tk->klass_is_exact()) {
+  if (improved_klass_ptr_type->klass_is_exact()) {
     // The following optimization tries to statically cast the speculative type of the object
     // (for example obtained during profiling) to the type of the superklass and then do a
     // dynamic check that the type of the object is what we expect. To work correctly
@@ -3571,7 +3572,7 @@ Node* GraphKit::gen_checkcast(Node *obj, Node* superklass, Node* *failure_contro
     // a speculative type use it to perform an exact cast.
     ciKlass* spec_obj_type = obj_type->speculative_type();
     if (spec_obj_type != nullptr || data != nullptr) {
-      cast_obj = maybe_cast_profiled_receiver(not_null_obj, tk, spec_obj_type, safe_for_replace);
+      cast_obj = maybe_cast_profiled_receiver(not_null_obj, improved_klass_ptr_type, spec_obj_type, safe_for_replace);
       if (cast_obj != nullptr) {
         if (failure_control != nullptr) // failure is now impossible
           (*failure_control) = top();
@@ -3583,8 +3584,14 @@ Node* GraphKit::gen_checkcast(Node *obj, Node* superklass, Node* *failure_contro
 
   if (cast_obj == nullptr) {
     // Generate the subtype check
-    Node* not_subtype_ctrl = gen_subtype_check(not_null_obj, superklass);
-
+    Node* improved_superklass = superklass;
+    if (improved_klass_ptr_type != klass_ptr_type && improved_klass_ptr_type->singleton()) {
+      // Only improve the super class for constants which allows subsequent sub type checks to possibly be commoned up.
+      // The other non-constant cases cannot be improved with a cast node here since they could be folded to top.
+      // Additionally, the benefit would only be minor in non-constant cases.
+      improved_superklass = makecon(improved_klass_ptr_type);
+    }
+    Node* not_subtype_ctrl = gen_subtype_check(not_null_obj, improved_superklass);
     // Plug in success path into the merge
     cast_obj = _gvn.transform(new CheckCastPPNode(control(), not_null_obj, toop));
     // Failure path ends in uncommon trap (or may be dead - failure impossible)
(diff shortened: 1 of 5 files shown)


src/hotspot/share/opto/type.cpp
```diff
@@ -992,6 +992,9 @@ class VerifyMeet {
 void Type::check_symmetrical(const Type* t, const Type* mt, const VerifyMeet& verify) const {
   Compile* C = Compile::current();
   const Type* mt2 = verify.meet(t, this);
+
+  // Verify that:
+  //      this meet t == t meet this
```
chhagedorn, line 997: Found these comments useful while verifying the type system changes.

src/hotspot/share/opto/graphKit.cpp
```diff
@@ -3446,8 +3446,9 @@ Node* GraphKit::gen_instanceof(Node* obj, Node* superklass, bool safe_for_replac
 // uncommon trap or exception is thrown.
 Node* GraphKit::gen_checkcast(Node *obj, Node* superklass, Node* *failure_control, bool null_free) {
   kill_dead_locals();           // Benefit all the uncommon traps
-  const TypeKlassPtr *tk = _gvn.type(superklass)->is_klassptr()->try_improve();
-  const TypeOopPtr *toop = tk->cast_to_exactness(false)->as_instance_type();
+  const TypeKlassPtr* klass_ptr_type = _gvn.type(superklass)->is_klassptr();
+  const TypeKlassPtr* improved_klass_ptr_type = klass_ptr_type->try_improve();
+  const TypeOopPtr* toop = improved_klass_ptr_type->cast_to_exactness(false)->as_instance_type();
```
chhagedorn, line 3451: Changes in this file taken from JDK-8328480.

src/hotspot/share/opto/type.cpp
```diff
@@ -4572,39 +4582,47 @@ template<class T> TypePtr::MeetResult TypePtr::meet_instptr(PTR& ptr, const Type
...
-      this_flat_in_array = other_flat_in_array = flat_array;
     } else if (above_centerline(this_ptr) && !above_centerline(other_ptr)) {
       this_type = other_type; // tinst is down; keep down man
       this_xk = other_xk;
-      this_flat_in_array = other_flat_in_array;
+      flat_in_array = other_flat_in_array;
     } else if (above_centerline(other_ptr) && !above_centerline(this_ptr)) {
       other_type = this_type; // this is down; keep down man
-      other_xk = this_xk;
-      other_flat_in_array = this_flat_in_array;
```
chhagedorn, line 4604: Assignments to `other*` are not used anymore later and are thus removed.

src/hotspot/share/opto/type.cpp
```diff
@@ -4572,39 +4582,47 @@ template<class T> TypePtr::MeetResult TypePtr::meet_instptr(PTR& ptr, const Type
...
+
   const T* subtype = nullptr;
   bool subtype_exact = false;
-  bool flat_array = false;
+  bool flat_in_array = false;
   if (this_type->is_same_java_type_as(other_type)) {
     subtype = this_type;
     subtype_exact = below_centerline(ptr) ? (this_xk && other_xk) : (this_xk || other_xk);
-    flat_array = below_centerline(ptr) ? (this_flat_in_array && other_flat_in_array) : (this_flat_in_array || other_flat_in_array);
-  } else if (!other_xk && this_type->is_meet_subtype_of(other_type) && (!other_flat_in_array || this_flat_in_array)) {
```
chhagedorn, line 4582: Extracted `is_meet_subtype_of()` + flat-in-array checks to separate methods to simplify the condition.

src/hotspot/share/opto/type.cpp
```diff
@@ -4521,10 +4530,11 @@ template<class T> TypePtr::MeetResult TypePtr::meet_instptr(PTR& ptr, const Type
...
   ciKlass* this_klass = this_type->klass();
   ciKlass* other_klass = other_type->klass();
-  bool this_flat_in_array = this_type->flat_in_array();
-  bool other_flat_in_array = other_type->flat_in_array();
-  bool this_flat_in_array_orig = this_flat_in_array;
-  bool other_flat_in_array_orig = other_flat_in_array;
+  const bool this_flat_in_array = this_type->flat_in_array();
+  const bool other_flat_in_array = other_type->flat_in_array();
+  const bool this_not_flat_in_array = this_type->not_flat_in_array();
+  const bool other_not_flat_in_array = other_type->not_flat_in_array();
```
chhagedorn, line 4536: `this_flat_in_array` and `flat_array` are both used to hold intermediate and final results about whether the type is going to be flat-in-array or not. This was somewhat hard to follow what was going on.

Changed this into: `this_flat_in_array` for the flat-in-array state of `this_type` and `flat_in_array` to hold the intermediate and eventually the final result assigned to `res_flat_in_array`.

test/hotspot/jtreg/compiler/valhalla/inlinetypes/TestFlatInArraysFolding.java
```diff
@@ -0,0 +1,239 @@
...
+ * questions.
+ */
+
+/*
+ * @test id=serialgc
+ * @bug 8321734
+ * @requires vm.gc.Serial
+ * @summary Test that CmpPNode::sub and SubTypeCheckNode::sub correctly identify unrelated classes based on the flat
+ *          in array property of the types. Additionally check that the type system properly handles the case of a
+ *         super class being flat in array while the sub klass could be flat in array.
```
TobiHartmann, line 30:
```suggestion
 *          super class being flat in array while the sub klass could be flat in array.
```

test/hotspot/jtreg/compiler/valhalla/inlinetypes/TestFlatInArraysFolding.java
```diff
@@ -0,0 +1,239 @@
...
+ *          java.base/jdk.internal.vm.annotation
+ * @run main compiler.valhalla.inlinetypes.TestFlatInArraysFolding serial
+ */
+
+/*
+ * @test
+ * @bug 8321734
+ * @summary Test that CmpPNode::sub and SubTypeCheckNode::sub correctly identify unrelated classes based on the flat
+ *          in array property of the types. Additionally check that the type system properly handles the case of a
+ *         super class being flat in array while the sub klass could be flat in array.
```
TobiHartmann, line 43:
```suggestion
 *          super class being flat in array while the sub klass could be flat in array.
```

src/hotspot/share/opto/type.cpp
TobiHartmann: Nice!

TobiHartmann (approved):
Thanks for the thorough investigation, Christian. Great job! The fixes look good to me.

chhagedorn:
Thanks Tobias for your review and the offline discussions!

/integrate

PR #1096, opened 2024-05-06
8330021: AArch64: Add backend support for FP16 add operation

Bhavana-Kilambi:
This commit [1] adds initial support for FP16 operations and adds backend support for FP16 add operation for X86. This task adds backend support for scalar and vector FP16 add operation on aarch64.

[1] https://github.com/openjdk/valhalla/commit/f03fb4e4ee4d59ed692d0c26ddce260511f544e7#diff-a799ce8da7f3062bb3699beb65aae504840c649942032e325c2a50f88a2869ad

---------
### Progress

### Issue
 * [JDK-8330021](https://bugs.openjdk.org/browse/JDK-8330021): AArch64: Add backend support for FP16 add operation (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/1096/head:pull/1096` \
`$ git checkout pull/1096`

`$ git checkout pull/1096` \
`$ git pull https://git.openjdk.org/valhalla.git pull/1096/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1096`

`$ git pr show -t 1096`

</details>

<a href="https://git.openjdk.org/valhalla/pull/1096.diff">https://git.openjdk.org/valhalla/pull/1096.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/1096#issuecomment-2097761627)


test/hotspot/jtreg/compiler/intrinsics/float16/TestFP16ScalarAdd.java
```diff
@@ -74,4 +90,45 @@ public void test2() {
         Float16 hf4 = Float16.valueOf((short)17408);
         res = Float16.sum(Float16.sum(Float16.sum(Float16.sum(hf0, hf1), hf2), hf3), hf4).float16ToRawShortBits();
     }
+
+    // Test for optimizing sequence - "dst (ReinterpretS2HF (ConvF2HF src)" in the backend to a single convert
+    // operation and do away with redundant moves
+    @Test
+    @IR(applyIfCPUFeature = {"avx512_fp16" , "true"}, counts = {IRNode.CONVF2HFANDS2HF, " >= 1"})
+    @IR(applyIfCPUFeatureAnd = {"fphp", "true", "asimdhp", "true"}, counts = {IRNode.CONVF2HFANDS2HF, " >= 1"})
+    public void test3() {
```
Bhavana-Kilambi, line 99: "test3" and "test4" are not really related to FP16 add operation and I can move them to another file if needed.

test/hotspot/gtest/aarch64/aarch64-asmtest.py
```diff
@@ -16,8 +16,8 @@
      = [0x1, 0x38, 0x7e, 0xff, 0x1fc, 0x1ff, 0x3f0,
         0x7e0, 0xfc0, 0x1f80, 0x3ff0, 0x7e00, 0x7e00,
         0x8000, 0x81ff, 0xc1ff, 0xc003, 0xc7ff, 0xdfff,
-        0xe03f, 0xe10f, 0xe1ff, 0xf801, 0xfc00, 0xfc07,
-        0xff03, 0xfffe]
+        0xe03f, 0xe1ff, 0xf801, 0xfc00, 0xfc07, 0xff03,
```
Bhavana-Kilambi, line 19: Removed "0xe10f" from the list as it is not a valid 16-imemdiate. 
Source : https://gist.github.com/dinfuehr/51a01ac58c0b23e4de9aac313ed6a06a

Bhavana-Kilambi:
Can I please ask for some reviews for this PR? Thank you in advance!

Bhavana-Kilambi:
Hi @jatin-bhateja @TobiHartmann Can I get some reviews please? Atleast for the c2 specific part. Thank you !

Bhavana-Kilambi:
Hello @jatin-bhateja , I noticed that the vector test - `test/hotspot/jtreg/compiler/vectorization/TestFloat16VectorSum.java` fails on aarch64/x86 machines that support the Float16 feature. Have you noticed this on any of your `avx512_fp16` supporting machines too? I suspect this failure does not show up on github actions for this PR or others because they might be running the tests on non-fp16 supporting hardware.

I digged a bit deeper and found out that the Float16 `add` operation was not getting unrolled in the first place because of which it doesn't get vectorized eventually. I have an internal implementation for other binary FP16 operations (like subtract, multiply, divide, min and max) and none of these operations get vectorized either. After some debugging, I noticed that the inner loop body -> `_body` at this point of code execution - `https://github.com/openjdk/valhalla/blob/404a4fae6fc099cd93cc73d57e29b08b5418adb9/src/hotspot/share/opto/loopTransform.cpp#L1069` contains 188 nodes for Half float add compared to 29 nodes for Float32 add and 31 nodes for FP16 add before the merge on June 9th. I think due to so many nodes in the loop, the unrolling policy decides not to unroll it. 
I printed the extra nodes that the loop (for FP16 add operation) contains and I can see nodes like - `DecodeNKlass`, `DecodeN`, `LoadNKlass`, `LoadN`, `CmpP`, `StoreP`, `StoreCM`, `MergeMem`, `EncodeP`, `CastPP` and other pointer related IR nodes and maybe some nodes related to decoding compressed class pointers. Although none of these exist in the FP32 code or the FP16 code before the merge. Do you have any idea about this failure or the extra nodes in the loop?

Bhavana-Kilambi:
Thanks @jatin-bhateja. I seem to have missed this RFE. I get it now.

Bhavana-Kilambi:
Hi @jatin-bhateja , I have rebased my patch to include your latest merge regarding flat array layout for FP16 operations. Could you please help review this patch? I can then post the patch that implements support for some more FP16 binary operations which is stalled until this patch is merged. Thanks!

test/hotspot/jtreg/compiler/vectorization/TestFloat16VectorSum.java
```diff
@@ -56,7 +56,8 @@ public TestFloat16VectorSum() {
 
     @Test
     @Warmup(10000)
-    @IR(applyIfCPUFeature = {"avx512_fp16" , "true"}, counts = {IRNode.ADD_VHF, " >= 1"})
+    @IR(applyIfCPUFeatureOr = {"avx512_fp16" , "true", "sve", "true"}, counts = {IRNode.ADD_VHF, " >= 1"})
+    @IR(applyIfCPUFeatureAnd = {"fphp", "true", "asimdhp", "true"}, counts = {IRNode.ADD_VHF, " >= 1"})
```
jatin-bhateja, line 60: Same as above,  multiple IR rules are enforced in cascaded fashion.

test/hotspot/jtreg/compiler/intrinsics/float16/TestFP16ScalarAdd.java
```diff
@@ -57,6 +55,7 @@ public TestFP16ScalarAdd() {
 
     @Test
     @IR(applyIfCPUFeature = {"avx512_fp16", "true"}, counts = {IRNode.ADD_HF, "> 0", IRNode.REINTERPRET_S2HF, "> 0", IRNode.REINTERPRET_HF2S, "> 0"})
+    @IR(applyIfCPUFeatureAnd = {"fphp", "true", "asimdhp", "true"}, counts = {IRNode.ADD_HF, "> 0", IRNode.REINTERPRET_S2HF, "> 0", IRNode.REINTERPRET_HF2S, "> 0"})
```
jatin-bhateja, line 58: Please have one IR rules with unified feature list for x86 and aarach64. 
Not sure, but asimdhp is only needed for SIMD instructions I guess.

test/hotspot/jtreg/compiler/intrinsics/float16/TestFP16ScalarAdd.java
```diff
@@ -67,6 +66,7 @@ public void test1() {
 
     @Test
     @IR(applyIfCPUFeature = {"avx512_fp16", "true"}, failOn = {IRNode.ADD_HF, IRNode.REINTERPRET_S2HF, IRNode.REINTERPRET_HF2S})
+    @IR(applyIfCPUFeatureAnd = {"fphp", "true", "asimdhp", "true"}, failOn = {IRNode.ADD_HF, IRNode.REINTERPRET_S2HF, IRNode.REINTERPRET_HF2S})
```
jatin-bhateja, line 69: Same as above.

jatin-bhateja (commented):
Hi @Bhavana-Kilambi ,  I have few comments on non-AARCH64 specific code changes.

test/hotspot/jtreg/compiler/intrinsics/float16/TestFP16ScalarAdd.java
Bhavana-Kilambi: Thanks @jatin-bhateja for your comments.
On Neon (non-SVE) machines or when we are testing scalar FP16 operations, Float16 is supported only if both `fphp` and `asimdhp` features are supported. So if we have to write it as a single rule, it needs to be something like - `avx512_fp16 == true || (fphp == true && asimdhp == true)` and I wasn't sure if this can be done in single IR rule here. Where ever I could write it as a single rule, I have done that for ex. clubbing avx_512_fp16 and sve in the same IR rule for `TestFloat16VectorSum.java`.

jatin-bhateja: I see, only way i can think of circumventing it currently is by creating a different test for AARCH16 as done for [some other tests](https://github.com/openjdk/jdk/tree/master/test/hotspot/jtreg/compiler/vectorapi/reshape)
 * @requires vm.cpu.features ~= ".*avx512bw.*"
 * @requires vm.cpu.features ~= ".*sve.*"

jatin-bhateja: Current rules looks ok to me, I just wrongly read the results on x86 where one of the IR rules failed, and that is bound to happene since these features are AARCH64 specific.

jatin-bhateja:
Non target specific part looks ok to me.

test/hotspot/jtreg/compiler/intrinsics/float16/TestFP16ScalarAdd.java
Bhavana-Kilambi: By "IR rules failed" you mean IR rule was not applied due to "feature constraints not met" right (for the second IR rule)? Hope the test did not fail. It should not ideally.

jatin-bhateja: Some rules were not enforced and showed up in debug log, Test passed 👍

src/hotspot/share/opto/vectornode.cpp
```diff
@@ -590,7 +590,9 @@ bool VectorNode::is_rotate_opcode(int opc) {
 bool VectorNode::is_float16_node(int opc) {
   switch (opc) {
   case Op_AddHF:
+  case Op_ConvF2HF:
   case Op_ReinterpretS2HF:
+  case Op_ReinterpretHF2S:
```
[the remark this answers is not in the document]
Bhavana-Kilambi: I actually stumbled upon a test failure when I was testing this pattern - `dst (ConvHF2F (ReinterpretHF2S src))` using this test case - 
 ```
// fin[] -> Float16 array
// flout[] -> Float array
public static void fp16Add() {
        for (int i = 0; i < SIZE; i++) {
            flout[i] = Float16.sum(fin[i], fin[i]).floatValue();
}
```

and with a debug build the test fails with the following error -
#
#
#  Internal Error (/home/bhakil01/valhalla_gerrit/valhalla/src/hotspot/share/opto/vectornode.cpp:1423), pid=2559308, tid=2559322
#  Error: assert(bt == T_SHORT) failed
#
# JRE version: OpenJDK Runtime Environment (23.0) (fastdebug build 23-internal-adhoc.bhakil01.valhalla)
# Java VM: OpenJDK 64-Bit Server VM (fastdebug 23-internal-adhoc.bhakil01.valhalla, mixed mode, compressed oops, compressed class ptrs, g1 gc, linux-aarch64)
# Problematic frame:
# V  [libjvm.so+0x18d00d8]  VectorCastNode::opcode(int, BasicType, bool)+0xf4
#

This happens because during auto-vectorization, the container type of `ReinterpretHF2S` is computed as `T_INT` and not `T_SHORT` while the input to ConvHF2F is supposed to be T_SHORT which is why the assertion at  https://github.com/openjdk/valhalla/blob/779d77f04810c137bb6a3374608b80cbe5cc9ef3/src/hotspot/share/opto/vectornode.cpp#L1422

The container type for `ReinterpretHF2S` is computed as `TypeInt::INT` here - https://github.com/openjdk/valhalla/blob/779d77f04810c137bb6a3374608b80cbe5cc9ef3/src/hotspot/share/opto/superword.cpp#L3352, which is propagating down to the VectorCastNode and leading to the assertion failure.

So I have included this node in `is_float16_node()` to return `TypeInt::SHORT` for `ReinterpretHF2S`.

If you feel this is not the right place for this node, maybe I can do something like - 

```
if (VectorNode::is_float16_node(n->Opcode()) || Op_ReinterpretHF2S) {
    return TypeInt::SHORT;
  }
```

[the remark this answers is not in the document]
Comment (jatin-bhateja) [reply src/hotspot/share/opto/vectornode.cpp:595]: Thanks for the explanation, yes this is a known issue,  we hit similar issue  earlier and [JDK-8333890](https://bugs.openjdk.org/browse/JDK-8333890?filter=-3) tracks it. I agree with above fix to move ReinterpretHF2S out of is_float16_node. 

Kindly add a test for the same.

**Comment by Bhavana-Kilambi** [comment]
Hi @jatin-bhateja , I have uploaded a new patch addressing your comments. Can you please review? Thanks!

File: test/hotspot/jtreg/compiler/vectorization/TestFP16VectorReinterpretConv.java [hunk 7]
```diff
@@ -0,0 +1,83 @@
+/*
+ * Copyright (c) 2022, 2024, Arm Limited. All rights reserved.
```
jatin-bhateja, line 2:
```suggestion
 * Copyright (c) 2024, Arm Limited. All rights reserved.
```

test/hotspot/jtreg/compiler/vectorization/TestFP16VectorReinterpretConv.java
jatin-bhateja, line 1: We have another test in the same directory with TestFloat16* nomenclature, I think we should pick one naming scheme.

test/hotspot/jtreg/compiler/vectorization/TestFP16VectorReinterpretConv.java
```diff
@@ -0,0 +1,83 @@
...
+    @IR(counts = {IRNode.VECTOR_CAST_HF2F, " >= 1", IRNode.VECTOR_REINTERPRET, " >= 1"},
+        applyIfCPUFeatureAnd = {"fphp", "true", "asimdhp", "true"})
+    public void test() {
+        for (int i = 0; i < LEN; i++) {
+            flout[i] = Float16.sum(fin[i], fin[i]).floatValue();
+        }
+        checkResult();
+    }
+
+    public void checkResult() {
```
jatin-bhateja, line 75:
```suggestion
    }

    @Check(test="test")
    public void checkResult() {
```
Should suffice since fin and flout are fields and accessible to all instance methods. We can also lift this limitation from https://github.com/openjdk/valhalla/pull/1096/files#diff-a4b825d43ecb78aad4d8f04965592dfdb9b56ac3da88de6d4117f39c74efc2ae

jatin-bhateja (commented):
Overall non-architecture part looks good to me adding some minor comments.

test/hotspot/jtreg/compiler/vectorization/TestFP16VectorReinterpretConv.java
Bhavana-Kilambi: I thought so too but was just trying to use a test name as small as possible as it was already geting bigger. But I don't mind replacing it with *Float16*. Will do that. thanks

jatin-bhateja (approved):
Looks good to me. Thanks for also fixing the vectorization issue along with the patch.

Bhavana-Kilambi:
@jatin-bhateja thank you. Hope it's ok for me to integrate? The aarch64 part should be ok as it's gone through internal code review in Arm anyway.

PR #1104, opened 2024-05-09
8331606: [lworld] javac should consider some library classes as value classes

vicente-romero-oracle:
For some library classes two compiled versions are produced, one is the legacy class file and the other one is a class file representing a value class. The version representing a value class is stored in a separate jar. During compilation of classes using one of these library classes javac loads the vanilla version. But if preview features are enabled, javac should act as if the loaded class was the one representing the value class. So basically javac's class loader needs to massage the flags in the loaded class so that the rest of the compiler pipeline can handle it as a value class. This PR is proposing an annotation to mark these classes so that they can be identified by javac without having to keep an internal list of classes that will need to be maintained,

TIA

---------
### Progress

### Issue
 * [JDK-8331606](https://bugs.openjdk.org/browse/JDK-8331606): [lworld] javac should consider some library classes as value classes (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/1104/head:pull/1104` \
`$ git checkout pull/1104`

`$ git checkout pull/1104` \
`$ git pull https://git.openjdk.org/valhalla.git pull/1104/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1104`

`$ git pr show -t 1104`

</details>

<a href="https://git.openjdk.org/valhalla/pull/1104.diff">https://git.openjdk.org/valhalla/pull/1104.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/1104#issuecomment-2101725097)

### DIFF
--- a/src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/ClassReader.java
+++ b/src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/ClassReader.java
@@ -1523,6 +1523,12 @@ public class ClassReader {
             } else if (proxy.type.tsym.flatName() == syms.valueBasedInternalType.tsym.flatName()) {
                 Assert.check(sym.kind == TYP);
                 sym.flags_field |= VALUE_BASED;
+            } else if (proxy.type.tsym.flatName() == syms.migratedValueClassInternalType.tsym.flatName()) {
+                Assert.check(sym.kind == TYP);
+                sym.flags_field |= MIGRATED_VALUE_CLASS;
+                if (needsValueFlag(sym, sym.flags_field)) {
+                    sym.flags_field |= VALUE_CLASS;
+                }
             } else if (proxy.type.tsym.flatName() == syms.restrictedType.tsym.flatName()) {
                 Assert.check(sym.kind == MTH);
                 sym.flags_field |= RESTRICTED;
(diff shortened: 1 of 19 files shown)


src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/ClassReader.java
```diff
@@ -3159,21 +3170,35 @@ long adjustMethodFlags(long flags) {
...
+    long adjustClassFlags(ClassSymbol c, long flags) {
         if ((flags & ACC_MODULE) != 0) {
             flags &= ~ACC_MODULE;
             flags |= MODULE;
         }
-        if ((flags & ACC_IDENTITY) != 0 || (majorVersion < V66.major && (flags & INTERFACE) == 0)) {
+        if (((flags & ACC_IDENTITY) != 0 && !isMigratedValueClass(flags)) || (majorVersion < V66.major && (flags & INTERFACE) == 0)) {
             flags |= IDENTITY_TYPE;
-        } else if ((flags & INTERFACE) == 0 && allowValueClasses && previewClassFile && majorVersion >= V66.major) {
+        } else if (needsValueFlag(c, flags)) {
```
mcimadamore, line 3180: isn't this redundant? E.g. isn't VALUE_CLASS set as soon as you see the annotation?

test/langtools/tools/javac/valhalla/value-objects/ValueObjectCompilationTests.java
```diff
@@ -308,6 +308,32 @@ void bar(Object o) {
                     }
                 }
                 """);
+        assertFail("compiler.err.type.found.req",
```
mcimadamore, line 311: Might be good to try the full set of options:
* --source 23, --enable-preview
* --source 23
* --release 23, --enable-preview
* --release 23

(since when release is specified, the code path is different)

src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/ClassReader.java
vicente-romero-oracle: no we need this path for the cases when we load a `real` value class

mcimadamore (approved):

vicente-romero-oracle:
thanks for the comments!

PR #1127, opened 2024-06-14
8333889: [lworld] C2: Hoisting speculative array access type check wrongly moves array access before its range check

chhagedorn:
#### Speculative Inline Type Checks for `aaload`

In `Parse::array_addressing()`, we emit speculative inline type checks (about non-flatness and non-null-freeness) with traps based on profiling information. These checks are emitted directly after the `RangeCheck`. This will lead to the situation that the `CastII` and actual `LoadN` nodes are disconnected from the `RangeCheck` true projection:

image

#### Known Problem when Load is Disconnected from Range Check

This disconnection of the load from its range check is a known general problem and could lead to the situation where the actual load ends up before the range check due to some optimizations. As a result, we could crash due to an unprotected out-of-bounds access when the load is actually scheduled before the range check with an invalid index. This was observed before on several occasions in mainline and was being fixed there (e.g. [JDK-8323274](https://bugs.openjdk.org/browse/JDK-8323274)). 

#### How this Problem Manifests in Valhalla

In the test case, the very same is happening:

1. Loop Predication hoists `237 If` out of the loop since it's invariant.
2. `108 RangeCheck` cannot be hoisted since it's not invariant and we have a non-counted loop with a `float` iv phi.
3. By hoisting `237 If` out of the loop, we also rewire the data dependencies (i.e. `140 CastII` and `146 LoadN`) out of the loop. These now end up before the range check which is still inside the loop.
4. When running with `-XX:+StressGCM`, the `LoadN` could be scheduled before the actual `RangeCheck` inside the loop and we crash when using an invalid out-of-bounds array index.

 #### Solution: Swap `RangeCheck` and Speculative Inline Type Traps
The fix is straight forward to first emit the speculative inline type checks with its traps and only then emit the `RangeCheck` such that it does not lose the connection to the actual load. 

I've additionally done some refactoring and added some comments which helped me to better understand the code.

Thanks,
Christian

---------
### Progress

### Issue
 * [JDK-8333889](https://bugs.openjdk.org/browse/JDK-8333889): [lworld] C2: Hoisting speculative array access type check wrongly moves array access before its range check (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/1127/head:pull/1127` \
`$ git checkout pull/1127`

`$ git checkout pull/1127` \
`$ git pull https://git.openjdk.org/valhalla.git pull/1127/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1127`

`$ git pr show -t 1127`

</details>

<a href="https://git.openjdk.org/valhalla/pull/1127.diff">https://git.openjdk.org/valhalla/pull/1127.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/1127#issuecomment-2167391279)

### DIFF
--- a/src/hotspot/share/opto/parse2.cpp
+++ b/src/hotspot/share/opto/parse2.cpp
@@ -436,13 +436,14 @@ void Parse::create_range_check(Node* idx, Node* ary, const TypeInt* sizetype) {
 }
 
 // For inline type arrays, we can use the profiling information for array accesses to speculate on the type, flatness,
-// and null-freeness. This avoids runtime checks and leads to a much simpler graph shape. We emit traps if the profiling
-// information turns out to be wrong at runtime.
+// and null-freeness. We can either prepare the speculative type for later uses or emit explicit speculative checks with
+// traps now. In the latter case, the speculative type guarantees can avoid additional runtime checks later (e.g.
+// non-null-free implies non-flat which allows us to remove flatness checks). This makes the graph simpler.
 Node* Parse::create_speculative_inline_type_array_checks(Node* array, const TypeAryPtr* array_type,
                                                          const Type*& element_type) {
   if (!array_type->is_flat() && !array_type->is_not_flat()) {
-    // For array accesses, it can be useful to speculate on flatness such that we can fix the data layout. We only want
-    // to do that when we know nothing about flatness since it requires a trap when profiling turns out to be wrong.
+    // For arrays that might be flat, speculate that the array has the exact type reported in the profile data such that
+    // we can rely on a fixed memory layout (i.e. either a flat layout or not).
     array = cast_to_speculative_array_type(array, array_type, element_type);
   } else if (UseTypeSpeculation && UseArrayLoadStoreProfile) {
     // Array is known to be either flat or not flat. If possible, update the speculative type by using the profile data
--- a/src/hotspot/share/opto/parse2.cpp
+++ b/src/hotspot/share/opto/parse2.cpp
@@ -373,169 +373,14 @@ Node* Parse::array_addressing(BasicType type, int vals, const Type*& elemtype) {
     return top();
   }
 
-  // Do the range check
-  if (need_range_check) {
-    Node* tst;
-    if (sizetype->_hi <= 0) {
-      // The greatest array bound is negative, so we can conclude that we're
-      // compiling unreachable code, but the unsigned compare trick used below
-      // only works with non-negative lengths.  Instead, hack "tst" to be zero so
-      // the uncommon_trap path will always be taken.
-      tst = _gvn.intcon(0);
-    } else {
-      // Range is constant in array-oop, so we can use the original state of mem
-      Node* len = load_array_length(ary);
+  ary = create_speculative_inline_type_array_checks(ary, arytype, elemtype);
 
-      // Test length vs index (standard trick using unsigned compare)
-      Node* chk = _gvn.transform( new CmpUNode(idx, len) );
-      BoolTest::mask btest = BoolTest::lt;
-      tst = _gvn.transform( new BoolNode(chk, btest) );
-    }
-    RangeCheckNode* rc = new RangeCheckNode(control(), tst, PROB_MAX, COUNT_UNKNOWN);
-    _gvn.set_type(rc, rc->Value(&_gvn));
-    if (!tst->is_Con()) {
-      record_for_igvn(rc);
-    }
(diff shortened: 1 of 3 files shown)


src/hotspot/share/opto/parse2.cpp
```diff
@@ -373,178 +373,215 @@ Node* Parse::array_addressing(BasicType type, int vals, const Type*& elemtype) {
...
-      }
+  // Make array address computation control dependent to prevent it
+  // from floating above the range check during loop optimizations.
+  Node* ptr = array_element_address(ary, idx, type, sizetype, control());
+  assert(ptr != top(), "top should go hand-in-hand with stopped");
+
+  return ptr;
+}
+
+void Parse::create_range_check(Node* idx, Node* ary, const TypeInt* sizetype) {
```
chhagedorn, line 393: Only extracted the code to a separate method without applying any refactoring to it.

src/hotspot/share/opto/parse2.cpp
```diff
@@ -373,178 +373,215 @@ Node* Parse::array_addressing(BasicType type, int vals, const Type*& elemtype) {
...
+    } else if (!slow_ctl->is_top()) {
+      { PreserveJVMState pjvms(this);
+        set_control(slow_ctl);
         uncommon_trap_exact(reason, Deoptimization::Action_maybe_recompile);
       }
-      assert(!stopped(), "flat array should have been caught earlier");
-      Node* better_ary = _gvn.transform(new CheckCastPPNode(control(), ary, arytype->cast_to_not_flat()));
-      replace_in_map(ary, better_ary);
-      ary = better_ary;
-      arytype = _gvn.type(ary)->is_aryptr();
```
chhagedorn, line 536: Dropped this line because `arytype` is not used anymore afterwards.

src/hotspot/share/opto/parse2.cpp
```diff
@@ -373,178 +373,215 @@ Node* Parse::array_addressing(BasicType type, int vals, const Type*& elemtype) {
...
+      builtin_throw(Deoptimization::Reason_range_check);
     }
+  }
+}
+
+// For inline type arrays, we can use the profiling information for array accesses to speculate on the type, flatness,
+// and null-freeness. This avoids runtime checks and leads to a much simpler graph shape. We emit traps if the profiling
+// information turns out to be wrong at runtime.
+Node* Parse::create_speculative_inline_type_array_checks(Node* array, const TypeAryPtr* array_type,
+                                                         const Type*& element_type) {
```
chhagedorn, line 442: I did not change the semantics and only applied simplifications like renaming, adding comments, extracted code to methods etc. This should improve the readability.

src/hotspot/share/opto/parse2.cpp
```diff
@@ -373,178 +373,215 @@ Node* Parse::array_addressing(BasicType type, int vals, const Type*& elemtype) {
...
-        BuildCutout unless(this, null_free_array_test(ary, /* null_free = */ false), PROB_MAX);
-        uncommon_trap_exact(reason, Deoptimization::Action_maybe_recompile);
-      }
-      assert(!stopped(), "null-free array should have been caught earlier");
-      Node* better_ary = _gvn.transform(new CheckCastPPNode(control(), ary, arytype->cast_to_not_null_free()));
-      replace_in_map(ary, better_ary);
-      ary = better_ary;
-      arytype = _gvn.type(ary)->is_aryptr();
-    }
+  // Even though the type does not tell us whether we have an inline type or not, we can still check the profile data
```
TobiHartmann, line 453:
```suggestion
  // Even though the type does not tell us whether we have an inline type array or not, we can still check the profile data
```
?

TobiHartmann (approved):
Nice analysis and great refactoring Christian!

TobiHartmann (approved):
Looks good, thanks!

chhagedorn:
Thanks Tobias for your review! As we've discussed offline, I've pushed an additional fix for a case that was missing: The load can still float above a range check when it's pinned at the If with a `FlatArrayCheck` which directly follows the range check. To avoid that, we pin it with `C2_UNKNOWN_CONTROL_LOAD` because we cannot set two nodes as control dependency.

TobiHartmann (approved):
Looks good!

chhagedorn:
Thanks Tobias!

/integrate

PR #1143, opened 2024-06-21
8334432: Refine Float16.fma

jddarcy:
Adding comments and test cases for Float16.fma.

---------
### Progress

### Issue
 * [JDK-8334432](https://bugs.openjdk.org/browse/JDK-8334432): Refine Float16.fma (**Enhancement** - P4)

### Reviewers
 * [Jatin Bhateja](https://openjdk.org/census#jbhateja) (@jatin-bhateja - Committer) ⚠️ Review applies to [172eea95](https://git.openjdk.org/valhalla/pull/1143/files/172eea9575dc56c0abe29405121a05a9d3a89bec)
 * [Raffaello Giulietti](https://openjdk.org/census#rgiulietti) (@rgiulietti - no project role) ⚠️ Review applies to [172eea95](https://git.openjdk.org/valhalla/pull/1143/files/172eea9575dc56c0abe29405121a05a9d3a89bec)

### Contributors
 * Raffaello Giulietti `<rgiulietti@openjdk.org>`

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/1143/head:pull/1143` \
`$ git checkout pull/1143`

`$ git checkout pull/1143` \
`$ git pull https://git.openjdk.org/valhalla.git pull/1143/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1143`

`$ git pr show -t 1143`

</details>

<a href="https://git.openjdk.org/valhalla/pull/1143.diff">https://git.openjdk.org/valhalla/pull/1143.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/1143#issuecomment-2181987696)


jddarcy:
I've asked my colleague @rgiulietti to help with the error analysis of the current implementation.

The commented-out code in the fma for an inexact product-sum does get executed for the expected test cases. If special-case code for an inexact product-sum  turns out need to be needed, this extra code will be deleted before a push.

rgiulietti:
The analysis done [here](https://github.com/openjdk/valhalla/pull/1117#issuecomment-2174342914) seems correct, but there is an implicit suspension moment in the last case.

> That leaves possibly non-exact product-sum with a combination of product in the subnormal range of Float16 and the c term to be added in being not-small. However, if this product-sum is non-exact, the smaller term from the product, with at most 22 exponent bit positions set, and the the 11 bits from c being summed in, must be separated by at least 53 - (22 + 11) = 20 bit positions otherwise the product-sum would fit in a double. I believe this implies at least one of the double-rounding scenarios cannot occur, in particular a half-way result in the smaller precision, Float16 in this case, rounding differently because sticky bit information from the higher precision was rounded away.

Here's a further analysis of this case.

Double rounding is usually harmless. It is harmful only in two situations:

- The first rounding from the exact value to the extended precision (here `double`) happens to be directed _toward_ 0 to a value exactly midway between two adjacent working precision (here `float16`) values, followed by a second rounding from there which again happens to be directed _toward_ 0 to one of these values (the one with lesser magnitude).
  A single rounding from the exact value to the working precision, in contrast, rounds to the value with larger magnitude.
- Symmetrically, the first rounding to the extended precision happens to be directed _away_ from 0 to a value exactly midway between two adjacent working precision values, followed by a second rounding from there which again happens to be directed _away_ from 0 to one of these values (the one with larger magnitude).
  However, a single rounding from the exact value to the working precision rounds to the value with lesser magnitude.

In any other case double rounding is innocuous, returning the same value as a single rounding to the working precision.
We only need to ensure that the first rounding to `double` does not produce the midpoint of two adjacent `float16` values.

- If a·b and c have the same sign, the sum a·b + c has a significand with a large gap of 20 or more 0s between the bits of the signifcand of c to the left (at most 11 bits) and those of the product a·b to the right (at most 22 bits).
The rounding bit for the final working precision of `float16` is the leftmost 0 in the gap.
  - If rounding to `double` is directed toward 0, all the 0s in the gap are preserved, thus the `float16` rounding bit is unaffected and remains 0. This means that the `double` value is _not_ the midpoint  of two adjacent `float16` values, so double rounding is harmless.
  - If rounding to `double` is directed away form 0, the rightmost 0 in the gap might be replaced by a 1, but the others are unaffected, including the `float16` rounding bit. Again, this shows that the `double` value is _not_ the midpoint  of two adjacent `float16` values, and double rounding is innocuous.

- If a·b and c have opposite signs, the sum a·b + c the long gap of 0s above is replaced by a long gap of 1s. The `float16` rounding bit is the leftmost 1 in the gap, or the second leftmost 1 iff c is a power of 2. In both cases, the rounding bit is followed by at least another 1.
  - If rounding to `double` is directed toward 0, the `float16` rounding bit and its follower are preserved and both 1, so the `double` value is _not_ the midpoint  of two adjacent `float16` values, and double rounding is harmless.
  - If rounding to `double` is directed away from 0, the `float16` rounding bit and its follower are either preserved (both 1), or both switch to 0. Either way, the `double` value is again _not_ the midpoint  of two adjacent `float16` values, and double rounding is harmless.

jddarcy:
> The analysis done [here](https://github.com/openjdk/valhalla/pull/1117#issuecomment-2174342914) seems correct, but there is an implicit suspension moment in the last case.
> 
> > That leaves possibly non-exact product-sum with a combination of product in the subnormal range of Float16 and the c term to be added in being not-small. However, if this product-sum is non-exact, the smaller term from the product, with at most 22 exponent bit positions set, and the the 11 bits from c being summed in, must be separated by at least 53 - (22 + 11) = 20 bit positions otherwise the product-sum would fit in a double. I believe this implies at least one of the double-rounding scenarios cannot occur, in particular a half-way result in the smaller precision, Float16 in this case, rounding differently because sticky bit information from the higher precision was rounded away.
> 
> Here's a further analysis of this case.
> 
> Double rounding is usually harmless. It is harmful only in two situations:
> 
>     * The first rounding from the exact value to the extended precision (here `double`) happens to be directed _toward_ 0 to a value exactly midway between two adjacent working precision (here `float16`) values, followed by a second rounding from there which again happens to be directed _toward_ 0 to one of these values (the one with lesser magnitude).
>       A single rounding from the exact value to the working precision, in contrast, rounds to the value with larger magnitude.
> 
>     * Symmetrically, the first rounding to the extended precision happens to be directed _away_ from 0 to a value exactly midway between two adjacent working precision values, followed by a second rounding from there which again happens to be directed _away_ from 0 to one of these values (the one with larger magnitude).
>       However, a single rounding from the exact value to the working precision rounds to the value with lesser magnitude.
> 
> 
> In any other case double rounding is innocuous, returning the same value as a single rounding to the working precision. We only need to ensure that the first rounding to `double` does not produce the midpoint of two adjacent `float16` values.
> 
>     * If a·b and c have the same sign, the sum a·b + c has a significand with a large gap of 20 or more 0s between the bits of the signifcand of c to the left (at most 11 bits) and those of the product a·b to the right (at most 22 bits).
>       The rounding bit for the final working precision of `float16` is the leftmost 0 in the gap.
>       
>       * If rounding to `double` is directed toward 0, all the 0s in the gap are preserved, thus the `float16` rounding bit is unaffected and remains 0. This means that the `double` value is _not_ the midpoint  of two adjacent `float16` values, so double rounding is harmless.
>       * If rounding to `double` is directed away form 0, the rightmost 0 in the gap might be replaced by a 1, but the others are unaffected, including the `float16` rounding bit. Again, this shows that the `double` value is _not_ the midpoint  of two adjacent `float16` values, and double rounding is innocuous.
> 
>     * If a·b and c have opposite signs, the sum a·b + c the long gap of 0s above is replaced by a long gap of 1s. The `float16` rounding bit is the leftmost 1 in the gap, or the second leftmost 1 iff c is a power of 2. In both cases, the rounding bit is followed by at least another 1.
>       
>       * If rounding to `double` is directed toward 0, the `float16` rounding bit and its follower are preserved and both 1, so the `double` value is _not_ the midpoint  of two adjacent `float16` values, and double rounding is harmless.
>       * If rounding to `double` is directed away from 0, the `float16` rounding bit and its follower are either preserved (both 1), or both switch to 0. Either way, the `double` value is again _not_ the midpoint  of two adjacent `float16` values, and double rounding is harmless.

Thanks; that agrees with the sketch of those cases I was working on. I'll update the explanatory comment and cleanup the implementation accordingly.

test/jdk/java/lang/Float16/BasicFloat16ArithTests.java
```diff
@@ -509,6 +510,56 @@ private static void testSimple() {
...
+                {0x1.ffcp-14f, 0x1.0p-24f, 0x1.0p14f, // *Cannot* be held exactly
+                 0x1.0p14f},
+
+                // Check values where the exact result cannot be
+                // exactly stored in a double.
+                {0x1.0p-24f, 0x1.0p-24f, 0x1.0p10f,
+                 0x1.0p10f},
+
+                {0x1.0p-24f, 0x1.0p-24f, 0x1.0p14f,
+                 0x1.0p14f},
```
jatin-bhateja, line 555: Your tests only covers overflow conditions, a few going below minimum sub-normal value (0x0.004P-14) and minimum normal value (0x1.000P-14) will also be useful.

jatin-bhateja (commented):
Hi @jddarcy , Some queries / comments.

test/jdk/java/lang/Float16/BasicFloat16ArithTests.java
jatin-bhateja: My query was in following context, citing excerpt from JVM Specification 22 section 2.11.3.

"The Java Virtual Machine does not indicate overflow or underflow during
operations on floating-point data types. That is, floating-point instructions never
cause the Java Virtual Machine to throw a run-time exception (not to be confused
with an IEEE 754 floating-point exception). **An operation that overflows produces
a signed infinity; an operation that underflows produces a subnormal value or a
signed zero;** an operation that has no unique mathematically defined result produces
NaN. All numeric operations with NaN as an operand produce NaN as a result"

jddarcy: > My query was in following context, citing excerpt from JVM Specification 22 section 2.11.3.
> 
> "The Java Virtual Machine does not indicate overflow or underflow during operations on floating-point data types. That is, floating-point instructions never cause the Java Virtual Machine to throw a run-time exception (not to be confused with an IEEE 754 floating-point exception). **An operation that overflows produces a signed infinity; an operation that underflows produces a subnormal value or a signed zero;** an operation that has no unique mathematically defined result produces NaN. All numeric operations with NaN as an operand produce NaN as a result"

Sure; that is reasonable.

The correctness of the Float16.fma method relies on the correctness of the double to Float16 conversion. As part of a larger test update, I was planned on expanding the double to Float16 testing. In particular, a test methodology like:

For each finite Float16 value, test the double values around the half-way point to make sure the rounding is correct in those cases. Since there are so few Float16 values, I expect this test to run quickly.

jatin-bhateja: Thanks again for explanation. sounds good to me. Yes full range of value are indeed covered in above exhaustive test including lowest sub-normal value (0x1.000P-24 == 0x0.004P-14 == 5.960464477539063E-8

jatin-bhateja (approved):

jddarcy:
Getting back to this PR, I pushed an update to the explanation about why Float16 fma is double arithmetic is sufficient and an initial update of the tests.

More test augmentation as previously discussed on the way.

jddarcy:
> Hi @jddarcy , Some queries / comments.

Hi @jatin-bhateja ,

Please re-review the current state of the PR and if it looks good, I'm integrate it for sponsorship. Thanks.

src/java.base/share/classes/java/lang/Float16.java
```diff
@@ -859,36 +873,188 @@ public static Float16 sqrt(Float16 radicand) {
...
+         * 1) Large exponent product, exponent > Float16.MAX_EXPONENT
+         *
+         * The magnitude of the overflow threshold for Float16 is:
+         *
+         * MAX_VALUE + 1/2 * ulp(MAX_VALUE) =  0x1.ffcp15 + 0x0.002p15 = 0x1.ffep15
+         *
+         * Therefore, any product greater than 0x1.ffep15 + MAX_VALUE
+         * = 0x1.ffdp16 will certainly overflow (under round to
+         * nearest) since adding in c = -MAX_VALUE will still be above
+         * the overflow threshold.
```
rgiulietti, line 954:
```suggestion
         * Therefore, for any product greater than or equal to 0x1.ffep15 + MAX_VALUE
         * = 0x1.ffdp16, the "fma" will certainly overflow (under round to
         * nearest), since adding in c = -MAX_VALUE will still be above
         * or at the overflow threshold.
```

src/java.base/share/classes/java/lang/Float16.java
```diff
@@ -859,36 +873,188 @@ public static Float16 sqrt(Float16 radicand) {
...
+         * *not* be exact and additional analysis is needed to justify
+         * not having any corrective computation to compensate for
+         * intermediate rounding errors.
+         *
+         * The following analysis will rely on the range of bit
+         * positions representable in the intermediate
+         * product-sum.
+         *
+         * For the product a*b of Float16 inputs, the range of
+         * exponents for nonzero finite results goes from 2^(-28)
```
rgiulietti, line 937:
```suggestion
         * exponents for nonzero finite results goes from 2^(-48)
```

src/java.base/share/classes/java/lang/Float16.java
```diff
@@ -859,36 +873,188 @@ public static Float16 sqrt(Float16 radicand) {
...
+         * exponent range goes from -1022 to 1023. The Float16 format
+         * has 11 bits of logical precision and its exponent range
+         * goes from -14 to 15. Therefore, the bit positions
+         * representable in the Float16 format range from the
+         * subnormal 2^(-24), MIN_VALUE, to 2^15, the leading bit
+         * position of MAX_VALUE.
+         *
+         * Consequently, a double can hold the exact sum of any two
+         * Float16 values as the maximum difference in exponents of
+         * Float16 values less than the precision width of double.
```
rgiulietti, line 917: This is correct, but doesn't seem to be relevant for what follows. We don't have a sum between two `Float16` values here.

src/java.base/share/classes/java/lang/Float16.java
```diff
@@ -859,36 +873,188 @@ public static Float16 sqrt(Float16 radicand) {
...
+         * including the propagation of the sign of zero if all three
+         * arguments are zero.
+         *
+         * The double format has 53 logical bits of precision and its
+         * exponent range goes from -1022 to 1023. The Float16 format
+         * has 11 bits of logical precision and its exponent range
+         * goes from -14 to 15. Therefore, the bit positions
+         * representable in the Float16 format range from the
+         * subnormal 2^(-24), MIN_VALUE, to 2^15, the leading bit
+         * position of MAX_VALUE.
```
rgiulietti, line 913: It's unclear what is meant by "bit position".

src/java.base/share/classes/java/lang/Float16.java
```diff
@@ -859,36 +873,188 @@ public static Float16 sqrt(Float16 radicand) {
...
+         *
+         * The following analysis will rely on the range of bit
+         * positions representable in the intermediate
+         * product-sum.
+         *
+         * For the product a*b of Float16 inputs, the range of
+         * exponents for nonzero finite results goes from 2^(-28)
+         * (from MIN_VALUE squared) to 2^31 (from the exact value of
+         * MAX_VALUE squared). This full range of exponent positions,
+         * (31 -(-28) + 1 ) = 60 exceeds the precision of
```
rgiulietti, line 940:
```suggestion
         * (31 -(-48) + 1 ) = 80 exceeds the precision of
```

src/java.base/share/classes/java/lang/Float16.java
```diff
@@ -859,36 +873,188 @@ public static Float16 sqrt(Float16 radicand) {
...
+         *
+         * The exact product has at most 22 contiguous bits in its
+         * logical significand. The third number being added in has
+         * at most 11 contiguous bits in its significand and the
+         * lowest bit position that could be set is
+         * 2^(-24). Therefore, when the product has the maximum
+         * in-range exponent, 2^15, a single double has enough
+         * precision to hold down to the smallest subnormal bit
+         * position, 15 - (-24) + 1 = 40 < 53. If the product was
+         * large and overflowed when the third operand was added, this
```
rgiulietti, line 971: In `double` arithmetic, as here, `a * b + c` cannot overflow, so this is a bit confusing.

src/java.base/share/classes/java/lang/Float16.java
```diff
@@ -859,36 +873,188 @@ public static Float16 sqrt(Float16 radicand) {
...
+         * would cause the exponent to increase to 16, which is within
+         * the range of double, so the product-sum is exact and will
+         * be correct when rounded to Float16.
+         *
+         * 3) Exponent of product is in the range of subnormal values or smaller,
+         * exponent < Float16.MIN_EXPONENT
+         *
+         * The smallest exponent possible in a product is 2^(-48).
+         * For moderately sized Float16 values added to the product,
+         * with a leading exponent of about 4, the sum will not be
```
rgiulietti, line 981:
```suggestion
         * with an exponent of about 4, the sum will not be
```

src/java.base/share/classes/java/lang/Float16.java
```diff
@@ -859,36 +873,188 @@ public static Float16 sqrt(Float16 radicand) {
...
+         * exact. Therefore, an analysis is needed to determine if the
+         * double-rounding is benign or would lead to a different
+         * final Float16 result. Double rounding an lead to a
```
rgiulietti, line 984:
```suggestion
         * final Float16 result. Double rounding can lead to a
```

test/jdk/java/lang/Float16/BasicFloat16ArithTests.java
```diff
@@ -375,11 +376,78 @@ private static void throwRE(String message) {
...
     }
 
+
+    private static void checkValueOfDouble() {
+        /*
+         * Check that double -> Float16 conversion rounds propertly
+         * around the widway point for each finite Float16 value.
+         */
+        for(int i = 0; i <= Short.MAX_VALUE; i++ ) {
+            // Start by just checking positive values...
```
rgiulietti, line 386: The test does, in fact, also check the negated value.
It doesn't check negative zero, though.

Maybe it's simpler to loop like so
```
        for(int i = Short.MIN_VALUE; i <= Short.MAX_VALUE; i++ ) {
```
and drop the logic for the negated value.

rgiulietti: Forget about this comment, except for the first line. The test is already complete

jddarcy:
Thanks for the review comments @rgiulietti ; the most recent push should address the points you raised.

rgiulietti (approved):
Looks good

jddarcy:
@jatin-bhateja , if you re-review the current state of the PR, I'll integrate the changes so they can be sponsored. Thanks.

test/jdk/java/lang/Float16/BasicFloat16ArithTests.java
```diff
@@ -375,11 +376,81 @@ private static void throwRE(String message) {
...
+                checkFloat16(add(f16,         ulp), roundedBack,    roundUpMsg);
+                checkFloat16(subtract(f16Neg, ulp), roundedBackNeg, roundUpMsg);
+            }
+
+            // Should always round down
+            double halfWayNextDown = Math.nextDown(halfWay);
+            checkFloat16(f16,    valueOf(halfWayNextDown).floatValue(),  roundDownMsg);
+            checkFloat16(f16Neg, valueOf(-halfWayNextDown).floatValue(), roundDownMsg);
+
+            // Should always round down up
```
jatin-bhateja, line 422:
```suggestion
            // Should always round up
```

jatin-bhateja (approved):
Hi @jddarcy ,
Thanks for very detailed educative comments. 
Looks good to me.

Best Regards,
Jatin

PR #1144, opened 2024-06-21
8334729: [lworld] CDS should no force the computation of identity hash code for value instances

fparain:
Simple fix to prevent the VM thread from calling back to Java code when archiving value instances.


DanHeidinga (approved):
Lgtm

PR #1160, opened 2024-07-09
8335545: [lworld] JNI MonitorEnter doesn't throw the right exception

fparain:
Fix JNI's MonitorEnter to throw IdentityException instead of IllegalMonitorStateException.

---------
### Progress

### Issue
 * [JDK-8335545](https://bugs.openjdk.org/browse/JDK-8335545): [lworld] JNI MonitorEnter doesn't throw the right exception (**Bug** - P3)

### Reviewers
 * [Coleen Phillimore](https://openjdk.org/census#coleenp) (@coleenp - no project role) ⚠️ Review applies to [18d72953](https://git.openjdk.org/valhalla/pull/1160/files/18d729533f410a216389f505ae4c27c038468f07)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/1160/head:pull/1160` \
`$ git checkout pull/1160`

`$ git checkout pull/1160` \
`$ git pull https://git.openjdk.org/valhalla.git pull/1160/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1160`

`$ git pr show -t 1160`

</details>

<a href="https://git.openjdk.org/valhalla/pull/1160.diff">https://git.openjdk.org/valhalla/pull/1160.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/1160#issuecomment-2217874309)

### DIFF
--- a/src/hotspot/share/runtime/synchronizer.cpp
+++ b/src/hotspot/share/runtime/synchronizer.cpp
@@ -306,6 +306,21 @@ jlong ObjectSynchronizer::_last_async_deflation_time_ns = 0;
 static uintx _no_progress_cnt = 0;
 static bool _no_progress_skip_increment = false;
 
+#define CHECK_THROW_NOSYNC_IE(obj) \
+  if (obj->klass()->is_inline_klass()) { \
+    ResourceMark rm(THREAD); \
+    const char* desc = "Cannot synchronize on an instance of value class "; \
+    const char* className = obj->klass()->external_name(); \
+    size_t msglen = strlen(desc) + strlen(className) + 1; \
+    char* message = NEW_RESOURCE_ARRAY(char, msglen); \
+    if (nullptr == message) { \
+      THROW_MSG(vmSymbols::java_lang_IdentityException(), className); \
+    } else { \
+      jio_snprintf(message, msglen, "%s%s", desc, className); \
+      THROW_MSG(vmSymbols::java_lang_IdentityException(), message); \
+    } \
+  }
+
 #define CHECK_THROW_NOSYNC_IMSE(obj)  \
   if (EnableValhalla && (obj)->mark().is_inline_type()) {  \
     JavaThread* THREAD = current;           \
@@ -540,8 +555,10 @@ void ObjectSynchronizer::enter_for(Handle obj, BasicLock* lock, JavaThread* lock
   // the locking_thread with respect to the current thread. Currently only used when
   // deoptimizing and re-locking locks. See Deoptimization::relock_objects
   assert(locking_thread == Thread::current() || locking_thread->is_obj_deopt_suspend(), "must be");
-  JavaThread* current = locking_thread;
-  CHECK_THROW_NOSYNC_IMSE(obj);
+  if (obj->klass()->is_inline_klass()) {
+    // JITed code should never have locked an instance of a value class
+    ShouldNotReachHere();
+  }
   if (!enter_fast_impl(obj, lock, locking_thread)) {
     // Inflated ObjectMonitor::enter_for is required
 
@@ -560,7 +577,8 @@ void ObjectSynchronizer::enter_for(Handle obj, BasicLock* lock, JavaThread* lock
 
 void ObjectSynchronizer::enter(Handle obj, BasicLock* lock, JavaThread* current) {
   assert(current == Thread::current(), "must be");
-  CHECK_THROW_NOSYNC_IMSE(obj);
+  JavaThread* THREAD = current;
+  CHECK_THROW_NOSYNC_IE(obj);
   if (!enter_fast_impl(obj, lock, current)) {
     // Inflated ObjectMonitor::enter is required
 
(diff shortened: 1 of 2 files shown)


src/hotspot/share/runtime/synchronizer.cpp
```diff
@@ -560,7 +577,8 @@ void ObjectSynchronizer::enter_for(Handle obj, BasicLock* lock, JavaThread* lock
 
 void ObjectSynchronizer::enter(Handle obj, BasicLock* lock, JavaThread* current) {
   assert(current == Thread::current(), "must be");
-  CHECK_THROW_NOSYNC_IMSE(obj);
+  JavaThread* THREAD = current;
+  CHECK_THROW_NOSYNC_IE(obj);
```
coleenp, line 581: We never get here.  The c1 slow path and interpreter code check for inline types before calling this.

I added an assert.

src/hotspot/share/runtime/synchronizer.cpp
```diff
@@ -306,6 +306,21 @@ jlong ObjectSynchronizer::_last_async_deflation_time_ns = 0;
 static uintx _no_progress_cnt = 0;
 static bool _no_progress_skip_increment = false;
 
+#define CHECK_THROW_NOSYNC_IE(obj) \
+  if (obj->klass()->is_inline_klass()) { \
+    ResourceMark rm(THREAD); \
+    const char* desc = "Cannot synchronize on an instance of value class "; \
+    const char* className = obj->klass()->external_name(); \
+    size_t msglen = strlen(desc) + strlen(className) + 1; \
+    char* message = NEW_RESOURCE_ARRAY(char, msglen); \
```
coleenp, line 315: If NEW_RESOURCE_ARRAY fails, it will call vm_exit_out_of_memory, so no need to check for nullptr.

src/hotspot/share/runtime/synchronizer.cpp
```diff
@@ -540,8 +555,10 @@ void ObjectSynchronizer::enter_for(Handle obj, BasicLock* lock, JavaThread* lock
   // the locking_thread with respect to the current thread. Currently only used when
   // deoptimizing and re-locking locks. See Deoptimization::relock_objects
   assert(locking_thread == Thread::current() || locking_thread->is_obj_deopt_suspend(), "must be");
-  JavaThread* current = locking_thread;
-  CHECK_THROW_NOSYNC_IMSE(obj);
+  if (obj->klass()->is_inline_klass()) {
+    // JITed code should never have locked an instance of a value class
+    ShouldNotReachHere();
+  }
```
coleenp, line 561: You could make this an assert instead.
assert(!EnableValhalla || !obj->klass()->is_inline_klass(), "^ above comment");

Then it's self documenting.  Better than ShouldNotReachHere.

src/hotspot/share/runtime/synchronizer.cpp
coleenp: Do you need EnableValhalla ?

src/hotspot/share/runtime/synchronizer.cpp
```diff
@@ -306,6 +306,21 @@ jlong ObjectSynchronizer::_last_async_deflation_time_ns = 0;
...
+    char* message = NEW_RESOURCE_ARRAY(char, msglen); \
+    if (nullptr == message) { \
+      THROW_MSG(vmSymbols::java_lang_IdentityException(), className); \
+    } else { \
+      jio_snprintf(message, msglen, "%s%s", desc, className); \
+      THROW_MSG(vmSymbols::java_lang_IdentityException(), message); \
+    } \
+  }
+
 #define CHECK_THROW_NOSYNC_IMSE(obj)  \
```
coleenp, line 324: I added a short comment why this is needed. like:

// These checks are required for wait, notify and exit to avoid inflating the monitor to
// find out this inline type object cannot be locked.

coleenp (changes requested):
I had a similar change, which I'll remove this part.

src/hotspot/share/runtime/synchronizer.cpp
```diff
@@ -560,7 +561,8 @@ void ObjectSynchronizer::enter_for(Handle obj, BasicLock* lock, JavaThread* lock
 
 void ObjectSynchronizer::enter(Handle obj, BasicLock* lock, JavaThread* current) {
   assert(current == Thread::current(), "must be");
-  CHECK_THROW_NOSYNC_IMSE(obj);
+  JavaThread* THREAD = current;
```
coleenp, line 564: You don't need the THREAD line anymore.

coleenp (approved):
This looks good.  One small line change.

src/hotspot/share/runtime/synchronizer.cpp
fparain: Line removed.

fparain:
Coleen,

Thank you for the suggestions and the review.

Fred

PR #1172, opened 2024-07-19
8336682: [lworld] java/lang/instrument/IsModifiableClassAgent.java test fails when run with migrates classes

coleenp:
The test uses the class file reconstituter to provide the redefined bytecodes, leaving out the LoadableDescriptors.  Adding them to the reconsistituted class file makes the test pass, also fix the verification code since these descriptors are utf8s now.  Tested with java/lang/instrument with --enable-preview.  Should write a test but I don't know how, and the code is already tested with permitted subclasses and other attributes.

---------
### Progress

### Issue
 * [JDK-8336682](https://bugs.openjdk.org/browse/JDK-8336682): [lworld] java/lang/instrument/IsModifiableClassAgent.java test fails when run with migrates classes (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/1172/head:pull/1172` \
`$ git checkout pull/1172`

`$ git checkout pull/1172` \
`$ git pull https://git.openjdk.org/valhalla.git pull/1172/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1172`

`$ git pr show -t 1172`

</details>

<a href="https://git.openjdk.org/valhalla/pull/1172.diff">https://git.openjdk.org/valhalla/pull/1172.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/1172#issuecomment-2239016488)

### DIFF
--- a/src/hotspot/share/prims/jvmtiClassFileReconstituter.cpp
+++ b/src/hotspot/share/prims/jvmtiClassFileReconstituter.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2005, 2023, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2005, 2024, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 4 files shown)


fparain:
Hi Coleen,

Thank you for the fix in ClassFileReconstituter, the handling of the LoadableDescriptors attribute was clearly missing.
However, I'm wondering if class redefinition should reject redefined classes with a missing or different LoadableDescriptors attribute. This attribute helps the VM optimizing this class, but it is not supposed to impact the semantic of the class.

Fred

DanHeidinga:
> However, I'm wondering if class redefinition should reject redefined classes with a missing or different LoadableDescriptors attribute. This attribute helps the VM optimizing this class, but it is not supposed to impact the semantic of the class.

I don't think we need to reject classes with missing or different LD attributes.  The attribute provides the VM an option to load classes earlier than it otherwise would but doesn't enforce a requirement.  I see this as being similar to redefining the `<clinit>` method: perfectly valid to do but will have no effect if the class has already been initialized.

Changing the LD attribute may give the VM permission to preload different classes but doesn't affect decisions already made (like changing `<clinit`>) and doesn't impact the integrity of the VM (like changing NestMates).   So I don't think it should be validated.

coleenp:
Okay, I've removed the check in redefinition.  Test still passes.

fparain (approved):
LGTM

coleenp:
Thanks Fred and Dan.

/integrate

PR #1183, opened 2024-07-25
8336747: [lworld] Test TestPrimitiveAndArrayModifiers.java fails because of the ACC_IDENTITY flag

RogerRiggs:
Include IDENTITY when checking the modifiers on arrays when --enable-preview


MrSimms (approved):
LGTM

PR #1190, opened 2024-07-31
8337585: [lworld] VM crashes if primitive type signatures are present in the LoadableDescriptors attribute

fparain:
Prevent VM crash when LoadableDescriptor attribute contains signature of primitive types.


coleenp (approved):
So you just ignore a non-class name in the LoadableDescriptors.  Looks good.

fparain:
Thank you for the review Coleen.

Fred

PR #1196, opened 2024-08-09
8338102: x86 backend support for newly added Float16 intrinsics.

jatin-bhateja:
This patch enables newly added Float16 intrinsicfication support added by [JDK-8336406](https://bugs.openjdk.org/browse/JDK-8336406) for x86 targets supporting AVX512_FP16 feature.

Kindly review and approve.

Best Regards,
Jatin

Hi @Bhavana-Kilambi, 
On a second thought, do you see a possibility of sharing the IR nodes by appending secondary opcode to shared IR node in applicable scenarios, so we can have one IR for each class of operations (unary / binary / ternary).  It may need defining following new matcher routines and some more interfaces:- 
```
   match_rule_supported_shared(int primary_opcode, int secondary_opcode)
   match_rule_supported_vector_shared (int primary_opcode, int secondary_opcode, int vlen, BasicType bt)
   VectorNode::opcode(int popc, int sopc, BasicType bt)

BinaryOpNode (Dst, Src1, Src2, immI_Opcode);   

```
Secondary opcode being a immediate operand can be accessed by encoding routines. WDYT ?

Another possibility could be to encode both primary and secondary opcodes in existing opcode without disturbing the interfaces and add relevant helper routines to extract primary / secondary opcodes, I think opcodes are never -ve values, hence secondary opcode could be accommodated into higher order bits starting from (MSB-1).

---------
### Progress

### Issue
 * [JDK-8338102](https://bugs.openjdk.org/browse/JDK-8338102): x86 backend support for newly added Float16 intrinsics. (**Enhancement** - P4)

### Reviewers
 * [Bhavana Kilambi](https://openjdk.org/census#bkilambi) (@Bhavana-Kilambi - no project role)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/1196/head:pull/1196` \
`$ git checkout pull/1196`

`$ git checkout pull/1196` \
`$ git pull https://git.openjdk.org/valhalla.git pull/1196/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1196`

`$ git pr show -t 1196`

</details>

<a href="https://git.openjdk.org/valhalla/pull/1196.diff">https://git.openjdk.org/valhalla/pull/1196.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/1196#issuecomment-2277668392)

### DIFF
--- a/test/hotspot/jtreg/compiler/vectorization/TestFloat16VectorOps.java
+++ b/test/hotspot/jtreg/compiler/vectorization/TestFloat16VectorOps.java
@@ -59,6 +59,8 @@ public class TestFloat16VectorOps {
     @Test
     @Warmup(10000)
     @IR(counts = {IRNode.ADD_VHF, ">= 1"},
+        applyIfCPUFeature = {"avx512_fp16", "true"})
+    @IR(counts = {IRNode.ADD_VHF, ">= 1"},
         applyIfCPUFeatureOr = {"avx512_fp16", "true", "sve", "true"})
     @IR(counts = {IRNode.ADD_VHF, ">= 1"},
         applyIfCPUFeatureAnd = {"fphp", "true", "asimdhp", "true"})
@@ -80,6 +82,8 @@ public class TestFloat16VectorOps {
 
     @Test
     @Warmup(10000)
+    @IR(counts = {IRNode.SUB_VHF, ">= 1"},
+        applyIfCPUFeature = {"avx512_fp16", "true"})
     @IR(counts = {IRNode.SUB_VHF, ">= 1"},
         applyIfCPUFeature = {"sve", "true"})
     @IR(counts = {IRNode.SUB_VHF, ">= 1"},
@@ -103,6 +107,8 @@ public class TestFloat16VectorOps {
     @Test
     @Warmup(10000)
     @IR(counts = {IRNode.MUL_VHF, ">= 1"},
+        applyIfCPUFeature = {"avx512_fp16", "true"})
+    @IR(counts = {IRNode.MUL_VHF, ">= 1"},
         applyIfCPUFeature = {"sve", "true"})
     @IR(counts = {IRNode.MUL_VHF, ">= 1"},
         applyIfCPUFeatureAnd = {"fphp", "true", "asimdhp", "true"})
@@ -124,6 +130,8 @@ public class TestFloat16VectorOps {
 
     @Test
     @Warmup(10000)
+    @IR(counts = {IRNode.DIV_VHF, ">= 1"},
+        applyIfCPUFeature = {"avx512_fp16", "true"})
     @IR(counts = {IRNode.DIV_VHF, ">= 1"},
         applyIfCPUFeature = {"sve", "true"})
     @IR(counts = {IRNode.DIV_VHF, ">= 1"},
@@ -147,6 +155,8 @@ public class TestFloat16VectorOps {
     @Test
     @Warmup(10000)
     @IR(counts = {IRNode.MIN_VHF, ">= 1"},
+        applyIfCPUFeature = {"avx512_fp16", "true"})
+    @IR(counts = {IRNode.MIN_VHF, ">= 1"},
         applyIfCPUFeature = {"sve", "true"})
     @IR(counts = {IRNode.MIN_VHF, ">= 1"},
         applyIfCPUFeatureAnd = {"fphp", "true", "asimdhp", "true"})
@@ -168,6 +178,8 @@ public class TestFloat16VectorOps {
 
     @Test
     @Warmup(10000)
+    @IR(counts = {IRNode.MAX_VHF, ">= 1"},
+        applyIfCPUFeature = {"avx512_fp16", "true"})
     @IR(counts = {IRNode.MAX_VHF, ">= 1"},
         applyIfCPUFeature = {"sve", "true"})
     @IR(counts = {IRNode.MAX_VHF, ">= 1"},
--- a/test/hotspot/jtreg/compiler/vectorization/TestFloat16VectorOps.java
+++ b/test/hotspot/jtreg/compiler/vectorization/TestFloat16VectorOps.java
@@ -59,8 +59,6 @@ public class TestFloat16VectorOps {
     @Test
     @Warmup(10000)
     @IR(counts = {IRNode.ADD_VHF, ">= 1"},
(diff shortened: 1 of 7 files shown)


jatin-bhateja:
FTR, other missing intrinsics are conversion operations (Float16.longValue/intValue/doubleValue etc), x86 FP16 ISA has direct instructions for conversions from / to Float16 values..

test/hotspot/jtreg/compiler/vectorization/TestFloat16VectorOps.java
```diff
@@ -58,6 +58,8 @@ public TestFloat16VectorOps() {
 
     @Test
     @Warmup(10000)
+    @IR(counts = {IRNode.ADD_VHF, ">= 1"},
+        applyIfCPUFeature = {"avx512_fp16", "true"})
     @IR(counts = {IRNode.ADD_VHF, ">= 1"},
         applyIfCPUFeatureOr = {"avx512_fp16", "true", "sve", "true"})
```
Bhavana-Kilambi, line 64: For add operation, avx512_fp16 feature check is already in place.

test/hotspot/jtreg/compiler/vectorization/TestFloat16VectorOps.java
```diff
@@ -80,6 +82,8 @@ public void checkResultAdd() {
 
     @Test
     @Warmup(10000)
+    @IR(counts = {IRNode.SUB_VHF, ">= 1"},
+        applyIfCPUFeature = {"avx512_fp16", "true"})
```
Bhavana-Kilambi, line 86: This rule can be merged with the next rule for sve - 
`applyIfCPUFeatureOr = {"avx512_fp16", "true", "sve", "true"}`. Same for other operations below.

Bhavana-Kilambi:
Hi @jatin-bhateja , what do you mean by - "shared IR node in applicable scenario". You mean nodes like MaxV, MinV which are shared between multiple types?
Also by - "unary / binary / secondary" do you mean ternary instead of secondary?

jatin-bhateja:
> Hi @jatin-bhateja , what do you mean by - "shared IR node in applicable scenario". You mean nodes like MaxV, MinV which are shared between multiple types? Also by - "unary / binary / secondary" do you mean ternary instead of secondary?

Hi @Bhavana-Kilambi , 

Yup, it was a typo.

As of now we have a seperate IR node for all half float operations. Do you see a benefit in sharing AddHF, SubHF, MulHF, DivHF, MaxHF and MinHF into one IR with secondary opcode ? We may need to define some new interfaces in matcher as mentioned in my PR description. Going forward we may also need to handle Int8 and FP8, creating a sepcialized IR may add overhead.

Bhavana-Kilambi:
> FTR, other missing intrinsics are conversion operations (Float16.longValue/intValue/doubleValue etc), x86 FP16 ISA has direct instructions for conversions from / to Float16 values..

I can add implementation for these conversions (if it's not already taken by anyone else or you :)

jatin-bhateja:
> I am not sure about the idea of having unified nodes for multiple classes of operations (unary, binary etc). I am wondering whether it will simplify or complicate the implementation of these operations? I'll just break down my thoughts into the following major points -
> 
> * The size of the resulting sea of nodes graph should not change. Instead of having AddHF/SubHF we will have a BinaryOpNode. Plus it needs to now store an extra field (secondary opcode) to differentiate between the various binary operations. Maybe the size of the JVM binary itself might reduce a bit but not sure about the c2 IR graph.

Yes, Sea Of Nodes is a compile time graph and is agnostic to common IR representation. 

Having a common IR translates into single instruction selection pattern,  as you know ADLC processes each pattern and generates code catering to different downstream passes.  By passing additional secondary opcode as an immI (constant) operand we can save lots of generate code.

> * How easy would it be to implement Ideal/Value optimizations for a BinaryOpNode (or any other unified node for that 

This is should be relatively straightforward since secondary opcode is a compile time constant, I guess your main concern looks like Value  / Identity routines which are inherited currently may not be usable, but one can always factor out the meaty parts of existing routines and then call it from new IR Value / Identity routines to avoid code-duplication..

> matter)? I don't think we can club optimizations for INT8 and FP8/FP16 as those optimizations would be vastly different between these types. Then should we have separate unified nodes for floating-point (FP8, FP16) and integer types (INT4, INT8)? 

What I meant was one Binary / Unary / Ternary node for each new reduced precision floating point type, as of now on x86 INT8 application is limited to dot product, its not a full blown ISA extension like FP16.

> In the current state, we are able to reuse FP32 optimizations very well for FP16 and where ever we do not want those optimizations to be applied, we just override it (in FP16 methods) and change accordingly.

Yes, that's our intention was while choosing this implementation strategy.  My intent here is to record our discussion on community mailing list so that we can accommodate suggestions from larger audience if any. 

> * From your description above, it looks like we will still have separate Vector Nodes right?

No, we can also have common IR for vectors.

> * Also, we might need some extra handling to decode the secondary opcode in the backend during matching.
> 

Matcher already handles constant integral operands, hence actual decision to emit specific instruction encoding based on secondary opcode can easily be deferred to encoding blocks without any hassle.
 
> I am not sure how much of benefit we might have with this approach. Also, do you know tentatively, by when is it planned to implement support for FP8/INT8 types in JDK? I think it will take quite some time for users to actually start using even FP16 and lot of future bugs/optimizations to be addressed for FP16.

FTR, here is link to our discussion on common vs separate 
IR https://github.com/jatin-bhateja/external_staging/blob/main/FP16Support/design_discussions.txt#L61

jatin-bhateja:
> > FTR, other missing intrinsics are conversion operations (Float16.longValue/intValue/doubleValue etc), x86 FP16 ISA has direct instructions for conversions from / to Float16 values..
> 
> I can add implementation for these conversions (if it's not already taken by anyone else or you :)

Hi @Bhavana-Kilambi , this will be really helpful. 

Also, your comments have been incorporated, kindly approve if looks ok.

Bhavana-Kilambi (approved):
The non-x86 part looks good to me.

Bhavana-Kilambi:
@jatin-bhateja Thanks for your responses for my questions above. I am still not sure how useful this approach would be for aarch64 as we still would have to generate separate match rules anyway for every operation (unlike x86). But this idea does sound interesting and  I am looking forward to a discussion with other members of OpenJDK community as well on this.

PR #1154, opened 2024-07-02
8333852: Allow flat array layout for value based classes.

jatin-bhateja:
Patch enables flat runtime layout for value based classes decorated with following annotations.
  1. @NullRestrictedArray:  Newly added annotation, ensures flat array elements are never assigned a null value during the lifetime of array, a violation to this rule results into a NullPointerException.
  2. @ImplicitylConstructible: Ensures default value assignment to flat array element at the time of new allocation.
  3. @LooselyConsistentValue:  Relaxes strict atomic update constraint of value objects.

In addition, all primitive box classes when annotated with #1 and #2 assume atomic update since they wrap around primitive type fields.

With this patch, an array of Float16 type has flat runtime layout which enables auto-vectorization of FP16 operations.

Best Regards,
Jatin

---------
### Progress
- [x] Change must not contain extraneous whitespace

### Issue
 * [JDK-8333852](https://bugs.openjdk.org/browse/JDK-8333852): Allow flat array layout for value based classes. (**Enhancement** - P4)

### Reviewers
 * [Bhavana Kilambi](https://openjdk.org/census#bkilambi) (@Bhavana-Kilambi - no project role)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/1154/head:pull/1154` \
`$ git checkout pull/1154`

Update a local copy of the PR: \
`$ git checkout pull/1154` \
`$ git pull https://git.openjdk.org/valhalla.git pull/1154/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 1154`

View PR using the GUI difftool: \
`$ git pr show -t 1154`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
<a href="https://git.openjdk.org/valhalla/pull/1154.diff">https://git.openjdk.org/valhalla/pull/1154.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/1154#issuecomment-2202974007)

PR #1211, opened 2024-08-19
8338061: Add support for FP16 unary and ternary operations

Bhavana-Kilambi:
This patch adds support for three unary operators - abs, neg and sqrt and a ternary operator - fma for FP16.
Both scalar and vector mid-end support along with aarch64 backend support are added.
Tested all FP16 tests successfully on aarch64.

---------
### Progress

### Issue
 * [JDK-8338061](https://bugs.openjdk.org/browse/JDK-8338061): Add support for FP16 unary and ternary operations (**Enhancement** - P4)

### Reviewers
 * [Jatin Bhateja](https://openjdk.org/census#jbhateja) (@jatin-bhateja - Committer) ⚠️ Review applies to [7538f860](https://git.openjdk.org/valhalla/pull/1211/files/7538f860113f934ffaa7f38b6680862737b76417)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/1211/head:pull/1211` \
`$ git checkout pull/1211`

`$ git checkout pull/1211` \
`$ git pull https://git.openjdk.org/valhalla.git pull/1211/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1211`

`$ git pr show -t 1211`

</details>

<a href="https://git.openjdk.org/valhalla/pull/1211.diff">https://git.openjdk.org/valhalla/pull/1211.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/1211#issuecomment-2296549981)


test/hotspot/jtreg/compiler/c2/irTests/ConvF2HFIdealizationTests.java
```diff
@@ -0,0 +1,62 @@
...
+
+    @Run(test = {"test1"})
+    public void runMethod() {
+        float f = RunInfo.getRandom().nextFloat();
+        Float16 f16 = Float16.valueOf(f);
+        assertResult(f16);
+    }
+
+    @DontCompile
+    public void assertResult(Float16 fp16) {
```
jatin-bhateja, line 49: We can use @Check annotation to comparing the actual and expected results on the lines of following existing test https://github.com/openjdk/valhalla/blob/lworld%2Bfp16/test/hotspot/jtreg/compiler/vectorization/TestFloat16VectorOps.java 

```
jshell> for (int i = 0; i < 65504; i++)  {
   ...>     float fi = Float.float16ToFloat(Float.floatToFloat16((float)i));
   ...>     if (Float16.float16ToRawShortBits(Float16.sqrt(Float16.valueOf(fi))) != Float.floatToFloat16((float)Math.sqrt(fi))) {
   ...>        System.out.println("Fist precision mismatch at = " + fi); break;
   ...>     }
   ...> }
```

We can even run through entire value ranges of Float16 type to empirically prove the equivalence..

src/hotspot/share/opto/compile.cpp
```diff
@@ -3677,13 +3677,18 @@ void Compile::final_graph_reshaping_impl(Node *n, Final_Reshape_Counts& frc, Uni
...
   case Op_SubHF:
   case Op_MulHF:
   case Op_DivHF:
   case Op_MaxHF:
   case Op_MinHF:
+  case Op_AbsHF:
+  case Op_NegHF:
+  case Op_SqrtHF:
+  case Op_FmaHF:
+  // Count all float operations that may use FPU
```
jatin-bhateja, line 3691: Hey @Bhavana-Kilambi , I am little worried about introducing specialized IR nodes as discussed in [pervious thread](https://github.com/openjdk/valhalla/pull/1196#issuecomment-2283627677),  even though our current strategy fits well with C2 compilation flow without causing much disturbance, going forward we may need to support other reduced precision types BF16 (AVX10.2 has several ISA extensions for BF16), FP8 and INT8, creating unique IR may not be desirable, lets revisit this after features completion with conversion ops.

jatin-bhateja (commented):
Looks good to me, apart from one test comment.

test/hotspot/jtreg/compiler/c2/irTests/ConvF2HFIdealizationTests.java
Bhavana-Kilambi: Sure, will do that. Thanks for reviewing.

src/hotspot/share/opto/compile.cpp
Bhavana-Kilambi: I have the conversion ops and isFinite/Infinite/NaN intrinsics still left to complete.

jatin-bhateja: I am in process of porting Float16 extension to VectorAPI into lworld+fp16, that will consolidate Flot16 effort on lworld+fp16 and leverage the unified backed implementation. With VectorAPI we will also need to support Reductions for Float16 vectors :-)

Bhavana-Kilambi: @jatin-bhateja Thanks for the effort. It will be good to test Float16 with vectorAPI and add the missing operations. I think we need to support Float16 reductions in auto-vectorization as well? I dont think those operations are added for FP16. 
Also about the design part, I was thinking instead of modifying the existing opcode or adding a separate secondary opcode for the operations, can we reuse the same existing FP32 nodes for FP16 as well? We can add flags for the FP32 IR nodes to check if it is a FP16 op or eventually FP8 op ? That way we can reuse the  same Value(), Ideal(), identity() methods and can modify those functions based on the value these flags returns in case we do not want a particular optimization to be applied for FP16/FP8? In the backend, we can use these flags to emit the correct instruction with the correct SIMD arrangement ? I guess this might introduce a lot of checks (to check value of the flags) in the compiler code.. What are your thoughts about it?

Bhavana-Kilambi: Also I will address review comments and post a new patch most likely after Aug 26th. I'll be on annual leave (plus UK bank holiday) from Aug 22nd to Aug 26th.

jatin-bhateja: > @jatin-bhateja Thanks for the effort. It will be good to test Float16 with vectorAPI and add the missing operations. I think we need to support Float16 reductions in auto-vectorization as well? I dont think those operations are added for FP16. Also about the design part, I was thinking instead of modifying the existing opcode or adding a separate secondary opcode for the operations, can we reuse the same existing FP32 nodes for FP16 as well? We can add flags for the FP32 IR nodes to check if it is a FP16 op or eventually FP8 op ? 

With new flag in existing IR approach we may end up adding complimentary predication checks in existing floating point instruction selection patterns.

> That way we can reuse the same Value(), Ideal(), identity() methods and can modify those functions based on the value these flags returns in case we do not want a particular optimization to be applied for FP16/FP8? 

Yes, idea is to reuse existing idealization routines, either by factoring out the meaty parts, or by inheritance. 

> In the backend, we can use these flags to emit the correct instruction with the correct SIMD arrangement ? I guess this might introduce a lot of checks (to check value of the flags) in the compiler code.. What are your thoughts about it?

I agree.
 
To me having a secondary opcode along with some code re-factoring around value / ideal / identity routines look cleaner, but we can think over alternatives which fits best in existing infrastructure.

test/jdk/java/foreign/enablenativeaccess/org/openjdk/foreigntest/libLinkerInvokerUnnamed.cpp
```diff
@@ -22,7 +22,7 @@
  */
 
 #include "jni.h"
-#include "testlib_threads.h"
+#include "testlib_threads.hpp"
```
Bhavana-Kilambi, line 25: I think this change was missed in the merge commit. I added it.

Bhavana-Kilambi:
Hi @jatin-bhateja , I have addressed review comments. Please take a look. Also, I have noticed that after the recent merge with `lworld` , the vector tests - `compiler/vectorization/TestFloat16VectorOps.java` and `compiler/vectorization/TestFloat16VectorReinterpretConv.java` failed. Looks like none of the FP16 operations are getting vectorized either. Any idea why? Initially I suspected if any of the code from this patch - https://github.com/openjdk/valhalla/pull/1154#issuecomment-2208350696 got botched up during the merge but it looks like all of that code is fine and the test to test flat array vectorization - `test/hotspot/jtreg/compiler/valhalla/inlinetypes/TestFlatArrayVectorization.java` also passes.

Bhavana-Kilambi:
Hi @jatin-bhateja , did you get a chance to go through my patch? Thanks!

jatin-bhateja:
> Hi @jatin-bhateja , I have addressed review comments. Please take a look. Also, I have noticed that after the recent merge with `lworld` , the vector tests - `compiler/vectorization/TestFloat16VectorOps.java` and `compiler/vectorization/TestFloat16VectorReinterpretConv.java` failed. Looks like none of the FP16 operations are getting vectorized either. Any idea why? Initially I suspected if any of the code from this patch - [#1154 (comment)](https://github.com/openjdk/valhalla/pull/1154#issuecomment-2208350696) got botched up during the merge but it looks like all of that code is fine and the test to test flat array vectorization - `test/hotspot/jtreg/compiler/valhalla/inlinetypes/TestFlatArrayVectorization.java` also passes.

Hi @Bhavana-Kilambi, I am working on fixing this at priority, @rgiulietti also faced problems in running the automated tests and pointed out the issue with multiple occurrence of libLinkerInvokerUnnamed.cpp.

Will address both these issues.

jatin-bhateja (approved):
Hi @Bhavana-Kilambi , Patch looks good to me, I did a merge and fixed the failing regressions. Kindly rebase and submit the patch, I will add x86 backed support for these new operations.

Best Regards,
Jatin

Bhavana-Kilambi:
Hi @jatin-bhateja , I have rebased my patch. Thanks for the approval. Could you please sponsor?

PR #1231, opened 2024-09-03
8339302: [lworld] C2: "assert(safepoints.length() == 0 || !res_type->is_inlinetypeptr()) failed: Inline type allocations should not have safepoint uses" with circular inline types

chhagedorn:
The circular inline type test cases show that two asserts are too strong: They assume that we can always replace inline types with `SafePointScalarObject` nodes before doing EA with its scalar replacement phase. If we fail to do so, then there must be a missed optimization opportunity. However, this only holds for non-circular inline types. When having circular inline types, we stop the (eager) scalarization at depth 1 because we could possibly do it endlessly in a recursion. Therefore, we could have a non-scalarized `CheckCastPP` inline type oop in a safepoint which triggers the asserts.

I propose to relax the assert by excluding circular inline types.

#### Additional Required Fix

I also had to adjust the code which creates a `SafePointScalarObject` in the scalar replacement after EA for such a circular inline type represented by a non-`InlineTypeNode` as follows.

Normally, we would call `InlineTypeNode::make_scalar_in_safepoint()` to scalarize an inline type. This adds an additional input to the safepoint for the `IsInit` input to check:

https://github.com/openjdk/valhalla/blob/884078f06a2bf15e4be768dd09d0790b63d8e015/src/hotspot/share/opto/inlinetypenode.cpp#L280-L287

For example, in `testCircularSafepointUse()`, `IsInit` is false for the inline type created for `v` at L3318 and we add a top input when creating `260 SafePointScalarObject`:

https://github.com/openjdk/valhalla/blob/d719d86c33f87bbdb37c5a4055fbbb5f81d7937f/test/hotspot/jtreg/compiler/valhalla/inlinetypes/TestNullableInlineTypes.java#L3315-L3321

image

The first field `v` is the `166 CheckCastPP` oop and the second field `i` has the value `188 ConI #23`. 

But when scalarizing the inline type field `v` later (i.e. `166 CheckCastPP`) after EA, without an `InlineTypeNode`, we miss to account for `IsInit`. We create a new `264 SafePointScalarObject` and directly append the fields after the last input (i.e. `188 ConI`) without `IsInit`:

image

The first field `v` is a `21 ConP #null` and the second field `i` is the value `74 ConI #34`.

Later, when processing the `264 SafePointScalarObject` at code generation, we try to read the `IsInit` input which was never added:

https://github.com/openjdk/valhalla/blob/884078f06a2bf15e4be768dd09d0790b63d8e015/src/hotspot/share/opto/output.cpp#L809-L816

The first input is `21 ConP #null` which is neither `top` nor an `int` and we crash at L816 when trying to convert the type to a `TypeInt`. 

The fix I propose for that is to add an int input with value 1 for circular inline type nodes when scalarizing them after EA to indicate that the buffer is always initialized (i.e. `IsInit` is true).

Thanks,
Christian

---------
### Progress

### Issue
 * [JDK-8339302](https://bugs.openjdk.org/browse/JDK-8339302): [lworld] C2: "assert(safepoints.length() == 0 || !res_type-&gt;is_inlinetypeptr()) failed: Inline type allocations should not have safepoint uses" with circular inline types (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/1231/head:pull/1231` \
`$ git checkout pull/1231`

`$ git checkout pull/1231` \
`$ git pull https://git.openjdk.org/valhalla.git pull/1231/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1231`

`$ git pr show -t 1231`

</details>

<a href="https://git.openjdk.org/valhalla/pull/1231.diff">https://git.openjdk.org/valhalla/pull/1231.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/1231#issuecomment-2326042611)

### DIFF
--- a/src/hotspot/share/opto/macro.cpp
+++ b/src/hotspot/share/opto/macro.cpp
@@ -865,7 +865,7 @@ SafePointScalarObjectNode* PhaseMacroExpand::create_scalarized_object_descriptio
       // InlineTypeNode::make_scalar_in_safepoint()). When having circular inline types, we stop scalarizing at depth 1
       // to avoid an endless recursion. Therefore, we do not have a SafePointScalarObjectNode node here, yet.
       // We are about to create a SafePointScalarObjectNode as if this is a normal object. Add an additional int input
-      // with value 1 to indicate that the buffer is always initialized (i.e. IsInit is true). This input is checked
+      // with value 1 which sets IsInit to true to indicate that the object is always non-null. This input is checked
       // later in PhaseOutput::filLocArray() for inline types.
       sfpt->add_req(_igvn.intcon(1));
     }
--- a/src/hotspot/share/opto/macro.cpp
+++ b/src/hotspot/share/opto/macro.cpp
@@ -708,7 +708,6 @@ bool PhaseMacroExpand::can_eliminate_allocation(PhaseIterGVN* igvn, AllocateNode
             }
           } else {
             // Add other uses to the worklist to process individually
-            // TODO will be fixed by 8328470
             worklist.push(use);
           }
         }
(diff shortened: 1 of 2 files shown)


src/hotspot/share/opto/macro.cpp
```diff
@@ -708,7 +708,6 @@ bool PhaseMacroExpand::can_eliminate_allocation(PhaseIterGVN* igvn, AllocateNode
             }
           } else {
             // Add other uses to the worklist to process individually
-            // TODO will be fixed by 8328470
```
chhagedorn, line 711: Just noticed that 8328470 was already fixed and thus the comment is outdated.

TobiHartmann (approved):
Great analysis, Christian! The fix looks good to me.

chhagedorn:
Thanks Tobias for the review and the offline suggestions!

/integrate

PR #1236, opened 2024-09-04
8331964: [lworld] C2: Support abstract value class fields

chhagedorn:
This patch adds support for abstract value class fields. To make it work, it required fixes in various places. The two main things to change was to account for fields in abstract value classes when collecting/accessing all fields which was not expected before and updating the way we handle larvals.

### Previous Larval Initialization
Before this change, larvals were always updated inside a single concrete value class constructor. It could have been done directly:
```
new MyValue -> MyValue.<init> (initialization here) -> Object.<init>
```
or by first delegating to another value type constructor inside the same value class which does the initiliazation:
```
new MyValue -> MyValue.<init> -> MyValue.<init> (initialization here) -> Object.<init>
```
It was not possible to distribute the initialization of an inline type to multiple constructors (i.e. at value class constructor cannot do a partial initialization).

Therefore, we could just detect any call from a value class constructor to an abstract constructor or the `Object` constructor with the same receiver and then mark the inline type as non-larval since it's completely initialized before calling the super constructor (abstract classes and `Object.<init>` cannot modify the inline type anymore). The only thing required after the call was to detect if we actually inlined another constructor with the same receiver within the same concrete value class. If that's the case, we need to update the exit map to propagate the initialized receiver to the caller.

This now changes: With fields in abstract value classes, a larval is not completely initialized after the concrete value class constructor is complete. We therefore needed to find a new way to ensure the correct initialization of larvals.

#### Problems
- A larval _could_ be fully initialized after calling the concrete value class constructor. But we could also have a situation where we still need to initialize a field in an abstract value class. Therefore, we cannot eagerly mark an inline type as non-larval once the concrete value class constructor is completed nor when an abstract value class constructor is completed (there could be multiple).
- The currently parsed method only has the map/JVM state from the direct caller. We therefore cannot simply detect the call to the `Object.<init>()` call at the end of chained value class constructor calls and then propagate the update to all the callers.

#### Solution
We therefore decided to only process larvals after the constructor call by looking at the just called method (regardless of whether is was inlined or not). If it was another concrete value class constructor, an abstract constructor or `Object.<init>` with the same receiver, then we know that after the call, the inline type is fully initialized. We only now mark it as non-larval and update the maps to propagate the update.

#### Other Fixes
I also needed to update other places where we update the receiver and the maps which often just checked if the holder class is a value type. This does not work anymore and needs to be updated (abstract value classes are not marked as inline type klasses).

#### Details
For more details, I refer to the code comments added in the patch for the various cases.

## Testing
I've copied the existing `TestValueConstructor` tests and created versions with abstract classes and added some new tests. The patch passed tier1-4 + stress except for one test failure in the stress testing: `TestNullableInlineTypes::test80()` failed with the very same assert and the same flags as `TestNullableInlineTypes::test81()` which was already disabled and tracked by [JDK-8325632](https://bugs.openjdk.org/browse/JDK-8325632). This suggests that it is most likely the same issue but triggered differently. To move forward with this patch, I also disabled `test80()`. We can get back to this later.

---------
### Progress

### Issue
 * [JDK-8331964](https://bugs.openjdk.org/browse/JDK-8331964): [lworld] C2: Support abstract value class fields (**Bug** - P2)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/1236/head:pull/1236` \
`$ git checkout pull/1236`

`$ git checkout pull/1236` \
`$ git pull https://git.openjdk.org/valhalla.git pull/1236/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1236`

`$ git pr show -t 1236`

</details>

<a href="https://git.openjdk.org/valhalla/pull/1236.diff">https://git.openjdk.org/valhalla/pull/1236.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/1236#issuecomment-2328574597)

### DIFF
--- a/src/hotspot/share/ci/ciInstanceKlass.hpp
+++ b/src/hotspot/share/ci/ciInstanceKlass.hpp
@@ -68,8 +68,9 @@ private:
   ciInstance*            _java_mirror;
 
   ciConstantPoolCache*   _field_cache;  // cached map index->field
+ public:
   GrowableArray<ciField*>* _nonstatic_fields;
-
+ private:
   int                    _has_injected_fields; // any non static injected fields? lazily initialized.
 
   // The possible values of the _implementor fall into following three cases:
--- a/src/hotspot/share/compiler/methodLiveness.cpp
+++ b/src/hotspot/share/compiler/methodLiveness.cpp
@@ -625,10 +625,10 @@ void MethodLiveness::BasicBlock::compute_gen_kill_single(ciBytecodeStream *instr
           (method->is_object_constructor() &&
            ((!holder->is_abstract() && holder->is_inlinetype()) ||
             holder->is_abstract()))) {
-        // Returning from Object.<init> implicitly registers a finalizer for the receiver if needed, so keep it alive.
+        // Returning from Object.<init> implicitly registers a finalizer for the receiver if needed, to keep it alive.
         // Value class constructors update the scalarized receiver. We need to keep it live so that we can find it after
         // (chained) constructor calls and propagate updates to the caller. If the holder of the constructor is abstract,
-        // we do not know if the constructor was called from a value class or not. We therefore keep the receiver of all
+        // we do not know if the constructor was called on a value class or not. We therefore keep the receiver of all
         // abstract constructors live.
         load_one(0);
       }
--- a/src/hotspot/share/compiler/methodLiveness.cpp
+++ b/src/hotspot/share/compiler/methodLiveness.cpp
@@ -624,7 +624,7 @@ void MethodLiveness::BasicBlock::compute_gen_kill_single(ciBytecodeStream *instr
       const bool abstract_klass = holder->is_abstract();
       const bool concrete_value_klass = !abstract_klass && holder->is_inlinetype();
       if (method->intrinsic_id() == vmIntrinsics::_Object_init ||
-          (method->is_object_constructor() && (concrete_value_klass || abstract_klass)) {
+          (method->is_object_constructor() && (concrete_value_klass || abstract_klass))) {
         // Returning from Object.<init> implicitly registers a finalizer for the receiver if needed, to keep it alive.
         // Value class constructors update the scalarized receiver. We need to keep it live so that we can find it after
         // (chained) constructor calls and propagate updates to the caller. If the holder of the constructor is abstract,
--- a/src/hotspot/share/compiler/methodLiveness.cpp
+++ b/src/hotspot/share/compiler/methodLiveness.cpp
@@ -618,18 +618,22 @@ void MethodLiveness::BasicBlock::compute_gen_kill_single(ciBytecodeStream *instr
       // These bytecodes have no effect on the method's locals.
       break;
 
-    case Bytecodes::_return:
-      if (instruction->method()->intrinsic_id() == vmIntrinsics::_Object_init ||
-          (instruction->method()->is_object_constructor() && instruction->method()->holder()->is_inlinetype())) {
-        // return from Object.init implicitly registers a finalizer
-        // for the receiver if needed, so keep it alive.
-        // Value class constructors update the scalarized receiver. Keep it live so that
-        // we can find it after (chained) constructor calls and propagate updates to the caller.
+    case Bytecodes::_return: {
+      ciMethod* method = instruction->method();
+      ciInstanceKlass* holder = method->holder();
+      if (method->intrinsic_id() == vmIntrinsics::_Object_init ||
+          (method->is_object_constructor() &&
+           ((!holder->is_abstract() && holder->is_inlinetype()) ||
+            holder->is_abstract()))) {
+        // Returning from Object.<init> implicitly registers a finalizer for the receiver if needed, so keep it alive.
+        // Value class constructors update the scalarized receiver. We need to keep it live so that we can find it after
+        // (chained) constructor calls and propagate updates to the caller. If the holder of the constructor is abstract,
+        // we do not know if the constructor was called from a value class or not. We therefore keep the receiver of all
+        // abstract constructors live.
         load_one(0);
       }
       break;
-
-
+    }
     case Bytecodes::_lload:
     case Bytecodes::_dload:
       load_two(instruction->get_index());
--- a/src/hotspot/share/compiler/methodLiveness.cpp
+++ b/src/hotspot/share/compiler/methodLiveness.cpp
@@ -621,10 +621,10 @@ void MethodLiveness::BasicBlock::compute_gen_kill_single(ciBytecodeStream *instr
     case Bytecodes::_return: {
       ciMethod* method = instruction->method();
       ciInstanceKlass* holder = method->holder();
+      const bool abstract_klass = holder->is_abstract();
+      const bool concrete_value_klass = !abstract_klass && holder->is_inlinetype();
       if (method->intrinsic_id() == vmIntrinsics::_Object_init ||
-          (method->is_object_constructor() &&
-           ((!holder->is_abstract() && holder->is_inlinetype()) ||
-            holder->is_abstract()))) {
+          (method->is_object_constructor() && (concrete_value_klass || abstract_klass)) {
         // Returning from Object.<init> implicitly registers a finalizer for the receiver if needed, to keep it alive.
         // Value class constructors update the scalarized receiver. We need to keep it live so that we can find it after
         // (chained) constructor calls and propagate updates to the caller. If the holder of the constructor is abstract,
--- a/src/hotspot/share/oops/inlineKlass.cpp
+++ b/src/hotspot/share/oops/inlineKlass.cpp
@@ -253,9 +253,9 @@ int InlineKlass::collect_fields(GrowableArray<SigEntry>* sig, int base_off) {
     if (fs.access_flags().is_static()) continue;
     int offset = base_off + fs.offset() - (base_off > 0 ? first_field_offset() : 0);
     // TODO 8284443 Use different heuristic to decide what should be scalarized in the calling convention
-    if (fs.current().is_flat()) {
+    if (fs.is_flat()) {
       // Resolve klass of flat field and recursively collect fields
-      Klass* vk = get_inline_type_field_klass(fs.current().index());
+      Klass* vk = get_inline_type_field_klass(fs.index());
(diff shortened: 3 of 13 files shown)


src/hotspot/share/ci/ciInstanceKlass.hpp
```diff
@@ -68,8 +68,9 @@ class ciInstanceKlass : public ciKlass {
   ciInstance*            _java_mirror;
 
   ciConstantPoolCache*   _field_cache;  // cached map index->field
+ public:
   GrowableArray<ciField*>* _nonstatic_fields;
```
chhagedorn, line 72: Could probably be improved in the future to avoid this direct public access.

src/hotspot/share/oops/inlineKlass.cpp
```diff
@@ -243,16 +243,19 @@ Klass* InlineKlass::value_array_klass_or_null() {
 // T_METADATA, drop everything until and including the closing
 // T_VOID) or the compiler point of view (each field of the inline
 // types is an argument: drop all T_METADATA/T_VOID from the list).
+//
+// Inline types could also have fields in abstract super value classes.
+// Use a HierarchicalFieldStream to get them as well.
 int InlineKlass::collect_fields(GrowableArray<SigEntry>* sig, int base_off) {
   int count = 0;
   SigEntry::add_entry(sig, T_METADATA, name(), base_off);
-  for (JavaFieldStream fs(this); !fs.done(); fs.next()) {
```
chhagedorn, line 249: `JavaFieldStream` failed to get the fields in the abstract class. Using `HierarchicalFieldStream` fixes this.

src/hotspot/share/compiler/methodLiveness.cpp
```diff
@@ -618,18 +618,22 @@ void MethodLiveness::BasicBlock::compute_gen_kill_single(ciBytecodeStream *instr
...
-        // for the receiver if needed, so keep it alive.
-        // Value class constructors update the scalarized receiver. Keep it live so that
-        // we can find it after (chained) constructor calls and propagate updates to the caller.
+    case Bytecodes::_return: {
+      ciMethod* method = instruction->method();
+      ciInstanceKlass* holder = method->holder();
+      if (method->intrinsic_id() == vmIntrinsics::_Object_init ||
+          (method->is_object_constructor() &&
+           ((!holder->is_abstract() && holder->is_inlinetype()) ||
+            holder->is_abstract()))) {
```
chhagedorn, line 627: We now keep more receivers live that might not be needed (non-value abstract classes). But I think this is the simplest/most straight forward solution since we do not have access to the actual receiver node here to check its bottom type as done in the code during parsing.

src/hotspot/share/opto/doCall.cpp
```diff
@@ -578,23 +578,6 @@ void Parse::do_call() {
   int       vtable_index       = Method::invalid_vtable_index;
   bool      call_does_dispatch = false;
 
-  // Detect the call to the object or abstract class constructor at the end of a value constructor to know when we are done initializing the larval
```
chhagedorn, line 581: Removed pre-call processing. Everything done post-call now.

src/hotspot/share/opto/graphKit.cpp
```diff
@@ -1970,13 +1990,18 @@ Node* GraphKit::set_results_for_java_call(CallJavaNode* call, bool separate_io_p
   }
 
   // We just called the constructor on a value type receiver. Reload it from the buffer
-  if (call->method()->is_object_constructor() && call->method()->holder()->is_inlinetype()) {
```
chhagedorn, line 1973: `call->method()->holder()->is_inlinetype()` false for abstract value classes -> use bottom type of receiver instead.

src/hotspot/share/opto/graphKit.cpp
```diff
@@ -1892,7 +1894,25 @@ void GraphKit::set_arguments_for_java_call(CallJavaNode* call, bool is_late_inli
       continue;
     } else if (arg->is_InlineType()) {
       // Pass inline type argument via oop to callee
-      arg = arg->as_InlineType()->buffer(this);
+      InlineTypeNode* inline_type = arg->as_InlineType();
```
chhagedorn, line 1897: Before this patch, we already marked an inline type as non-larval before the non-inlined super constructor call. So, we would always initialize the buffer when calling the super constructor. But now, we could still have marked larvals on which we call the super constructors. We therefore need to force an initialization in this case.

src/hotspot/share/opto/inlinetypenode.cpp
```diff
@@ -1327,3 +1339,11 @@ const Type* InlineTypeNode::Value(PhaseGVN* phase) const {
   }
   return t;
 }
+
+#ifndef PRODUCT
+void InlineTypeNode::dump_spec(outputStream* st) const {
+  if (_is_larval) {
+    st->print(" #larval");
+  }
+}
```
chhagedorn, line 1348: Added for easier debugging.

src/hotspot/share/opto/inlinetypenode.cpp
```diff
@@ -258,9 +258,15 @@ bool InlineTypeNode::field_is_null_free(uint index) const {
...
-      sfpt->as_CallJava()->method()->holder()->is_inlinetype() && sfpt->in(TypeFunc::Parms) == this) {
+  // We should not scalarize larvals in their own inline type constructor methods because their fields could still be
+  // updated. If we scalarize now and update the fields, we cannot propagate the update out of the constructor call
+  // because the scalarized fields are local to this method. We need to use the buffer to make the update visible to
+  // the outside (see also CompiledEntrySignature::compute_calling_conventions()).
+  if (is_larval() && sfpt->is_CallJava() && sfpt->as_CallJava()->method() != nullptr &&
+      sfpt->as_CallJava()->method()->is_object_constructor() && bottom_type()->is_inlinetypeptr() &&
+      sfpt->in(TypeFunc::Parms) == this) {
+    // Receiver of larval is always buffered in its constructor call because it was initially created outside the
+    // constructor.
```
TobiHartmann, line 269:
```suggestion
    // Receiver is always buffered because it's passed as oop, see special case in CompiledEntrySignature::compute_calling_conventions().
```

src/hotspot/share/opto/inlinetypenode.cpp
```diff
@@ -584,9 +590,15 @@ InlineTypeNode* InlineTypeNode::buffer(GraphKit* kit, bool safe_for_replace) {
...
-        // Larval will be initialized later
+
+      // Larval?
+      // - must_init set: We are about to call an abstract value class constructor or the Object constructor which is
+      //                  not inlined. It is therefore escaping and we must initialize the buffer because we have not
+      //                  done this, yet, for larvals (see second case).
+      // - Other larvals: We do not need to initialize the buffer because a larval could still be updated which will
+      //                  create a new buffer. Once the larval escapes, we will initialize the buffer (must_init set).
+      assert(!must_init || is_larval(), "must_init should only be set for larval");
+      if (!is_larval() || must_init) {
```
TobiHartmann, line 601: This is confusing: Shouldn't the default of `must_init` be true because that's the common case? And then we can assert here that it can only ever be false if `is_larval()` is true.

src/hotspot/share/opto/graphKit.cpp
```diff
@@ -1892,7 +1894,25 @@ void GraphKit::set_arguments_for_java_call(CallJavaNode* call, bool is_late_inli
...
+          (holder->is_abstract() || holder->is_java_lang_Object())) {
+        // We normally do not want to initialize the buffer of a larval because we are going to update it anyway which
+        // requires us to create a new buffer. But at this point, we are going to call the super constructor of this
+        // concrete or abstract inline type without inlining it. After that, the larval is completely initialized and
+        // thus not a larval anymore. We therefore need to force an initialization of the buffer to not lose all the
+        // field writes so far in case the buffer needs to be used (e.g. to read from when deoptimizing at runtime) or
+        // further updated in abstract super value class constructors which could have more fields to be initialized.
+        must_init_buffer = true;
+      }
+      arg = inline_type->buffer(this, true, must_init_buffer);
```
TobiHartmann, line 1915: I think `must_init_buffer` should be true by default and set to false here for larvals, see comments below.

TobiHartmann (approved):
Excellent work, Christian! Great analysis and proper fix. I know how tedious it was to find all these tests to trigger all construction variants and investigate the correct fix.

I just left a few code style comments.

chhagedorn:
Thanks a lot Tobias for your careful review, the suggestions, and all the offline discussions! I addressed your review comments.

src/hotspot/share/opto/graphKit.cpp
```diff
@@ -1895,22 +1895,25 @@ void GraphKit::set_arguments_for_java_call(CallJavaNode* call, bool is_late_inli
...
-        // further updated in abstract super value class constructors which could have more fields to be initialized.
-        must_init_buffer = true;
+      const bool is_receiver = (i == TypeFunc::Parms);
+      const bool is_abstract_or_object_klass_constructor = method->is_object_constructor() &&
+                                                           (holder->is_abstract() || holder->is_java_lang_Object());
+      const bool is_larval_receiver_on_super_constructor = is_receiver && is_abstract_or_object_klass_constructor;
+      bool must_init_buffer = true;
+      // We always need to buffer inline types when they are escaping. However, we can skip the actual initialization
+      // of the buffer if the inline type is a larval because we are going to update the buffer anyway which requires
+      // us to create a one. But there is one special case where we are still required to initialize the buffer: When
```
TobiHartmann, line 1907:
```suggestion
      // us to create a new one. But there is one special case where we are still required to initialize the buffer: When
```

TobiHartmann (approved):
That looks good, thanks for the updates!

TobiHartmann (approved):

chhagedorn:
As discussed offline, I've pushed another fix to correctly update the receiver after the `super()` call when an inline type is no longer a larval. We wrongly updated the original receiver in the map instead of the updated one which led to a wrong result when calling another method after `super()` which queries the receiver.

TobiHartmann (approved):
Good catch!

chhagedorn:
Thanks for the review, testing looked good!

/integrate

PR #1239, opened 2024-09-11
8339473: Add support for FP16 isFinite, isInfinite and isNaN

Bhavana-Kilambi:
This patch adds intrinsic support for FP16 isNaN, isFinite and isInfinite methods and also adds aarch64 backend for these intrinsics.

Tested all FP16 related tests successfully on aarch64.

---------
### Progress

### Issue
 * [JDK-8339473](https://bugs.openjdk.org/browse/JDK-8339473): Add support for FP16 isFinite, isInfinite and isNaN (**Enhancement** - P4)

### Reviewers
 * [Jatin Bhateja](https://openjdk.org/census#jbhateja) (@jatin-bhateja - Committer) ⚠️ Review applies to [a116fbe8](https://git.openjdk.org/valhalla/pull/1239/files/a116fbe887da51923c742a65b4eedb184fc785e5)
 * [Joe Darcy](https://openjdk.org/census#darcy) (@jddarcy - Committer)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/1239/head:pull/1239` \
`$ git checkout pull/1239`

`$ git checkout pull/1239` \
`$ git pull https://git.openjdk.org/valhalla.git pull/1239/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1239`

`$ git pr show -t 1239`

</details>

<a href="https://git.openjdk.org/valhalla/pull/1239.diff">https://git.openjdk.org/valhalla/pull/1239.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/1239#issuecomment-2343021222)


Bhavana-Kilambi:
Hi @jatin-bhateja , I have implemented these intrinsics in floating point in this patch. I also have another version where I tried implementing isInfinite in integer arithmetic for aarch64. I have created a draft PR here - https://github.com/openjdk/valhalla/pull/1242/commits/e095b2d47a4c28bece8fc0fcbc656da713feb132

I avoided generating the ReinterpretS2HF nodes in the mid-end which will make the "src" to be in a GPR and made the implementation in integer arithmetic easier and resulted in significant gains which can be seen in the commit message of the draft PR.

I have also tried implementing similar integer logic on x86 (in the draft PR) and I could see ~2.6x better performance than the case without the intrinsic (I have however not tried to implement the vpclass instruction for FP16 and compare performance).

Keeping the RenterpretS2HF nodes would mean that we need to move the `src` from an FPR to a GPR in the backend, before executing the integer operations which proved to be costly and not very beneficial.

What do you think of not using ReinterpretS2HF node for these intrinsics and keeping the `src` in a GPR instead? I know that technically FP16 is floating point (even if storage is `short`) and we have always implemented the `src` in an FPR so as to generate FP instructions but I was wondering if this tweak is ok if we see good performance gains? Looking forward to hearing your feedback. Thanks!

jatin-bhateja:
> Hi @jatin-bhateja , I have implemented these intrinsics in floating point in this patch. I also have another version where I tried implementing isInfinite in integer arithmetic for aarch64. I have created a draft PR here - [e095b2d](https://github.com/openjdk/valhalla/commit/e095b2d47a4c28bece8fc0fcbc656da713feb132)
> 
> I avoided generating the ReinterpretS2HF nodes in the mid-end which will make the "src" to be in a GPR and made the implementation in integer arithmetic easier and resulted in significant gains which can be seen in the commit message of the draft PR.
> 
> I have also tried implementing similar integer logic on x86 (in the draft PR) and I could see ~2.6x better performance than the case without the intrinsic (I have however not tried to implement the vpclass instruction for FP16 and compare performance).
> 
> Keeping the RenterpretS2HF nodes would mean that we need to move the `src` from an FPR to a GPR in the backend, before executing the integer operations which proved to be costly and not very beneficial.
> 
> What do you think of not using ReinterpretS2HF node for these intrinsics and keeping the `src` in a GPR instead? I know that technically FP16 is floating point (even if storage is `short`) and we have always implemented the `src` in an FPR so as to generate FP instructions but I was wondering if this tweak is ok if we see good performance gains? Looking forward to hearing your feedback. Thanks!

Hi @Bhavana-Kilambi , Thanks for the patch, I will add the corresponding x86 backend implimentation.

On taking the intrinsic route, c2 instruction selector is very flexible can can folding bigger graph pallets,  we can skip over folding graph fragment into a specialized IR node and directly matcher rule covering a bigger pattern and emit optimized instruction sequence for it. 

So what you have in this patch looks good to me.

jatin-bhateja:
Hi @Bhavana-Kilambi , 
Did you trying modifying the Java implementation and measuring the performance with intrinsic solution ? 
e.g.  isNaNTest :=  (shortValue & 0x7e00) == 0x7e00.

Bhavana-Kilambi:
> Hi @Bhavana-Kilambi , Did you trying modifying the Java implementation and measuring the performance with intrinsic solution ? e.g. isNaNTest := (Float16.float16ToRawShortValue() & 0x7e00) == 0x7e00 etc..

Hi @jatin-bhateja , I just tried for isNaN and the Java version is 13% faster than intrinsic (both with FP16 instructions and integer arithmetic). Do you think it's better to default to the java version instead of using intrinsics? I havent' tried for isFinite and isInfinite.

Bhavana-Kilambi:
Hi @jatin-bhateja , just tested the other two intrinsics as well. I do get huge performance gains (on aarch64) with the java versions. I used these - 

```
@IntrinsicCandidate
    public static boolean isNaN(Float16 f16) {
        return ((float16ToRawShortBits(f16) & 0x7e00) == 0x7e00);
    }
```

```
@IntrinsicCandidate
    public static boolean isInfinite(Float16 f16) {
        return ((float16ToRawShortBits(f16) ^ float16ToRawShortBits(POSITIVE_INFINITY)) & 0x7fff) == 0;
    }
```

```
@IntrinsicCandidate
    public static boolean isFinite(Float16 f16) {
        return !(isInfinite(f16) || isNaN(f16));
     }
```

I see these gains compared to the intrinsic version uploaded in the latest patch on this PR - 
```
Benchmark | (SIZE) | Gain
FP16Ops.isFiniteHF | 65504 | 1.18409353876134
FP16Ops.isInfiniteHF | 65504 | 1.57611028955779
FP16Ops.isNaNHF | 65504 | 1.48276431718062
```

jatin-bhateja:
> Hi @jatin-bhateja , just tested the other two intrinsics as well. I do get huge performance gains (on aarch64) with the java versions. I used these -
> 
> ```
> @IntrinsicCandidate
>     public static boolean isNaN(Float16 f16) {
>         return ((float16ToRawShortBits(f16) & 0x7e00) == 0x7e00);
>     }
> ```
> 
> ```
> @IntrinsicCandidate
>     public static boolean isInfinite(Float16 f16) {
>         return ((float16ToRawShortBits(f16) ^ float16ToRawShortBits(POSITIVE_INFINITY)) & 0x7fff) == 0;
>     }
> ```
> 
> ```
> @IntrinsicCandidate
>     public static boolean isFinite(Float16 f16) {
>         return !(isInfinite(f16) || isNaN(f16));
>      }
> ```
> 
> I see these gains compared to the intrinsic version uploaded in the latest patch on this PR -
> 
> ```
> Benchmark               (SIZE)    Gain
> FP16Ops.isFiniteHF      65504     1.18409353876134
> FP16Ops.isInfiniteHF    65504     1.57611028955779
> FP16Ops.isNaNHF         65504     1.48276431718062
> ```
> 
> Do you think it's better to not have any intrinsics at all for these operations?

Thanks @Bhavana-Kilambi , yes it will be optimal, your java versions still carries @IntrinsicCandidate, hope we are not accedently running intrinsics, if that is not the case then making only a java side change should suffice as its also easily auto-vectorizatable

Bhavana-Kilambi:
> > Hi @jatin-bhateja , just tested the other two intrinsics as well. I do get huge performance gains (on aarch64) with the java versions. I used these -
> > ```
> > @IntrinsicCandidate
> >     public static boolean isNaN(Float16 f16) {
> >         return ((float16ToRawShortBits(f16) & 0x7e00) == 0x7e00);
> >     }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ```
> > @IntrinsicCandidate
> >     public static boolean isInfinite(Float16 f16) {
> >         return ((float16ToRawShortBits(f16) ^ float16ToRawShortBits(POSITIVE_INFINITY)) & 0x7fff) == 0;
> >     }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ```
> > @IntrinsicCandidate
> >     public static boolean isFinite(Float16 f16) {
> >         return !(isInfinite(f16) || isNaN(f16));
> >      }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I see these gains compared to the intrinsic version uploaded in the latest patch on this PR -
> > ```
> > Benchmark               (SIZE)    Gain
> > FP16Ops.isFiniteHF      65504     1.18409353876134
> > FP16Ops.isInfiniteHF    65504     1.57611028955779
> > FP16Ops.isNaNHF         65504     1.48276431718062
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Do you think it's better to not have any intrinsics at all for these operations?
> 
> Thanks @Bhavana-Kilambi , yes it will be optimal, your java versions still carries @IntrinsicCandidate, hope we are not accedently running intrinsics, if that is not the case then making only a java side change should suffice as its also easily auto-vectorizatable

Thanks, I will make the necessary changes. I ran the tests with the `-XX:DisableIntrinsic` flag to test the java code and not run the intrinsics

Bhavana-Kilambi:
Hi @jatin-bhateja , I have uploaded a new patch with changes in the java implementation instead of using intrinsics for these three operations. I have new performance results with this patch here - 

```
Benchmark               Gain vs FP16 insns      Gain vs default
FP16Ops.isFiniteHF      1.64                    2.36
FP16Ops.isInfiniteHF    1.56                    2.97
FP16Ops.isNaNHF	        1.49                    1.62
```

The second column shows the ratio of thrpt between this patch and theprevious one which implements these operations as intrinsics (either in FP16 or integer arithmetic) and the third column is the ratio of thrpt between this patch and the default JDK implementation which requires the result to first be converted to FP32 before computing the result.

jatin-bhateja (approved):
LGTM.

Best Regards,
Jatin

PS: I have done some test refactoring, let me know if its ok with you.
[test_refactoring.txt](https://github.com/user-attachments/files/17059007/test_refactoring.txt)

Bhavana-Kilambi:
> LGTM.
> 
> Best Regards, Jatin
> 
> PS: I have done some test refactoring, let me know if its ok with you. [test_refactoring.txt](https://github.com/user-attachments/files/17059007/test_refactoring.txt)

Thanks for the refactor. I think this looks cleaner. I will update the patch.

src/java.base/share/classes/java/lang/Float16.java
```diff
@@ -444,7 +444,7 @@ public static Float16 valueOf(BigDecimal v) {
      * @see Double#isNaN(double)
      */
     public static boolean isNaN(Float16 f16) {
```
jddarcy, line 446: With changing the implementation if isNaN, isFinite, etc. to bit-based, the regression tests should be updated to cover more cases. In particular, different NaN bit patterns should be tested.

Bhavana-Kilambi: Hi @jddarcy , thank you for the review. I will add more input patterns to be tested.

Bhavana-Kilambi:
Hi @jddarcy @jatin-bhateja , I have updated the patch to test for the full range of NaN values for FP16. Please review. Thanks !

src/java.base/share/classes/java/lang/Float16.java
jddarcy: Thanks for the update; looks fine.

jddarcy (approved):

Bhavana-Kilambi:
Hi @jatin-bhateja , Hope the changes you suggested for the JTREG testcase are ok? I am marking it ready for integration -

/integrate

jatin-bhateja:
> Hi @jatin-bhateja , Hope the changes you suggested for the JTREG testcase are ok? I am marking it ready for integration -
> 
> /integrate

Hi @Bhavana-Kilambi ,  I think we can take same approach for unary operations - Float16.negate and Float16.abs and remove explicit intrinsification, java side implimentation is auto-vectorizable and performance should be at par with explicit intrinsic.

jatin-bhateja:
Since Float16 is a value type, hence C2 will always create a scalarized IR "InlineTypeNode" for it even if we don't intrinsify it. 
Consider following byte code sequence of Float16.abs
image

getfiled will directly return the scalar 'short' field which after subsequent logical and will return absolute value. I check SVE does have direct instruction to negate IEEE 754 binary16 encoded value, but don't know about its latency and performance compared to logical operations.  On x86 FP16 ISA does not have a direct negation instruction so java implementation should be equally performant.

Bhavana-Kilambi:
@jatin-bhateja Sure, I'll look into replacing intrinsics with java code for other unary operations too (or wherever there's bit manipulation possible). Thanks for the explanation!

PR #1251, opened 2024-09-24
8340125: [lw5] null checks are not being generated for method invocations and array writes

vicente-romero-oracle:
javac is not generating null checks for runtime execution for method invocations and array writes

---------
### Progress

### Issue
 * [JDK-8340125](https://bugs.openjdk.org/browse/JDK-8340125): [lw5] null checks are not being generated for method invocations and array writes (**Bug** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/1251/head:pull/1251` \
`$ git checkout pull/1251`

`$ git checkout pull/1251` \
`$ git pull https://git.openjdk.org/valhalla.git pull/1251/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1251`

`$ git pr show -t 1251`

</details>

<a href="https://git.openjdk.org/valhalla/pull/1251.diff">https://git.openjdk.org/valhalla/pull/1251.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/1251#issuecomment-2374370505)


src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/Gen.java
```diff
@@ -1977,8 +1969,8 @@ public void visitApply(JCMethodInvocation tree) {
         // the parameters of the method's external type (that is, any implicit
         // outer instance of a super(...) call appears as first parameter).
         MethodSymbol msym = (MethodSymbol)TreeInfo.symbol(tree.meth);
-        genArgs(tree.args,
-                msym.externalType(types).getParameterTypes());
+        List<Type> erasedPts = msym.externalType(types).getParameterTypes();
```
mcimadamore, line 1972: Should we revert these lines to whatever is in mainline?

vicente-romero-oracle: sure will do

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/TransTypes.java
```diff
@@ -485,6 +518,11 @@ public void visitMethodDef(JCMethodDecl tree) {
 
     public void visitVarDef(JCVariableDecl tree) {
         tree.vartype = translate(tree.vartype, null);
+        if (tree.init != null) {
+            if (types.isNonNullable(tree.type) && !types.isNonNullable(tree.init.type)) {
```
mcimadamore, line 522: I see why you did this. The synthetic casts are added by `TransTypes` in the various visitor methods. This makes sense because the casts are only required for method calls and field accesses.

My feeling is that most of this remains true also for null checks. E.g. you need to insert a check when you are translating a _variable_ (field, local, method parameter), or a _method call_ - in these cases there might be some nullness markers (on the variable, or on the method return) that should be taken into account, and compared with the "expected" nullness. If a "nullness" mismatch is found, a null-check should be inserted.

So I wonder if the main `translate` method should go from this:

```
translate(expr, expectedErasedType)
```

To this:

```
translate(expr, expectedErasedType, expectedNullness)
```

And the conversions "pushed" down in the visitor, in a similar way as what we do today for synthetic casts.

PR #1271, opened 2024-10-09
8315003: [lworld] C2: Implement full support for JDK-8287061 and JDK-8316991

chhagedorn:
This patch makes the improved allocation merge at EA work with inline types.

Before this patch, the assumption was that we do **not** have non-scalarized inline types at safepoints which EA is going to scalarize (and if we do, then there is a missed opportunity which can be considered a bug). We should have pushed all `InlineTypeNodes` down through phis in order to have them scalarized before EA.

This, however, changes with the improved allocation merge RFEs that have been merged into Valhalla ([JDK-8287061](https://bugs.openjdk.org/browse/JDK-8287061) and [JDK-8316991](https://bugs.openjdk.org/browse/JDK-8316991)). We could now the following graph as found in the test case:

image

Here, `272 Phi` merges the non-value type allocation `238 CheckCastPP` with the value type allocation `217 InlineType`. Before the improved allocation merge RFEs, we would not be able to remove the allocations. But now, we can do it at EA: We create a special `SafePointScalarMergeNode` which in case has `SafePointScalarObjectNodes` as inputs for the merged allocations. Additionally, each field of the merged allocations are added to the safepoint, which also includes the value type field `v2` of `V`:

image

We should not have safepoint uses for `InlineType` and thus now replace them with `SafePointScalarObjectNodes` as well (newly done with this patch). As a result, we have the following graph after EA:

image

Note that the safepoint node inputs for fields are well-defined to keep track of which field belong to each of the `SafePointScalarObjectNodes` of the `SafePointScalarMergeNode`.

### Additional Code Changes
- There was a bug in mainline with the new allocation merge implementation that is already fixed in mainline but not integrated in Valhalla, yet: [JDK-8335220](https://bugs.openjdk.org/browse/JDK-8335220). I cherry-picked these changes (L589-605 in escape.cpp).
- I added assertion code to verify that when we are going to scalar replace an inline type allocation that is an input into a phi at EA, we would not have been able to push the `InlineTypeNode` down (which should have been done earlier if possible).
- Re-enabled uncommented tests and removed problem-listed `AllocationMergesTests.java` that had been added with the allocation merge RFE.

### Future Work
There is currently an open bug for the allocation merges ([JDK-8335977](https://bugs.openjdk.org/browse/JDK-8335977)) which has been found by extending `AllocationMergesTests.java` with various deoptimization points and checking for correct values (similar to what we do in `TestValueConstruction.java`). Once this bug is fixed in mainline (it should extend this test and add these deoptimization checks), we should extend `AllocationMergesTests.java` and add tests for value types as well. I filed [JDK-8341856](https://bugs.openjdk.org/browse/JDK-8341856) to keep track of that.

Thanks,
Christian

### DIFF
--- a/src/hotspot/share/opto/escape.cpp
+++ b/src/hotspot/share/opto/escape.cpp
@@ -37,6 +37,7 @@
 #include "opto/cfgnode.hpp"
 #include "opto/compile.hpp"
 #include "opto/escape.hpp"
+#include "opto/inlinetypenode.hpp"
 #include "opto/macro.hpp"
 #include "opto/locknode.hpp"
 #include "opto/phaseX.hpp"
@@ -567,8 +568,7 @@ bool ConnectionGraph::can_reduce_check_users(Node* n, uint nesting) const {
     } else if (nesting > 0) {
       NOT_PRODUCT(if (TraceReduceAllocationMerges) tty->print_cr("Can NOT reduce Phi %d on invocation %d. Unsupported user %s at nesting level %d.", n->_idx, _invocation, use->Name(), nesting);)
       return false;
-    // TODO 8315003 Re-enable
-    } else if (use->is_CastPP() && false) {
+    } else if (use->is_CastPP()) {
       const Type* cast_t = _igvn->type(use);
       if (cast_t == nullptr || cast_t->make_ptr()->isa_instptr() == nullptr) {
 #ifndef PRODUCT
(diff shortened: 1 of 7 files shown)


src/hotspot/share/opto/escape.cpp
```diff
@@ -586,18 +586,23 @@ bool ConnectionGraph::can_reduce_check_users(Node* n, uint nesting) const {
...
-                n->dump(5);
-              }
-#endif
-              return false;
+            if (TraceReduceAllocationMerges) {
+              tty->print_cr("Can NOT reduce Phi %d on invocation %d. CastPP %d doesn't have simple control.", n->_idx, _invocation, use->_idx);
+              n->dump(5);
             }
+#endif
+            return false;
```
chhagedorn, line 605: cherry-picked mainline fix

test/hotspot/jtreg/compiler/valhalla/inlinetypes/TestAllocationMergeAndFolding.java
```diff
@@ -0,0 +1,100 @@
...
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+/*
+ * @test
+ * @bug 8315003
+ * @summary Test that removing allocation merges of non-value and value types at EA is working properly.
```
TobiHartmann, line 27:
```suggestion
 * @summary Test that removing allocation merges of non-value and value object at EA is working properly.
```

TobiHartmann (approved):
Great work, Christian. The detailed description in the PR is useful for future reference. Looks good to me.

chhagedorn:

PR #1275, opened 2024-10-11
8341757: Field layout computation allowing atomic and nullable flattening

fparain:
Initial implementation of fields layouts in order to support heap flattening with JEP 401 model.
The patch includes the generation of two new flat fields layouts: atomic and nullable.
The patch also includes the support in the interpreter (through the access API) to access the new layouts while respecting the new semantic (on x86_64 and aarch64).
Some areas do not support the new layouts yet, like JITs, Unsafe and arrays.

There's no automatic tests yet, but considering the impact those new layouts will have on the JVM, the priority was to make this code available to all developers as soon as possible. Automatic tests will be added in a following CR.

Fred

---------
### Progress

### Issue
 * [JDK-8341757](https://bugs.openjdk.org/browse/JDK-8341757): Field layout computation allowing atomic and nullable flattening (**Bug** - P4)

### Reviewers
 * [Dan Heidinga](https://openjdk.org/census#heidinga) (@DanHeidinga - no project role) ⚠️ Review applies to [0eacc549](https://git.openjdk.org/valhalla/pull/1275/files/0eacc54946f5a0ea9e2aa15c028f7e39ddeaf43c)
 * [David Simms](https://openjdk.org/census#dsimms) (@MrSimms - Committer) ⚠️ Review applies to [0eacc549](https://git.openjdk.org/valhalla/pull/1275/files/0eacc54946f5a0ea9e2aa15c028f7e39ddeaf43c)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/1275/head:pull/1275` \
`$ git checkout pull/1275`

`$ git checkout pull/1275` \
`$ git pull https://git.openjdk.org/valhalla.git pull/1275/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1275`

`$ git pr show -t 1275`

</details>

<a href="https://git.openjdk.org/valhalla/pull/1275.diff">https://git.openjdk.org/valhalla/pull/1275.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/1275#issuecomment-2431933639)


src/hotspot/share/interpreter/interpreterRuntime.cpp
```diff
@@ -273,7 +273,14 @@ JRT_ENTRY(void, InterpreterRuntime::uninitialized_static_inline_type_field(JavaT
...
       assert(field_k != nullptr, "Should have been loaded or an exception thrown above");
-      klass->set_inline_type_field_klass(index, InlineKlass::cast(field_k));
+      if (!field_k->is_inline_klass()) {
+        THROW_MSG(vmSymbols::java_lang_IncompatibleClassChangeError(),
+                  err_msg("class %s expects class %s to be a concrete value class but it is not",
+                  klass->name()->as_C_string(), field_k->external_name()));
+      }
+      InlineLayoutInfo* li = klass->inline_layout_info_adr(index);
+      li->set_klass(InlineKlass::cast(field_k));
+      li->set_kind(LayoutKind::REFERENCE);
```
coleenp, line 283: Do you always set these together?  Might be good to have one function like fill_in() that fills both values in.

src/hotspot/share/oops/inlineKlass.cpp
```diff
@@ -73,31 +73,32 @@ void InlineKlass::init_fixed_block() {
...
+  *((int*)adr_reset_value_offset()) = 0;
   *((address*)adr_value_array_klasses()) = nullptr;
+  *((int*)adr_first_field_offset()) = -1;
+  *((int*)adr_payload_size_in_bytes()) = -1;
+  *((int*)adr_payload_alignment()) = -1;
+  *((int*)adr_non_atomic_size_in_bytes()) = -1;
+  *((int*)adr_non_atomic_alignment()) = -1;
+  *((int*)adr_atomic_size_in_bytes()) = -1;
+  *((int*)adr_nullable_size_in_bytes()) = -1;
+  *((int*)adr_null_marker_offset()) = -1;
```
coleenp, line 85: hm don't know what this is.

src/hotspot/share/oops/instanceKlass.hpp
```diff
@@ -144,15 +144,54 @@ class InlineKlassFixedBlock {
...
+    return _null_marker_offset;
+  }
+  void set_null_marker_offset(int o) { _null_marker_offset = o; }
+
+  void metaspace_pointers_do(MetaspaceClosure* it);
+  MetaspaceObj::Type type() const { return InlineLayoutInfoType; }
+
+  static ByteSize klass_offset() { return in_ByteSize(offset_of(InlineLayoutInfo, _klass)); }
+  static ByteSize null_marker_offset_offset() { return in_ByteSize(offset_of(InlineLayoutInfo, _null_marker_offset)); }
+
```
coleenp, line 191: Should this be in inlineKlass.hpp ?  Is that included here?  It seems better there and then #include that.
Edit: it is complicated though isn't it?  Maybe it is better here but I don't like extra classes in header files, especially this one since it's sort of key.  Maybe it's better in its own header file?

coleenp (commented):
This is sort of a surface review of all but the actual field layout part (the meat of the change) until I ran out of steam.

src/hotspot/share/interpreter/interpreterRuntime.cpp
```diff
@@ -299,76 +306,71 @@ JRT_ENTRY(void, InterpreterRuntime::uninitialized_static_inline_type_field(JavaT
...
-  assert(field_holder->is_instance_klass(), "Sanity check");
-  InstanceKlass* klass = InstanceKlass::cast(field_holder);
-
-  assert(klass->field_is_flat(index), "Sanity check");
+  InstanceKlass* holder = InstanceKlass::cast(entry->field_holder());
+  assert(entry->field_holder()->field_is_flat(entry->field_index()), "Sanity check");
 
-  InlineKlass* field_vklass = InlineKlass::cast(klass->get_inline_type_field_klass(index));
+  InlineLayoutInfo* layout_info = holder->inline_layout_info_adr(entry->field_index());
+  InlineKlass* field_vklass = layout_info->klass();
```
DanHeidinga, line 317: Why `vklass`?  Valhalla Klass?

src/hotspot/share/oops/inlineKlass.hpp
```diff
@@ -166,7 +187,17 @@ class InlineKlass: public InstanceKlass {
...
+  static InlineKlass* cast(Klass* k) {
+    return const_cast<InlineKlass*>(cast(const_cast<const Klass*>(k)));
+  }
+
+  static const InlineKlass* cast(const Klass* k) {
+    assert(k != nullptr, "k should not be null");
+    assert(k->is_inline_klass(), "cast to InlineKlass");
+    return static_cast<const InlineKlass*>(k);
+  }
+  // static InlineKlass* cast(Klass* k);
```
DanHeidinga, line 200:
```suggestion
```

src/hotspot/share/oops/instanceKlass.cpp
```diff
@@ -175,7 +180,8 @@ bool InstanceKlass::field_is_null_free_inline_type(int index) const {
 bool InstanceKlass::is_class_in_loadable_descriptors_attribute(Symbol* name) const {
   if (_loadable_descriptors == nullptr) return false;
   for (int i = 0; i < _loadable_descriptors->length(); i++) {
-        Symbol* class_name = _constants->klass_at_noresolve(_loadable_descriptors->at(i));
+        // Symbol* class_name = _constants->klass_at_noresolve(_loadable_descriptors->at(i));
```
DanHeidinga, line 183:
```suggestion
```

DanHeidinga (commented):
General question - does the Valhalla CDS support need to be updated to handle the Klass pointers in InlineLayoutInfo when this is merged?

src/hotspot/share/classfile/fieldLayoutBuilder.cpp
```diff
@@ -810,57 +901,36 @@ void FieldLayoutBuilder::inline_class_field_sorting() {
...
+        group->add_flat_field(idx, vk, lk, size, alignment);
+        _inline_layout_info_array->adr_at(fieldinfo.index())->set_kind(lk);
+        _nonstatic_oopmap_count += vk->nonstatic_oop_map_count();
+        field_alignment = alignment;
+        _field_info->adr_at(idx)->field_flags_addr()->update_flat(true);
+        _field_info->adr_at(idx)->set_layout_kind(lk);
+        // default is atomic, but class file parsing could have set _must_be_atomic to false (@LooselyConsistentValue + checks)
+        // Presence of a must_be_atomic field must revert the _must_be_atomic flag of the holder to true
+        if (vk->must_be_atomic()) {
+          _must_be_atomic = true;
```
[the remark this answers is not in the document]
fparain: So far, we have avoided logging individual layout decisions. Only the final result is available as a diagnostic feature. If we start with this one, we could have many other decisions to add to the logging.

src/hotspot/share/classfile/fieldLayoutBuilder.cpp
```diff
@@ -1032,43 +1105,133 @@ void FieldLayoutBuilder::compute_inline_class_layout() {
...
     _payload_size_in_bytes = 0;
   }
-  if (_layout->first_field_block() == nullptr) {
-    assert(_is_abstract_value, "Concrete inline types must have at least one field");
-    _payload_size_in_bytes = 0;
-  } else {
-    _payload_size_in_bytes = _layout->last_block()->offset() - _layout->first_field_block()->offset();
+
+  // Determining if the value class is naturally atomic:
+  if ((!_layout->super_has_fields() && _declared_non_static_fields_count <= 1 && !_has_non_naturally_atomic_fields)
```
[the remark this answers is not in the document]
fparain: Empty abstract values have no instance fields. The ".empty" field is only added to concrete value classes.

src/hotspot/share/classfile/fieldLayoutBuilder.hpp
```diff
@@ -59,32 +64,31 @@ class LayoutRawBlock : public ResourceObj {
     REGULAR,               // primitive or oop field (including not flat inline type fields)
     FLAT,                  // flat field
     INHERITED,             // field(s) inherited from super classes
-    NULL_MARKER,           // stores the null marker for a flat field
-    INHERITED_NULL_MARKER  // a super-class used this slot to store a null marker
+    NULL_MARKER            // stores the null marker for a flat field
```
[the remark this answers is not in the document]
fparain: Correct. The null-marker is only added to concrete value classes. In fact, all the flat layouts are computed only for non-abstract value classes. If this is going to be an issue in the future, we should discuss another design before pushing those changes.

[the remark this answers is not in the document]
DanHeidinga: We have hypothesized that sealed hierarchies may be a target for future flattening optimizations as users will target the sealed type (typically an interface) for use as fields in their classes with a limited number (sometimes one!) of implementations.

These kinds of optimizations will require additional book-keeping for tracking the null marker, determining the klass of the value, and maybe additional info.  I don't think we want to delay this work for those future optimizations.... we can figure them out under a future RFE

src/hotspot/share/classfile/fieldLayoutBuilder.cpp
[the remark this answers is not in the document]
DanHeidinga: Fair enough.  This comment can be resolved.

src/hotspot/share/classfile/javaClasses.cpp
```diff
@@ -3782,11 +3782,11 @@ int java_lang_boxing_object::_value_offset;
 int java_lang_boxing_object::_long_value_offset;
 
 #define BOXING_FIELDS_DO(macro) \
-  macro(_value_offset,      integerKlass, "value", int_signature, false); \
+  macro(_value_offset,      byteKlass, "value", byte_signature, false); \
```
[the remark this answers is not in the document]
fparain: JavaClasses::check_offsets() verifies the offset of the boxing classes. The code used to handle two offsets, one taken rom j.l.Integer and used for Byte, Char, Short, Int and Float, and another one taken from j.l.Long for Long and Double. If you think of an object with compress class pointer, the first group has its field aligned on a 32bits boundary, and the second group has a field aligned on a 64bits boundary. The addition of ATOMIC_NULLABLE_FLAT layout changes of the layout of j.l.Integer and j.l.Float which now must have their field aligned on a 64bits boundary to be copiable atomically with their null marker. The new code now uses the j.l.Byte class as a reference for Byte, Char and Short, and has special cases to handle j.l.Integer and j.l.Float.

src/hotspot/share/gc/shared/barrierSetRuntime.cpp
```diff
@@ -28,12 +28,21 @@
...
 
+
 JRT_LEAF(void, BarrierSetRuntime::value_copy(void* src, void* dst, InlineKlass* md))
   assert(md->is_inline_klass(), "invariant");
-  HeapAccess<>::value_copy(src, dst, md);
+  HeapAccess<>::value_copy(src, dst, md, LayoutKind::PAYLOAD);  // FIXME Hard coded value for the transition
+JRT_END
+
+JRT_LEAF(void, BarrierSetRuntime::value_copy2(void* src, void* dst, InlineLayoutInfo* li))
+  HeapAccess<>::value_copy(src, dst, li->klass(), li->kind());
```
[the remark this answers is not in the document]
fparain: This is temporary. The original value_copy() method is still there because of arrays. Once arrays have been updated to support the new flat layouts, value_copy() will be removed, and value_copy2() will be renamed value_copy().

src/hotspot/share/classfile/fieldLayoutBuilder.cpp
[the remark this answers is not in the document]
DanHeidinga: Thanks for confirming.  This can be resolved,

src/hotspot/share/classfile/javaClasses.cpp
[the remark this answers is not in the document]
DanHeidinga: Sorry, I may be slow today but I would expect j.l.Byte and j.l.Short/Char to have different field layouts - given a 12byte header, the `value` field of j.l.Byte can start at the first free byte.  Using hand-hacked versions of the print fields output:
```
 @0 RESERVED 12/-
 @12 REGULAR 1/1 "value" B
```
and the `value` field of a j.l.Short needs additional alignment:
```
 @0 RESERVED 12/-
 @12 PADDING 4/1
 @16 REGULAR 2/2 "value" S
```

What am I missing?

src/hotspot/share/interpreter/interpreterRuntime.cpp
fparain: Not always. For instance fields, the class of those fields is usually known while parsing the class file (LoadableDescriptors attribute, @NullRestricted annotation), but the layout kind is only known after the field layout computation (post-processing of the parsed stream).

src/hotspot/share/classfile/classFileParser.cpp
```diff
@@ -1663,40 +1560,23 @@ void ClassFileParser::parse_fields(const ClassFileStream* const cfs,
...
-    FieldInfo::FieldFlags fflags(0);
-    fflags.update_injected(true);
-    AccessFlags aflags;
-    FieldInfo fi(aflags,
-                 (u2)vmSymbols::as_int(VM_SYMBOL_ENUM_NAME(empty_marker_name)),
-                 (u2)vmSymbols::as_int(VM_SYMBOL_ENUM_NAME(byte_signature)),
+    int idx = _temp_field_info->append(fi);
+    _temp_field_info->adr_at(idx)->set_index(idx);
+    _static_oop_count++;
+    // Inject static ".reset" field
```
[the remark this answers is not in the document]
coleenp: null_exemplar I like that!

src/hotspot/share/interpreter/interpreterRuntime.cpp
fparain: Value Klass. InstanceKlass pointers are often named ik, but now those two letters can be interpreted as either InstanceKlass or InlineKlass. So, I often use vk or vklass to designate a pointer to an InlineKlass.

src/hotspot/share/oops/instanceKlass.hpp
fparain: The InlineLayoutInfo array is a side data structure associated with InstanceKlass instances (any class with value fields needs this array, and this includes identity classes and abstract classes). This is why it is declared here, like the OopMapBlock. It could be declared in its own header file, but this would add a new file for just 30 lines of code.

src/hotspot/share/classfile/javaClasses.cpp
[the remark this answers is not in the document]
fparain: I don't know why, but we don't have the same layouts.
Here's the ones I have without support for NULLABLE_ATOMIC_FLAT:

```
Layout of class java/lang/Byte@0x7eae3c1357a0 extends java/lang/Number@0x7eae3c1357a0
Instance fields:
 @0 RESERVED 12/-
 @12 REGULAR 1/1 "value" B
```

```
Layout of class java/lang/Short@0x7eae3c1357a0 extends java/lang/Number@0x7eae3c1357a0
Instance fields:
 @0 RESERVED 12/-
 @12 REGULAR 2/2 "value" S
```

```
Layout of class java/lang/Integer@0x7eae3c1357a0 extends java/lang/Number@0x7eae3c1357a0
Instance fields:
 @0 RESERVED 12/-
 @12 REGULAR 4/4 "value" I
```

And the layouts of the same classes with NULLABLE_ATOMIC_FLAT enabled:

```
Layout of class java/lang/Byte@0x7bd87c1357a0 extends java/lang/Number@0x7bd87c1357a0
Instance fields:
 @0 RESERVED 12/-
 @12 REGULAR 1/1 "value" B
 @13 NULL_MARKER 1/1
```

```
Layout of class java/lang/Short@0x7bd87c1357a0 extends java/lang/Number@0x7bd87c1357a0
Instance fields:
 @0 RESERVED 12/-
 @12 REGULAR 2/2 "value" S
 @14 NULL_MARKER 1/1
```

```
Layout of class java/lang/Integer@0x7bd87c1357a0 extends java/lang/Number@0x7bd87c1357a0
Instance fields:
 @0 RESERVED 12/-
 @12 PADDING 4/1
 @16 REGULAR 4/4 "value" I
 @20 NULL_MARKER 1/1
```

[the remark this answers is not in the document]
DanHeidinga: Thanks Fred.  I think I get it now.  With the pre-valhalla layouts, the j.l.Integer takes 32 bits for the payload so it can start at `@12` and have the correct alignment.  In the Valhalla case, we need the 4bytes of padding to get the 64 bit payload (32 bit int + null marker) aligned correctly to be read atomically.

fparain:
> General question - does the Valhalla CDS support need to be updated to handle the Klass pointers in InlineLayoutInfo when this is merged?

The InstanceKlass' metaspace_pointers_do() method has been updated to process the _inline_layout_info_array, and the InlineLayoutInfo class has its own metaspace_pointers_do() method to process its _klass pointer. Did I miss something?

src/hotspot/share/oops/instanceKlass.cpp
```diff
@@ -1356,6 +1359,22 @@ void InstanceKlass::initialize_impl(TRAPS) {
...
+            Handle e(THREAD, PENDING_EXCEPTION);
+            CLEAR_PENDING_EXCEPTION;
+            {
+                EXCEPTION_MARK;
+                add_initialization_error(THREAD, e);
+                // Locks object, set state, and notify all waiting threads
+                set_initialization_state_and_notify(initialization_error, THREAD);
+                CLEAR_PENDING_EXCEPTION;
+            }
+            THROW_OOP(e());
```
DanHeidinga, line 1374: If I understand this code correctly, an OutOfMemoryError here will make the class unusable, which is a bit surprising.  Is it possible to instead disable all layouts except the REFERENCE if an OOM is thrown here?

DanHeidinga:
> > General question - does the Valhalla CDS support need to be updated to handle the Klass pointers in InlineLayoutInfo when this is merged?
> 
> The InstanceKlass' metaspace_pointers_do() method has been updated to process the _inline_layout_info_array, and the InlineLayoutInfo class has its own metaspace_pointers_do() method to process its _klass pointer. Did I miss something?

No, I missed `it->push(&_inline_layout_info_array, MetaspaceClosure::_writable);` in the instanceKlass.  Thanks for confirming

src/hotspot/share/oops/instanceKlass.cpp
fparain: Only the NULLABLE_ATOMIC_FLAT layout would need to be disabled. But if an OOM is thrown at this stage (class initialization), does it really worth the effort to try to recover the class? It is unlikely the application will be able to allocate an instance of this class anyway.

src/hotspot/share/gc/shared/barrierSetRuntime.hpp
```diff
@@ -38,7 +38,9 @@ class BarrierSetRuntime: public AllStatic {
 public:
   // Template interpreter...
   static void value_copy(void* src, void* dst, InlineKlass* md);
+  static void value_copy2(void* src, void* dst, InlineLayoutInfo* layout_info);
```
MrSimms, line 41: Are there not better names (e.g _ilo) than 2 ? Or is this temp naming ?

src/hotspot/cpu/x86/gc/shared/barrierSetAssembler_x86.hpp
```diff
@@ -50,6 +50,8 @@ class BarrierSetAssembler: public CHeapObj<mtGC> {
 
   virtual void value_copy(MacroAssembler* masm, DecoratorSet decorators,
                           Register src, Register dst, Register value_klass);
+  virtual void flat_field_copy(MacroAssembler* masm, DecoratorSet decorators,
+                          Register src, Register dst, Register inline_layout_info);
```
MrSimms, line 54: Does new flat_field_copy eventually replace value_copy ? Yes, this checks out (re-read the PR comments)

MrSimms (approved):
That's a lot of work, good job !

DanHeidinga (approved):
All my comments have been addressed

src/hotspot/share/gc/shared/barrierSetRuntime.hpp
fparain: This is a temporary naming, once flat arrays have been updated, we can remove the old value_copy() method, and reuse the name for the new method.

fparain:
Coleen, Dan, MrSimms, 
Thank you for your reviews of this huge patch.

Fred

PR #1283, opened 2024-10-23
8341414: Add support for FP16 conversion routines

Bhavana-Kilambi:
This patch adds intrinsic support for FP16 conversion routines to int/long/double and also the aarch64 backend support. This patch implements both scalar and vector versions for these conversions.

Performance numbers on aarch64 machine with SVE support :

```
Benchmark                         (vectorDim)   Gain
Float16OpsBenchmark.fp16ToDouble  1024          18.23
Float16OpsBenchmark.fp16ToInt     1024          1.93
Float16OpsBenchmark.fp16ToLong    1024          3.95
```

The Gain column is the ratio between thrpt of this patch and the thrpt with the intrinsics disabled (which generates FP32 arithmetic).

---------
### Progress

### Issue
 * [JDK-8341414](https://bugs.openjdk.org/browse/JDK-8341414): Add support for FP16 conversion routines (**Enhancement** - P4)

### Reviewers
 * [Jatin Bhateja](https://openjdk.org/census#jbhateja) (@jatin-bhateja - Committer) ⚠️ Review applies to [0809cedc](https://git.openjdk.org/valhalla/pull/1283/files/0809cedc7f46c82a403a67cd1acef0d2735691b2)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/1283/head:pull/1283` \
`$ git checkout pull/1283`

`$ git checkout pull/1283` \
`$ git pull https://git.openjdk.org/valhalla.git pull/1283/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1283`

`$ git pr show -t 1283`

</details>

<a href="https://git.openjdk.org/valhalla/pull/1283.diff">https://git.openjdk.org/valhalla/pull/1283.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/1283#issuecomment-2432346721)
</details>


src/java.base/share/classes/java/lang/Float16.java
```diff
@@ -649,6 +650,7 @@ public int intValue() {
      * @jls 5.1.3 Narrowing Primitive Conversion
      */
     @Override
+    @IntrinsicCandidate
```
jatin-bhateja, line 653: Should this be handled trough Idealization xform ?

ConvHF2F + ConvF2L => ConvHF2L

src/java.base/share/classes/java/lang/Float16.java
```diff
@@ -638,6 +638,7 @@ public short shortValue() {
...
     public int intValue() {
         return (int)floatValue();
```
jatin-bhateja, line 643: Should it be handled trough Idealization xform ?

ConvHF2F + ConvF2I => ConvHF2I

src/java.base/share/classes/java/lang/Float16.java
```diff
@@ -679,6 +681,7 @@ public float floatValue() {
      * @jls 5.1.2 Widening Primitive Conversion
      */
     @Override
+    @IntrinsicCandidate
     public double doubleValue() {
         return (double)floatValue();
```
jatin-bhateja, line 686: Should it be handled trough Idealization step ?

ConvHF2F + ConvF2D => ConvHF2D

src/java.base/share/classes/java/lang/Float16.java
jatin-bhateja: Matcher pattern may also suffice, but the problem will if ConvHF2F has multiple users.

jatin-bhateja: If you check JDK mainline Float16 C2 compiler support PR, John has suggested us to go with pattern matching, I have added more details to that draft PR.

Bhavana-Kilambi: Hi Jatin, thanks for the review comments. I did go through the comments on your draft PR earlier but I wasn't sure if we would be following their comments on the Valhalla branch as yet but I do agree with him about having too many intrinsics being an overkill. I'll remove the `@IntrinsicCandidate` here and add the code to do pattern matching in the mid-end.

Bhavana-Kilambi:
Hi @jatin-bhateja , I have uploaded a patch addressing your comments. Please review.

jatin-bhateja:
Hi @Bhavana-Kilambi , should also consider float16 to integral conversion, i.e. to int/short/long.
ISA supports direct instruction corresponding to these APIs.

Bhavana-Kilambi:
Hi Jatin, I have added support for float16 to int and long. Apologies for not missing the conversion to `short`. Will add that. This patch adds support for fp16 -> double as well. fp16->float is already taken care of.

Bhavana-Kilambi:
Hi @jatin-bhateja , Thanks for the reminder. I remember asking you in a previous email about the reverse conversions and I forgot about that myself. 
I am thinking if we have to intrinsify, can we not directly intrinsify Float16 valueOf() routines in Float16 itself instead of defining a new routine in Integer.java and then calling it in the Float16.valueOf() method and intrinsifying the one in Integer.java?

Bhavana-Kilambi:
I would prefer if I can do this in a separate patch please? I feel this patch is big enough. I will add some Ideal/Identity transformations as required for the new IR (for ex. ConvHF2I <-> ConvI2HF return the half float etc) in the new patch.

jatin-bhateja:
> Hi @jatin-bhateja , Thanks for the reminder. I remember asking you in a previous email about the reverse conversions and I forgot about that myself. I am thinking if we have to intrinsify, can we not directly intrinsify Float16 valueOf() routines in Float16 

Idea here is to avoid complexifying scalar intrinsic by delegating boxing to expander, otherwise we will also have to pass additional box type argument. Instead, we can rely explicit boxing happening in Java side and bank on escape analysis for its elimination thus directly exposing ConvI2HF to its user.

> new routine in Integer.java and then calling it in the Float16.valueOf() method and intrinsifying the one in Integer.java?
No, I am not suggesting adding <Primitive_Box_Type>.float16Value() APIs in existing primitive classes for time being, let Joe decide that. If you intrinsify leaf level wrapper routine, then we just need to plug that into Integer.float16Value(), we will lose this flexibility if we intrinsify Float16.valueOf(int).

jatin-bhateja:
> I would prefer if I can do this in a separate patch please? I feel this patch is big enough. I will add some Ideal/Identity transformations as required for the new IR (for ex. ConvHF2I <-> ConvI2HF return the half float etc) in the new patch.

Sounds good!

Bhavana-Kilambi:
> No, I am not suggesting to add <Primitive_Box_Type>.float16Value() API in existing primitive classes for time being, let Joe decide that. If you intrinsify leaf level wrapper routine, then we just need to plug that into Integer.float16Value(), we will lose this flexibility if we intrinsify Float16.valueOf(int).


Thanks for the explanation. So from what I understand - we currently have four valueOf() routines in Float16.java, float16 to int/long/float/double. The valueOf(long) calls valueOf(float) inside it which contains an intrinsified routine already so we have ConvL2F -> ConvF2HF being generated for that. I can add an Ideal optimization to generate ConvL2HF for this sequence.

For `valueOf(int)` which calls `valueOf(double)` -> I will add a new leaf routine something like `short d2hf(double d)` in Float16.java which I will intrinsify to generate ConvD2HF and this routine will be called in valueOf(int i) - 
```
public static Float16 valueOf(int value) {
return Float16(d2hf((double) value);
}
````
This would probably generate a ConvI2D -> ConvD2HF which can be optimized to ConvI2HF. The same routine can be called in `Float16 valueOf(double value)` as well which should generate ConvD2HF. Is this ok?

**Comment by Bhavana-Kilambi** [comment]
Can you also please review this patch? If it's all ok for you then this can be integrated and I can work on the next conv patch on top of this.

File: src/hotspot/share/opto/cfgnode.cpp [hunk 4]
```diff
@@ -1897,10 +1897,14 @@ static Node* split_flow_path(PhaseGVN *phase, PhiNode *phi) {
...
+static BasicType get_convert_type(Node* convert, const Type* type, const bool is_source=false) {
   int convert_op = convert->Opcode();
-  if (type->isa_int() && (convert_op == Op_ConvHF2F || convert_op == Op_ConvF2HF)) {
-    return T_SHORT;
+  if (type->isa_int()) {
+    if (is_source && convert_op == Op_ConvHF2I) {
+      return T_SHORT;
+    } else if (convert_op == Op_ConvHF2F || convert_op == Op_ConvF2HF || convert_op == Op_ConvHF2D || convert_op == Op_ConvHF2L) {
+      return T_SHORT;
+    }
```
jatin-bhateja, line 1907: This transformation is moving conversion across the Phi node thereby saving expensive conversion instructions.

`Phi (ConvHF2F src1) (ConvHF2F src2) =>   ConvHF2F  (Phi src1  src2)
`

For the Op_ConvF2HF case, type parameter will be of float type and thus we will never enter control flow between L1905 - L1907. 

I think it will be harmless if we remove convert_op == Op_ConvF2HF check...

jatin-bhateja: We can also update existing test for new conversion

test/hotspot/jtreg/compiler/c2/irTests/TestPhiDuplicatedConversion.java

jatin-bhateja (commented):
Hi @Bhavana-Kilambi ,
I have added few comments, patch looks good otherwise.

Best Regards,
Jatin

Bhavana-Kilambi:
Thank you Jatin for your comments. I will address them in a new patch soon.

jatin-bhateja (approved):
Hi @Bhavana-Kilambi , I am working on adding x86 backend support.
Kindly address the pending concerns in follow-up patch.

src/hotspot/cpu/aarch64/stubGenerator_aarch64.cpp
```diff
@@ -5493,14 +5493,35 @@ class StubGenerator: public StubCodeGenerator {
...
-    StubCodeMark mark(this, "StubRoutines", "float16ToFloat");
+    StubCodeMark mark(this, "StubRoutines", "float16 to float/double");
     address entry = __ pc();
     BLOCK_COMMENT("Entry:");
-    __ flt16_to_flt(v0, r0, v1);
+    switch (bt) {
+      case T_FLOAT:  __ flt16_to_flt(v0, r0, v1, T_FLOAT);  break;
+      case T_DOUBLE: __ flt16_to_flt(v0, r0, v1, T_DOUBLE); break;
+      default: ShouldNotReachHere();
+    }
```
jatin-bhateja, line 5507: Ok, I re-visited this, so conversion stubs for constant folding call direct FP16 to INT/LONG/DOUBLE instructions. 

This looks reasonable.  Though constant folding is something that happens at compile time so it may not result in any runtime penalty even if we remove the stubs and directly cast to target type after hf2f stub.

src/hotspot/share/opto/cfgnode.cpp
Bhavana-Kilambi: > This transformation is moving conversion across the Phi node thereby saving expensive conversion instructions.
> 
> `Phi (ConvHF2F src1) (ConvHF2F src2) => ConvHF2F (Phi src1 src2) `
> 
> For the Op_ConvF2HF case, type parameter will be of float type and thus we will never enter control flow between L1905 - L1907.
> 
> I think it will be harmless if we remove convert_op == Op_ConvF2HF check...

Hi @jatin-bhateja , I think that check should be there. The lines 1905-1907 will be reachable when get_convert_type() is called again with the dest_type (`short`) when it is supposed to return `short`. If this check is not there, it will return `T_INT`. For the following example - 
`sout[i] = rng.nextBoolean() ? Float.floatToFloat16(a) : Float.floatToFloat16(b)
`
it will generate `ConvF2I` instead of `ConvF2HF`.

Bhavana-Kilambi: I have added tests in `TestPhiDuplicatedConversion.java`

src/hotspot/share/opto/cfgnode.cpp
```diff
@@ -1897,10 +1897,14 @@ static Node* split_flow_path(PhaseGVN *phase, PhiNode *phi) {
 
 // Returns the BasicType of a given convert node and a type, with special handling to ensure that conversions to
 // and from half float will return the SHORT basic type, as that wouldn't be returned typically from TypeInt.
```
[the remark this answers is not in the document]
Bhavana-Kilambi: Done

test/jdk/java/lang/Float16/FP16ScalarOperations.java
```diff
@@ -146,6 +146,44 @@ public static void test_fin_inf_nan() {
...
+
+    public static void test_fp16ToDouble(Float16 fp16) {
+        double expected, actual;
+        actual = fp16.doubleValue();
+        expected = Float.valueOf(fp16.floatValue()).doubleValue();
+
+        if (expected != actual) {
+            System.out.println("Incorrest result for FP16 to double conversion. Expected value:" + expected + " Actual value: " + actual);
+        }
+    }
```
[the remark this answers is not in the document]
Bhavana-Kilambi: Some of these values are already being tested through the `special_values` array. I have modified to include a few more FP16 special values.

Bhavana-Kilambi:
Hi @jatin-bhateja , I have addressed your comments. Please review. Thanks!

Bhavana-Kilambi:
Hi @jatin-bhateja , can I integrate these changes if you feel these are good to go?

jatin-bhateja:
> Hi @jatin-bhateja , can I integrate these changes if you feel these are good to go?

Hi @Bhavana-Kilambi ,

Kindly integrate this, you now have committer's rights :-) 

Best Regards,
Jatin

FTR, upcasting from float16 to float conversions using existing runtime helpers is precision preserving, and constant folding for newly introduced scalar IR can be performed by subsequent integral casting thereby avoiding the need for newly introduced helpers.

Bhavana-Kilambi:
Yes :) I just wanted to make sure you are ok with the changes before I integrate.

So for constant folding we would do something like - convert half float value to float and then to integral/double right? But that would mean extra instructions in the backend when we have direct instructions to cast from half float to integral/double?

jatin-bhateja:
> Yes :) I just wanted to make sure you are ok with the changes before I integrate.
> 
> So for constant folding we would do something like - convert half float value to float and then to integral/double right? But that would mean extra instructions in the backend when we have direct instructions to cast from half float to integral/double?

Since constants are folded at compile time, adding runtime helpers just for constant folding looks like an overkill. Post folding compile will directly operate over constant IR.

Best Regards,
Jatin

PR #1314, opened 2024-12-06
8345250: [lworld] C2: Array loads and stores on inexact flat arrays cause crashes

chhagedorn:
This PR fixes crashes that occur when trying to load/store from/to a flat array from which we statically know it is flat but we do not know the exact layout because the klass is inexact.

### The Layout of a Flat Array is only Statically Known If the Klass is Exact
Currently, the code for getting the array element address in `GraphKit::array_element_address()` assumes that whenever we have an array from which we know that it is flat, then we also know the exact array layout (i.e. assume the value klass is exact):
https://github.com/openjdk/valhalla/blob/5e876d607aba68e921cb8fa6a1a012acd1735825/src/hotspot/share/opto/graphKit.cpp#L1835-L1838

This does not need to be true as shown in the following test case where we have a flat array with an inexact klass/type:
https://github.com/openjdk/valhalla/blob/95fc9c71bffcf5b92a4cdbfb34ad9e3149266f79/test/hotspot/jtreg/compiler/valhalla/inlinetypes/TestLWorld.java#L4568-L4578

### Emit Runtime Call to Load/Store from/to the Correct Array Element Address
What we actually would need for a flat array with an inexact klass is to emit a runtime call to load/store from/to the correct array element address. We cannot do better statically because we do not know the exact layout.

### Challenges
However, the problem is that in `GraphKit::array_element_address()` we somehow need to bail out back to `Parse::array_load/store()` such that we can emit the runtime call there. We could use a sentinel node because we do not actually need the element address node. But this could mess with some GVN transformations in between. What we do instead is to just fall to the `else` case in `GraphKit::array_element_address()` for non-flat arrays and still construct the element address node, even though it's not needed (will be folded away later):
https://github.com/openjdk/valhalla/blob/0858ce110da58ed7f1b634ae381e7c4eb0f37c3a/src/hotspot/share/opto/graphKit.cpp#L1840-L1849

### Additionally Required Fixes
1. `TypeAryKlassPtr::cast_to_exactness()` asserts that all null-free or flat arrays are exact which is not the case as shown above. Fix: Adjust assert.
2. `ciArrayKlass::make()` assumes that all null-free arrays are concrete value classes which does not need to be the case as shown above. Fix: Add check whether the klass is an inline type (which is only true for concrete inline type klasses).
3. `Parse::array_store_check()` needs to pass `null_free` to the `gen_checkcast()` in case we have a null-free flat array that is inexact (i.e. either `Object` or an abstract value class) and thus `is_inlinetypeptr()` is false. If we don't do that, we will hit this assert with `-XX:-UncommonNullCast` that ensures that flat arrays are null-free:
https://github.com/openjdk/valhalla/blob/265189e2b311daaa8cac3d387c3a70051b1e419a/src/hotspot/share/opto/parse2.cpp#L235-L237
The right/complete way to fix this would be to change `TypeAry::is_null_free()` to also handle non-exact inline types. However, this is invasive since we use this check in many places. Changing this method could have a bigger impact due to possibly having more places where we assume that only exact inline types are null-free. I filed [JDK-8345681](https://bugs.openjdk.org/browse/JDK-8345681) to revisit this. For this PR, I just implemented a point fix in `array_store_check()` for now.

### Applied some Additional Renamings and Simple Refactorings
While trying to better understand the `array_load/store()`, I've applied some renamings and simple refactorings. I'm planning to do some more refactorings to reduce the size of `array_load/store()`. But I will do that separately with [JDK-8345696](https://bugs.openjdk.org/browse/JDK-8345696).

Thanks,
Christian

---------
### Progress

### Issue
 * [JDK-8345250](https://bugs.openjdk.org/browse/JDK-8345250): [lworld] C2: Array loads and stores on inexact flat arrays cause crashes (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/1314/head:pull/1314` \
`$ git checkout pull/1314`

`$ git checkout pull/1314` \
`$ git pull https://git.openjdk.org/valhalla.git pull/1314/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1314`

`$ git pr show -t 1314`

</details>

<a href="https://git.openjdk.org/valhalla/pull/1314.diff">https://git.openjdk.org/valhalla/pull/1314.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/1314#issuecomment-2523795743)
</details>

### DIFF
--- a/src/hotspot/share/ci/ciArrayKlass.cpp
+++ b/src/hotspot/share/ci/ciArrayKlass.cpp
@@ -105,23 +105,25 @@ bool ciArrayKlass::is_leaf_type() {
 ciArrayKlass* ciArrayKlass::make(ciType* element_type, bool null_free) {
   if (element_type->is_primitive_type()) {
     return ciTypeArrayKlass::make(element_type->basic_type());
-  } else {
-    ciKlass* klass = element_type->as_klass();
-    if (null_free && klass->is_loaded()) {
-      GUARDED_VM_ENTRY(
-        EXCEPTION_CONTEXT;
-        Klass* ak = InlineKlass::cast(klass->get_Klass())->value_array_klass(THREAD);
-        if (HAS_PENDING_EXCEPTION) {
-          CLEAR_PENDING_EXCEPTION;
-        } else if (ak->is_flatArray_klass()) {
-          return CURRENT_THREAD_ENV->get_flat_array_klass(ak);
-        } else if (ak->is_objArray_klass()) {
-          return CURRENT_THREAD_ENV->get_obj_array_klass(ak);
-        }
-      )
-    }
-    return ciObjArrayKlass::make(klass);
   }
+
+  ciKlass* klass = element_type->as_klass();
+  assert(!null_free || !klass->is_loaded() || klass->is_inlinetype() || klass->is_abstract() ||
+         klass->is_java_lang_Object(), "only value classes are null free");
+  if (null_free && klass->is_loaded() && klass->is_inlinetype()) {
+    GUARDED_VM_ENTRY(
+      EXCEPTION_CONTEXT;
+      Klass* ak = InlineKlass::cast(klass->get_Klass())->value_array_klass(THREAD);
+      if (HAS_PENDING_EXCEPTION) {
+        CLEAR_PENDING_EXCEPTION;
+      } else if (ak->is_flatArray_klass()) {
+        return CURRENT_THREAD_ENV->get_flat_array_klass(ak);
+      } else if (ak->is_objArray_klass()) {
+        return CURRENT_THREAD_ENV->get_obj_array_klass(ak);
+      }
+    )
+  }
+  return ciObjArrayKlass::make(klass);
 }
 
 int ciArrayKlass::array_header_in_bytes() {
--- a/src/hotspot/share/opto/graphKit.cpp
+++ b/src/hotspot/share/opto/graphKit.cpp
@@ -3504,7 +3504,7 @@ Node* GraphKit::gen_checkcast(Node *obj, Node* superklass, Node* *failure_contro
   const TypeKlassPtr* improved_klass_ptr_type = klass_ptr_type->try_improve();
   const TypeOopPtr* toop = improved_klass_ptr_type->cast_to_exactness(false)->as_instance_type();
   bool safe_for_replace = (failure_control == nullptr);
-  assert(!null_free || toop->is_inlinetypeptr(), "must be an inline type pointer");
+  assert(!null_free || toop->can_be_inline_type(), "must be an inline type pointer");
 
   // Fast cutout:  Check the case that the cast is vacuously true.
   // This detects the common cases where the test will short-circuit
--- a/src/hotspot/share/opto/graphKit.cpp
+++ b/src/hotspot/share/opto/graphKit.cpp
@@ -1835,7 +1835,18 @@ void GraphKit::access_clone(Node* src, Node* dst, Node* size, bool is_array) {
 Node* GraphKit::array_element_address(Node* ary, Node* idx, BasicType elembt,
                                       const TypeInt* sizetype, Node* ctrl) {
   const TypeAryPtr* arytype = _gvn.type(ary)->is_aryptr();
-  uint shift = arytype->is_flat() ? arytype->flat_log_elem_size() : exact_log2(type2aelembytes(elembt));
+  assert(!arytype->is_flat() || elembt == T_OBJECT, "element type of flat arrays are T_OBJECT");
+  uint shift;
+  if (arytype->is_flat() && arytype->klass_is_exact()) {
+    // We can only determine the flat array layout statically if the klass is exact. Otherwise, we could have different
+    // value classes at runtime with a potentially different layout. The caller needs to fall back to call
+    // load/store_unknown_inline_Type() at runtime. We could return a sentinel node for the non-exact case but that
+    // might mess with other GVN transformations in between. Thus, we just continue in the else branch normally, even
+    // though we don't need the address node in this case and throw it away again.
+    shift = arytype->flat_log_elem_size();
+  } else {
+    shift = exact_log2(type2aelembytes(elembt));
+  }
   uint header = arrayOopDesc::base_offset_in_bytes(elembt);
 
   // short-circuit a common case (saves lots of confusing waste motion)
--- a/src/hotspot/share/opto/parse2.cpp
+++ b/src/hotspot/share/opto/parse2.cpp
@@ -76,57 +76,64 @@ void Parse::array_load(BasicType bt) {
   Node* adr = array_addressing(bt, 0, elemtype);
   if (stopped())  return;     // guaranteed null or range check
 
-  Node* idx = pop();
-  Node* ary = pop();
+  Node* array_index = pop();
+  Node* array = pop();
 
   // Handle inline type arrays
-  const TypeOopPtr* elemptr = elemtype->make_oopptr();
-  const TypeAryPtr* ary_t = _gvn.type(ary)->is_aryptr();
-  if (ary_t->is_flat()) {
+  const TypeOopPtr* element_ptr = elemtype->make_oopptr();
+  const TypeAryPtr* array_type = _gvn.type(array)->is_aryptr();
+  if (array_type->is_flat()) {
     // Load from flat inline type array
-    Node* vt = InlineTypeNode::make_from_flat(this, elemtype->inline_klass(), ary, adr);
-    push(vt);
+    Node* inline_type;
+    if (element_ptr->klass_is_exact()) {
+      inline_type = InlineTypeNode::make_from_flat(this, elemtype->inline_klass(), array, adr);
+    } else {
+      // Element type of flat array is not exact. Therefore, we cannot determine the flat array layout statically.
+      // Emit a runtime call to load the element from the flat array.
+      inline_type = load_from_unknown_flat_array(array, array_index, element_ptr);
+      inline_type = record_profile_for_speculation_at_array_load(inline_type);
+    }
+    push(inline_type);
     return;
-  } else if (ary_t->is_null_free()) {
-    // Load from non-flat inline type array (elements can never be null)
-    bt = T_OBJECT;
-  } else if (!ary_t->is_not_flat()) {
+  }
+
+  if (!array_type->is_not_flat()) {
     // Cannot statically determine if array is a flat array, emit runtime check
-    assert(UseFlatArray && is_reference_type(bt) && elemptr->can_be_inline_type() && !ary_t->is_not_null_free() &&
-           (!elemptr->is_inlinetypeptr() || elemptr->inline_klass()->flat_in_array()), "array can't be flat");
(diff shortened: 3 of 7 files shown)


src/hotspot/share/ci/ciArrayKlass.cpp
```diff
@@ -105,23 +105,25 @@ bool ciArrayKlass::is_leaf_type() {
...
-        }
-      )
-    }
-    return ciObjArrayKlass::make(klass);
   }
+
+  ciKlass* klass = element_type->as_klass();
+  assert(!null_free || !klass->is_loaded() || klass->is_inlinetype() || klass->is_abstract() ||
+         klass->is_java_lang_Object(), "only value classes are null free");
+  if (null_free && klass->is_loaded() && klass->is_inlinetype()) {
```
chhagedorn, line 113: Added `klass->is_inlinetype()` to be able to get the `InlineKlass`.

src/hotspot/share/opto/graphKit.cpp
```diff
@@ -3493,7 +3504,7 @@ Node* GraphKit::gen_checkcast(Node *obj, Node* superklass, Node* *failure_contro
   const TypeKlassPtr* improved_klass_ptr_type = klass_ptr_type->try_improve();
   const TypeOopPtr* toop = improved_klass_ptr_type->cast_to_exactness(false)->as_instance_type();
   bool safe_for_replace = (failure_control == nullptr);
-  assert(!null_free || toop->is_inlinetypeptr(), "must be an inline type pointer");
+  assert(!null_free || toop->can_be_inline_type(), "must be an inline type pointer");
```
chhagedorn, line 3507: When we pass now `null_free == true` for inexact inline types, `is_inlinetypeptr()` is false (only true for concrete value classes). `can_be_inline_type()` is the next better thing we could check.

src/hotspot/share/opto/parse2.cpp
```diff
@@ -76,57 +76,64 @@ void Parse::array_load(BasicType bt) {
...
-    Node* vt = InlineTypeNode::make_from_flat(this, elemtype->inline_klass(), ary, adr);
-    push(vt);
+    Node* inline_type;
+    if (element_ptr->klass_is_exact()) {
+      inline_type = InlineTypeNode::make_from_flat(this, elemtype->inline_klass(), array, adr);
+    } else {
+      // Element type of flat array is not exact. Therefore, we cannot determine the flat array layout statically.
+      // Emit a runtime call to load the element from the flat array.
+      inline_type = load_from_unknown_flat_array(array, array_index, element_ptr);
+      inline_type = record_profile_for_speculation_at_array_load(inline_type);
```
chhagedorn, line 94: Cover missing case, other changes in `array_load()` are just renaming/refactoring

src/hotspot/share/opto/parse2.cpp
```diff
@@ -135,37 +142,11 @@ void Parse::array_load(BasicType bt) {
...
-
-        insert_mem_bar_volatile(Op_MemBarCPUOrder, C->get_alias_index(TypeAryPtr::INLINES));
-
-        // Keep track of the information that the inline type is in flat arrays
-        const Type* unknown_value = elemptr->is_instptr()->cast_to_flat_in_array();
-        buffer = _gvn.transform(new CheckCastPPNode(control(), buffer, unknown_value));
-
+        // Element type is unknown, and thus we cannot statically determine the exact flat array layout. Emit a
+        // runtime call to correctly load the inline type element from the flat array.
+        Node* inline_type = load_from_unknown_flat_array(array, array_index, element_ptr);
```
chhagedorn, line 147: Extracted to new method such that we can reuse this above.

src/hotspot/share/opto/parse2.cpp
```diff
@@ -76,57 +76,64 @@ void Parse::array_load(BasicType bt) {
...
+      // Emit a runtime call to load the element from the flat array.
+      inline_type = load_from_unknown_flat_array(array, array_index, element_ptr);
+      inline_type = record_profile_for_speculation_at_array_load(inline_type);
+    }
+    push(inline_type);
     return;
-  } else if (ary_t->is_null_free()) {
-    // Load from non-flat inline type array (elements can never be null)
-    bt = T_OBJECT;
-  } else if (!ary_t->is_not_flat()) {
```
chhagedorn, line 93: Swapped cases and inserted null free case further down after this case.

src/hotspot/share/opto/parse2.cpp
```diff
@@ -175,98 +156,143 @@ void Parse::array_load(BasicType bt) {
...
+        // Re-execute flat array store if buffering triggers deoptimization
+        PreserveReexecuteState preexecs(this);
+        inc_sp(3);
+        jvms()->set_should_reexecute(true);
+        stored_value_casted->as_InlineType()->store_flat(this, array, adr, nullptr, 0, MO_UNORDERED | IN_HEAP | IS_ARRAY);
+      } else {
+        // Element type of flat array is not exact. Therefore, we cannot determine the flat array layout statically.
+        // Emit a runtime call to store the element to the flat array.
+        store_to_unknown_flat_array(array, array_index, stored_value_casted);
+      }
```
chhagedorn, line 267: New case, other code is just refactoring in `array_store()`.

src/hotspot/share/opto/parse2.cpp
```diff
@@ -175,98 +156,143 @@ void Parse::array_load(BasicType bt) {
...
       return;
-    } else if (ary_t->is_null_free()) {
-      // Store to non-flat inline type array (elements can never be null)
-      assert(!tval->maybe_null(), "should be guaranteed by array store check");
+    }
+    if (array_type->is_null_free()) {
```
chhagedorn, line 270: `else if` can be turned into simple `if` because the other case returns.

src/hotspot/share/opto/parse2.cpp
```diff
@@ -275,60 +301,67 @@ void Parse::array_store(BasicType bt) {
...
-          // Element type is known, cast and store to flat representation
-          assert(vk->flat_in_array() && elemtype->maybe_null(), "never/always flat - should be optimized");
-          ciArrayKlass* array_klass = ciArrayKlass::make(vk, /* null_free */ true);
-          const TypeAryPtr* arytype = TypeOopPtr::make_from_klass(array_klass)->isa_aryptr();
-          casted_ary = _gvn.transform(new CheckCastPPNode(control(), casted_ary, arytype));
-          Node* casted_adr = array_element_address(casted_ary, idx, T_OBJECT, arytype->size(), control());
-          if (!val->is_InlineType()) {
-            assert(!gvn().type(val)->maybe_null(), "inline type array elements should never be null");
-            val = InlineTypeNode::make_from_oop(this, val, vk);
+        if (!stopped()) {
```
chhagedorn, line 310: Simplified cases since both cases checked `!stopped()` before.

src/hotspot/share/opto/parse2.cpp
```diff
@@ -275,60 +301,67 @@ void Parse::array_store(BasicType bt) {
...
+                                                                               inline_Klass);
+            }
+            // Re-execute flat array store if buffering triggers deoptimization
+            PreserveReexecuteState preexecs(this);
+            inc_sp(3);
+            jvms()->set_should_reexecute(true);
+            null_checked_stored_value_casted->as_InlineType()->store_flat(this, casted_array, casted_adr, nullptr, 0, MO_UNORDERED | IN_HEAP | IS_ARRAY);
+          } else {
+            // Element type is unknown, emit a runtime call since the flat array layout is not statically known.
+            store_to_unknown_flat_array(array, array_index, null_checked_stored_value_casted);
```
chhagedorn, line 331: Can now call extracted method which is shared with new code inserted above.

src/hotspot/share/opto/parseHelper.cpp
```diff
@@ -262,11 +262,34 @@ Node* Parse::array_store_check(Node*& adr, const Type*& elemtype) {
...
   const TypeAryPtr* arytype = _gvn.type(ary)->is_aryptr();
-  bool null_free = false;
+  const TypePtr* elem_ptr = elemtype->make_ptr();
+  bool null_free;
   if (elemtype->make_ptr()->is_inlinetypeptr()) {
     // We statically know that this is an inline type array, use precise klass ptr
     null_free = arytype->is_flat() || !elemtype->make_ptr()->maybe_null();
     a_e_klass = makecon(TypeKlassPtr::make(elemtype->inline_klass()));
+  } else {
+    // TODO: Should move to TypeAry::is_null_free() with JDK-8345681
```
chhagedorn, line 272: See PR description under additional fixes point 3.

src/hotspot/share/opto/parseHelper.cpp
```diff
@@ -262,11 +262,34 @@ Node* Parse::array_store_check(Node*& adr, const Type*& elemtype) {
 
   // If we statically know that this is an inline type array, use precise element klass for checkcast
   const TypeAryPtr* arytype = _gvn.type(ary)->is_aryptr();
-  bool null_free = false;
+  const TypePtr* elem_ptr = elemtype->make_ptr();
```
TobiHartmann, line 265: Should this be moved to the else branch?

src/hotspot/share/opto/parseHelper.cpp
```diff
@@ -262,11 +262,34 @@ Node* Parse::array_store_check(Node*& adr, const Type*& elemtype) {
...
+  bool null_free;
   if (elemtype->make_ptr()->is_inlinetypeptr()) {
     // We statically know that this is an inline type array, use precise klass ptr
     null_free = arytype->is_flat() || !elemtype->make_ptr()->maybe_null();
     a_e_klass = makecon(TypeKlassPtr::make(elemtype->inline_klass()));
+  } else {
+    // TODO: Should move to TypeAry::is_null_free() with JDK-8345681
+    TypePtr::PTR ptr = elem_ptr->ptr();
+    null_free = ptr == TypePtr::NotNull || ptr == TypePtr::AnyNull;
```
TobiHartmann, line 274:
```suggestion
    null_free = ((ptr == TypePtr::NotNull) || (ptr == TypePtr::AnyNull));
```

src/hotspot/share/opto/parse2.cpp
```diff
@@ -175,98 +156,143 @@ void Parse::array_load(BasicType bt) {
...
+                             OptoRuntime::load_unknown_inline_Java(),
+                             nullptr, TypeRawPtr::BOTTOM,
+                             array, array_index);
+  }
+  make_slow_call_ex(call, env()->Throwable_klass(), false);
+  Node* buffer = _gvn.transform(new ProjNode(call, TypeFunc::Parms));
+
+  insert_mem_bar_volatile(Op_MemBarCPUOrder, C->get_alias_index(TypeAryPtr::INLINES));
+
+  // Keep track of the information that the inline type is in flat arrays
```
TobiHartmann, line 203:
```suggestion
  // Keep track of the information that the inline type is flat in arrays
```

TobiHartmann (approved):
Great work Christian! The fix looks good and it's nice to have better test coverage now.

chhagedorn (commented):
Thanks a lot for your review!

chhagedorn:
Thanks Tobias!

/integrate

PR #1316, opened 2024-12-11
8346035: [lworld] javadoc for value objects when enable preview

RogerRiggs:
Updated javadoc for value objects when preview is enabled.

Added text and decoration to highlight preview behavior of identity classes that are value classes with --enable-preview
The decorations are similar to the existing preview highlights for Preview classes and methods.

The updates include all of the wrapper classes, java.time, and java.util.Optional* and other classed annotated with \@ValueBased.

The ValueBased is updated to describe the behavior when --enable-preview.

---------
### Progress

### Issue
 * [JDK-8346035](https://bugs.openjdk.org/browse/JDK-8346035): [lworld] javadoc for value objects when enable preview (**Bug** - P4)

### Reviewers
 * [Chen Liang](https://openjdk.org/census#liach) (@liach - no project role) ⚠️ Review applies to [1d1e6234](https://git.openjdk.org/valhalla/pull/1316/files/1d1e6234949fdff3ec74badc2f711f34062ac287)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/1316/head:pull/1316` \
`$ git checkout pull/1316`

`$ git checkout pull/1316` \
`$ git pull https://git.openjdk.org/valhalla.git pull/1316/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1316`

`$ git pr show -t 1316`

</details>

<a href="https://git.openjdk.org/valhalla/pull/1316.diff">https://git.openjdk.org/valhalla/pull/1316.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/1316#issuecomment-2537302155)
</details>

### DIFF
--- a/src/java.base/share/classes/java/lang/Boolean.java
+++ b/src/java.base/share/classes/java/lang/Boolean.java
@@ -46,10 +46,18 @@ import java.util.Optional;
  * {@code boolean}.
  *
  * <p>This is a <a href="{@docRoot}/java.base/java/lang/doc-files/ValueBased.html">value-based</a>
- * class; programmers should treat instances that are
- * {@linkplain #equals(Object) equal} as interchangeable and should not
- * use instances for synchronization, or unpredictable behavior may
- * occur. For example, in a future release, synchronization may fail.
+ * class; programmers should treat instances that are {@linkplain #equals(Object) equal}
+ * as interchangeable and should not use instances for synchronization, mutexes, or
+ * with {@linkplain java.lang.ref.Reference object references}.
+ *
+ * <div class="preview-block">
+ *      <div class="preview-comment">
+ *          When preview features are enabled, {@code Boolean} is a {@linkplain Class#isValue value class}.
+ *          Use of value class instances for synchronization, mutexes, or with
+ *          {@linkplain java.lang.ref.Reference object references} result in
+ *          {@link IdentityException}.
+ *      </div>
+ * </div>
  *
  * @author  Arthur van Hoff
  * @since   1.0
--- a/src/java.base/share/classes/java/lang/Record.java
+++ b/src/java.base/share/classes/java/lang/Record.java
@@ -71,7 +71,7 @@ package java.lang;
  *      <div class="preview-comment">
  *          When preview features are enabled, {@code Record} is
  *          an abstract {@linkplain Class#isValue value class}.
- *          Subclasses of {@code Number} can be either an {@linkplain Class#isIdentity identity class}
+ *          Subclasses of {@code Record} can be either an {@linkplain Class#isIdentity identity class}
  *          or a {@linkplain Class#isValue value class}.
  *          See {@jls The Java Language Specification 8.1.1.5 Value Classes}.
  *      </div>
--- a/src/java.base/share/classes/java/lang/Record.java
+++ b/src/java.base/share/classes/java/lang/Record.java
@@ -67,6 +67,16 @@ package java.lang;
  * </pre>
  * then it must be the case that {@code r.equals(copy)}.
  *
+ * <div class="preview-block">
+ *      <div class="preview-comment">
+ *          When preview features are enabled, {@code Record} is
+ *          an abstract {@linkplain Class#isValue value class}.
+ *          Subclasses of {@code Number} can be either an {@linkplain Class#isIdentity identity class}
+ *          or a {@linkplain Class#isValue value class}.
+ *          See {@jls The Java Language Specification 8.1.1.5 Value Classes}.
+ *      </div>
+ * </div>
+ *
  * @apiNote
  * A record class that {@code implements} {@link java.io.Serializable} is said
  * to be a <i>serializable record</i>. Serializable records are serialized and
--- a/src/java.base/share/classes/java/lang/doc-files/ValueBased.html
+++ b/src/java.base/share/classes/java/lang/doc-files/ValueBased.html
@@ -65,10 +65,10 @@ A value-based class has the following properties:
 <p>Synchronization on instances of value-based classes is strongly discouraged,
     because the programmer cannot guarantee exclusive ownership of the
     associated monitor. If the class is declared {@code value class},
-    {@code IdentityException} is thrown.</p>
+    {@link java.lang.IdentityException} is thrown.</p>
 
-<p>Identity-related behavior of value-based classes may changes when preview features are enabled.
-    For example, synchronization may fail.</p>
+<p>When preview features are enabled, use of value class instances for synchronization, mutexes, or with
+    {@linkplain java.lang.ref.Reference object references} result in {@link IdentityException}.</p>
 
 </body>
 </html>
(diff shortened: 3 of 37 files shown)


src/java.base/share/classes/java/time/ZoneId.java
```diff
@@ -162,11 +162,9 @@
...
  * <p>
  * This is a <a href="{@docRoot}/java.base/java/lang/doc-files/ValueBased.html">value-based</a>
- * class; programmers should treat instances that are
- * {@linkplain #equals(Object) equal} as interchangeable and should not
- * use instances for synchronization, or unpredictable behavior may
- * occur. For example, in a future release, synchronization may fail.
- * The {@code equals} method should be used for comparisons.
+ * class; programmers should treat instances that are {@linkplain #equals(Object) equal}
+ * as interchangeable and should not use instances for synchronization, mutexes, or
+ * with {@linkplain java.lang.ref.Reference object references}.
```
liach, line 167: Should we add the preview warning with rewords, like
```
When preview features are enabled, all implementations of {@code ZoneId} are {@linkplain Class#isValue value class}. ...
```

src/java.base/share/classes/java/lang/doc-files/ValueBased.html
```diff
@@ -30,43 +30,45 @@
...
 
 <p>Synchronization on instances of value-based classes is strongly discouraged,
     because the programmer cannot guarantee exclusive ownership of the
-    associated monitor.</p>
+    associated monitor. If the class is declared {@code value class},
+    {@link java.lang.IdentityException} is thrown.</p>
 
-<p>Identity-related behavior of value-based classes may change in a future release.
-    For example, synchronization may fail.</p>
+<p>When preview features are enabled, use of value class instances for synchronization, mutexes, or with
```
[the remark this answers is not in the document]
RogerRiggs: In this context the preview-block looks out of place.

src/java.base/share/classes/java/time/ZoneId.java
RogerRiggs: There's more to do before ZoneId can be a Valhalla value class.
Its subtype ZoneOffset has a non-final field that caches ZoneRules. So ZoneOffset can't be a value class.

liach (approved):
Thanks. Looks good to me.

src/java.base/share/classes/java/lang/Record.java
```diff
@@ -67,6 +67,16 @@
  * </pre>
  * then it must be the case that {@code r.equals(copy)}.
  *
+ * <div class="preview-block">
+ *      <div class="preview-comment">
+ *          When preview features are enabled, {@code Record} is
+ *          an abstract {@linkplain Class#isValue value class}.
+ *          Subclasses of {@code Number} can be either an {@linkplain Class#isIdentity identity class}
```
dansmithcode, line 74: s/Number/Record/

src/java.base/share/classes/java/lang/doc-files/ValueBased.html
```diff
@@ -30,43 +30,43 @@
...
         of the class's instance fields (and the members of the objects they
         reference), not from the instance's identity;</li>
     <li>the class's methods treat instances as <em>freely substitutable</em>
-        when equal, meaning that interchanging any two instances <code>x</code> and
-        <code>y</code> that are equal according to <code>equals()</code> produces no
+        when equal, meaning that interchanging any two instances {@code x} and
+        {@code y} that are equal according to {@code equals()} produces no
         visible change in the behavior of the class's methods;</li>
     <li>the class performs no synchronization using an instance's monitor;</li>
-    <li>the class does not declare (or has deprecated any) accessible constructors;</li>
```
dansmithcode, line 48: This one is still relevant, as a concrete case of the next bullet, right? A value-based class should not let clients control identity creation.

src/java.base/share/classes/java/lang/doc-files/ValueBased.html
```diff
@@ -30,43 +30,43 @@
...
+    equality {@code ==} or indirectly via an appeal to synchronization, identity hashing,
     serialization, or any other identity-sensitive mechanism.</p>
 
 <p>Synchronization on instances of value-based classes is strongly discouraged,
     because the programmer cannot guarantee exclusive ownership of the
     associated monitor.</p>
 
-<p>Identity-related behavior of value-based classes may change in a future release.
-    For example, synchronization may fail.</p>
+<p>When preview features are enabled, use of value class instances for synchronization, mutexes, or with
```
dansmithcode, line 68: A value-based class is not necessarily a value class under preview.

I think this "behavior may change" disclaimer is trying to cover two cases, which should probably be identified directly:
- The class may choose to allocate/cache instances differently
- The class may choose to become a value class, and then [things may fail]

src/java.base/share/classes/java/lang/Boolean.java
```diff
@@ -46,10 +46,18 @@
  * {@code boolean}.
  *
  * <p>This is a <a href="{@docRoot}/java.base/java/lang/doc-files/ValueBased.html">value-based</a>
- * class; programmers should treat instances that are
- * {@linkplain #equals(Object) equal} as interchangeable and should not
- * use instances for synchronization, or unpredictable behavior may
- * occur. For example, in a future release, synchronization may fail.
+ * class; programmers should treat instances that are {@linkplain #equals(Object) equal}
+ * as interchangeable and should not use instances for synchronization, mutexes, or
+ * with {@linkplain java.lang.ref.Reference object references}.
```
dansmithcode, line 51: We have a preview feature disclaimer for References.

Is there a similar one for mutexes? (What exactly do you mean by "mutexes"? Is it distinct from "synchronization"?)

src/java.base/share/classes/java/lang/doc-files/ValueBased.html
dansmithcode: Could maybe soften to "(or discourages use of any)"...

RogerRiggs: I understood we've reversed the thinking of deprecating constructors for value classes, i.e. Integer.
They don't need to be deprecated. Saying `new Integer(3)` does not give it identity (with --enable-preview).
In the short term, the deprecation warning raises awareness in the non-enable-review case.
This text is trying to cover two similar but not identical states, one as a recommendation for non-preview and the other for preview enabled.

src/java.base/share/classes/java/lang/Boolean.java
RogerRiggs: Object.wait* and Object.notify are the most obvious APIs to add a disclaimer.
IllegalMonitorStateException is thrown with the class name.  
IdentityException might be preferred and easier for developers to diagnose.

src/java.base/share/classes/java/lang/doc-files/ValueBased.html
dansmithcode: This document is primarily meant for API users who are not using preview features.

The general rule, covered by the next bullet, is that a value-based (identity) class API will not expose identity creation. As applied to constructors, that means these classes shouldn't have public constructors. However, we want to allow for classes like Integer that have public constructors, but have deprecated them.

In contrast, yes, a _value class_ doesn't need to worry about exposing constructors, because those don't expose any kind of identity creation anymore. After migrating Integer to be a value class, we can un-deprecate its constructors. (And thus, for now we can undo "forRemoval", since we know we don't intend to remove them.)

I'd be happy to say "the class does not declare (or discourages use of any) accessible constructors;".

PR #1327, opened 2025-01-06
8346307: [lworld] Clarify identity vs value in Class, Objects, and document limitations of value objects

RogerRiggs:
Add APINote  and javadoc for IdentityException where it will be useful to know that identity or value objects are treated differently. 
Simplified WeakHashMap javadoc updates for IdentityException. 
Added note to System.identityHashCode to include value objects.
Added to class javadoc for IdentityHashMap for value objects.

---------
### Progress

### Issue
 * [JDK-8346307](https://bugs.openjdk.org/browse/JDK-8346307): [lworld] Clarify identity vs value in Class, Objects, and document limitations of value objects (**Bug** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/1327/head:pull/1327` \
`$ git checkout pull/1327`

`$ git checkout pull/1327` \
`$ git pull https://git.openjdk.org/valhalla.git pull/1327/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1327`

`$ git pr show -t 1327`

</details>

<a href="https://git.openjdk.org/valhalla/pull/1327.diff">https://git.openjdk.org/valhalla/pull/1327.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/1327#issuecomment-2573839484)
</details>


RogerRiggs:
I've been trying out the various incarnations of the javadoc support to see how they work, but that can be a separate PR when it settles.

src/java.base/share/classes/java/lang/Class.java
```diff
@@ -683,36 +683,34 @@ private static Class<?> forName(Module module, String name, Class<?> caller) {
...
      */
     @PreviewFeature(feature = PreviewFeature.Feature.VALUE_OBJECTS, reflective=true)
     public boolean isValue() {
-        if (!PreviewFeatures.isEnabled()) {
-            return false;
-        }
-         if (isPrimitive() || isArray() || isInterface())
-             return false;
-        return ((getModifiers() & Modifier.IDENTITY) == 0);
+        return PreviewFeatures.isEnabled() ? !isIdentity() : false;
```
Sm0keySa1m0n, line 713: Is there a reason this isn't just `PreviewFeatures.isEnabled() && !isIdentity()`?

liach: Indeed, this API is returning `false` for interfaces and primitive types if preview features are not enabled, which is wrong. Making it simply return `!isIdentity()` may be better and conforms to the javadoc specification here.

RogerRiggs: Corrected in 68dd3fb675d0f87da8de2bc7525d5bcc2a87dac1

liach:
PreviewNote isn't merged yet; we can revisit this one once that taglet is available here.

test/jdk/java/lang/Class/GenericStringTest.java
```diff
@@ -57,7 +57,7 @@ public static void main(String... args) throws ReflectiveOperationException {
                    new PlatformTestCase(java.lang.Enum.class,
                                         "public abstract class java.lang.Enum<E extends java.lang.Enum<E>>"),
                    new PlatformTestCase(java.util.Map.class,
-                                        "public abstract interface java.util.Map<K,V>"),
+                                        "public abstract value interface java.util.Map<K,V>"),
```
liach, line 60: Hmm, isn't value redundant on interfaces?

ExE-Boss (commented):
This is still incorrect:

RogerRiggs:
> This is still incorrect:

Unless Preview is enabled, there is no such thing as a value class (according to the spec).
Until JEP 401 is enabled, all classes are identity classes.
So it would be incorrect to report that primitive classes, interfaces, and void.class are value classes.

liach:
We should document explicitly that all classes, including primitives and void, are considered identity when preview features are not enabled.

RogerRiggs:
> We should document explicitly that all classes, including primitives and void, are considered identity when preview features are not enabled.

Primitives and void are not classes and never have identity. 
The primitive *wrappers*  and Void  *are* classes have identity if that's what you're meaning.
Interface classes never have identity (do not have ACC_SUPER/ACC_IDENTITY).  
Interface classes with --enable-preview are reported as value classes.

src/java.base/share/classes/java/lang/System.java
```diff
@@ -471,6 +471,21 @@ public static native void arraycopy(Object src,  int  srcPos,
      * hashCode().
      * The hash code for the null reference is zero.
      *
+     * <div class="preview-block">
+     *      <div class="preview-comment">
+     *          The "identity hash code" of a {@linkplain Class#isValue() value object}
+     *          is computed by combining the hash codes of the value object's fields recursively.
```
ExE-Boss, line 477: Doesn’t this work by recursively calling the `identityHashCode` algorithm?
```suggestion
     *          The "identity hash code" of a {@linkplain Class#isValue() value object}
     *          is computed by combining the identity hash codes of the value object's
     *          fields recursively.
```

RogerRiggs:
If there are no more comments, I'll push this into Valhalla tomorrow.

PR #1333, opened 2025-01-21
8343846: [lworld] implement spec changes to stack map table

vicente-romero-oracle:
still experimental


src/java.base/share/classes/java/lang/classfile/attribute/StackMapFrameInfo.java
```diff
@@ -61,17 +61,24 @@ public sealed interface StackMapFrameInfo
      */
     List<VerificationTypeInfo> stack();
 
+    /**
+     * {@return the expanded unset fields}
+     */
+    List<Integer> unSetFields();
```
liach, line 67: Use `List<NameAndTypeEntry>` please

src/java.base/share/classes/java/lang/classfile/attribute/StackMapFrameInfo.java
```diff
@@ -61,17 +61,24 @@ public sealed interface StackMapFrameInfo
...
      * @param target the location of the frame
      * @param locals the complete list of frame locals
      * @param stack the complete frame stack
+     * @param unSetFields the complete list of unset fields
      */
     public static StackMapFrameInfo of(Label target,
             List<VerificationTypeInfo> locals,
-            List<VerificationTypeInfo> stack) {
+            List<VerificationTypeInfo> stack,
+            List<Integer> unSetFields) {
```
liach, line 79: Can keep the original one untouched for compatibility and add a new factory `ofUnsetFields(List<NameAndTypeEntry>)`

src/jdk.jdeps/share/classes/com/sun/tools/javap/AttributeWriter.java
```diff
@@ -572,7 +572,7 @@ public void write(Attribute<?> a, CodeAttribute lr) {
                         int offsetDelta = lr.labelToBci(frame.target()) - lastOffset - 1;
                         switch (frameType) {
                             case 247 -> {
```
liach, line 574: Should print the assert unset here

src/java.base/share/classes/java/lang/classfile/attribute/StackMapFrameInfo.java
vicente-romero-oracle: thanks for taking a look, will do

PR #1358, opened 2025-02-10
8335256: [lworld] compiler/valhalla/inlinetypes/TestValueConstruction.java fails intermittently

merykitty:
Hi,

The assert fires because in this place:

    if (is_osr_parse()) {
      Node* osr_buf = entry_map->in(TypeFunc::Parms+0);
      entry_map->set_req(TypeFunc::Parms+0, top());
      set_map(entry_map);
      load_interpreter_state(osr_buf);
    } else {
      set_map(entry_map);
      do_method_entry();
    }

`load_interpreter_state` assumes that all incoming inline types are non-larval, while in the following place:

    if (t->is_inlinetypeptr()) {
      // Create InlineTypeNode from the oop and replace the parameter
      bool is_larval = (i == 0) && method()->is_object_constructor() && !method()->holder()->is_java_lang_Object();
      Node* vt = InlineTypeNode::make_from_oop(this, parm, t->inline_klass(), !t->maybe_null(), is_larval);
      replace_in_map(parm, vt);
    }

`is_larval == true` because we are osr-compiling the constructor of a value class. As a result, we pass a non-larval object into `InlineTypeNode::make_from_oop` and want to make it larval, triggering the assert.

Thinking more deeply into this, I think there is no way we can know the larval-ness of an incoming object in an osr-compilation unless `ciTypeFlow` does so. For example, consider this bytecode sequence:

    ValueObject x = new ValueObject;
    for {
    }
    x.<init>();

Then, when we osr-compile the method from the loop, `x` will be a larval object and it can be at any place on the JVM stack or in the local slots.

As a result, I propose optimistically assuming all osr parameters be non-larval, then we can change them to larval when encountering a constructor invocation or a field store instruction. The verifier should ensure that we cannot do anything else on a larval object as well as do these actions on a non-larval object, so we should recognize the correct state of the value object if any operation is performed on them. Otherwise, they should be dead and not apear here.

Please take a look and give your advices, thanks a lot.

---------
### Progress

### Issue
 * [JDK-8335256](https://bugs.openjdk.org/browse/JDK-8335256): [lworld] compiler/valhalla/inlinetypes/TestValueConstruction.java fails intermittently (**Bug** - P3)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/1358/head:pull/1358` \
`$ git checkout pull/1358`

`$ git checkout pull/1358` \
`$ git pull https://git.openjdk.org/valhalla.git pull/1358/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1358`

`$ git pr show -t 1358`

</details>

<a href="https://git.openjdk.org/valhalla/pull/1358.diff">https://git.openjdk.org/valhalla/pull/1358.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/1358#issuecomment-2648910762)
</details>


src/hotspot/share/opto/doCall.cpp
```diff
@@ -581,6 +581,23 @@ void Parse::do_call() {
   ciKlass* speculative_receiver_type = nullptr;
   if (is_virtual_or_interface) {
     Node* receiver_node             = stack(sp() - nargs);
+
+    // For osr compilation, we have no way to know which parameter is larval and which is not. We
+    // optimistically assume all parameters are non-larval. Then, a parameter is changed to larval
+    // if we encounter a store into one of its fields, or if we encounter a constructor invocation
+    // with it being the first argument
```
jatin-bhateja, line 588: The implicit larval transition has a limited life span, it begins from the new allocation and ends just before making a call to the super class constructor. 

[https://github.com/openjdk/valhalla/blob/d94c58952d2f9a0193af4cd9df66868fd0a6e3db/src/hotspot/share/opto/doCall.cpp#L830](https://github.com/openjdk/valhalla/blob/d94c58952d2f9a0193af4cd9df66868fd0a6e3db/src/hotspot/share/opto/doCall.cpp#L830
)

The only possibility to create a larval value that gets passed on to OSR buffer is by explicitly marking it as larval using Unsafe.makePrivateBuffer API before the loop.

May I request you to kindly add a new or identify an existing test case for this scenario

TobiHartmann: > The only possibility to create a larval value that gets passed on to OSR buffer is by explicitly marking it as larval using Unsafe.makePrivateBuffer API before the loop.

Why is that? We can have a loop triggering OSR within that limited life span, right?

jatin-bhateja: Hi @TobiHartmann ,
The case mentioned above has the value type constructor call out of loop,  lifespan of larval is limited till construction.   
For OSR compilation, we begin compilation starting from loop entry and generate the downstream graph.

In practice, an OSR compiled method should always receive a value in non-larval state in its argument buffer.

The case I am referring to is the one where we have an Unsafe.makePrivateBuffer() call outside the loop to explicitly transition value instance to larval state,  if this mutable value is used within the loop then OSR buffer should hold the value in larval state.

merykitty:
Please take a glimpse at my new attempt to tackle this issue and the rationale #1405

src/hotspot/share/opto/doCall.cpp
merykitty: > In practice, an OSR compiled method should always receive a value in non-larval state in its argument buffer.

@jatin-bhateja This is wrong, if you look at the constructor of `TestValueConstruction.MyValue10`

    public MyValue10(int x, int cnt) {
        this.x = x;
        int res = 0;
        for (int i = 0; i < cnt; ++i) {
            res += x;
        }
        this.y = res;
        super();
    }

Then `this` is a larval value type until the end of `super()`. As a result, an OSR compilation starting at the loop will see a larval value object in its argument buffer.

merykitty: Please note that while `javac` often emits the constructor call right after the `new` instruction. There is no restriction and it can insert arbitrary things between them. As a result, the Hotspot compiler must be able to deal with arbitrary bytecode shapes, not only those emitted by `javac`

PR #1363, opened 2025-02-12
8349068: [lworld] C2 compilation fails with "not enough operands for reexecution"

merykitty:
Hi,

The issue here is that `GraphKit::make_runtime_call` is often used to execute a bytecode. As a result, it expects that it may deoptimize and thus need to reexecute the current bytecode in the interpreter. This is the intention of the `assert`, to verify that we are having enough operands for reexecution of the bytecode. However, in the failing case, `GraphKit::make_runtime_call` is not used to execute the bytecode, but to handle the exceptions thrown by that bytecode. In this case, the bytecode itself has finished executing and must not be reexecuted, we can see that in `Parse::catch_inline_exceptions`:

    // Oops, need to call into the VM to resolve the klasses at runtime.
    // Note:  This call must not deoptimize, since it is not a real at this bci!
    kill_dead_locals();

    make_runtime_call(RC_NO_LEAF | RC_MUST_THROW,
                      OptoRuntime::rethrow_Type(),
                      OptoRuntime::rethrow_stub(),
                      nullptr, nullptr,
                      ex_node);

As a result, reexecution is impossible and we don't need to worry about the operand stack, I propose removing the `assert` as it seems to be the cleanest fix.

The reason this only fails on `lworld` is because here the execution of `aastore` involves a static Java call, resulting in a potential exception that needs catching.

Please share your thoughts, thanks a lot.

### DIFF
--- a/src/hotspot/share/opto/graphKit.cpp
+++ b/src/hotspot/share/opto/graphKit.cpp
@@ -938,7 +938,6 @@ void GraphKit::add_safepoint_edges(SafePointNode* call, bool must_throw) {
     int inputs = 0, not_used; // initialized by GraphKit::compute_stack_effects()
     assert(method() == youngest_jvms->method(), "sanity");
     assert(compute_stack_effects(inputs, not_used), "unknown bytecode: %s", Bytecodes::name(java_bc()));
-    assert(out_jvms->sp() >= (uint)inputs, "not enough operands for reexecution");
 #endif // ASSERT
     out_jvms->set_should_reexecute(true); //NOTE: youngest_jvms not changed
   }


TobiHartmann (commented):
The fix looks reasonable to me. Just wondering, as an alternative, could we weaken the assert by checking if the call is to `OptoRuntime::rethrow_stub`?

TobiHartmann:
I just wanted to add the same comment. If this is indeed the same issue as [JDK-8350208](https://bugs.openjdk.org/browse/JDK-8350208), the fix should be done in mainline and [JDK-8349068](https://bugs.openjdk.org/browse/JDK-8349068) should be closed as duplicate.

src/hotspot/share/opto/graphKit.cpp
```diff
@@ -938,7 +938,6 @@ void GraphKit::add_safepoint_edges(SafePointNode* call, bool must_throw) {
     int inputs = 0, not_used; // initialized by GraphKit::compute_stack_effects()
     assert(method() == youngest_jvms->method(), "sanity");
     assert(compute_stack_effects(inputs, not_used), "unknown bytecode: %s", Bytecodes::name(java_bc()));
-    assert(out_jvms->sp() >= (uint)inputs, "not enough operands for reexecution");
```
jatin-bhateja, line 941: Removal of this assertion looks incorrect, given that stack holds the active expressions this sanity check ensures that required number of inputs needed to re-construction of interpreter frame are present.

TobiHartmann: This will be fixed by [JDK-8350208](https://bugs.openjdk.org/browse/JDK-8350208) in mainline.

PR #1367, opened 2025-02-14
8348961: [lworld] C2: Joining "Object:flat_in_array" with "Object:exact" does not properly work

chhagedorn:
This patch fixes the wrong joining operation of two types of the same klass with regard to flat in array. We have the following shape in the test case (how we end up with such a graph can be found as comments in the added tests):
 
image

The type of `341 CheckCastPP` is `Object:exact`, so we definitely know that the type is `Object` and nothing else and thus also not flat in array (Object is not a value class). For `339 CheckCastPP`, we know that it is some kind of value class but definitely not `Object` itself - it's just chosen because we do not know more about the klass.

Joining these two should result in an empty set since they are unrelated. However, the type system is not handling it correctly and we conclude that one is a sub type of the other and we later fail with an assert.

### Reworking flat in array Meet Operation
I revisited the way we meet two instruction types with regard to the flat in array property. The code was a little hard to understand, so I added a summary how the flat in array property should be treated when doing a meet or join:

https://github.com/openjdk/valhalla/blob/b6b30301998e09e61b9c05282878f0596bc2f61b/src/hotspot/share/opto/type.cpp#L4604-L4630

I added a new case for joining the same klasses (see `(FiA-J-Same)`) which was missing before and resulting in an assertion failure.

### Adding dual of `_flat_in_array`?
I kept asking myself if we should introduce a proper lattice with a dual for flat in array:

https://github.com/openjdk/valhalla/blob/b6b30301998e09e61b9c05282878f0596bc2f61b/src/hotspot/share/opto/type.cpp#L4593-L4597

While it's probably the proper way to address this long term, I think we should investigate this separately as we would need to touch quite some code. I therefore left the idea as a comment and went with a point fix on top - keeping the way we currently handle `_flat_in_array` in meet operations without a proper dual.

### Including Changes
- More detailed summary how the meet/join of `flat_in_array` is done .
- Refactored the code in `meet_instptr()` to make it more readable/clear how the `flat_in_array` property is treated.
- For better debugging:
  - Dumping `(flat in array)` also for `TypeInstKlassPtr` and not only for `TypeInstPtr`.
  - Adapting "Show types" IGV filter to also show the flat in array property (see image of graph above).

### Additionally Required Fixes
- (i) With the type system fix above, we sanely remove the conflicting data nodes but we fail to remove the corresponding control path as well. The reason is that `TypeInstKlassPtr::not_flat_in_array()` does not take exactness information into account:
https://github.com/openjdk/valhalla/blob/991ec927a2a111e5021b12be3413dfc96efbc74a/src/hotspot/share/opto/type.hpp#L1787

   `_klass->can_be_inline_klass()` will treat the klass as inexact (note that `TypeInstPtr::not_flat_in_array()` takes exactness into account). As a result, `TypeInstKlassPtr::not_flat_in_array()` called on `Object:exact` returns false even though it is statically known that this should be true (i.e. not flat in array). This is fixed by passing `klass_is_exact()` to `_klass->can_be_inline_klass()`:
https://github.com/openjdk/valhalla/blob/b6b30301998e09e61b9c05282878f0596bc2f61b/src/hotspot/share/opto/type.hpp#L1793-L1795

- (ii) Applying (i) resulted in some assertion failures because we now wrongly treat a sub type check of the form `X <: Object` as false even though this should be true for any class `X`. We check in `SubTypeCheckNode::Value()` -> `SybTypeCheckNode::sub()` the following:
https://github.com/openjdk/valhalla/blob/991ec927a2a111e5021b12be3413dfc96efbc74a/src/hotspot/share/opto/subtypenode.cpp#L53-L54

  In a failing case, `subk` is a value class that is flat in array while `superk` is `Object:exact`. Due to now passing the exactness information in `TypeInstKlassPtr::not_flat_in_array()` with (i), this returns true. We wrongly assume the classes are unrelated  due to conflicting `flat_in_array` information and thus the check is wrongly treated as false. 

  The fix I propose for this is to introduce a special `not_flat_in_array_inexact()` that reflects the old behavior of `not_flat_in_array()` to ignore exactness information.
 
I added some comments in the PR to help understanding the change better.

Thanks,
Christian

---------
### Progress

### Issue
 * [JDK-8348961](https://bugs.openjdk.org/browse/JDK-8348961): [lworld] C2: Joining "Object:flat_in_array" with "Object:exact" does not properly work (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/1367/head:pull/1367` \
`$ git checkout pull/1367`

`$ git checkout pull/1367` \
`$ git pull https://git.openjdk.org/valhalla.git pull/1367/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1367`

`$ git pr show -t 1367`

</details>

<a href="https://git.openjdk.org/valhalla/pull/1367.diff">https://git.openjdk.org/valhalla/pull/1367.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/1367#issuecomment-2659074709)
</details>

### DIFF
--- a/src/hotspot/share/opto/subtypenode.cpp
+++ b/src/hotspot/share/opto/subtypenode.cpp
@@ -51,8 +51,8 @@ const Type* SubTypeCheckNode::sub(const Type* sub_t, const Type* super_t) const 
   // Similar to logic in CmpPNode::sub()
   bool unrelated_classes = false;
   // Handle inline type arrays
-  if (subk->flat_in_array() && superk->not_flat_in_array()) {
-    // The subtype is in flat arrays and the supertype is not in flat arrays. Must be unrelated.
+  if (subk->flat_in_array() && superk->not_flat_in_array_inexact()) {
+    // The subtype is in flat arrays and the supertype is not in flat arrays and no subklass can be. Must be unrelated.
     unrelated_classes = true;
   } else if (subk->is_not_flat() && superk->is_flat()) {
     // The subtype is a non-flat array and the supertype is a flat array. Must be unrelated.
@@ -68,18 +68,16 @@ const Type* SubTypeCheckNode::sub(const Type* sub_t, const Type* super_t) const 
     }
   }
 
-  if (subk != nullptr) {
-    switch (Compile::current()->static_subtype_check(superk, subk, false)) {
-      case Compile::SSC_always_false:
-        return TypeInt::CC_GT;
-      case Compile::SSC_always_true:
-        return TypeInt::CC_EQ;
-      case Compile::SSC_easy_test:
-      case Compile::SSC_full_test:
-        break;
-      default:
-        ShouldNotReachHere();
-    }
+  switch (Compile::current()->static_subtype_check(superk, subk, false)) {
+    case Compile::SSC_always_false:
+      return TypeInt::CC_GT;
+    case Compile::SSC_always_true:
+      return TypeInt::CC_EQ;
+    case Compile::SSC_easy_test:
+    case Compile::SSC_full_test:
+      break;
+    default:
+      ShouldNotReachHere();
   }
 
   return bottom_type();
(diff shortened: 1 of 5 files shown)


TobiHartmann (approved):
Nice analysis and fix, Christian! Looks good to me.

chhagedorn:
Thanks Tobias!

/integrate

PR #1371, opened 2025-02-19
8349923: Refactor StackMapTable constructor and StackMapReader

matias9927:
This patch upstreams a commit from mainline which is needed for the implementation of strict fields.
Originally reviewed-by: coleenp, dholmes

### DIFF
--- a/src/hotspot/share/classfile/stackMapTable.cpp
+++ b/src/hotspot/share/classfile/stackMapTable.cpp
@@ -29,38 +29,49 @@
 #include "oops/oop.inline.hpp"
 #include "runtime/handles.inline.hpp"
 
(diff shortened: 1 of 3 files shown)


DanHeidinga (approved):
lgtm

liach:
I think @MrSimms periodically syncs master into lworld. Should we wait for the next sync?

matias9927:
> I think @MrSimms periodically syncs master into lworld. Should we wait for the next sync?

This patch was done because the strict field implementation relies on it. This shouldn't impact the syncs in a negative way but it should smooth the process by having this patch upstreamed early.

matias9927:
Thanks for the review @DanHeidinga!
/integrate

PR #1383, opened 2025-02-27
8350901: [lworld] Substitutability check is broken for non-flat fields

RogerRiggs:
Revert "8348904: [lworld] Remove concept of default value object and zeroInstance"

This reverts commit 38041d7572e853348691af5b384a08f245e2354d.

---------
### Progress

### Issue
 * [JDK-8350901](https://bugs.openjdk.org/browse/JDK-8350901): [lworld] Substitutability check is broken for non-flat fields (**Bug** - P2)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/1383/head:pull/1383` \
`$ git checkout pull/1383`

`$ git checkout pull/1383` \
`$ git pull https://git.openjdk.org/valhalla.git pull/1383/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1383`

`$ git pr show -t 1383`

</details>

<a href="https://git.openjdk.org/valhalla/pull/1383.diff">https://git.openjdk.org/valhalla/pull/1383.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/1383#issuecomment-2689192893)
</details>

### DIFF
--- a/src/hotspot/share/prims/unsafe.cpp
+++ b/src/hotspot/share/prims/unsafe.cpp
@@ -410,6 +410,13 @@ UNSAFE_ENTRY(jboolean, Unsafe_IsFlatArray(JNIEnv *env, jobject unsafe, jclass c)
   return k->is_flatArray_klass();
 } UNSAFE_END
 
+UNSAFE_ENTRY(jobject, Unsafe_UninitializedDefaultValue(JNIEnv *env, jobject unsafe, jclass vc)) {
+  Klass* k = java_lang_Class::as_Klass(JNIHandles::resolve_non_null(vc));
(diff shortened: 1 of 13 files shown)


liach:
I don't think this is the right way to resolve this problem; this "broken" only happens through null-restricted arrays not properly running value class constructors. The right way to resolve this is to encapsulate `ValueClass.newNullRestrictedArray` to fill the array with valid instances, which can be running the component type constructor.

RogerRiggs:
This resolves the short term blocking of the VM work.
When the VM support is in place, this will be re-re-verted and the arrays will be properly initialized.

PR #1387, opened 2025-03-02
8351006: [lworld] Refactor LoadNode::Identity of an InlineType into can_see_stored_value

merykitty:
Hi,

This patch moves the logic of `LoadNode::Identity` that tries to look through an `InlineTypeNode` to `can_see_stored_value`. This also fixes 2 issues:

- `Identity` should not return a new node
- `LoadUS` from a `short` is not a mismatched access and hence can return a wrong value

### DIFF
--- a/src/hotspot/share/opto/memnode.cpp
+++ b/src/hotspot/share/opto/memnode.cpp
@@ -1098,6 +1098,15 @@ Node* LoadNode::can_see_arraycopy_value(Node* st, PhaseGVN* phase) const {
   return nullptr;
 }
 
+static Node* see_through_inline_type(PhaseValues* phase, const MemNode* load, Node* base, int offset) {
+  if (!load->is_mismatched_access() && base != nullptr && base->is_InlineType() && offset > oopDesc::klass_offset_in_bytes()) {
+    Node* value = base->as_InlineType()->field_value_by_offset(offset, true);
+    assert(value != nullptr, "must see some value");
+    return value;
+  }
+
+  return nullptr;
+}
 
 //---------------------------can_see_stored_value------------------------------
 // This routine exists to make sure this set of tests is done the same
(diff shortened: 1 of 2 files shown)


src/hotspot/share/opto/memnode.cpp
```diff
@@ -1098,6 +1098,17 @@ Node* LoadNode::can_see_arraycopy_value(Node* st, PhaseGVN* phase) const {
   return nullptr;
 }
 
+static Node* see_through_inline_type(PhaseValues* phase, const MemNode* load, Node* base, int offset) {
```
chhagedorn, line 1101: Is there a specific reason that you made this `static` and not a member function of `MemNode` (then you can use `this` instead of `load`)?

chhagedorn (commented):
Nice refactoring!

> LoadUS from a short is not a mismatched access and hence can return a wrong value

Do you have a test case for that? If so, it would be great to add it as well :-)

merykitty:
@chhagedorn Thanks for your reviews

> Do you have a test case where we get a wrong execution? If so, it would be great to add it as well :-)

Unfortunately, I cannot, this pops up when I'm working on another change.

> Is there a specific reason that you made this `static` and not a member function of `MemNode` (then you can use `this` instead of `load`)?

I think using `static` offers a more local scope, less visibility, and reduce convoluting the class declaration with helper methods.

chhagedorn:
> @chhagedorn Thanks for your reviews
> 
> > Do you have a test case where we get a wrong execution? If so, it would be great to add it as well :-)
> 
> Unfortunately, I cannot, this pops up when I'm working on another change.

Okay, no problem.

> > Is there a specific reason that you made this `static` and not a member function of `MemNode` (then you can use `this` instead of `load`)?
> 
> I think using `static` offers a more local scope, less visibility, and reduce convoluting the class declaration with helper methods.

I understand. I'm not quite sure though about the visibility. It's now local to `memnode.cpp` which is huge. Moreover, it's not immediately clear to what class this method belongs to or on what node it should be called on. Moving it to `MemNode` makes the intention clearer I think. But of course, it's just a personal opinion :-)

merykitty:
> It's now local to memnode.cpp which is huge.

If it is a method in `MemNode`, it needs to be `protected` since it is used by `LoadNNode`, so it is still effectively visible to the whole `memnode.cpp`, though.

chhagedorn (approved):
> > It's now local to memnode.cpp which is huge.
> 
> If it is a method in `MemNode`, it needs to be `protected` since it is used by `LoadNNode`

That's true, it would need to be `protected` but I guess that's okay.

> so it is still effectively visible to the whole `memnode.cpp`, though.

Not entirely, there are other methods of classes in `memnode.cpp` that do not inherit from `MemNode` like `MergeMemNode` or `MemBarNode`. That's why I thought it would be less visible/clearer when moving it to the `MemNode` class instead where the method is called from (or from sub classes). But I don't have a strong opinion and leave it up to you to make the final decision.

chhagedorn:
I'm running some testing, will report the results back tomorrow.

chhagedorn:
There's been some failures which are probably unrelated - will double check with Tobias on Monday to be sure.

chhagedorn:
Testing looked good!

merykitty:
Thanks a lot for your reviews and testings, I assume 1 review is enough :p
/integrate

PR #1392, opened 2025-03-07
8351441: [lworld] ClassFile API support generating stack maps for strict field assignment

liach:
Works for a basic case, running tests, need more tests to ensure other features added in this patch

---------
### Progress

### Issue
 * [JDK-8351441](https://bugs.openjdk.org/browse/JDK-8351441): [lworld] ClassFile API support generating stack maps for strict field assignment (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/1392/head:pull/1392` \
`$ git checkout pull/1392`

`$ git checkout pull/1392` \
`$ git pull https://git.openjdk.org/valhalla.git pull/1392/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1392`

`$ git pr show -t 1392`

</details>

<a href="https://git.openjdk.org/valhalla/pull/1392.diff">https://git.openjdk.org/valhalla/pull/1392.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/1392#issuecomment-2709361627)
</details>


liach:
Fixed an issue from vicente's review and moved test per riggs' suggestion.

src/java.base/share/classes/jdk/internal/classfile/impl/WritableField.java
```diff
@@ -0,0 +1,90 @@
...
+
+        public static UnsetField[] copyArray(UnsetField[] incoming, int resultLen) {
+            assert resultLen <= incoming.length : resultLen + " > " + incoming.length;
+            return resultLen == 0 ? EMPTY_ARRAY : Arrays.copyOf(incoming, resultLen, UnsetField[].class);
+        }
+
+        public static boolean mismatches(UnsetField[] one, int sizeOne, UnsetField[] two, int sizeTwo) {
+            if (sizeOne != sizeTwo)
+                return true;
+            for (int i = 0; i < sizeOne; i++) {
```
vicente-romero-oracle, line 82: there could be a mismatch even if both arrays have the same length and same content but the order of the fields is different in each array, not sure if this could happen but we should consider it

liach: Originally this matches because I am only removing unsets. Now Dan told me that we need to consider adding more to unset fields if we have loops, so I will do a sort for the arrays to ensure they have some definite order.

liach: I have added more test cases for control flow merge with different DA results. In the new version, I sort the fields by their CP indices (their UTF8s are from the same CP).

test/jdk/jdk/classfile/StrictStackMapsTest.java
```diff
@@ -0,0 +1,260 @@
...
+        var classModel = ClassFile.of().parse(classBytes);
+        var ctorModel = classModel.methods().getFirst();
+        var stackMaps = ctorModel.code().orElseThrow().findAttribute(Attributes.stackMapTable()).orElseThrow();
+        assertEquals(2 + 1, stackMaps.entries().size()); // if -> else, then -> end of if, just one assert
+        var assertFrame = stackMaps.entries().get(0); // else jump from if
+        assertEquals(246, assertFrame.frameType());
+        assertEquals(List.of(ConstantPoolBuilder.of().nameAndTypeEntry("fsf", CD_int)), assertFrame.unsetFields());
+    }
+
+    @Test
```
vicente-romero-oracle, line 119: nit: it would be great if a comment with the corresponding java code being generated was included in these tests, this will make easier future changes, fixing bugs etc. But I let that to your consideration

vicente-romero-oracle (approved):

src/java.base/share/classes/jdk/internal/classfile/impl/StackMapGenerator.java
```diff
@@ -1197,6 +1254,51 @@ private Type merge(Type me, Type[] toTypes, int i, Frame target) {
             return newTo;
         }
 
+        // Merge this frame's unset fields into the target frame
+        private void mergeUnsetFields(Frame target) {
```
vicente-romero-oracle, line 1258: what test case is stressing this code?

liach: allowMultiAssignTest

PR #1396, opened 2025-03-12
8351550: [lworld] Some bootstrap methods/lambda factories fail with -XX:+UseNullableValueFlattening

fparain:
Fix handling of offsets for boxing classes.
Thanks to Chen Liang and John Rose for their help identifying the root cause of the problem.

Fred

---------
### Progress

### Issue
 * [JDK-8351550](https://bugs.openjdk.org/browse/JDK-8351550): [lworld] Some bootstrap methods/lambda factories fail with -XX:+UseNullableValueFlattening (**Bug** - P3)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/1396/head:pull/1396` \
`$ git checkout pull/1396`

`$ git checkout pull/1396` \
`$ git pull https://git.openjdk.org/valhalla.git pull/1396/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1396`

`$ git pr show -t 1396`

</details>

<a href="https://git.openjdk.org/valhalla/pull/1396.diff">https://git.openjdk.org/valhalla/pull/1396.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/1396#issuecomment-2718693606)
</details>


src/hotspot/share/classfile/javaClasses.cpp
```diff
@@ -3987,15 +3987,18 @@ bool java_lang_ref_Reference::is_referent_field(oop obj, ptrdiff_t offset) {
   return is_reference;
 }
 
-int java_lang_boxing_object::_value_offset;
-int java_lang_boxing_object::_long_value_offset;
+int java_lang_boxing_object::_sub32bits_value_offset;
```
rose00, line 3990: An existing term of art in our code for `sub32bits` is "subword".
The `_32` and `_64` are distracting; suggest simply `int` and `long`.
That combined with `subword` will make clear enough what is happening.
Suggestion: `s/sub32bits/subword/; s/32bits/int/; s/64bits/long/`

Also, are we sure the subwords will always be all the same, and the floats and non-floats will correspond?  This logic assumes that.  I suppose the adjusted asserts will catch any future problems.

A fuller fix would be to make a local array indexed by `T_FOO` code, as the VM has in many other places.   (The length is `T_VOID`.)  Then each offset could be stored in its own place, customized to each `T_FOO` type.  Why not move all the way there, and forget about these ad hoc classifications?

(But did the existing asserts catch this problem in the first place?  If not why not?)

rose00 (commented):
Good.  As it stands it needs some cleanup when Valhalla is integrated.
Put a comment to that effect, something like "when we exit preview,
this code must be simplified."

I suggested some name changes.

I also suggested a deeper cut, a more robust fix that would remember the offset separately for each `T_FOO` type.  I would prefer that, but I won't insist, since this code is going to be reworked again.

On the other hand, why not fix it in such a way, using a table indexed by `T_FOO`, that no further rework will be necessary?  It may be that with a full table-based approach, you won't need to look at the preview flag.

src/hotspot/share/classfile/javaClasses.cpp
fparain: I've hesitated getting rid of the classification and doing the check for each wrapper class individually, but I was wondering why it was implemented that way initially, and that I might be missing something.
But I agree it would be a much cleaner and robust fix.

PR #1400, opened 2025-03-14
8351979: Improve verification error messages for strict fields

matias9927:
This patch adds more detail to the verifier errors throws when there are problems with strict instance fields. The class name and one field name is now printed. Verified with tier 1-3 tests.

### DIFF
--- a/src/hotspot/share/classfile/stackMapFrame.cpp
+++ b/src/hotspot/share/classfile/stackMapFrame.cpp
@@ -47,6 +47,30 @@ StackMapFrame::StackMapFrame(u2 max_locals, u2 max_stack, AssertUnsetFieldTable*
   }
 }
 
+void StackMapFrame::unsatisfied_strict_fields_error(InstanceKlass* klass, int bci) {
+  Symbol* name;
+  Symbol* sig;
+  int num_uninit_fields = 0;
+
+  auto find_unset = [&] (const NameAndSig& key, const bool& value) {
+    if (!value) {
+      name = key._name;
+      sig = key._signature;
+      num_uninit_fields++;
+    }
+  };
+  assert_unset_fields()->iterate_all(find_unset);
+
+  verifier()->verify_error(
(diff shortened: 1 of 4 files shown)


src/hotspot/share/classfile/stackMapFrame.cpp
```diff
@@ -47,6 +47,30 @@ StackMapFrame::StackMapFrame(u2 max_locals, u2 max_stack, AssertUnsetFieldTable*
...
+  assert_unset_fields()->iterate_all(find_unset);
+
+  verifier()->verify_error(
+    ErrorContext::bad_strict_fields(bci, this),
+    "All strict final fields must be initialized before super(): %d field(s), %s:%s in %s",
+    num_uninit_fields,
+    name->as_C_string(),
+    sig->as_C_string(),
+    klass->name()->as_C_string()
+  );
```
coleenp, line 71: I wonder if you need a ResourceMark here.  Also, I wonder if you could use a stringStream here and concatenate all the names and signatures of the unset fields?

matias9927: Placing a ResourceMark here messes with the error message so I opted not to use one.

matias9927: Regarding stringStream, there's a chance that approach would flood the log output in some cases. Although most realistic cases would have a few strict instance fields per class, it is still possible to have a large amount and create a massive error message that could be hard to trace. 

I think a minimal but descriptive error message is sufficient here and then users can use `-Xlog:verification` for the complete list of fields.

coleenp: Okay, that makes sense.  Also ResourceMarks are tricky in this code since so much is resource allocated.

coleenp (approved):
This looks good!

matias9927:
Thanks for the review @coleenp!
/integrate

PR #1402, opened 2025-03-19
8351569: [lworld] Revisit atomic access modes in flat var handles

mcimadamore:
This PR is an attempt to put var handle support for flat values on a more solid footing. Some more notes in a comment below.

---------
### Progress

### Issue
 * [JDK-8351569](https://bugs.openjdk.org/browse/JDK-8351569): [lworld] Revisit atomic access modes in flat var handles (**Bug** - P3)

### Reviewers
 * [Chen Liang](https://openjdk.org/census#liach) (@liach - no project role)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/1402/head:pull/1402` \
`$ git checkout pull/1402`

`$ git checkout pull/1402` \
`$ git pull https://git.openjdk.org/valhalla.git pull/1402/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1402`

`$ git pr show -t 1402`

</details>

<a href="https://git.openjdk.org/valhalla/pull/1402.diff">https://git.openjdk.org/valhalla/pull/1402.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/1402#issuecomment-2739950558)
</details>

### DIFF
--- a/make/modules/java.base/gensrc/GensrcVarHandles.gmk
+++ b/make/modules/java.base/gensrc/GensrcVarHandles.gmk
@@ -49,6 +49,7 @@ define GenerateVarHandle
     $1_ARGS += -KAtomicAdd
     $1_ARGS += -KNonPlainAccess
     $1_ARGS += -KStatic
+    $1_ARGS += -KArray
   endif
 
   ifneq ($$(findstring $$($1_InputType), Byte Short Char Int Long), )
@@ -59,6 +60,7 @@ define GenerateVarHandle
     $1_ARGS += -KBitwise
     $1_ARGS += -KNonPlainAccess
     $1_ARGS += -KStatic
+    $1_ARGS += -KArray
   endif
 
   ifneq ($$(findstring $$($1_InputType), Byte Short Char), )
@@ -69,12 +71,14 @@ define GenerateVarHandle
     $1_ARGS += -KReference
     $1_ARGS += -KNonPlainAccess
     $1_ARGS += -KStatic
+    $1_ARGS += -KArray
   endif
 
   ifeq ($$($1_InputType), NonAtomicReference)
       $1_ARGS += -KReference
       $1_ARGS += -KStatic
       $1_Type := Reference
+      $1_ARGS += -KArray
   endif
 
   ifeq ($$($1_InputType), FlatValue)
--- a/make/modules/java.base/gensrc/GensrcVarHandles.gmk
+++ b/make/modules/java.base/gensrc/GensrcVarHandles.gmk
@@ -40,19 +40,14 @@ VARHANDLES_SRC_DIR := $(MODULE_SRC)/share/classes/java/lang/invoke
 define GenerateVarHandle
 
   $1_Type := $2
-  $1_TypeAtomic := $2
 
   $1_FILENAME := $(VARHANDLES_GENSRC_DIR)/VarHandle$$($1_Type)s.java
 
   $1_ARGS += -KCAS
 
-  ifneq ($$(findstring $$($1_Type), Boolean Byte Short Char Int Long Float Double Reference), )
-      $1_ARGS += -KNonPlain
-      $1_ARGS += -KStatic
-  endif
-
   ifneq ($$(findstring $$($1_Type), Byte Short Char Int Long Float Double), )
     $1_ARGS += -KAtomicAdd
+    $1_ARGS += -KNonPlainAccess
   endif
 
   ifneq ($$(findstring $$($1_Type), Boolean Byte Short Char Int Long), )
@@ -65,24 +60,35 @@ define GenerateVarHandle
 
   ifeq ($$($1_Type), Reference)
     $1_ARGS += -KReference
+    $1_ARGS += -KNonPlainAccess
+    $1_ARGS += -KCovariantArrayArg
   endif
 
-  ifeq ($$($1_Type), FlatValue)
+  ifeq ($$($1_Type), FlatAtomicValue)
     $1_ARGS += -KFlatValue
-    $1_TypeAtomic := Reference
+    $1_ARGS += -KFlatAtomicValue
+    $1_ARGS += -KNonPlainAccess
+    $1_ARGS += -KCovariantArrayArg
+  endif
+
+  ifeq ($$($1_Type), FlatNonAtomicValue)
+    $1_ARGS += -KFlatValue
+    $1_ARGS += -KFlatNonAtomicValue
   endif
 
   $$($1_FILENAME): $(VARHANDLES_SRC_DIR)/X-VarHandle.java.template $(BUILD_TOOLS_JDK)
         ifeq ($$($1_Type), Reference)
 	  $$(eval $1_type := Object)
-        else ifeq ($$($1_Type), FlatValue)
+        else ifeq ($$($1_Type), FlatAtomicValue)
+	  $$(eval $1_type := Object)
+        else ifeq ($$($1_Type), FlatNonAtomicValue)
 	  $$(eval $1_type := Object)
         else
 	  $$(eval $1_type := $$$$(shell $(TR) '[:upper:]' '[:lower:]' <<< $$$$($1_Type)))
         endif
 	$$(call MakeDir, $$(@D))
 	$(RM) $$@
-	$(TOOL_SPP) -nel -K$$($1_type) -Dtype=$$($1_type) -DType=$$($1_Type) -DTypeAtomic=$$($1_TypeAtomic) \
+	$(TOOL_SPP) -nel -K$$($1_type) -Dtype=$$($1_type) -DType=$$($1_Type) \
 	    $$($1_ARGS) -i$$< -o$$@
 
   GENSRC_VARHANDLES += $$($1_FILENAME)
@@ -301,7 +307,7 @@ endef
 ################################################################################
 
 # List the types to generate source for, with capitalized first letter
-VARHANDLES_TYPES := Boolean Byte Short Char Int Long Float Double Reference FlatValue
+VARHANDLES_TYPES := Boolean Byte Short Char Int Long Float Double Reference FlatAtomicValue FlatNonAtomicValue
 $(foreach t, $(VARHANDLES_TYPES), \
   $(eval $(call GenerateVarHandle,VAR_HANDLE_$t,$t)))
 
--- a/make/modules/java.base/gensrc/GensrcVarHandles.gmk
+++ b/make/modules/java.base/gensrc/GensrcVarHandles.gmk
@@ -40,10 +40,16 @@ VARHANDLES_SRC_DIR := $(MODULE_SRC)/share/classes/java/lang/invoke
 define GenerateVarHandle
 
   $1_Type := $2
+  $1_TypeAtomic := $2
 
   $1_FILENAME := $(VARHANDLES_GENSRC_DIR)/VarHandle$$($1_Type)s.java
 
   $1_ARGS += -KCAS
+
+  ifneq ($$(findstring $$($1_Type), Boolean Byte Short Char Int Long Float Double Reference), )
+      $1_ARGS += -KNonPlain
+      $1_ARGS += -KStatic
+  endif
 
   ifneq ($$(findstring $$($1_Type), Byte Short Char Int Long Float Double), )
     $1_ARGS += -KAtomicAdd
@@ -63,6 +69,7 @@ define GenerateVarHandle
 
   ifeq ($$($1_Type), FlatValue)
     $1_ARGS += -KFlatValue
+    $1_TypeAtomic := Reference
   endif
 
   $$($1_FILENAME): $(VARHANDLES_SRC_DIR)/X-VarHandle.java.template $(BUILD_TOOLS_JDK)
@@ -75,7 +82,7 @@ define GenerateVarHandle
         endif
 	$$(call MakeDir, $$(@D))
 	$(RM) $$@
-	$(TOOL_SPP) -nel -K$$($1_type) -Dtype=$$($1_type) -DType=$$($1_Type) \
+	$(TOOL_SPP) -nel -K$$($1_type) -Dtype=$$($1_type) -DType=$$($1_Type) -DTypeAtomic=$$($1_TypeAtomic) \
 	    $$($1_ARGS) -i$$< -o$$@
 
   GENSRC_VARHANDLES += $$($1_FILENAME)
--- a/make/modules/java.base/gensrc/GensrcVarHandles.gmk
+++ b/make/modules/java.base/gensrc/GensrcVarHandles.gmk
@@ -39,56 +39,73 @@ VARHANDLES_SRC_DIR := $(MODULE_SRC)/share/classes/java/lang/invoke
 # Param 2 - Type with first letter capitalized
 define GenerateVarHandle
 
-  $1_Type := $2
+  $1_InputType := $2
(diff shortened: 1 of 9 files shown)


mcimadamore:
#### Problem

The VM has [several](https://github.com/openjdk/valhalla/blob/lworld/src/hotspot/share/oops/layoutKind.hpp#L28) different flat layouts to pick from. Some layouts support atomic load/stores (but require the value payload, including the null channel -- where present -- to fit in 64 bits) -- we call these layouts *atomic layouts*. Others do not.

When value fields/array elements are accessed through the VarHandle API, a question arises: under which conditions can the API support the various access modes provided by the API? Should all access modes be available, regardless of the JVM layout? Or should non-plain access modes only be supported by atomic JVM layouts?

The current implementation attempts to do the former -- all access modes are supported all the time. This is done by surrounding atomic operations such as CAS with a `synchronized` block. While this strategy is good for prototyping, it is not efficient, and only works correctly _as long as all the accesses_ occur via the `synchronized `block. For instance, if CAS code is mixed with plain access, the CAS code might observe a torn read -- which seems bad.

#### Solution

The VarHandle API should only provide access to non-plain access mode if either a layout is non-flat, or if the layout is flat and *atomic*. In the former case, we can implement access using `Unsafe`'s reference access primitives. In the latter case we can implement access using a combination of `Unsafe::getFlatValue`, memory barriers (for non-plain access modes such as acquire, or volatile) and weak CAS loops. In other words, by limiting support to atomic flat layouts, no `synchronized` block is ever required.

One complication is that we want the set of access modes supported by a `VarHandle` to be stable and predictable. That is, given that there are many factors influencing which JVM layout is picked for a field or an array element, how do user knows if a CAS operation is going to fail or not?

To provide for more stability I've opted for a simple rule: a field var handle supports atomic operations if either

* it is declared `volatile`
* its type is nullable
* the field type is not a `@WeaklyConsistent` class

For arrays, things are more convoluted: as arrays are covariant, it is always possible to create an array for `Object[].class` but then use it against a flat array. In such cases some dynamic dispatch is unavoidable: the array var handle code has to detect covariant cases, and switch to a "sharper" var handle that models the accessed array. An accessed array instance supports atomic operations if either:

* `ValueClass::isAtomicArray` returns true for the accessed array
* `ValueClass::isNullRestrictedArray` returns false for the accessed array
* the component type of the accessed array is not a `@WeaklyConsistent` class

src/java.base/share/classes/java/lang/invoke/VarHandles.java
```diff
@@ -50,11 +54,31 @@ static VarHandle makeFieldHandle(MemberName f, Class<?> refc, boolean isWriteAll
...
+                                    ? new VarHandleNonAtomicFlatValues.FieldInstanceReadOnly(refc, foffset, type, f.getCheckedFieldType(), layout)
+                                    : new VarHandleNonAtomicFlatValues.FieldInstanceReadWrite(refc, foffset, type, f.getCheckedFieldType(), layout));
+                        }
+                    } else {
+                        if (isAtomic) {
+                            return maybeAdapt(f.isFinal() && !isWriteAllowedOnFinalFields
+                                    ? new VarHandleReferences.FieldInstanceReadOnly(refc, foffset, type, f.getCheckedFieldType())
+                                    : new VarHandleReferences.FieldInstanceReadWrite(refc, foffset, type, f.getCheckedFieldType()));
+                        } else {
+                            return maybeAdapt(f.isFinal() && !isWriteAllowedOnFinalFields
```
mcimadamore, line 77: While this case is not terribly important (typically, if something is non-flat, then it will be "atomic" from a programming model perspective), there are various JVM flags that can affect whether something is flattened or not. It's important here that we don't return a var handles that "overpromises" and support more access modes (just because we know that, underneath, we can use reference `Unsafe` primitives) as such behavior might not be stable,

src/java.base/share/classes/java/lang/invoke/VarHandles.java
```diff
@@ -217,14 +306,9 @@ static VarHandle makeArrayElementHandle(Class<?> arrayClass) {
...
-            if (UNSAFE.isFlatArray(arrayClass)) {
-                int layout = UNSAFE.arrayLayout(arrayClass);
-                vh = new VarHandleFlatValues.Array(aoffset, ashift, arrayClass, layout);
-            } else {
-                vh = new VarHandleReferences.Array(aoffset, ashift, arrayClass);
-            }
-            return maybeAdapt(vh);
+            // do not check for atomicity now -- atomicity will be checked on the accessed array object
+            // a sharper var handle will be obtained (dynamically) from flatArrayElementHandleFor(Object)
+            return maybeAdapt(new VarHandleReferences.Array(aoffset, ashift, arrayClass));
```
mcimadamore, line 311: Note: this works because:
* the reference Array plain access uses array syntax (which works no matter the array passed in)
* non-plain access modes always delegate to sharper var handles based on the type of the accessed array

mcimadamore: Another note: at the language level there's no mechanism to obtain a class mirror for a null-restricted, or atomic array. This means we can't really create an array var handle that works only on one array shape (which would allow us to eliminate the dynamic dispatch completely, at least for some array var handles).

src/java.base/share/classes/jdk/internal/misc/Unsafe.java
```diff
@@ -2836,6 +2844,50 @@ public final void putDoubleOpaque(Object o, long offset, double x) {
         putDoubleVolatile(o, offset, x);
     }
 
+    private boolean compareAndSetFlatValueAsBytes(Object o, long offset, int layout, Class<?> valueType, Object expected, Object x) {
+        // try nullable atomic array first
+        Object expectedArray = ValueClass.newNullableAtomicArray(valueType, 1);
+        Object xArray = ValueClass.newNullableAtomicArray(valueType, 1);
+        if (arrayLayout(expectedArray.getClass()) != layout) {
```
mcimadamore, line 2851: This part feels a bit dirty: we are receiving a request for layout X - we need to be able to create an array with _that same layout_ -- but there's no primitive for doing that. Which means I have to try using different factories and make sure that I get something compatible (which isn't always guaranteed because of VM flags). All this guesswork should not be necessary -- that is, there should be an unsafe API to create an array with the layout I want (assuming that layout is supported of course)

mcimadamore: I have considered avoiding the array creation -- but that is messy because it means you have to deal with nulls directly. E.g. if either `expected` or `x` are `null`, what is their corresponding numeric configuration? Going through an array element allows the code to remain agnostic as to how nulls are represented.

test/jdk/valhalla/valuetypes/NullRestrictedArraysTest.java
```diff
@@ -171,65 +172,157 @@ public void testVarHandle() {
...
+        assertEquals(vh.get(array, 1), value2);
+        vh.set(array, 1, value1);
 
+        assertTrue(vh.weakCompareAndSetRelease(array, 1, value1, value2));
+        assertEquals(vh.get(array, 1), value2);
+        vh.set(array, 1, value1);
+
+        // test atomic set with null witness
+
+        assertFalse(vh.compareAndSet(array, 2, null, value1));
```
mcimadamore, line 304: I realized that the VH code we have is liberal w.r.t. `null` witness values in CAS-like operations -- in the sense that it allows them, even though the underlying variable is null-restricted. For now I decided to preserve this behavior (instead e.g. throwing an IAE)

make/modules/java.base/gensrc/GensrcVarHandles.gmk
```diff
@@ -39,43 +39,77 @@ VARHANDLES_SRC_DIR := $(MODULE_SRC)/share/classes/java/lang/invoke
 # Param 2 - Type with first letter capitalized
 define GenerateVarHandle
 
-  $1_Type := $2
+  $1_InputType := $2
```
mcimadamore, line 42: We need to do a bit of trickery here. Basically we have two pair of input values, which should use the _same_ unsafe access primitives:
* Reference, NonAtomicReference -> get/putReference
* FlatValue, NonAtomicFlatValue -> get/putFlatValue

So, we now use `InputType` to model the input type to the template, and from there we derive `Type` (which is really the Unsafe access type).

There's also some other changes:
* Not all generated classes need non-plain operations (e.g. NonAtomicReference, NonAtomicFlatValue do not)
* Not all generated classes need to handle static fields (e.g. FlatValues and NonAtomicFlatValues do not, as static fields are never flattened)
* Not all generated classes need to deal with arrays (e.g. FlatValues and NonAtomicFlatValues do not -- access to value arrays is always handled in the Reference impl class)

src/java.base/share/classes/jdk/internal/misc/Unsafe.java
```diff
@@ -2726,8 +2730,10 @@ public final Object getReferenceOpaque(Object o, long offset) {
         return getReferenceVolatile(o, offset);
     }
 
-    public final <V> Object getFlatValueOpaque(Object base, long offset, int layout, Class<?> valueType) {
-        return getFlatValueVolatile(base, offset, layout, valueType);
+    @ForceInline
+    public final <V> Object getFlatValueOpaque(Object o, long offset, int layout, Class<?> valueType) {
+        // this is stronger that opaque semantics
```
liach, line 2735:
```suggestion
        // this is stronger than opaque semantics
```

src/java.base/share/classes/jdk/internal/misc/Unsafe.java
```diff
@@ -2784,8 +2790,10 @@ public final void putReferenceOpaque(Object o, long offset, Object x) {
         putReferenceVolatile(o, offset, x);
     }
 
+    @ForceInline
     public final <V> void putFlatValueOpaque(Object o, long offset, int layout, Class<?> valueType, V x) {
-        putFlatValueVolatile(o, offset, layout, valueType, x);
+        // this is stronger that opaque semantics
```
liach, line 2795:
```suggestion
        // this is stronger than opaque semantics
```

src/java.base/share/classes/jdk/internal/misc/Unsafe.java
liach, line 2735:
```suggestion
        // this is stronger than opaque semantics
```

src/java.base/share/classes/jdk/internal/misc/Unsafe.java
liach, line 2795:
```suggestion
        // this is stronger than opaque semantics
```

src/java.base/share/classes/jdk/internal/misc/Unsafe.java
liach: I think runtime support is better for down the road for this part.

src/java.base/share/classes/jdk/internal/misc/Unsafe.java
```diff
@@ -3222,13 +3274,21 @@ public final Object getAndSetReference(Object o, long offset, Object newValue) {
...
+    @ForceInline
+    public final Object getAndSetReference(Object o, long offset, Class<?> valueType, Object newValue) {
+        Object v;
+        do {
+            v = getReferenceVolatile(o, offset);
+        } while (!compareAndSetReference(o, offset, valueType, v, newValue));
+        return v;
+    }
+
+    public Object getAndSetFlatValue(Object o, long offset, int layoutKind, Class<?> valueType, Object newValue) {
```
liach, line 3286: Why no more `@ForceInline`?

src/java.base/share/classes/jdk/internal/misc/Unsafe.java
mcimadamore: We have decided to address this by adding a primitive to create a flat array with given known layout:
https://github.com/openjdk/valhalla/pull/1404

Once that API is integrated, I will rebase this PR on top of that.

src/java.base/share/classes/jdk/internal/misc/Unsafe.java
mcimadamore: Good catch

src/java.base/share/classes/jdk/internal/misc/Unsafe.java
liach: That API is now in. This patch looks fine to me otherwise.

src/java.base/share/classes/java/lang/invoke/VarHandles.java
liach: Yep, now it is crucial for us to benchmark these var handles - I suspect they are very sensitive to profile pollutions.

src/java.base/share/classes/java/lang/invoke/X-VarHandle.java.template
```diff
@@ -433,7 +438,7 @@ final class VarHandle$Type$s {
         final CheckedType checkedFieldType;
 #end[Object]
 #if[FlatValue]
-        final int layout;
+        final int layout;
```
liach, line 441: extra spaces

src/java.base/share/classes/jdk/internal/misc/Unsafe.java
liach: Still missing forceinline here
```suggestion
    @ForceInline
    public Object getAndSetFlatValue(Object o, long offset, int layoutKind, Class<?> valueType, Object newValue) {
```

test/jdk/valhalla/valuetypes/NullRestrictedArraysTest.java
```diff
@@ -25,8 +25,7 @@
 /*
  * @test
  * @enablePreview
- * @run junit/othervm NullRestrictedArraysTest
- * @run junit/othervm -XX:-UseArrayFlattening NullRestrictedArraysTest
```
liach, line 29: Why remove the no flattening configurations?

src/java.base/share/classes/jdk/internal/misc/Unsafe.java
mcimadamore: I've now simplified the code to use the new primitive.

test/jdk/valhalla/valuetypes/NullRestrictedArraysTest.java
mcimadamore: If array flattening is disabled, but field flattening is enabled, we might run into situation where the new flat array primitive fails -- that primitive requires same level of flattening for both arrays and fields.

mcimadamore: I've now re-enabled this combo -- but disabled both array flattening and nullable flattening.

src/java.base/share/classes/jdk/internal/misc/Unsafe.java
```diff
@@ -3225,13 +3275,21 @@ public final Object getAndSetReference(Object o, long offset, Object newValue) {
...
+        Object v;
```
liach, line 3288:
```suggestion

    @ForceInline
    public Object getAndSetFlatValue(Object o, long offset, int layoutKind, Class<?> valueType, Object newValue) {
        Object v;
```
Just because how stupid github comment thread is :(

liach (approved):
Looks good.

PR #1403, opened 2025-03-19
8334484: [lworld] new translation strategy for instance field initializers

vicente-romero-oracle:
Proxy locals are added as a new compiler phase which is executed just before code generation.

Q - what it does?
A - basically for code like:
```
abstract value class Super {
    Super(String s) {}
}

value class V extends Super {
    int i;
    V(String s) {
        if (s != null) {
            i = 100;
        } else {
            i = 200;
        }
        super("100" + s);
    }
}
```
javac will generate for V's constructor something in the lines of:
```
    V(String s) {
        /*synthetic*/ int local$i;
        if (s != null) {
            local$i = 100;
        } else {
            local$i = 200;
        }
        {
            /*synthetic*/ final String local$0 = "100" + s;
            i = local$i;
            super(local$0);
        }
    }
```
so given a constructor for which any strict field is assigned to more than once, javac will generate synthetic local variables that will correspond to it and any read or write done to this strict field will be done on the synthetic local variable.

Javac will also generate additional synthetic local variables to hold the arguments, if any, of the super constructor invocation and will assign the strict fields with the current value of the corresponding synthetic local just before invoking the super constructor.

Q - does it interacts with other valhalla features?
A - yes it has a direct interaction with the new stackmap table as the new assert_unset_fields entry, in most cases, is not generated now
Q- is it on by default?
A- yes but there is a hidden option to tell javac not to generate local proxy variables: `noLocalProxyVars` this will allow us to continue using javac for test cases that check for the generation of the assert_unset_fields entry in the stackmap table attribute until we have time to migrate those tests

TIA for any comments

---------
### Progress

### Issue
 * [JDK-8334484](https://bugs.openjdk.org/browse/JDK-8334484): [lworld] new translation strategy for instance field initializers (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/1403/head:pull/1403` \
`$ git checkout pull/1403`

`$ git checkout pull/1403` \
`$ git pull https://git.openjdk.org/valhalla.git pull/1403/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1403`

`$ git pr show -t 1403`

</details>

<a href="https://git.openjdk.org/valhalla/pull/1403.diff">https://git.openjdk.org/valhalla/pull/1403.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/1403#issuecomment-2761614791)
</details>


dansmithcode:
Based on the description, looks like the trigger for this treatment is being a strict field assigned to with nontrivial control flow.

That's not an issue—we've got StackMapTable entries to manage those fields (and we _want_ to be exercising those StackMapTable entries, to ensure they're working properly).

Where we need a proxy local is when a field is _read_ somewhere in the early construction phase. In that case, the generated `getfield` would always generate a verifier error, and javac needs to translate to a local variable read instead.

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/LocalProxyVarsGen.java
```diff
@@ -0,0 +1,322 @@
...
+            strictInstanceFields = tree.defs.stream()
+                    .filter(t -> t.hasTag(VARDEF))
+                    .map(t -> (JCVariableDecl)t)
+                    .filter(vd -> vd.sym.isStrict() && !vd.sym.isStatic())
+                    .collect(List.collector());
+            if (!strictInstanceFields.isEmpty()) {
+                for (JCTree t : tree.defs) {
+                    if (t.hasTag(JCTree.Tag.METHODDEF)) {
+                        JCMethodDecl md = (JCMethodDecl) t;
+                        // ignore telescopic and generated constructors
```
mcimadamore, line 147: I think TreeInfo::hasAnyConstructorCall might check this for you?

mcimadamore: Uhmmm probably that doesn't deal with GENERATEDCONSTR (although -- a default constructor should not do much damage -- just maybe some redundant analysis?)

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/LocalProxyVarsGen.java
```diff
@@ -0,0 +1,322 @@
...
+        }
+    }
+
+    // the idea of this class is to find if there are assignments to fields that are not in the same
+    // nesting level as the super invocation, those are the ones for which we need a new local variable
+    // it is not clear if we will allow free assignment to final fields, in that case we could need to add
+    // proxy locals regardless of the nesting level
+    private class ConstructorScanner extends TreeScanner {
+        // Match this scan stack: 1=JCMethodDecl, 2=JCExpressionStatement, 3=JCMethodInvocation
+        private int scanDepth = 0;              // current scan recursion depth in method body
```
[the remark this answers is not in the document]
vicente-romero-oracle: yep removed it, dead code, thanks

vicente-romero-oracle:
> Based on the description, looks like the trigger for this treatment is being a strict field assigned to with nontrivial control flow.
> 
> That's not an issue—we've got StackMapTable entries to manage those fields (and we _want_ to be exercising those StackMapTable entries, to ensure they're working properly).
> 
> Where we need a proxy local is when a field is _read_ somewhere in the early construction phase. In that case, the generated `getfield` would always trigger a verifier error, and javac needs to translate to a local variable read instead.

ok my bad, I will update the implementation then, thanks

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/LocalProxyVarsGen.java
vicente-romero-oracle: well but TreeInfo::hasAnyConstructorCall will return true for `this()` which I want to ignore

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/LocalProxyVarsGen.java
```diff
@@ -0,0 +1,271 @@
...
+                    .collect(List.collector());
+            super.visitClassDef(tree);
+        } finally {
+            currentClass = prevCurrentClass;
+            strictInstanceFields = prevStrictInstanceFields;
+        }
+    }
+
+    public void visitMethodDef(JCMethodDecl tree) {
+        if (strictFieldsReadInPrologue.get(tree) != null) {
```
mcimadamore, line 151: Should this call `remove`, so that, as we visit, we get rid of entries (and keep memory usage under control) ?

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/LocalProxyVarsGen.java
```diff
@@ -0,0 +1,271 @@
...
+            Set<Symbol> fieldSet = strictFieldsReadInPrologue.get(tree);
+            if (!strictInstanceFields.isEmpty()) {
```
mcimadamore, line 153: Do we need this code? After all, we know from `Resolve` that a strict instance field was accessed -- so what is left to verify here?

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/LocalProxyVarsGen.java
```diff
@@ -0,0 +1,271 @@
...
+            newBody = newBody.append(fr.translate(st));
+        }
+        localDeclarations.addAll(newBody);
+        ListBuffer<JCStatement> assigmentsBeforeSuper = new ListBuffer<>();
+        for (Symbol vsym : fieldToLocalMap.keySet()) {
+            Symbol local = fieldToLocalMap.get(vsym);
+            assigmentsBeforeSuper.append(make.at(constructor.pos()).Assignment(vsym, make.at(constructor.pos()).Ident(local)));
+        }
+        constructor.body.stats = localDeclarations.toList();
+        if (!assigmentsBeforeSuper.isEmpty()) {
```
mcimadamore, line 200: Can this ever be empty (given we end up here only when `Resolve` found something) ?

mcimadamore (commented):
It seems like there's few more cleanup opportunities now that we detect "early" strict field instance reads in `Resolve` (which make things a lot simpler).

Note: a possible different implementation tactic could be to add some info to JCIdent/JCFieldAccess, so that we can record that the accessed field was instance strict, and that access was "early" (or, we could create a special variable symbol to communicate this). This would eliminate the need for a map. And would also be a bit more precise when rewriting -- because we can only rewrite field accesses that have the special tree/symbol -- and leave the other in place.

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/LocalProxyVarsGen.java
```diff
@@ -0,0 +1,271 @@
...
+        Map<Symbol, Symbol> fieldToLocalMap;
+
+        public FieldRewriter(JCMethodDecl md, Map<Symbol, Symbol> fieldToLocalMap) {
+            this.md = md;
+            this.fieldToLocalMap = fieldToLocalMap;
+        }
+
+        @Override
+        public void visitIdent(JCTree.JCIdent tree) {
+            if (fieldToLocalMap.get(tree.sym) != null) {
```
mcimadamore, line 245: How do we tell this not to rewrite a field access if there's no need (e.g. not in early context) ?

vicente-romero-oracle:
> It seems like there's few more cleanup opportunities now that we detect "early" strict field instance reads in `Resolve` (which make things a lot simpler).
> 
> Note: a possible different implementation tactic could be to add some info to JCIdent/JCFieldAccess, so that we can record that the accessed field was instance strict, and that access was "early" (or, we could create a special variable symbol to communicate this). This would eliminate the need for a map. And would also be a bit more precise when rewriting -- because we can only rewrite field accesses that have the special tree/symbol -- and leave the other in place.

not sure what would be the benefit of rewriting some accesses and some not, also there could be two sources of truth for a given field

vicente-romero-oracle:
> It seems like there's few more cleanup opportunities now that we detect "early" strict field instance reads in `Resolve` (which make things a lot simpler).
> 
> Note: a possible different implementation tactic could be to add some info to JCIdent/JCFieldAccess, so that we can record that the accessed field was instance strict, and that access was "early" (or, we could create a special variable symbol to communicate this). This would eliminate the need for a map. And would also be a bit more precise when rewriting -- because we can only rewrite field accesses that have the special tree/symbol -- and leave the other in place.

if we eliminate the map then we will need to traverse every constructor's body to extract the info already in the map

vicente-romero-oracle:
thanks for the review

PR #1405, opened 2025-03-23
8335256: [lworld] compiler/valhalla/inlinetypes/TestValueConstruction.java fails intermittently

merykitty:
Draft only!

The root issue is that a larval value object is fundamentally different from a non-larval one. The most important thing is that it has a unique identity and it expects any modification on 1 reference observable by all other equivalent references. As a result, we need a mechanism to track the identity of a larval object, which `InlineTypeNode` does not provide. My current proposal to fix this issue is to abandon larval `InlineTypeNode` and use the oop like other objects. This, in turn, leads to me having to make several other adjustments to keep the tests happy. I will then break this PR down into several smaller PRs.

It is probably beneficial to have another mechanism to make it easier optimizing larval inline type nodes, but I think it can be a follow-up RFE.

An example regarding the issue with tracking the identity of a larval `InlineTypeNode`:

Consider this pseudobytecode sequence (I will add this to `TestValueConstruction` when finalizing this PR):

    new MyValue;
    dup;
    loop;
    invokespecial MyValue::<init>;
    areturn;

There are 2 equivalent references in the stack at the loop entry. When `Parse::merge` encounters them, it cannot know that these are the same because the back-edge has not been processed yet. As a result, it creates 2 separate phi for these references. When, `invokespecial` then will only make the top of the stack a non-larval object, not the next one, which is the one returned to the caller. As a result, we fail with `assert(!value->is_InlineType() || !value->as_InlineType()->is_larval(), "returning a larval")`. Worse, if the method is osr-compiled at the loop head, we have 2 separate references fed into the compiled function and there is no way we may know that they are of the same object.

---------
### Progress

### Issue
 * [JDK-8335256](https://bugs.openjdk.org/browse/JDK-8335256): [lworld] compiler/valhalla/inlinetypes/TestValueConstruction.java fails intermittently (**Bug** - P3)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/1405/head:pull/1405` \
`$ git checkout pull/1405`

`$ git checkout pull/1405` \
`$ git pull https://git.openjdk.org/valhalla.git pull/1405/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1405`

`$ git pr show -t 1405`

</details>

<a href="https://git.openjdk.org/valhalla/pull/1405.diff">https://git.openjdk.org/valhalla/pull/1405.diff</a>

</details>


TobiHartmann:
Great work @merykitty! Thanks **a lot** for the investigation. When working on the new value object construction model, I was concerned that we might end up with adding identity back to the larvals but I was still naively hoping that we could get around that. Looks like we can't.

Thanks for working on properly fixing this. Please let me know once you want me to have a closer look at your draft.

Just a heads-up: @fparain and I are working on the removal of the default value in the VM code, see https://github.com/openjdk/valhalla/pull/1389 which will cause some conflicts with your changes.

PR #1406, opened 2025-03-23
8352647: [lworld] Remove larval InlineTypeNode in Unsafe intrinsics

merykitty:
Hi,

As a step of #1405, I would like to remove the usage of larval value objects in `Unsafe` intrinsics. This patch also tries to make sure that `Unsafe::makePrivateBuffer` and `Unsafe::finishPrivateBuffer` are always inlined, so that larval objects do not cross call boundaries except for invoking object constructors.

Please take a look and review this PR, thanks a lot.

---------
### Progress

### Issue
 * [JDK-8352647](https://bugs.openjdk.org/browse/JDK-8352647): [lworld] Remove larval InlineTypeNode in Unsafe intrinsics (**Enhancement** - P4)

### Reviewers
 * [Jatin Bhateja](https://openjdk.org/census#jbhateja) (@jatin-bhateja - Committer) ⚠️ Review applies to [3c3082ae](https://git.openjdk.org/valhalla/pull/1406/files/3c3082ae01f45124f503902f0f89f280e13cf733)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/1406/head:pull/1406` \
`$ git checkout pull/1406`

`$ git checkout pull/1406` \
`$ git pull https://git.openjdk.org/valhalla.git pull/1406/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1406`

`$ git pr show -t 1406`

</details>

<a href="https://git.openjdk.org/valhalla/pull/1406.diff">https://git.openjdk.org/valhalla/pull/1406.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/1406#issuecomment-2746323608)
</details>

### DIFF
--- a/src/hotspot/share/opto/doCall.cpp
+++ b/src/hotspot/share/opto/doCall.cpp
@@ -154,6 +154,7 @@ CallGenerator* Compile::call_generator(ciMethod* callee, int vtable_index, bool 
     CallGenerator* cg = find_intrinsic(callee, call_does_dispatch);
 
     if (cg == nullptr) {
+      // This is probably because the intrinsics is disabled from the command line
       char reason[256];
       jio_snprintf(reason, sizeof(reason), "cannot find an intrinsics for %s", callee->name()->as_utf8());
       C->record_method_not_compilable(reason);
--- a/src/hotspot/share/opto/doCall.cpp
+++ b/src/hotspot/share/opto/doCall.cpp
@@ -29,6 +29,7 @@
 #include "compiler/compileBroker.hpp"
 #include "compiler/compileLog.hpp"
 #include "interpreter/linkResolver.hpp"
+#include "jvm_io.h"
 #include "logging/log.hpp"
 #include "logging/logLevel.hpp"
 #include "logging/logMessage.hpp"
@@ -146,7 +147,20 @@ CallGenerator* Compile::call_generator(ciMethod* callee, int vtable_index, bool 
   // methods.  If these methods are replaced with specialized code,
   // then we return it as the inlined version of the call.
   CallGenerator* cg_intrinsic = nullptr;
-  if (allow_inline && allow_intrinsics) {
+  if (callee->intrinsic_id() == vmIntrinsics::_makePrivateBuffer || callee->intrinsic_id() == vmIntrinsics::_finishPrivateBuffer) {
+    // These methods must be inlined so that we don't have larval value objects crossing method
+    // boundaries
+    assert(!call_does_dispatch, "callee should not be virtual %s", callee->name()->as_utf8());
+    CallGenerator* cg = find_intrinsic(callee, call_does_dispatch);
+
+    if (cg == nullptr) {
+      char reason[256];
+      jio_snprintf(reason, sizeof(reason), "cannot find an intrinsics for %s", callee->name()->as_utf8());
+      C->record_method_not_compilable(reason);
+      return nullptr;
+    }
+    return cg;
+  } else if (allow_inline && allow_intrinsics) {
     CallGenerator* cg = find_intrinsic(callee, call_does_dispatch);
     if (cg != nullptr) {
       if (cg->is_predicated()) {
@@ -644,6 +658,10 @@ void Parse::do_call() {
   // This call checks with CHA, the interpreter profile, intrinsics table, etc.
   // It decides whether inlining is desirable or not.
   CallGenerator* cg = C->call_generator(callee, vtable_index, call_does_dispatch, jvms, try_inline, prof_factor(), speculative_receiver_type);
+  if (failing()) {
+    return;
+  }
+  assert(cg != nullptr, "must find a CallGenerator for callee %s", callee->name()->as_utf8());
 
   // NOTE:  Don't use orig_callee and callee after this point!  Use cg->method() instead.
   orig_callee = callee = nullptr;
--- a/src/hotspot/share/opto/library_call.cpp
+++ b/src/hotspot/share/opto/library_call.cpp
@@ -2683,6 +2683,7 @@ bool LibraryCallKit::inline_unsafe_make_private_buffer() {
   ciInlineKlass* vk = type->inline_klass();
   Node* klass = makecon(TypeKlassPtr::make(vk));
   Node* obj = new_instance(klass);
+  AllocateNode::Ideal_allocation(obj)->_larval = true;
 
   assert(value->is_InlineType(), "must be an InlineTypeNode");
   value->as_InlineType()->store(this, obj, obj, vk);
@@ -2711,6 +2712,11 @@ bool LibraryCallKit::inline_unsafe_finish_private_buffer() {
   if (stopped()) {
     return true;
   }
+
+  // Unset the larval bit in the object header
+  Node* old_header = make_load(control(), buffer, TypeX_X, TypeX_X->basic_type(), MemNode::unordered, LoadNode::Pinned);
+  Node* new_header = gvn().transform(new AndXNode(old_header, MakeConX(~markWord::larval_bit_in_place)));
+  access_store_at(buffer, buffer, type->is_ptr(), new_header, TypeX_X, TypeX_X->basic_type(), MO_UNORDERED | IN_HEAP);
 
   // We must ensure that the buffer is properly published
   insert_mem_bar(Op_MemBarStoreStore, alloc->proj_out(AllocateNode::RawAddress));
--- a/src/hotspot/share/opto/library_call.cpp
+++ b/src/hotspot/share/opto/library_call.cpp
@@ -2663,40 +2663,59 @@ bool LibraryCallKit::inline_unsafe_access(bool is_store, const BasicType type, c
 bool LibraryCallKit::inline_unsafe_make_private_buffer() {
   Node* receiver = argument(0);
   Node* value = argument(1);
-  if (!value->is_InlineType()) {
+
+  const Type* type = gvn().type(value);
+  if (!type->is_inlinetypeptr()) {
+    C->record_method_not_compilable("value passed to Unsafe::makePrivateBuffer is not of a constant value type");
     return false;
   }
 
-  receiver = null_check(receiver);
+  null_check(receiver);
   if (stopped()) {
     return true;
   }
 
-  set_result(value->as_InlineType()->make_larval(this, true));
+  value = null_check(value);
+  if (stopped()) {
+    return true;
+  }
+
+  ciInlineKlass* vk = type->inline_klass();
+  Node* klass = makecon(TypeKlassPtr::make(vk));
+  Node* obj = new_instance(klass);
+
+  assert(value->is_InlineType(), "must be an InlineTypeNode");
+  value->as_InlineType()->store(this, obj, obj, vk);
+
+  set_result(obj);
   return true;
 }
 
 bool LibraryCallKit::inline_unsafe_finish_private_buffer() {
   Node* receiver = argument(0);
   Node* buffer = argument(1);
-  if (!buffer->is_InlineType()) {
-    return false;
-  }
-  InlineTypeNode* vt = buffer->as_InlineType();
-  if (!vt->is_allocated(&_gvn)) {
-    return false;
-  }
-  // TODO 8239003 Why is this needed?
-  if (AllocateNode::Ideal_allocation(vt->get_oop()) == nullptr) {
+
+  const Type* type = gvn().type(buffer);
+  if (!type->is_inlinetypeptr()) {
+    C->record_method_not_compilable("value passed to Unsafe::finishPrivateBuffer is not of a constant value type");
     return false;
   }
 
-  receiver = null_check(receiver);
+  AllocateNode* alloc = AllocateNode::Ideal_allocation(buffer);
+  if (alloc == nullptr) {
+    C->record_method_not_compilable("value passed to Unsafe::finishPrivateBuffer must be allocated by Unsafe::makePrivateBuffer");
+    return false;
+  }
+
+  null_check(receiver);
   if (stopped()) {
     return true;
   }
 
-  set_result(vt->finish_larval(this));
+  // We must ensure that the buffer is properly published
+  insert_mem_bar(Op_MemBarStoreStore, alloc->proj_out(AllocateNode::RawAddress));
+  assert(!type->maybe_null(), "result of an allocation should not be null");
+  set_result(InlineTypeNode::make_from_oop(this, buffer, type->inline_klass(), false));
   return true;
 }
 
--- a/test/hotspot/jtreg/compiler/valhalla/inlinetypes/TestIntrinsics.java
+++ b/test/hotspot/jtreg/compiler/valhalla/inlinetypes/TestIntrinsics.java
@@ -1682,6 +1682,28 @@ public class TestIntrinsics {
         }
     }
 
+    /*
+    TODO: 8335256: Properly handle merging of value object oops
+    @Test
+    @IR(failOn = {CALL_UNSAFE, ALLOC})
+    public MyValue1 test84(MyValue1 v) {
+        v = U.makePrivateBuffer(v);
+        for (int i = 0; i < 10; i++) {
+            U.putInt(v, X_OFFSET, i);
+        }
+        U.putInt(v, X_OFFSET, rI);
+        v = U.finishPrivateBuffer(v);
+        return v;
+    }
+
+    @Run(test = "test84")
+    public void test84_verifier() {
+        MyValue1 v1 = MyValue1.createWithFieldsInline(rI, rL);
+        MyValue1 v2 = test84(MyValue1.setX(v1, 0));
+        Asserts.assertEQ(v1.hash(), v2.hash());
+    }
+    */
+
     static value class MyValueClonable implements Cloneable {
         int x;
 
--- a/src/hotspot/share/opto/graphKit.cpp
+++ b/src/hotspot/share/opto/graphKit.cpp
@@ -3503,9 +3503,22 @@ Node* GraphKit::gen_instanceof(Node* obj, Node* superklass, bool safe_for_replac
 // If failure_control is supplied and not null, it is filled in with
 // the control edge for the cast failure.  Otherwise, an appropriate
 // uncommon trap or exception is thrown.
-Node* GraphKit::gen_checkcast(Node *obj, Node* superklass, Node* *failure_control, bool null_free) {
+Node* GraphKit::gen_checkcast(Node* obj, Node* superklass, Node* *failure_control, bool null_free) {
   kill_dead_locals();           // Benefit all the uncommon traps
   const TypeKlassPtr* klass_ptr_type = _gvn.type(superklass)->is_klassptr();
+  const Type* obj_type = _gvn.type(obj);
+  if (obj_type->is_inlinetypeptr() && !obj_type->maybe_null() && klass_ptr_type->klass_is_exact() && obj_type->inline_klass() == klass_ptr_type->exact_klass(true)) {
+    // Special case: larval inline objects must not be scalarized. They are also generally not
+    // allowed to participate in most operations except as the first operand of putfield, or as an
+    // argument to a constructor invocation with it being a receiver, Unsafe::putXXX with it being
+    // the first argument, or Unsafe::finishPrivateBuffer. This allows us to aggressively scalarize
+    // value objects in all other places. This special case comes from the limitation of the Java
+    // language, Unsafe::makePrivateBuffer returns an Object that is checkcast-ed to the concrete
+    // value type. We must do this first because C->static_subtype_check may do nothing when
+    // StressReflectiveCode is set.
+    return obj;
+  }
+
   const TypeKlassPtr* improved_klass_ptr_type = klass_ptr_type->try_improve();
   const TypeOopPtr* toop = improved_klass_ptr_type->cast_to_exactness(false)->as_instance_type();
   bool safe_for_replace = (failure_control == nullptr);
@@ -3519,26 +3532,14 @@ Node* GraphKit::gen_checkcast(Node *obj, Node* superklass, Node* *failure_contro
   // for example, in some objArray manipulations, such as a[i]=a[j].)
   if (improved_klass_ptr_type->singleton()) {
     const TypeKlassPtr* kptr = nullptr;
-    const Type* t = _gvn.type(obj);
-    if (t->isa_oop_ptr()) {
-      kptr = t->is_oopptr()->as_klass_type();
+    if (obj_type->isa_oop_ptr()) {
+      kptr = obj_type->is_oopptr()->as_klass_type();
     } else if (obj->is_InlineType()) {
-      ciInlineKlass* vk = t->inline_klass();
+      ciInlineKlass* vk = obj_type->inline_klass();
       kptr = TypeInstKlassPtr::make(TypePtr::NotNull, vk, Type::Offset(0));
     }
 
     if (kptr != nullptr) {
-      if (t->is_inlinetypeptr() && !t->maybe_null() && kptr == improved_klass_ptr_type) {
-        // Special case: larval inline objects must not be scalarized. They are also generally not
-        // allowed to participate in most operations except as an argument to Unsafe::putXXX or
-        // Unsafe::finishPrivateBuffer. This allows us to aggressively scalarize value objects in
-        // all other places. This special case comes from the limitation of the Java language,
-        // Unsafe::makePrivateBuffer returns an Object that is checkcast-ed to the concrete value
-        // type. We must do this first because C->static_subtype_check may do nothing when
-        // StressReflectiveCode is set.
-        return obj;
-      }
-
       switch (C->static_subtype_check(improved_klass_ptr_type, kptr)) {
       case Compile::SSC_always_true:
         // If we know the type check always succeed then we don't use
@@ -3557,7 +3558,7 @@ Node* GraphKit::gen_checkcast(Node *obj, Node* superklass, Node* *failure_contro
           obj = null_check(obj);
         }
         // It needs a null check because a null will *pass* the cast check.
-        if (t->isa_oopptr() != nullptr && !t->is_oopptr()->maybe_null()) {
+        if (obj_type->isa_oopptr() != nullptr && !obj_type->is_oopptr()->maybe_null()) {
           bool is_aastore = (java_bc() == Bytecodes::_aastore);
           Deoptimization::DeoptReason reason = is_aastore ?
             Deoptimization::Reason_array_check : Deoptimization::Reason_class_check;
--- a/src/hotspot/share/opto/graphKit.cpp
+++ b/src/hotspot/share/opto/graphKit.cpp
@@ -3537,7 +3537,6 @@ Node* GraphKit::gen_checkcast(Node *obj, Node* superklass, Node* *failure_contro
           assert(safe_for_replace, "must be");
           obj = null_check(obj);
         }
-        assert(stopped() || !toop->is_inlinetypeptr() || obj->is_InlineType(), "should have been scalarized");
         return obj;
       case Compile::SSC_always_false:
         if (null_free) {
--- a/src/hotspot/share/opto/graphKit.cpp
+++ b/src/hotspot/share/opto/graphKit.cpp
@@ -3526,7 +3526,19 @@ Node* GraphKit::gen_checkcast(Node *obj, Node* superklass, Node* *failure_contro
       ciInlineKlass* vk = t->inline_klass();
       kptr = TypeInstKlassPtr::make(TypePtr::NotNull, vk, Type::Offset(0));
     }
+
     if (kptr != nullptr) {
+      if (t->is_inlinetypeptr() && !t->maybe_null() && kptr == improved_klass_ptr_type) {
+        // Special case: larval inline objects must not be scalarized. They are also generally not
+        // allowed to participate in most operations except as an argument to Unsafe::putXXX or
+        // Unsafe::finishPrivateBuffer. This allows us to aggressively scalarize value objects in
+        // all other places. This special case comes from the limitation of the Java language,
+        // Unsafe::makePrivateBuffer returns an Object that is checkcast-ed to the concrete value
+        // type. We must do this first because C->static_subtype_check may do nothing when
+        // StressReflectiveCode is set.
+        return obj;
+      }
+
       switch (C->static_subtype_check(improved_klass_ptr_type, kptr)) {
       case Compile::SSC_always_true:
         // If we know the type check always succeed then we don't use
@@ -3537,6 +3549,7 @@ Node* GraphKit::gen_checkcast(Node *obj, Node* superklass, Node* *failure_contro
           assert(safe_for_replace, "must be");
           obj = null_check(obj);
         }
+        assert(stopped() || !toop->is_inlinetypeptr() || obj->is_InlineType(), "should have been scalarized");
         return obj;
       case Compile::SSC_always_false:
         if (null_free) {


src/hotspot/share/opto/doCall.cpp
```diff
@@ -146,7 +147,20 @@ CallGenerator* Compile::call_generator(ciMethod* callee, int vtable_index, bool
...
   // then we return it as the inlined version of the call.
   CallGenerator* cg_intrinsic = nullptr;
-  if (allow_inline && allow_intrinsics) {
+  if (callee->intrinsic_id() == vmIntrinsics::_makePrivateBuffer || callee->intrinsic_id() == vmIntrinsics::_finishPrivateBuffer) {
+    // These methods must be inlined so that we don't have larval value objects crossing method
+    // boundaries
+    assert(!call_does_dispatch, "callee should not be virtual %s", callee->name()->as_utf8());
+    CallGenerator* cg = find_intrinsic(callee, call_does_dispatch);
+
+    if (cg == nullptr) {
```
jatin-bhateja, line 156: Please add comment when will cg be null in this context.

src/hotspot/share/opto/library_call.cpp
```diff
@@ -2663,40 +2663,59 @@ bool LibraryCallKit::inline_unsafe_access(bool is_store, const BasicType type, c
...
 
-  set_result(value->as_InlineType()->make_larval(this, true));
+  value = null_check(value);
+  if (stopped()) {
+    return true;
+  }
+
+  ciInlineKlass* vk = type->inline_klass();
+  Node* klass = makecon(TypeKlassPtr::make(vk));
+  Node* obj = new_instance(klass);
```
jatin-bhateja, line 2685: We have adpoted a similar strategy in lworld-vector 

https://github.com/openjdk/valhalla/blob/lworld%2Bvector/src/hotspot/share/opto/inlinetypenode.cpp#L1071

jatin-bhateja: We should also emit the IR to mark the larval bit in the object header.
https://github.com/openjdk/valhalla/blob/lworld%2Bvector/src/hotspot/share/opto/inlinetypenode.cpp#L1075

TobiHartmann:
I just had a look at https://github.com/openjdk/valhalla/pull/1405, thanks for working on this @merykitty. 

> so that larval objects do not cross call boundaries except for invoking object constructors

But this will not prevent users from passing larval objects created via `makePrivateBuffer` to other methods, right? I.e., we still have the issues described in [JDK-8239003](https://bugs.openjdk.org/browse/JDK-8239003) (and linked issues).

src/hotspot/share/opto/library_call.cpp
merykitty: Thanks for noticing, I forgot about that, I believe setting the `_larval` field in the corresponding `AllocateNode` is adequate, since macro expansion will do the right thing. I also add code to unset the larval bit in `finishPrivateBuffer`.

merykitty:
@TobiHartmann

> But this will not prevent users from passing larval objects created via makePrivateBuffer to other methods, right? I.e., we still have the issues described in [JDK-8239003](https://bugs.openjdk.org/browse/JDK-8239003) (and linked issues).

My stance on this matter is that we should simply disallow passing larval objects except to `Unsafe::putXXX` and `Unsafe::finishPrivateBuffer`. What do you think?

TobiHartmann:
> My stance on this matter is that we should simply disallow passing larval objects except to Unsafe::putXXX and Unsafe::finishPrivateBuffer. What do you think?

Right, I agree, maybe we should put some safeguards in place for this at some point. Deoptimization is still an issue thought (see [JDK-8322547](https://bugs.openjdk.org/browse/JDK-8322547)). Anyway, it's out-of-scope of this patch.

jatin-bhateja:
Hi @merykitty ,
For attached test case I am seeing crash at state merge point b/w larval and non-larval value
 
[unsafe_access.txt](https://github.com/user-attachments/files/19501609/unsafe_access.txt)

```
CPROMPT>javac --enable-preview -source 25 --add-exports=java.base/jdk.internal.misc=ALL-UNNAMED unsafe_access.java
Note: unsafe_access.java uses preview features of Java SE 25.
Note: Recompile with -Xlint:preview for details.
CPROMPT>
CPROMPT>java -XX:-PauseAtStartup --enable-preview --add-exports=java.base/jdk.internal.misc=ALL-UNNAMED -cp . unsafe_access
#
#
#  Internal Error (/home/jatinbha/sandboxes/lworld/src/hotspot/share/opto/parse1.cpp:1790), pid=2453805, tid=2453825
#  assert(gvn().type(n)->is_zero_type()) failed: Should have been scalarized
#
# JRE version: OpenJDK Runtime Environment (25.0) (slowdebug build 25-internal-adhoc.root.lworld)
# Java VM: OpenJDK 64-Bit Server VM (slowdebug 25-internal-adhoc.root.lworld, mixed mode, sharing, tiered, compressed oops, compressed class ptrs, g1 gc, linux-amd64)
# Problematic frame:
# V  [libjvm.so+0x178efa5]  Parse::merge_common(Parse::Block*, int)+0x49f
#
# Core dump will be written. Default location: Core dumps may be processed with "/usr/share/apport/apport -p%p -s%s -c%c -d%d -P%P -u%u -g%g -- %E" (or dumping to /home/jatinbha/code/java/valhalla/core.2453805)
#
# An error report file with more information is saved as:
# /home/jatinbha/code/java/valhalla/hs_err_pid2453805.log
10001399990
#
# Compiler replay data is saved as:
# /home/jatinbha/code/java/valhalla/replay_pid2453805.log
#
# If you would like to submit a bug report, please visit:
#   https://bugreport.java.com/bugreport/crash.jsp
#
Aborted (core dumped)
```

merykitty:
@jatin-bhateja Merging a larval object and a non-larval object is undefined behaviour, and crashing should be an acceptable outcome then. There are issues with merges of larval objects, though, for example:

    MyValue v = UNSAFE.makePrivateBuffer(v0);
    for (int i = 0; i < 100; i++) {}
    return UNSAFE.finishPrivateBuffer(v);

But generally, the situation of merging larval objects are messy not just with larval objects created by `Unsafe` and I want to defer the fix to after #1405.

jatin-bhateja:
> @jatin-bhateja Merging a larval object and a non-larval object is undefined behaviour, and crashing should be an acceptable outcome then. There are issues with merges of larval objects, though, for example:
> 
> ```
> MyValue v = UNSAFE.makePrivateBuffer(v0);
> for (int i = 0; i < 100; i++) {}
> return UNSAFE.finishPrivateBuffer(v);
> ```
> 
> But generally, the situation of merging larval objects are messy not just with larval objects created by `Unsafe` and I want to defer the fix to after #1405.

Easier solution will be to buffer InlineTypeNode if other input of the Phi node is a larval object i.e. in oop form,  all the users of this Phi will see the value in oop form with no value forwarding. 

> @jatin-bhateja Merging a larval object and a non-larval object is undefined behaviour, and crashing should be an acceptable outcome then. There are issues with merges of larval objects, though, for example:
> 
> ```
> MyValue v = UNSAFE.makePrivateBuffer(v0);
> for (int i = 0; i < 100; i++) {}
> return UNSAFE.finishPrivateBuffer(v);
> ```
> 
> But generally, the situation of merging larval objects are messy not just with larval objects created by `Unsafe` and I want to defer the fix to after #1405.

Thanks, @merykitty. I think MEET over larval and non-larval should result in a larval value, so InlineTypeNode, which is the other input of the Phi node, should be buffered and treated as a larval object, too, until we hit upon a finishPrivate buffer call that brings this merged oop value out of the larval state and re-materialize an InlineTypeNode. 

Kindly include this test with this patch. for now, we can add it to ProblemList.txt, which can be enabled after #1405 .

merykitty:
@jatin-bhateja A larval object created by `Unsafe` should adhere to all the rules that are put on a larval object created by the `new` bytecode. In the type verification system, each larval object has a distinct type being either `uninitializedThis` or `uninitialized(Offset)`. Meeting of a larval object with another larval object or with a non-larval object, thus, will result in a non-typed `reference`. As a result, the program is ill-formed if the merge is used. Please refer to [the JVMS, section 4.10.1](https://docs.oracle.com/javase/specs/jvms/se21/html/jvms-4.html#jvms-4.10.1).

jatin-bhateja:
> @jatin-bhateja A larval object created by `Unsafe` should adhere to all the rules that are put on a larval object created by the `new` bytecode. In the type verification system, each larval object has a distinct type being either `uninitializedThis` or `uninitialized(Offset)`. Meeting of a larval object with another larval object or with a non-larval object, thus, will result in a non-typed `reference`. As a result, the program is ill-formed if the merge is used. Please refer to [the JVMS, section 4.10.1](https://docs.oracle.com/javase/specs/jvms/se21/html/jvms-4.html#jvms-4.10.1).

Hi @merykitty ,
Larval transitions are either implicit, i.e. when a new value type instance is created, it begins in larval state and then post initilization transition out of larval state. The second kind of larval transition is explicit when we use UNSAFE APIs.

While an object is in the larval state, it should not exist in scalarized form, thus, a larval object should be buffered; it still is a legitimate value instance, but because it is assigned a buffer location, we can say for now it has a pseudo-identity. The compiler cannot directly forward the field initialization of larvals to its user; a read of a field value of a larval object is similar to getfield, i.e., the compiler needs to explicitly load the field value at the corresponding field offset from the base.

Given these semantics, a MEET b/w a larval and non-larval value should result in a laval value when mering the states of two basic blocks. I don't fully follow your comment on non-typed reference, as larval and non-larval instances are of value type.
Your approach to buffer the value type instance within the makePrivate buffer looks correct to me. Please include the test case that I shared in ProblemList.txt

merykitty:
@jatin-bhateja
Note that for the verifier, `uninitialized` and `MyValue` are completely different types that have their closest common ancestor being `reference` ([section 4.10.1.2](https://docs.oracle.com/javase/specs/jvms/se21/html/jvms-4.html#jvms-4.10.1.2)).

When a `new MyValue` instruction is executed, it pushes onto the stack a value of type `uninitialized(Offset)`, not a value of type `MyValue` ([section 4.10.1.9.new](https://docs.oracle.com/javase/specs/jvms/se21/html/jvms-4.html#jvms-4.10.1.9.new)). The only action that can be done with a value of this type is to pass it into a constructor, after which all instances of type `uninitialized(Offset)` on the stack are replaced with the a value of type `MyValue` ([section 4.10.1.9.invokespecial](https://docs.oracle.com/javase/specs/jvms/se21/html/jvms-4.html#jvms-4.10.1.9.invokespecial)).

Inside the constructor `MyValue::<init>`, the first parameter has its type being `uninitializedThis` ([section 4.10.1.6](https://docs.oracle.com/javase/specs/jvms/se21/html/jvms-4.html#jvms-4.10.1.6)). There is nothing that can be done with a value of this type except passing it to a constructor of the superclass or executing `putfield` instructions. This value will be replaced by a value of type `MyValue` after the execution of the constructor of the superclass.

`Unsafe` lets you bypass the restriction that final fields can only be set inside the constructor of its holder class. To achieve this, it lies to the bytecode verifier that you are having a value of type `MyValue` when in reality the returned value of `Unsafe::makePrivateBuffer` is a value of type `uninitialized`. However, you still must adhere to the rules that are put on the values of type `uninitialized`, such as:

- You cannot pass a larval object to any method except to `Unsafe::putXXX` or `Unsafe::finishPrivateBuffer`, which are designed to allow working with larval objects.
- You cannot load from a field of a larval object.
- Meeting of an `uninitialized` with a `MyValue` results in a value of type `reference`, which you cannot possibly work with. As a result, you cannot assign to a non-larval local variable with a larval value.

jatin-bhateja:
> @jatin-bhateja Note that for the verifier, `uninitialized` and `MyValue` are completely different types that have their closest common ancestor being `reference` ([section 4.10.1.2](https://docs.oracle.com/javase/specs/jvms/se21/html/jvms-4.html#jvms-4.10.1.2)).
> 
> When a `new MyValue` instruction is executed, it pushes onto the stack a value of type `uninitialized(Offset)`, not a value of type `MyValue` ([section 4.10.1.9.new](https://docs.oracle.com/javase/specs/jvms/se21/html/jvms-4.html#jvms-4.10.1.9.new)). The only action that can be done with a value of this type is to pass it into a constructor, after which all instances of type `uninitialized(Offset)` on the stack are replaced with the a value of type `MyValue` ([section 4.10.1.9.invokespecial](https://docs.oracle.com/javase/specs/jvms/se21/html/jvms-4.html#jvms-4.10.1.9.invokespecial)).
> 
> Inside the constructor `MyValue::<init>`, the first parameter has its type being `uninitializedThis` ([section 4.10.1.6](https://docs.oracle.com/javase/specs/jvms/se21/html/jvms-4.html#jvms-4.10.1.6)). There is nothing that can be done with a value of this type except passing it to a constructor of the superclass or executing `putfield` instructions. This value will be replaced by a value of type `MyValue` after the execution of the constructor of the superclass.
> 
> `Unsafe` lets you bypass the restriction that final fields can only be set inside the constructor of its holder class. To achieve this, it lies to the bytecode verifier that you are having a value of type `MyValue` when in reality the returned value of `Unsafe::makePrivateBuffer` is a value of type `uninitialized`. However, you still must adhere to the rules that are put on the values of type `uninitialized`, such as:
> 
> * You cannot pass a larval object to any method except to `Unsafe::putXXX` or `Unsafe::finishPrivateBuffer`, which are designed to allow working with larval objects.
> * You cannot load from a field of a larval object.
> * Meeting of an `uninitialized` with a `MyValue` results in a value of type `reference`, which you cannot possibly work with. As a result, you cannot assign to a non-larval local variable with a larval value.

Hi @merykitty ,

>> * Meeting of an `uninitialized` with a `MyValue` results in a value of type `reference`, "

Yes, I agree,  this should be seen  on the lines of a primitive (value) meeting an identity object (boxes),  

https://openjdk.org/projects/valhalla/design-notes/state-of-valhalla/01-background#:~:text=Java%E2%80%99s%20eight%20built%2Din%20primitive%20types%20are%20not%20objects%2C%20but%20they%20have%20boxes.%20When%20primitives%20want%20to%20interact%20in%20the%20world%20of%20objects%2C%20we%20transparently%20convert%20them%20to%20and%20from%20their%20corresponding%20box

https://openjdk.org/projects/valhalla/design-notes/state-of-valhalla/01-background#:~:text=Objects%20have%20identity%2C%20whereas%20primitives%20do%20not%3B%20boxing%20is%20not%20able%20to%20fully%20paper%20over%20this%20gap.%20Each%20time%20we%20convert%20from%20Integer%20to%20int%20the%20identity%20is%20lost%2C%20and%20each%20time%20we%20convert%20from%20int%20to%20an%20Integer%2C%20a%20fresh%20(but%20accidental)%20identity%20is%20created

Laraval value carries an identity while value (MyValue) does not,  larval value still adheres to semantic restrictions of value type, i.e should not be used for synchronization, non-tearable etc.

Thus, here at merge point, MyValue should be buffered, since this is MEETing a larval value so new oop (Phi) should be larval, similar to following
image

>> "Merging a larval object and a non-larval object is undefined behaviour, and crashing should be an acceptable outcome then. "

Do you think we should get the semantics documented in the specification first or let the spec evolve from the implementation you choose?

merykitty:
> Yes, I agree, this should be seen on the lines of a primitive (value) meeting an identity object (boxes),

I think you are misunderstanding, the type `reference` here refers to a verification type that represents the most general reference type, it is even higher than `java.lang.Object`. From the JVMS:

    Verification type hierarchy:

                                 top
                     ____________/\____________
                    /                          \
                   /                            \
                oneWord                       twoWord
               /   |   \                     /       \
              /    |    \                   /         \
            int  float  reference        long        double             <--- meeting of a larval value and a non-larval value results in the `reference` type here
                         /     \
                        /       \_____________
                       /                      \
                      /                        \
               uninitialized                    +------------------+
                /         \                     |  Java reference  |    <--- `java.lang.Object` and its subclasses are here
               /           \                    |  type hierarchy  |    <--- larval objects have their verification type being `uninitialized`
    uninitializedThis  uninitialized(Offset)    +------------------+  
                                                         |
                                                         |
                                                        null

> Thus, here at merge point, MyValue should be buffered, since this is MEETing a larval value so new oop (Phi) should be larval, similar to following

Note that this implicit boxing and unboxing only happens at the language level, `javac` will automatically translate `ret = genInt(val)` into `ret = Integer.valueOf(genInt(val))`. So, at the VM level, we do not ever encounter a meet of a `java.lang.Integer` and an `int`. As a result, C2 does not have to take auto boxing and unboxing into consideration.

merykitty:
To be clear, your second example is irrelevant also because a `java.lang.Integer` is not a larval `int`, it is a fully constructed `int` and its identity comes from the lack of value classes, which will eventually be removed once value classes land. A more direct translation of your first example to non-unsafe would be something like:

    p = new Point;
    p.<init>(val, val + 10);
    if (val > 5000) {
        q = new Point;
        q.f1 = p.f1;
        q.f2 = p.f2;
        p = q;
        p.f1 = val + 10;
    }
    last_hash = p.hashCode();
    if (val > 5000) {
        p.<init>();
    }
    return p;

And I am sure the verifier would reject a bytecode sequence like this.

jatin-bhateja:
> > Yes, I agree, this should be seen on the lines of a primitive (value) meeting an identity object (boxes),
> 
> I think you are misunderstanding, the type `reference` here refers to a verification type that represents the most general reference type, it is even higher than `java.lang.Object`. From the JVMS:
> 
> ```
> Verification type hierarchy:
> 
>                              top
>                  ____________/\____________
>                 /                          \
>                /                            \
>             oneWord                       twoWord
>            /   |   \                     /       \
>           /    |    \                   /         \
>         int  float  reference        long        double             <--- meeting of a larval value and a non-larval value results in the `reference` type here
>                      /     \
>                     /       \_____________
>                    /                      \
>                   /                        \
>            uninitialized                    +------------------+    <--- larval objects have their verification type being `uninitialized`
>             /         \                     |  Java reference  |    <--- `java.lang.Object` and its subclasses are here
>            /           \                    |  type hierarchy  |
> uninitializedThis  uninitialized(Offset)    +------------------+  
>                                                      |
>                                                      |
>                                                     null
> ```
> 
> > Thus, here at merge point, MyValue should be buffered, since this is MEETing a larval value so new oop (Phi) should be larval, similar to following
> 
> Note that this implicit boxing and unboxing only happens at the language level, `javac` will automatically translate `ret = genInt(val)` into `ret = Integer.valueOf(genInt(val))`. So, at the VM level, we do not ever encounter a meet of a `java.lang.Integer` and an `int`. As a result, C2 does not have to take auto boxing and unboxing into consideration.

Exactly,  MEET b/w two reference type results in a type which is LCA of both, and in the worst cast it could be j.l.o

Here, we are dealing with larval and non-larval values, which are different representations of same value type instance.

Our discussion context is around the VM model as language semantics are clear on make/finishPrivate buffer semantics.

In the above example, I drew an analogy b/w boxes as larval values VS primitives as value type instances. 

```
  public static Point micro(int val) throws Exception {
       Point p = new Point(val, val + 10);
       if (val > 50000) {
           p = UNSAFE.makePrivateBuffer(p);
           UNSAFE.putInt(p, UNSAFE.objectFieldOffset(Point.class.getDeclaredField("f1")), val + 10);
       }
       last_hash = p.hashCode();
       if (val > 50000) {
           p = UNSAFE.finishPrivateBuffer(p);
       }
       return p;
   }
```

>> "Merging a larval object and a non-larval object is undefined behaviour, and crashing should be an acceptable outcome then. "

Please mention your strategy to handle the above case and what you mean by crash as an acceptable outcome.

We can think through this as address this as a larger fix in PR 1405, for now I am ok with you fix and if can checkin the test and add it to ProblemList.txt

merykitty:
> Exactly, MEET b/w two reference type results in a type which is LCA of both, and in the worst case it could be j.l.o

No, it is wrong, from the verification type system, the highest reference type in the type hierarchy is `reference`, which comprises `java.lang.Object` and larval objects ('uninitialized`). Larval objects have their types being subtypes of `uninitialized` and non-larval objects have their types being subtypes of `java.lang.Object`. These 2 are separate under the bytecode verifier.

jatin-bhateja:
> > Exactly, MEET b/w two reference type results in a type which is LCA of both, and in the worst case it could be j.l.o
> 
> No, it is wrong, from the verification type system, the highest reference type in the type hierarchy is `reference`, which comprises `java.lang.Object` and larval objects (`uninitialized`). Larval objects have their types being subtypes of `uninitialized` and non-larval objects have their types being subtypes of `java.lang.Object`. These 2 are separate under the bytecode verifier.

Bytecode verification should be same for larval and non-larval objects, you are mixing explicit and implicit larval here which is where the confusion arises. 

MEET operation semantics are in accordance with semantics defined in following logic https://github.com/openjdk/valhalla/blob/lworld/src/hotspot/share/opto/type.cpp#L4734

With UNSAFE we explicit transition the value to a mutable state and thus need to handle them at merge point if it meets a non-larval value.

Let's think through this and address it as your larger fix for https://github.com/openjdk/valhalla/pull/1405

>> Crashing is an acceptable outcome when you are dealing with Unsafe. Using Unsafe means you are lying to the VM about the >> actual thing you are doing and have to rely on your own verification to make sure the program is valid. If the program using >> Unsafe is invalid (e.g. writing arbitrary value to a reference field), there is nothing the VM can do.

Crash occurs at the state merge point. If you run following test with C1 compiler it works fine. So we need to address this crash in C2 compiler.
https://github.com/user-attachments/files/19501609/unsafe_access.txt

CPROMPT>java --enable-preview --add-exports=java.base/jdk.internal.misc=ALL-UNNAMED -XX:TieredStopAtLevel=3 -cp . unsafe_access
10001399990

For now I am ok with your fix, please also emit the IR to set larval bit in mark work during allocation expansion.

merykitty:
> Bytecode verification should be same for larval and non-larval objects, you are mixing explicit and implicit larval here which is where the confusion arises.

I have addressed this point above: "`Unsafe` lets you bypass the restriction that final fields can only be set inside the constructor of its holder class. To achieve this, it lies to the bytecode verifier that you are having a value of type `MyValue` when in reality the returned value of `Unsafe::makePrivateBuffer` is a value of type `uninitialized`. However, you still must adhere to the rules that are put on the values of type `uninitialized`". Bytecode verification being the same for `Unsafe::makePrivateBuffer` and non-larval objects is due to the fact that you are lying to the bytecode verifier when using `Unsafe`, not because they are fundamentally the same. Explicit and implicit larval objects should be fundamentally the same.

> Crash occurs at the state merge point. If you run following test with C1 compiler it works fine. So we need to address this crash in C2 compiler.

Undefined behaviour means that any behaviour is an acceptable behaviour. C1 can execute just fine and it is an acceptable outcome, C2 crashes and it is also an acceptable outcome.

> For now I am ok with your fix, please also emit the IR to set larval bit in mark word during allocation expansion.

I believe setting `AllocateNode._larval` to true will make the expansion include setting the larval bit.

src/hotspot/share/opto/library_call.cpp
```diff
@@ -2663,40 +2663,65 @@ bool LibraryCallKit::inline_unsafe_access(bool is_store, const BasicType type, c
 bool LibraryCallKit::inline_unsafe_make_private_buffer() {
   Node* receiver = argument(0);
   Node* value = argument(1);
-  if (!value->is_InlineType()) {
+
+  const Type* type = gvn().type(value);
+  if (!type->is_inlinetypeptr()) {
+    C->record_method_not_compilable("value passed to Unsafe::makePrivateBuffer is not of a constant value type");
```
jatin-bhateja, line 2669:
```suggestion
    C->record_method_not_compilable("Non value type argument passed to Unsafe::makePrivateBuffer");
```

jatin-bhateja:
> Undefined behaviour means that any behaviour is an acceptable behaviour. C1 can execute just fine and it is an acceptable outcome, C2 crashes and it is also an acceptable outcome.

From user standpoint be it C1 or C2 compiler, JVM needs to be consistent w.r.t to undefined behavior and crash is not acceptable.

> I believe setting `AllocateNode._larval` to true will make the expansion include setting the larval bit.

Yes, AllocateNode::make_ideal_mark handles this.

>> No, it is wrong, from the verification type system, the highest reference type in the type hierarchy is reference, which comprises java.lang.Object and larval objects (uninitialized). Larval objects have their types being subtypes of uninitialized and non-larval objects have their types being subtypes of java.lang.Object. These 2 are separate under the bytecode verifier.

I see your point , but the contention here is that value types have a default value see
https://github.com/openjdk/valhalla/blob/lworld/src/hotspot/share/oops/instanceKlass.cpp#L1459

So, technically they don't qualify as uninitialized objects.

jatin-bhateja:
What if we return a larval value to caller ? i.e. value returned by makePrivateBuffer is returned from a method, since the return type will still be a concrete value type hence caller will expect to receive the arguments in scalarized form. 
https://github.com/openjdk/valhalla/blob/lworld/src/hotspot/share/opto/graphKit.cpp#L2009

merykitty:
@jatin-bhateja FYI the default value of a value class is created in a previous round of experiment when we have to construct a value object with `aconst_init` and `withfield`. With the introduction of larval objects the notion of default value is no longer necessary and is in progress of removal (#1389). Additionally, all objects value or not are created at the bytecode `new` with their default value and they are still `uninitialized`.

> Not asking to address in this patch, but what if we return a larval value to caller?

`uninitialized` is not a valid operand for `areturn` so a normal larval object cannot be returned from a method. As a result, returning an object created by `Unsafe::makePrivateBuffer` should also not allowed to be returned without invoking `Unsafe::finishPrivateBuffer` first.

merykitty:
> From user standpoint be it C1 or C2 compiler, JVM needs to be consistent w.r.t to undefined behavior and crash is not acceptable.

Using `Unsafe` is inherently unsafe and we cannot have any promise regarding the behaviour of a program. For example, this will crash the VM:

    UNSAFE.getInt(null, 0);

Worse, consider this program, given `v0 == v1` and `v0.v == 0`:

    static class Int {
        int v;
    }

    void test(int[] array, Int v0, int v1) {
        int x = v0.v;
        Unsafe.putInt(array, someOffset, 1);
        int y = v1.v;
        int z = v0.v;
    }

Now if `array + someOffset` somehow points to `v0.v`, we will see that 3 consecutive loads from the same memory location return impossible results with `x == z == 0` and `y == 1`.

jatin-bhateja:
> @jatin-bhateja FYI the default value of a value class is created in a previous round of experiment when we have to construct a value object with `aconst_init` and `withfield`. With the introduction of larval objects the notion of default value is no longer necessary and is in progress of removal (#1389). Additionally, all objects value or not are created at the bytecode `new` with their default value and they are still `uninitialized`.
> 

Correct. With the new scheme, we no longer use the factory constructor, and construction is aligned with the identity instance.
[@ImplicitlyConstructible](https://github.com/openjdk/valhalla/blob/lworld/src/java.base/share/classes/jdk/internal/vm/annotation/ImplicitlyConstructible.java#L31) annotations can still be used to enforces default value initialization 

Refer to the following comment in istanceKlass.hpp

  // Query if this class can be implicitly constructed, meaning the VM is allowed
  // to create instances without calling a constructor
  // Applies to inline classes and their super types

https://github.com/openjdk/valhalla/blob/lworld/src/hotspot/share/oops/instanceKlass.hpp#L423

Just noticed that https://github.com/openjdk/valhalla/pull/1389 is in draft state.

> > Not asking to address in this patch, but what if we return a larval value to caller?
> 
> `uninitialized` is not a valid operand for `areturn` so a normal larval object cannot be returned from a method. As a result, returning an object created by `Unsafe::makePrivateBuffer` should also not allowed to be returned without invoking `Unsafe::finishPrivateBuffer` first.

We need to enforce these semantics in C2,  I feel that a better place to document these is https://bugs.openjdk.org/browse/JDK-8046159

All execution engines must then adhere to clearly spelled semantics.

Similar to "areturn" ,  comparing uninitialized values using "acmp" is also undefined.

merykitty:
> We need to enforce these semantics in C2

Sure, I think adding best-effort mechanisms to catch undefined behaviour would be appreciated. We would also need to spell out the restrictions more clearly in the javadoc of `Unsafe` Valhalla intrinsics. For this PR, do you have any other concern? Thanks a lot.

jatin-bhateja:
Hi @merykitty , 
Thanks for clearly spelling out the UNSAFE API semantics it chops out ambiguities about
context in which larval values can be used.

The approach looks correct to me and aligns with larval handling in lworld+vector.

I see you have not addressed some minor comments around comment reformatting, but we can address those along with PR 1405 :-)

Best Regards,
Jatin

merykitty:
I have added a test that uses `Unsafe` in a proper way and crashes.

src/hotspot/share/opto/library_call.cpp
merykitty: I believe we do often not capitalize the first letter passed into `C->record_method_not_compilable`. Additionally, what is passed into `Unsafe::makePrivateBuffer` is a value object, not a concrete value type, and we cannot compile this method because we cannot statically determine the value type this object belongs to, not because the object is not actually a value object. If it is not a value object we will just crash here instead.

jatin-bhateja: A reference of identity class, or an abstract value class does not carry an InlineType. Only a concrete value type is valid argument for makePrivateBuffer.

test/hotspot/jtreg/compiler/valhalla/inlinetypes/TestIntrinsics.java
```diff
@@ -1682,6 +1682,28 @@ public void test83_verifier(RunInfo info) {
...
+    /*
+    TODO: 8335256: Properly handle merging of value object oops
+    @Test
+    @IR(failOn = {CALL_UNSAFE, ALLOC})
+    public MyValue1 test84(MyValue1 v) {
+        v = U.makePrivateBuffer(v);
+        for (int i = 0; i < 10; i++) {
+            U.putInt(v, X_OFFSET, i);
+        }
+        U.putInt(v, X_OFFSET, rI);
```
jatin-bhateja, line 1694: Here, crash occurs becausethe compiler expects to receive an InlineTypeNode for  value type objects, since we are now buffering this value in make private buffer hence we hit following assertion failure

https://github.com/openjdk/valhalla/blob/lworld/src/hotspot/share/opto/parse1.cpp#L1789

This test does not capture a scenario where we merge a larval and non-larval value.
https://github.com/user-attachments/files/19501609/unsafe_access.txt

jatin-bhateja (approved):
Hi @merykitty ,
Patch looks good to me.

Thanks for clarifications and clearly spelling out UNSAFE semantics in  https://bugs.openjdk.org/browse/JDK-8353236

Best Regards,
Jatin

merykitty:
Thank @jatin-bhateja for the reviews and profound discussion.

@TobiHartmann Do I need to wait for you to run some tests?

TobiHartmann:
Thanks for asking, @merykitty. I'll run some testing and will report back.

TobiHartmann:
`compiler/valhalla/inlinetypes/TestIntrinsics.java` fails with `-XX:+UnlockDiagnosticVMOptions -XX:-TieredCompilation -XX:+StressArrayCopyMacroNode -XX:+StressCallingConvention -XX:+StressLCM -XX:+StressGCM -XX:+StressCCP -XX:+StressIGVN -XX:+StressReflectiveCode -XX:+StressMethodHandleLinkerInlining -XX:+StressCompiledExceptionHandlers -XX:MaxNodeLimit=100000` in all scenarios due to:
```
Custom Run Test: @Run: test22_verifier - @Test: test22:
compiler.lib.ir_framework.shared.TestRunException: Method public void compiler.valhalla.inlinetypes.TestIntrinsics.test22(compiler.valhalla.inlinetypes.MyValue1) not compilable (anymore) at level C2
	at compiler.lib.ir_framework.shared.TestRun.check(TestRun.java:32)
	at compiler.lib.ir_framework.test.AbstractTest.enqueueMethodForCompilation(AbstractTest.java:156)
	at compiler.lib.ir_framework.test.AbstractTest.compileMethod(AbstractTest.java:131)
	at compiler.lib.ir_framework.test.CustomRunTest.compileSingleTest(CustomRunTest.java:113)
	at compiler.lib.ir_framework.test.CustomRunTest.compileTest(CustomRunTest.java:101)
	at compiler.lib.ir_framework.test.AbstractTest.run(AbstractTest.java:101)
	at compiler.lib.ir_framework.test.CustomRunTest.run(CustomRunTest.java:89)
	at compiler.lib.ir_framework.test.TestVM.runTests(TestVM.java:867)
	at compiler.lib.ir_framework.test.TestVM.start(TestVM.java:255)
	at compiler.lib.ir_framework.test.TestVM.main(TestVM.java:168)


Custom Run Test: @Run: test39_verifier - @Test: test39:
compiler.lib.ir_framework.shared.TestRunException: Method public compiler.valhalla.inlinetypes.MyValue1 compiler.valhalla.inlinetypes.TestIntrinsics.test39(compiler.valhalla.inlinetypes.MyValue1) not compilable (anymore) at level C2
	at compiler.lib.ir_framework.shared.TestRun.check(TestRun.java:32)
	at compiler.lib.ir_framework.test.AbstractTest.enqueueMethodForCompilation(AbstractTest.java:156)
	at compiler.lib.ir_framework.test.AbstractTest.compileMethod(AbstractTest.java:131)
	at compiler.lib.ir_framework.test.CustomRunTest.compileSingleTest(CustomRunTest.java:113)
	at compiler.lib.ir_framework.test.CustomRunTest.compileTest(CustomRunTest.java:101)
	at compiler.lib.ir_framework.test.AbstractTest.run(AbstractTest.java:101)
	at compiler.lib.ir_framework.test.CustomRunTest.run(CustomRunTest.java:89)
	at compiler.lib.ir_framework.test.TestVM.runTests(TestVM.java:867)
	at compiler.lib.ir_framework.test.TestVM.start(TestVM.java:255)
	at compiler.lib.ir_framework.test.TestVM.main(TestVM.java:168)


	at compiler.lib.ir_framework.test.TestVM.runTests(TestVM.java:903)
	at compiler.lib.ir_framework.test.TestVM.start(TestVM.java:255)
	at compiler.lib.ir_framework.test.TestVM.main(TestVM.java:168)
```

TobiHartmann:
Thanks! I'll re-run testing.

TobiHartmann (approved):
All tests passed. The changes look good to me.

src/hotspot/share/opto/library_call.cpp
merykitty: @TobiHartmann I'm not sure, currently the larval bit does not seem to have any use, but maybe in the future the runtime could try to deduplicate value objects. In that case, IMO we don't actually need to have an exact value of the larval-ness but only a conservative value answering the question "Is this object definitely a non-larval value object?"

merykitty:
@TobiHartmann Thanks a lot for your reviews and testings.
/integrate

PR #1407, opened 2025-03-24
Adding flattening of strict final fields

fparain:
Strict final instance fields are not subject to concurrent writes during a read access, so they can be flattened even if they are nullable and bigger than 64 bits. The NULLABLE_NON_ATOMIC_FLAT layout is added for this particular case.

### DIFF
--- a/src/hotspot/share/classfile/fieldLayoutBuilder.cpp
+++ b/src/hotspot/share/classfile/fieldLayoutBuilder.cpp
@@ -38,7 +38,12 @@
 #include "utilities/powerOfTwo.hpp"
 
 static LayoutKind field_layout_selection(FieldInfo field_info, Array<InlineLayoutInfo>* inline_layout_info_array,
-                                         bool use_atomic_flat) {
+                                         bool can_use_atomic_flat) {
+
+  // The can_use_atomic_flat argument indicates if an atomic flat layout can be used for this field.
+  // This argument will be false if the container is a loosely consistent value class. Using an atomic layout
+  // in a container that has no atomicity guarantee creates a risk to see this field's value be subject to
+  // tearing even if the field's class was declared atomic (non loosely consistent).
 
   if (!UseFieldFlattening) {
     return LayoutKind::REFERENCE;
@@ -66,13 +71,18 @@ static LayoutKind field_layout_selection(FieldInfo field_info, Array<InlineLayou
     assert(vk->is_implicitly_constructible(), "null-free fields must be implicitly constructible");
     if (vk->must_be_atomic() || AlwaysAtomicAccesses) {
       if (vk->is_naturally_atomic() && vk->has_non_atomic_layout()) return LayoutKind::NON_ATOMIC_FLAT;
-      return (vk->has_atomic_layout() && use_atomic_flat) ? LayoutKind::ATOMIC_FLAT : LayoutKind::REFERENCE;
+      return (vk->has_atomic_layout() && can_use_atomic_flat) ? LayoutKind::ATOMIC_FLAT : LayoutKind::REFERENCE;
     } else {
       return vk->has_non_atomic_layout() ? LayoutKind::NON_ATOMIC_FLAT : LayoutKind::REFERENCE;
     }
   } else {
+    // To preserve the consistency between the null-marker and the field content, the NULLABLE_NON_ATOMIC_FLAT
+    // can only be used in containers that have atomicity quarantees (can_use_atomic_flat argument set to true)
+    if (field_info.access_flags().is_strict() && field_info.access_flags().is_final() && can_use_atomic_flat) {
+      if (vk->has_nullable_non_atomic_layout()) return LayoutKind::NULLABLE_NON_ATOMIC_FLAT;
+    }
     if (UseNullableValueFlattening && vk->has_nullable_atomic_layout()) {
-      return use_atomic_flat ? LayoutKind::NULLABLE_ATOMIC_FLAT : LayoutKind::REFERENCE;
+      return can_use_atomic_flat ? LayoutKind::NULLABLE_ATOMIC_FLAT : LayoutKind::REFERENCE;
     } else {
       return LayoutKind::REFERENCE;
     }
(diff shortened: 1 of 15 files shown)


fparain:
> This would forbid the class which contains the `NULLABLE_NON_ATOMIC_FLAT` from being flattened into a class which contains it. I think we should only do so if the holder of the flattened strict final field should not be flattened if it is not itself a strict final field:
> 
>     * The holder is an identity class.
> 
>     * The holder is a large value class without `@LooselyConsistentValue`.
> 
> 
> This is equally valid to null-restricted field of non-`@LooselyConsistentValue`, too.

Yes, there are some restrictions where the NULLABLE_NON_ATOMIC_FLAT can be used, see the comments and the `can_use_atomic_flat` argument in the `field_layout_selection()` method.

merykitty:
I assume this PR is for flattening of nullable fields at immutable memory (strict final fields) only. But I think the same flattening can be done for null-free field, right?

src/hotspot/share/oops/inlineKlass.cpp
```diff
@@ -154,17 +159,21 @@ int InlineKlass::layout_size_in_bytes(LayoutKind kind) const {
...
+      assert(has_null_free_atomic_layout(), "Layout not available");
+      return null_free_atomic_size_in_bytes();
       break;
     case LayoutKind::NULLABLE_ATOMIC_FLAT:
       assert(has_nullable_atomic_layout(), "Layout not available");
       return nullable_atomic_size_in_bytes();
       break;
+    case LayoutKind::NULLABLE_NON_ATOMIC_FLAT:
+      assert(has_nullable_non_atomic_layout(), "Layout not available");
+      return null_free_non_atomic_alignment();
```
coleenp: Should this be nullable_non_atomic_alignment() ? Or is it the same alignment as null_free_non_atomic_alignment?  (if so pls comment why)

src/hotspot/share/oops/inlineKlass.hpp
```diff
@@ -88,10 +88,11 @@ class InlineKlass: public InstanceKlass {
...
     int _payload_alignment;        // alignment required for payload
-    int _non_atomic_size_in_bytes; // size of null-free non-atomic flat layout
-    int _non_atomic_alignment;     // alignment requirement for null-free non-atomic layout
-    int _atomic_size_in_bytes;     // size and alignment requirement for a null-free atomic layout, -1 if no atomic flat layout is possible
-    int _nullable_size_in_bytes;   // size and alignment requirement for a nullable layout (always atomic), -1 if no nullable flat layout is possible
+    int _null_free_non_atomic_size_in_bytes; // size of null-free non-atomic flat layout
+    int _null_free_non_atomic_alignment;     // alignment requirement for null-free non-atomic layout
+    int _null_free_atomic_size_in_bytes;     // size and alignment requirement for a null-free atomic layout, -1 if no atomic flat layout is possible
+    int _nullable_atomic_size_in_bytes;   // size and alignment requirement for a nullable layout (always atomic), -1 if no nullable flat layout is possible
+    int _nullable_non_atomic_size_in_bytes; // size and alignment requirement for a nullable non-atomic layout, -1 if not available
```
coleenp: Here is something that I'm confused about.  Does an InlineKlass have multiple possible layouts and sizes depending on the container?  Is that why the InlineKlass needs to save all these size non-static data members?

coleenp: Can you align the // s please?

src/hotspot/share/oops/layoutKind.hpp
```diff
@@ -66,6 +66,17 @@
 //             null marker). The reset value instance is needed because the VM needs an instance guaranteed to
 //             always be filled with zeros, and the default value could have its null marker set to non-zero if
 //             it is used as a source to update a NULLABLE_ATOMIC_FLAT field.
+// NULLABLE_NON_ATOMIC_FLAT: this is a special layout, only used for strict final non-static fields. Because strict
```
coleenp: Can you capitalize 't' in "this" in these descriptions?

coleenp (commented):
I have a couple questions and comments.  Looks good as an extension to the existing field layout design.

fparain:
> This would forbid the class which contains the `NULLABLE_NON_ATOMIC_FLAT` from being flattened into a class which contains it. I think we should only do so if the holder of the flattened strict final field should not be flattened if it is not itself a strict final field:
> 
>     * The holder is an identity class.
> 
>     * The holder is a large value class without `@LooselyConsistentValue`.
> 
> 
> This is equally valid to null-restricted field of non-`@LooselyConsistentValue`, too.

Yes, there are some restrictions where the NULLABLE_NON_ATOMIC_FLAT can be used, see the comments and the `can_use_atomic_flat` argument in the `field_layout_selection()` method.

> I assume this PR is for flattening of nullable fields at immutable memory (strict final fields) only. But I think the same flattening can be done for null-free field, right?

Strict final null-free fields could also be flattened without having to be concerned about atomicity, but null-freeness is not part of JEP 401. So this case is not supported in this PR. It can be added later when the null-restriction JEP has been finalized.

src/hotspot/share/oops/inlineKlass.hpp
fparain: > Here is something that I'm confused about. Does an InlineKlass have multiple possible layouts and sizes depending on the container? Is that why the InlineKlass needs to save all these size non-static data members?

An InlineKlass supports a set of layout according to its own characteristics (number of fields, atomicity, etc.).
When computing the layout of a container with a value field, the JVM looks at the characteristics of the field (static, final , strict, null-free, etc.) and the layouts available for this InlineKlass and selects the best one that satisfies all requirements.

src/hotspot/share/classfile/fieldLayoutBuilder.cpp
```diff
@@ -1182,54 +1201,62 @@ void FieldLayoutBuilder::compute_inline_class_layout() {
...
-      int nullable_size = round_up_power_of_2(new_raw_size);
-      if (nullable_size <= (int)MAX_ATOMIC_OP_SIZE) {
-        _nullable_layout_size_in_bytes = nullable_size;
+      if (UseNullableNonAtomicValueFlattening) {
+        _nullable_non_atomic_layout_size_in_bytes = new_raw_size;
         _null_marker_offset = null_marker_offset;
-      } else {
+        _null_free_non_atomic_layout_alignment = _payload_alignment;
+      }
+      if (UseNullableValueFlattening) {
```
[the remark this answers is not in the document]
fparain: They are not mutually exclusive. Each layout can be enabled/disabled individually, mostly for testing/debugging purpose.

coleenp (approved):
Looks good!

fparain:
Thanks @merykitty @coleenp for your comments and reviews.

PR #1409, opened 2025-03-26
8352750: Use jimage for preview enabled value classes

david-beaumont:
Draft pull request. Final description TBD.

---------
### Progress

### Integration blocker
&nbsp;⚠️ Title mismatch between PR and JBS for issue [JDK-8352750](https://bugs.openjdk.org/browse/JDK-8352750)

### Issue
 * [JDK-8352750](https://bugs.openjdk.org/browse/JDK-8352750): [lworld] Use jimage for preview enabled value classes (**Enhancement** - P3) ⚠️ Title mismatch between PR and JBS.

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/1409/head:pull/1409` \
`$ git checkout pull/1409`

`$ git checkout pull/1409` \
`$ git pull https://git.openjdk.org/valhalla.git pull/1409/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1409`

`$ git pr show -t 1409`

</details>

<a href="https://git.openjdk.org/valhalla/pull/1409.diff">https://git.openjdk.org/valhalla/pull/1409.diff</a>

</details>

### DIFF
--- a/src/jdk.jlink/share/classes/jdk/tools/jmod/JmodTask.java
+++ b/src/jdk.jlink/share/classes/jdk/tools/jmod/JmodTask.java
@@ -731,8 +731,9 @@ public class JmodTask {
             return jf.stream()
                     .filter(e -> !e.isDirectory())
                     .map(JarEntry::getName)
-                    .map(name -> toPackageName(name))
-                    .flatMap(Optional::stream)
+                    .map(this::toPackageName)
+                    // Less overhead than "flatMap(Optional::stream)"
+                    .<String>mapMulti(Optional::ifPresent)
                     .collect(Collectors.toSet());
         }
 


src/hotspot/share/classfile/classLoader.cpp
```diff
@@ -97,9 +97,18 @@ static JImageClose_t                   JImageClose            = nullptr;
 static JImageFindResource_t            JImageFindResource     = nullptr;
 static JImageGetResource_t             JImageGetResource      = nullptr;
 
-// JimageFile pointer, or null if exploded JDK build.
```
david-beaumont, line 100: Typo fix.

src/hotspot/share/classfile/classLoader.cpp
```diff
@@ -97,9 +97,18 @@ static JImageClose_t                   JImageClose            = nullptr;
...
 
+// JImageMode status to control preview behaviour. JImage_file is unusable
+// for normal lookup until (JImage_mode != JIMAGE_MODE_UNINITIALIZED).
+enum JImageMode {
+  JIMAGE_MODE_UNINITIALIZED = 0,
+  JIMAGE_MODE_DEFAULT = 1,
+  JIMAGE_MODE_ENABLE_PREVIEW = 2
+};
+static JImageMode                      JImage_mode            = JIMAGE_MODE_UNINITIALIZED;
+
```
david-beaumont, line 111: Private static state. I don't like it being static, but everything else in ClassLoader is, so I figured put it next to the thing it affects.

src/hotspot/share/classfile/classLoader.cpp
```diff
@@ -154,7 +163,7 @@ void ClassLoader::print_counters(outputStream *st) {
 
 GrowableArray<ModuleClassPathList*>* ClassLoader::_patch_mod_entries = nullptr;
 GrowableArray<ModuleClassPathList*>* ClassLoader::_exploded_entries = nullptr;
-ClassPathEntry* ClassLoader::_jrt_entry = nullptr;
```
david-beaumont, line 157: Reflect the true type of _jrt_entry (this is useful later).

src/hotspot/share/classfile/classLoader.cpp
```diff
@@ -367,20 +376,21 @@ ClassFileStream* ClassPathZipEntry::open_stream(JavaThread* current, const char*
...
+  assert(JImage_file != nullptr, "should have been opened by ClassLoader::lookup_vm_options "
                            "and remained throughout normal JVM lifetime");
-  return jimage();
+  return JImage_file;
+}
+
+bool ClassPathImageEntry::is_preview_enabled() const {
+  assert(JImage_mode != JIMAGE_MODE_UNINITIALIZED, "uninitialized JImage reference");
+  return JImage_mode == JIMAGE_MODE_ENABLE_PREVIEW;
 }
```
david-beaumont, line 389: Helper function to encapsulate the static field read and ensure it's only used inside ClassPathImageEntry.
While other code could directly access JImage_mode, it's easier to audit when going through this function (and would be easier to refactor later if it stops being a static field). It also adds a guard to prevent use-before-init.

src/hotspot/share/classfile/classLoader.cpp
```diff
@@ -389,6 +399,8 @@ ClassPathImageEntry::ClassPathImageEntry(JImageFile* jimage, const char* name) :
   ClassPathEntry() {
   guarantee(jimage != nullptr, "jimage file is null");
   guarantee(name != nullptr, "jimage file name is null");
+  guarantee(JImage_mode != JIMAGE_MODE_UNINITIALIZED, "jimage is uninitialized");
+
```
david-beaumont, line 403: I made this a guarantee to match the null pointer checks above (this should be the only place, other than is_preview_enabled() where JImage_mode is read.

src/hotspot/share/classfile/classLoader.cpp
```diff
@@ -407,16 +419,19 @@ ClassFileStream* ClassPathImageEntry::open_stream(JavaThread* current, const cha
 //     2. A package is in at most one module in the jimage file.
 //
 ClassFileStream* ClassPathImageEntry::open_stream_for_loader(JavaThread* current, const char* name, ClassLoaderData* loader_data) {
+  JImageFile* jimage = this->jimage_non_null();
+  bool is_preview = this->is_preview_enabled();
+
```
david-beaumont, line 424: Read related fields together and only once. Now this is the only place where either of these fields are obtained (and guards are present to ensure it happens after final initialization of flags) we know it's the only bit of code that needs modifying later to handle is_preview. For now the flag is ignored however.

src/hotspot/share/classfile/classLoader.cpp
```diff
@@ -438,9 +453,9 @@ ClassFileStream* ClassPathImageEntry::open_stream_for_loader(JavaThread* current
       ClassLoader::perf_sys_classfile_bytes_read()->inc(size);
     }
     char* data = NEW_RESOURCE_ARRAY(char, size);
-    (*JImageGetResource)(jimage_non_null(), location, data, size);
+    (*JImageGetResource)(jimage, location, data, size);
     // Resource allocated
-    assert(this == (ClassPathImageEntry*)ClassLoader::get_jrt_entry(), "must be");
```
david-beaumont, line 443: No longer needs the cast since the return type is specific.

src/hotspot/share/classfile/classLoader.cpp
```diff
@@ -678,9 +693,11 @@ void ClassLoader::setup_bootstrap_search_path_impl(JavaThread* current, const ch
           const char* canonical_path = get_canonical_path(path, current);
           assert(canonical_path != nullptr, "canonical_path issue");
 
+          // Hand over lifecycle control of JImage_file to the _jrt_entry singleton (see
+          // ClassPathImageEntry::close_jimage). The image must be initialized by now.
+          assert(JImage_mode != JIMAGE_MODE_UNINITIALIZED, "init_jimage() must be called before here");
           _jrt_entry = new ClassPathImageEntry(JImage_file, canonical_path);
           assert(_jrt_entry != nullptr && _jrt_entry->is_modules_image(), "No java runtime image present");
-          assert(_jrt_entry->jimage() != nullptr, "No java runtime image");
         } // else it's an exploded build.
```
david-beaumont, line 701: This final assert is very very redundant given the asserts around it and in the ClassPathImageEntry constructor. I removed it because it is otherwise a call to the old jimage() method outside the ClassPathImageEntry namespace (which means I would need to leave a *public* method that returns the `JImageFile*`, which I think is quite bad.

src/hotspot/share/classfile/classLoader.hpp
```diff
@@ -107,19 +105,29 @@ class ClassPathZipEntry: public ClassPathEntry {
 };
 
 
-// For java image files
+// A singleton path entry which takes ownership of the initialized JImageFile
+// reference. Not used for exploded builds.
 class ClassPathImageEntry: public ClassPathEntry {
```
david-beaumont, line 110: I added some comments while working since there is subtlety here and things weren't very clear to me. It's important, for example, to know that this is explicitly intended as a singleton instance.

src/hotspot/share/classfile/classLoader.hpp
```diff
@@ -107,19 +105,29 @@ class ClassPathZipEntry: public ClassPathEntry {
...
   DEBUG_ONLY(static ClassPathImageEntry* _singleton;)
+
+  // Private to guarantee only ClassPathImageEntry can access the JImage structure,
+  // and ensure we completely control the behaviour around '--enable-preview'.
+  //
+  // Returns the non-null, initialized JImage file reference.
+  JImageFile* jimage_non_null() const;
+  // Returns whether the JImage file will return resources suitable for a preview JVM.
+  bool is_preview_enabled() const;
 public:
```
david-beaumont, line 122: Having these functions private guarantees nobody else can read (and potentially) misuse the `JImageFile*` outside the `ClassPathImageEntry` class. This is very important for improving encapsulation, which was somewhat lacking before.

src/hotspot/share/classfile/classLoader.hpp
```diff
@@ -107,19 +105,29 @@ class ClassPathZipEntry: public ClassPathEntry {
...
   bool is_modules_image() const;
   const char* name() const { return _name == nullptr ? "" : _name; }
-  JImageFile* jimage() const;
-  JImageFile* jimage_non_null() const;
+  // Called to closes the JImage during os::abort (normally not called).
   void close_jimage();
```
david-beaumont, line 126: Comment added here because it's otherwise confusing to see a "close" method in a C++ object, because you naturally expect a destructor to be used. This is optional, but I think it helps people understand that this isn't "normal" semantics.

src/hotspot/share/classfile/classLoader.hpp
```diff
@@ -63,8 +63,6 @@ class ClassPathEntry : public CHeapObj<mtClass> {
   // Is this entry created from the "Class-path" attribute from a JAR Manifest?
   virtual bool from_class_path_attr() const { return false; }
   virtual const char* name() const = 0;
-  virtual JImageFile* jimage() const { return nullptr; }
-  virtual void close_jimage() {}
```
RogerRiggs, line 67: I'm not sure but this provides close_jimage for ClassPathZipEntry;  though it is a no-op, as usual for zip files.

david-beaumont: Yes, but it seems nobody calls it virtually and everything compiles fine without it, so I'm calling that "dead code".
Much better encapsulation without this.

make/GenerateLinkOptData.gmk
```diff
@@ -66,7 +66,7 @@ endif
 #   default classlist is minimal, let's filter out the '@cp' lines until we can
 #   find a proper solution.
 CLASSLIST_FILE_VM_OPTS = \
-    -Duser.language=en -Duser.country=US
+    -Duser.language=en -Duser.country=US -Dvalue.bsm.debug=true
```
RogerRiggs, line 69: Can be removed.

src/jdk.jlink/share/classes/jdk/tools/jmod/JmodTask.java
```diff
@@ -682,67 +682,79 @@ Set<String> findPackages(List<Path> classpath) {
...
          */
         Set<String> findPackages(Path dir) {
             try (Stream<Path> stream = Files.find(dir, Integer.MAX_VALUE,
                                   (path, attrs) -> attrs.isRegularFile(),
                                   FileVisitOption.FOLLOW_LINKS)) {
                 return stream.map(dir::relativize)
-                        .filter(path -> isResource(path.toString()))
                         .map(path -> toPackageName(path))
-                        .filter(pkg -> pkg.length() > 0)
+                        .flatMap(Optional::stream)
```
liach, line 695: Maybe `mapMulti(Optional::ifPresent)`?

david-beaumont: That doesn't actually work, since it ends up with a sequence of Object, not String.

liach: What if `.<String>mapMulti(Optional::ifPresent)`?

david-beaumont: I mean maybe (I'm sure it can be made to work), but is that any clearer at this stage?

david-beaumont:
Obsolete now. Well open a new PR when the latest changes are available.

PR #1411, opened 2025-03-26
8352750: More encapsulation of JImage file. Too much?

david-beaumont:
---------
### Progress

### Integration blocker
&nbsp;⚠️ Title mismatch between PR and JBS for issue [JDK-8352750](https://bugs.openjdk.org/browse/JDK-8352750)

### Issue
 * [JDK-8352750](https://bugs.openjdk.org/browse/JDK-8352750): [lworld] Use jimage for preview enabled value classes (**Enhancement** - P3) ⚠️ Title mismatch between PR and JBS.

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/1411/head:pull/1411` \
`$ git checkout pull/1411`

`$ git checkout pull/1411` \
`$ git pull https://git.openjdk.org/valhalla.git pull/1411/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1411`

`$ git pr show -t 1411`

</details>

<a href="https://git.openjdk.org/valhalla/pull/1411.diff">https://git.openjdk.org/valhalla/pull/1411.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/classfile/classLoader.cpp
+++ b/src/hotspot/share/classfile/classLoader.cpp
@@ -253,9 +253,10 @@ Symbol* ClassLoader::package_from_class_name(const Symbol* name, bool* bad_class
 // The following jimage_xxx static functions encapsulate all JImage_file and JImage_mode access.
 // This is done to make it easy to reason about the JImage file state (exists vs initialized etc.).
 
-// Opens the named JImage file.
+// Opens the named JImage file and sets the JImage file reference.
+// Returns true if opening the JImage file was successful (see also jimage_exists()).
 static bool jimage_open(const char* modules_path) {
-  // Current 'error' is not set to anything useful so ignore it here.
+  // Currently 'error' is not set to anything useful, so ignore it here.
   jint error;
   JImage_file = (*JImageOpen)(modules_path, &error);
   return JImage_file != nullptr;
--- a/src/hotspot/share/classfile/classLoader.cpp
+++ b/src/hotspot/share/classfile/classLoader.cpp
@@ -97,8 +97,17 @@ static JImageClose_t                   JImageClose            = nullptr;
 static JImageFindResource_t            JImageFindResource     = nullptr;
 static JImageGetResource_t             JImageGetResource      = nullptr;
 
-// JimageFile pointer, or null if exploded JDK build.
+// JImageFile pointer, or null if exploded JDK build.
 static JImageFile*                     JImage_file            = nullptr;
+
+// JImageMode status to control preview behaviour. JImage_file is unusable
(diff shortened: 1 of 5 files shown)


src/hotspot/share/classfile/classLoader.cpp
```diff
@@ -240,6 +249,74 @@ Symbol* ClassLoader::package_from_class_name(const Symbol* name, bool* bad_class
   return SymbolTable::new_symbol(name, pointer_delta_as_int(start, base), pointer_delta_as_int(end, base));
 }
 
+// --------------------------------
```
david-beaumont, line 252: The section below look like a lot, but it's been really useful for handling (image file / access mode) state.
The rest of the code reads more semantically cleaner because of these.

src/hotspot/share/classfile/classLoader.cpp
```diff
@@ -367,28 +444,15 @@ ClassFileStream* ClassPathZipEntry::open_stream(JavaThread* current, const char*
...
-
 void ClassPathImageEntry::close_jimage() {
-  if (jimage() != nullptr) {
-    (*JImageClose)(jimage());
-    JImage_file = nullptr;
-  }
+  jimage_close();
 }
 
-ClassPathImageEntry::ClassPathImageEntry(JImageFile* jimage, const char* name) :
```
david-beaumont, line 388: It really didn't seem useful or helpful to pass in what looks like an instance field value, only to have it ignored by the constructor (other than testing it's the thing it should be). It seems better to just use the static function to get and reason about the jimage file directly. Everything still compiles, so it's not like it was even useful for testing either.

src/hotspot/share/classfile/classLoader.cpp
```diff
@@ -450,16 +516,9 @@ ClassFileStream* ClassPathImageEntry::open_stream_for_loader(JavaThread* current
   return nullptr;
 }
 
-JImageLocationRef ClassLoader::jimage_find_resource(JImageFile* jf,
```
david-beaumont, line 453: Replaced by jimage_find_resource() static function which doesn't require callers to get and pass the JImage file reference (making it clear that this code is unavoidably static).

src/hotspot/share/classfile/classLoader.cpp
```diff
@@ -1520,49 +1580,51 @@ void ClassLoader::initialize(TRAPS) {
   setup_bootstrap_search_path(THREAD);
 }
 
-static char* lookup_vm_resource(JImageFile *jimage, const char *jimage_version, const char *path) {
```
david-beaumont, line 1523: Only actually called once below, so felt better to inline it, since only "string" resources should be treated like C-strings and by inlining it, it's clear that we're talking about the options *string*.

PR #1425, opened 2025-04-09
8353236: [lworld] Better documentation for Valhalla Unsafe intrinsics

merykitty:
Hi,

This patch clarifies the meaning of undefined behavior when working with `Unsafe` and specifies the requirements when working with `Unsafe::makePrivateBuffer` and `Unsafe::finishPrivateBuffer`.

Please take a look and leave your suggestions, thanks a lot.

---------
### Progress

### Issue
 * [JDK-8353236](https://bugs.openjdk.org/browse/JDK-8353236): [lworld] Better documentation for Valhalla Unsafe intrinsics (**Enhancement** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/1425/head:pull/1425` \
`$ git checkout pull/1425`

`$ git checkout pull/1425` \
`$ git pull https://git.openjdk.org/valhalla.git pull/1425/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1425`

`$ git pr show -t 1425`

</details>

<a href="https://git.openjdk.org/valhalla/pull/1425.diff">https://git.openjdk.org/valhalla/pull/1425.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/1425#issuecomment-2789698891)
</details>

### DIFF
--- a/src/java.base/share/classes/jdk/internal/misc/Unsafe.java
+++ b/src/java.base/share/classes/jdk/internal/misc/Unsafe.java
@@ -60,6 +60,7 @@ import static jdk.internal.misc.UnsafeConstants.*;
  * behaviors may include but not be limited to:
  *
  * <ul>
+ * <li>Working as expected.
  * <li>Crashing the VM.
  * <li>Corruption of the heap or JVM memory.
  * <li>Nonsensical variable value. E.g. an {@code int} may appear to be
@@ -89,9 +90,6 @@ import static jdk.internal.misc.UnsafeConstants.*;
  * {@code Unsafe} invocation, these assumptions may propagate backward to
  * previous statements, leading to wrong executions if the assumptions are
  * invalid.
- * <p>
- * By default, usage of all methods in this class exhibits undefined behavior,
- * unless otherwise explicitly specified.
  *
  * @author John R. Rose
  * @see #getUnsafe
--- a/src/java.base/share/classes/jdk/internal/misc/Unsafe.java
+++ b/src/java.base/share/classes/jdk/internal/misc/Unsafe.java
@@ -41,14 +41,39 @@ import static jdk.internal.misc.UnsafeConstants.*;
  * Although the class and all methods are public, use of this class is
  * limited because only trusted code can obtain instances of it.
  *
- * <em>Note:</em> It is the responsibility of the caller to make sure
- * arguments are checked before methods of this class are
- * called. While some rudimentary checks are performed on the input,
- * the checks are best effort and when performance is an overriding
- * priority, as when methods of this class are optimized by the
- * runtime compiler, some or all checks (if any) may be elided. Hence,
- * the caller must not rely on the checks and corresponding
- * exceptions!
+ * <h2><a id="undefined-behavior">Undefined Behavior</a></h2>
+ * For performance reasons, {@code Unsafe} is allowed to work outside the
+ * restrictions enforced by the JVM. As a result, it is the responsibility of
+ * the caller to ensure that an invocation of an {@code Unsafe} method is
+ * conformant, and failure to do so will result in undefined behavior. When a
+ * program exhibits undefined behavior, there is no restrictions on its
+ * behaviors. Such behaviors may include but not be limited to:
+ *
+ * <ul>
+ * <li>Crashing the VM.
+ * <li>Corruption of the heap or JVM memory.
+ * <li>Nonsensical variable value. E.g. an {@code int} may appear to be
+ * simultaneously 0 and 1.
+ * <li>Impossible code execution. E.g. the branches of an {@code if} are
+ * both executed or both not executed.
+ * <li>Wiping out the hard drive.
+ * </ul>
+ *
+ * Some methods (e.g. {@link #getInt}) exhibit undefined behavior if they
+ * are invoked at runtime with illegal arguments. This means that they will
+ * never exhibit undefined behavior if they are not actually reachable at
+ * runtime. On the other hands, other methods (e.g.
+ * {@link #makePrivateBuffer(Object)}) exhibit undefined behavior if they are
+ * used incorrectly, even if the invocation may not be reachable at runtime.
+ * <p>
+ * For methods exhibiting undefined behavior if they are invoked at runtime
+ * with illegal arguments, undefined behavior may time travel. That is, if a
+ * control path may eventually reach an invocation of an {@code Unsafe} method
+ * with illegal arguments, the symptoms of undefined behavior may be present
+ * even before the invocation of the {@code Unsafe} method.
+ * <p>
+ * By default, usage of all methods in this class exhibits undefined behavior,
+ * unless otherwise explicitly specified.
  *
  * @author John R. Rose
  * @see #getUnsafe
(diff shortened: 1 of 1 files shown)


src/java.base/share/classes/jdk/internal/misc/Unsafe.java
```diff
@@ -41,14 +41,39 @@
...
+ * never exhibit undefined behavior if they are not actually reachable at
+ * runtime. On the other hands, other methods (e.g.
+ * {@link #makePrivateBuffer(Object)}) exhibit undefined behavior if they are
+ * used incorrectly, even if the invocation may not be reachable at runtime.
+ * <p>
+ * For methods exhibiting undefined behavior if they are invoked at runtime
+ * with illegal arguments, undefined behavior may time travel. That is, if a
+ * control path may eventually reach an invocation of an {@code Unsafe} method
+ * with illegal arguments, the symptoms of undefined behavior may be present
+ * even before the invocation of the {@code Unsafe} method.
```
liach, line 73: I think you can talk in a more JIT-oriented manner, that while Unsafe is mostly set up to be the hooks between core libraries and hotspot runtime, hotspot compiler may interpret the input values and perform aggressive optimizations with contracts that cannot be explicitly expressed by Java program semantics.

src/java.base/share/classes/jdk/internal/misc/Unsafe.java
```diff
@@ -41,14 +41,57 @@
...
+ * control path may eventually reach an invocation of an {@code Unsafe} method
+ * with illegal arguments, the symptoms of undefined behavior may be present
+ * even before the invocation of the {@code Unsafe} method. This is because the
+ * JIT compiler may have certain assumptions about the inputs of an
+ * {@code Unsafe} invocation, these assumptions may propagate backward to
+ * previous statements, leading to wrong executions if the assumptions are
+ * invalid.
+ * <p>
+ * By default, usage of all methods in this class exhibits undefined behavior,
+ * unless otherwise explicitly specified.
```
RogerRiggs, line 94: The "by default" and "unless otherwise specified" seems to be redundant.

```suggestion
 * Unless otherwise explicitly specified, all methods in this class can exhibit undefined behavior.
 * .
```

src/java.base/share/classes/jdk/internal/misc/Unsafe.java
```diff
@@ -381,8 +424,30 @@ public native Object[] newSpecialArray(Class<?> componentType,
...
+     * result, it is the responsibility of the caller to ensure that this
+     * method is used in a conformant manner. This method exhibits
+     * {@linkplain Unsafe undefined behavior} unless all the following
+     * conditions are true:
+     * <ul>
+     * <li>The argument must be a value object at runtime.
+     * <li>The return object must be assigned to a local variable that is
+     * effectively final as well as definitely assigned to with the return
+     * value of this method. The object must also be not assigned to another
+     * local variable.
```
RogerRiggs, line 442: Why this restriction? What assumption is violated if there is a second assignment?

src/java.base/share/classes/jdk/internal/misc/Unsafe.java
```diff
@@ -381,8 +424,30 @@ public native Object[] newSpecialArray(Class<?> componentType,
...
+     * <li>The return object can only be used as the first argument passed to
+     * {@code Unsafe::putXXX} or to {@link #finishPrivateBuffer(Object)}, any
+     * other usage, such as loading from or returning it, is illegal. The only
+     * exception is the implicit check cast inserted by the compiler on the
+     * return value of this method. Explicit check casts are not allowed.
```
RogerRiggs, line 447: "not allowed"?  But there is no enforcement; unpredictable behavior yes.

src/java.base/share/classes/jdk/internal/misc/Unsafe.java
```diff
@@ -381,8 +424,30 @@ public native Object[] newSpecialArray(Class<?> componentType,
...
+     * </ul>
+     * Illegal usage of this method exhibits undefined behavior even if the
```
RogerRiggs, line 449: may or can; but not always.

src/java.base/share/classes/jdk/internal/misc/Unsafe.java
```diff
@@ -392,6 +457,19 @@ public native Object[] newSpecialArray(Class<?> componentType,
...
+     * {@linkplain Unsafe undefined behavior} unless all the following
+     * conditions are true:
+     * <ul>
+     * <li>The argument must be a larval value object created by
+     * {@link #makePrivateBuffer(Object)}.
+     * <li>After the invocation of this method, the variable that holds the
+     * argument passed into this method must not be used.
+     * </ul>
+     * Illegal usage of this method exhibits undefined behavior even if the
+     * illegal statements are never actually reached at runtime.
```
RogerRiggs, line 472: Another can/may exhibit case; improve the phrase everywhere.

src/java.base/share/classes/jdk/internal/misc/Unsafe.java
```diff
@@ -392,6 +457,19 @@ public native Object[] newSpecialArray(Class<?> componentType,
...
+     * This method mimics the action of passing a larval object to a do-nothing
+     * constructor to obtain a non-larval object. This method exhibits
+     * {@linkplain Unsafe undefined behavior} unless all the following
+     * conditions are true:
+     * <ul>
+     * <li>The argument must be a larval value object created by
+     * {@link #makePrivateBuffer(Object)}.
+     * <li>After the invocation of this method, the variable that holds the
+     * argument passed into this method must not be used.
+     * </ul>
```
RogerRiggs, line 470: I would add the exhortation that every field of the value object must have a value.

src/java.base/share/classes/jdk/internal/misc/Unsafe.java
merykitty: I decide to just remove it as each method specifies its behaviours already.

src/java.base/share/classes/jdk/internal/misc/Unsafe.java
merykitty: Not allowed here in the sense that failure to do so will result in undefined behavior.

src/java.base/share/classes/jdk/internal/misc/Unsafe.java
merykitty: Illegal usage ALWAYS exhibits undefined behavior, the uncertainty here is that whether this undefined behavior manifests itself as an unexpected operation.

jatin-bhateja:
Hi @merykitty ,

Just to complement your documentation update, I have created test scenarios for various verification errors around implicit larval objects, i.e. the ones created using new operator but are still pending initialization,  explicit larvals generated using Unsafe.makePrivateBuffer should adhere to the same semantics.

[Implicit larval objects verification.docx](https://github.com/user-attachments/files/20243499/Implicit.larval.objects.verification.docx)

Overall, your changes look good to me; finer refinements can be done in due course, but specs are much clear now, thank you so much!

src/java.base/share/classes/jdk/internal/misc/Unsafe.java
```diff
@@ -41,14 +41,55 @@
...
+ *
+ * <ul>
+ * <li>Working as expected.
+ * <li>Crashing the VM.
+ * <li>Corruption of the heap or JVM memory.
+ * <li>Nonsensical variable value. E.g. an {@code int} may appear to be
+ * simultaneously 0 and 1.
+ * <li>Impossible code execution. E.g. the branches of an {@code if} are
+ * both executed or both not executed.
+ * <li>Wiping out the hard drive.
```
liach, line 70: I think these are just for alerting people; this is not that valuable to create a huge list to specifically highlight. I think what we should highlight instead is the best practices.

liach:
Should we change these long paragraphs to a design document, as suggested in  https://github.com/openjdk/jdk/pull/24777#pullrequestreview-2864975777?

PR #1426, opened 2025-04-10
8353717: [lworld] Complete merge of 8350756

chhagedorn:
This patch fixes some failures after merging JDK-25+14 in:

1. [JDK-8333393](https://bugs.openjdk.org/browse/JDK-8333393) improved the "split Phi through MergeMem" optimization and was merged in:
    1. This should fix [JDK-8327465](https://bugs.openjdk.org/browse/JDK-8327465) (it does not reproduce anymore but I think it's still an issue before this patch): When adding `Phi` nodes to the `OuterStripMinedLoop` node, it should match the number of `Phi` nodes at the `CountedLoop` node. However, during IGVN, we could split some phis of the `CountedLoop` node through `MergeMems` which created more phi nodes at the `OuterStripMinedLoop` node. But these phis could not be split further and we hit an assertion failure due to this phi number mismatch. Only with JDK-8333393, we can further split these phis at the `OuterStripMinedLoop` node and we no longer hit the assert. I've therefore closed JDK-8327465 as a dup.
    2. The merge of JDK-8333393 was not correct:
        1. We ended up splitting fewer phi nodes through mergemems as before and thus hitting case `1.` even more often.
        2. The Valhalla specific "Always split `Phi` when only having `MergeMem` inputs" optimization was broken. This leads to some IR rule failures since the splitting optimization is no longer performed which blocked some loop optimizations (e.g. turning a loop into a counted one and then apply more optimizations).
     
        I fixed both issues.
2. With JDK-25+14, some tests to check Loop Unswitching started to fail which matched on the number of `CountedLoop` nodes. The reason is that we can now eliminate one of the unswitched loop versions which we previously could not. Therefore, the number of `CountedLoop` nodes is one less. To fix that, I've changed the IR matching to `CCP1` where we still have both unswitched loop versions. This required the following:
    1. I had to adjust `INLINE_ARRAY_NULL_GUARD` to match on ideal graphs instead of the opto assembly. This can easily be done since all information is there in the ideal graph.
    2. I removed the `applyIf*` conditions for `UseG1GC` because we are not running any scenario with another GC.

Thanks,
Christian

---------
### Progress

### Issue
 * [JDK-8353717](https://bugs.openjdk.org/browse/JDK-8353717): [lworld] Complete merge of 8350756 (**Bug** - P2)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/1426/head:pull/1426` \
`$ git checkout pull/1426`

`$ git checkout pull/1426` \
`$ git pull https://git.openjdk.org/valhalla.git pull/1426/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1426`

`$ git pr show -t 1426`

</details>

<a href="https://git.openjdk.org/valhalla/pull/1426.diff">https://git.openjdk.org/valhalla/pull/1426.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/1426#issuecomment-2791667604)
</details>

### DIFF
--- a/src/hotspot/share/opto/cfgnode.cpp
+++ b/src/hotspot/share/opto/cfgnode.cpp
@@ -2565,7 +2565,7 @@ Node *PhiNode::Ideal(PhaseGVN *phase, bool can_reshape) {
             }
           }
         }
-      } else if (split_always_terminates) {
+      } else if (mergemem_only || split_always_terminates) {
         // If all inputs reference this phi (directly or through data nodes) -
         // it is a dead loop.
         bool saw_safe_input = false;
--- a/src/hotspot/share/opto/loopnode.cpp
+++ b/src/hotspot/share/opto/loopnode.cpp
@@ -2492,8 +2492,7 @@ void LoopNode::verify_strip_mined(int expect_skeleton) const {
       }
       // Late optimization of loads on backedge can cause Phi of outer loop to be eliminated but Phi of inner loop is
       // not guaranteed to be optimized out.
-      // TODO: 8353717
-      //assert(outer->outcnt() >= phis + 2 - be_loads && outer->outcnt() <= phis + 2 + stores + 1, "only phis");
+      assert(outer->outcnt() >= phis + 2 - be_loads && outer->outcnt() <= phis + 2 + stores + 1, "only phis");
     }
     assert(sfpt->outcnt() == 1, "no data node");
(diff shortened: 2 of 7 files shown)


TobiHartmann (approved):
Looks good to me. Thanks for fixing this Christian!

chhagedorn:
Thanks for the review!

/integrate

PR #1449, opened 2025-05-01
8352068: [lworld] test StrictFinalInstanceFieldsTest.java needs to be updated after fix for JDK-8351951

matias9927:
Updates to javac prevent it from creating classes with improper uses of strict fields. Additionally, non-final strict fields are now allowed so `StrictFinalInstanceFieldsTest.java` needs to updated with new test cases. This patch adds new cases for final and non-final stricts as well as negative test cases using asmtools that check improper use of strict fields and early larval stack map frames. 

During testing, it was revealed that the new verify errors added in [JDK-8354694](https://bugs.openjdk.org/browse/JDK-8354694) do not work properly so a fix is included in this patch.

Note: This cannot be integrated until jtreg is updated with a recent build of asmtools which can handle strict fields and can generate early_larval frames.

---------
### Progress

### Issue
 * [JDK-8352068](https://bugs.openjdk.org/browse/JDK-8352068): [lworld] test StrictFinalInstanceFieldsTest.java needs to be updated after fix for JDK-8351951 (**Bug** - P4)

### Reviewers
 * [Chen Liang](https://openjdk.org/census#liach) (@liach - no project role) ⚠️ Review applies to [86fcbf48](https://git.openjdk.org/valhalla/pull/1449/files/86fcbf486e8ba831f9668ae87d197c4591ca86ee)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/1449/head:pull/1449` \
`$ git checkout pull/1449`

`$ git checkout pull/1449` \
`$ git pull https://git.openjdk.org/valhalla.git pull/1449/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1449`

`$ git pr show -t 1449`

</details>

<a href="https://git.openjdk.org/valhalla/pull/1449.diff">https://git.openjdk.org/valhalla/pull/1449.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/1449#issuecomment-2845349031)
</details>


test/hotspot/jtreg/ProblemList.txt
```diff
@@ -139,6 +139,7 @@ runtime/AccModule/ConstModule.java 8294051 generic-all
 runtime/valhalla/inlinetypes/CircularityTest.java 8349037 generic-all
 runtime/valhalla/inlinetypes/PreloadCircularityTest.java 8349631 linux-aarch64-debug
 runtime/valhalla/inlinetypes/ValuePreloadTest.java 8349630 linux-aarch64-debug
+runtime/valhalla/inlinetypes/verifier/StrictInstanceFieldsTest.java generic-all
```
coleenp, line 142: Does this require a bug number?

coleenp: Maybe CODETOOLS-7903081 will work.

liach (approved):
Just noticed these jasm and jcod are from StrictInstanceFieldsTest. Should be all good.

DanHeidinga (commented):
Maintaining jasm and jcod files can be a pain.  The less those files contain, the better as it is less classfile / bytecode to try to build in your head while debugging.

One option for set of common tests like this that all have x & y fields, as well as getter methods and a complicated toString, is to use an abstract parent class:
```
abstract JasmParent extends Parent {
  abstract int get_x();
  abstract int get_y();
  
  public String toString() {
          return "x: " + get_x() + "\n" + "y: " + get_y() + "\n" + super.toString();
    }
}
```
and have all the .jasm file class extend this class instead of using Parent directly.  This will move all the bootstrap methods & indy related code out of jasm files and hopefully result in something simpler to look at.

Another option is to keep Parent as it is, but have its toString() use Reflection like: `this.getClass().getDeclaredFeilds()` to find all the fields and print them.  Any solution that moves code out of jasm / jcod files makes the code easier to maintain and understand longterm

test/hotspot/jtreg/runtime/valhalla/inlinetypes/verifier/StrictFieldsNotSubset.jcod
```diff
@@ -0,0 +1,273 @@
...
+                  21  126;
+                  29  129;
+                  39  131;
+                  43  132;
+              }
+            }                              // end of LineNumberTable
+            ;
+            Attr(#36) {                    // StackMapTable
+              [] {                         //
+                246b, []{#8; #12}, {            // early_larval_frame
```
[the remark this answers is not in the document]
matias9927: The issue is that there is a field, or rather a NameAndType, that doesn't exist in the initial set of strict fields. The early_larval frame offers a set of fields that are not a subset of the initial set of strict fields generated in the VM. This is certainly one of the larger and more confusing tests so I'll add comments to this test as well as the others.

matias9927:
Thanks for the reviews @DanHeidinga and @liach!
/integrate

PR #1465, opened 2025-05-20
8349945: Implement strict static fields (proposed JVM feature)

matias9927:
This patch enables and implements verification for fields with the ACC_STATIC and ACC_STRICT modifiers. To enforce strictness on static fields, the reads and writes on the field are tracked dynamically to ensure that the field is written before being read and written to before.

---------
### Progress

### Issue
 * [JDK-8349945](https://bugs.openjdk.org/browse/JDK-8349945): Implement strict static fields (proposed JVM feature) (**Enhancement** - P4)

### Reviewers
 * [Dan Heidinga](https://openjdk.org/census#heidinga) (@DanHeidinga - no project role) ⚠️ Review applies to [95c76e6c](https://git.openjdk.org/valhalla/pull/1465/files/95c76e6c319ebb529d7dfe72708e8c30550a8d94)
 * [Chen Liang](https://openjdk.org/census#liach) (@liach - Committer) ⚠️ Review applies to [95c76e6c](https://git.openjdk.org/valhalla/pull/1465/files/95c76e6c319ebb529d7dfe72708e8c30550a8d94)
 * [Frederic Parain](https://openjdk.org/census#fparain) (@fparain - Committer) ⚠️ Review applies to [8631c3c0](https://git.openjdk.org/valhalla/pull/1465/files/8631c3c0e2613b9a304bc34e6eb87a54703ef19e)

### Contributors
 * John R Rose `<jrose@openjdk.org>`

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/1465/head:pull/1465` \
`$ git checkout pull/1465`

`$ git checkout pull/1465` \
`$ git pull https://git.openjdk.org/valhalla.git pull/1465/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1465`

`$ git pr show -t 1465`

</details>

<a href="https://git.openjdk.org/valhalla/pull/1465.diff">https://git.openjdk.org/valhalla/pull/1465.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/1465#issuecomment-2895962657)
</details>


test/hotspot/jtreg/runtime/valhalla/inlinetypes/verifier/Brbefore_BAD.jasm
```diff
@@ -0,0 +1,50 @@
...
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+identity class Brbefore_BAD version 69:65535
```
liach, line 24: Can't this be compiled with javac? I don't think we are going to add any checks for strict initialization on the javac side as pre-write reads cannot be detected because they can easily happen in nested method calls.

Even if we decide to use jasm, I think the best practice seems to leave the source to generate the jasm in a .test file, and comment in the jasm what was changed.

src/hotspot/share/c1/c1_Runtime1.cpp
```diff
@@ -1165,6 +1166,9 @@ JRT_ENTRY(void, Runtime1::patch_code(JavaThread* current, C1StubId stub_id ))
     // field because it was unknown at compile time.
     deoptimize_for_flat = result.is_flat();
 
+    // Strict statics may require tracking if their class is not fully initialized.
+    // For now we can bail out of the compiler and let the interpreter handle it.
+    deoptimize_for_strict_static = result.is_strict_static_unset();
```
DanHeidinga, line 1171: Deoptimizing on strict fields makes sense when the field is unset as an initial implementation.

As a possible future optimization, we could make some additional checks here to decide whether we need to deopt or not and instead simulate the strict static is_set bits here.  `current` Thread is either the thread initializing the class and therefore able to read/update the set/unset bits or it is some other thread that would need to wait.

If `current` is the initializing thread, and this is a strict static field write, we know we're currently executing the field set or we wouldn't be patching the code.  Given that, it would be possible to avoid the deopt by calling a helper from this code to toggle the bit.

Something to think about for the future when strict statics are more common to avoid spurious deopts

src/hotspot/share/oops/instanceKlass.cpp
```diff
@@ -1561,6 +1591,88 @@ void InstanceKlass::set_initialization_state_and_notify(ClassState state, TRAPS)
...
+      // clear the "unset" bit, since the field is actually going to be written
+      fs.update_strict_static_unset(false);
+    } else {
+      // throw an IllegalStateException, since we are reading before writing
+      // see also InstanceKlass::initialize_impl, Step 8 (at end)
+      Symbol* bad_strict_static = field(field_index).name(constants());
+      throw_strict_static_exception(bad_strict_static, "is unset before first read in", CHECK);
+    }
+  } else {
+    // Experimentally, enforce additional proposed conditions after the first write.
```
DanHeidinga, line 1641:
```suggestion
    // Ensure no write after read for final strict statics
```
I think this comment may be left over from the multi-option initial prototype.  It might be worth going over all the comments and making sure they still apply now that we've settled on a single approach

src/hotspot/share/oops/instanceKlass.cpp
```diff
@@ -1561,6 +1591,88 @@ void InstanceKlass::set_initialization_state_and_notify(ClassState state, TRAPS)
...
+    FieldInfo fi = field(field_index);
+    bool is_final = fi.access_flags().is_final();
+    if (is_final) {
+      // no final write after read, so observing a constant freezes it, as if <clinit> ended early
+      // (maybe we could trust the constant a little earlier, before <clinit> ends)
+      if (is_writing && !fs.is_strict_static_unread()) {
+        Symbol* bad_strict_static = fi.name(constants());
+        throw_strict_static_exception(bad_strict_static, "is set after read (as final) in", CHECK);
+      } else if (!is_writing && fs.is_strict_static_unread()) {
+        // log the read (this requires an extra status bit, with extra tests on it)
```
DanHeidinga, line 1651:
```suggestion
```
This comment also seems like a holdover from the prototype

src/hotspot/share/oops/instanceKlassFlags.hpp
```diff
@@ -59,6 +59,8 @@ class InstanceKlassFlags {
     flag(is_naturally_atomic                , 1 << 16) /* loaded/stored in one instruction*/ \
     flag(must_be_atomic                     , 1 << 17) /* doesn't allow tearing */ \
     flag(has_loosely_consistent_annotation  , 1 << 18) /* the class has the LooselyConsistentValue annotation WARNING: it doesn't automatically mean that the class allows tearing */ \
+    flag(is_implicitly_constructible        , 1 << 19) /* the class has the ImplicitlyConstrutible annotation */ \
```
DanHeidinga, line 62: Did this sneak into this patch from somewhere else?

src/hotspot/share/prims/jni.cpp
```diff
@@ -2055,8 +2055,11 @@ JNI_ENTRY(jfieldID, jni_GetStaticFieldID(JNIEnv *env, jclass clazz,
   k->initialize(CHECK_NULL);
 
   fieldDescriptor fd;
+  // strict fields need special tracking during <clinit>; do not hand them out so early
   if (!k->is_instance_klass() ||
-      !InstanceKlass::cast(k)->find_field(fieldname, signame, true, &fd)) {
+      !InstanceKlass::cast(k)->find_field(fieldname, signame, true, &fd) ||
+      // strict fields need special tracking during <clinit>; do not hand them out so early
+      (fd.access_flags().is_strict() && !InstanceKlass::cast(k)->is_initialized())) {
```
DanHeidinga, line 2062: Should this be an assert rather than a runtime check given that we force the class to be initialized a few lines above?

test/hotspot/jtreg/runtime/valhalla/inlinetypes/verifier/Brbefore_BAD.jasm
matias9927: Javac can catch this because the static field hasn't been initialized yet:
`error: variable F2__STRICT might not have been initialized
        int x = F2__STRICT; //FAIL`
It may be possible to write a java program with nested calls that does this but I think it's best to keep the test cases simple.

I agree with the last comment, I'll add the equivalent java code to the jasm files.

liach:
I think reflective write for static final is already impossible - at least that is what I know about Method Handles and core reflection. So we probably just need to care about reflective reads.

liach (approved):
I can keep an eye on the verification part of classfile and investigate that from the language and tooling side. Please simplify the debug output of a ClassModel as I suggested above. Otherwise, changes to java.lang.invoke looks good.

liach:
I think this is ready for integration; we can fine tune the test and other things later.

matias9927:
Thank you for the reviews @DanHeidinga @fparain and @liach and thank you @rose00 for the prototype!
/integrate

PR #1478, opened 2025-05-31
8353432: [lworld] Deoptimization needs to handle nullable, flat fields in non-value class holders

merykitty:
Hi,

This patch adds handling of nullable flat fields in non-value class holder. Currently, we add all the fields to the safepoint, then add all the null markers. I decide to change this to adding the null marker of a flat field right after all the subfields of it. It makes the shape similar to how we pass nullable flat fields in the ABI, as well as simplifies the implementation since we do not have to keep track of all null markers when traversing the object payload.

### DIFF
--- a/src/hotspot/share/opto/inlinetypenode.cpp
+++ b/src/hotspot/share/opto/inlinetypenode.cpp
@@ -273,15 +273,16 @@ int InlineTypeNode::field_null_marker_offset(uint index) const {
   return field->null_marker_offset();
 }
 
-uint InlineTypeNode::add_fields_to_safepoint(Unique_Node_List& worklist, Node_List& null_markers, SafePointNode* sfpt) {
+uint InlineTypeNode::add_fields_to_safepoint(Unique_Node_List& worklist, SafePointNode* sfpt) {
   uint cnt = 0;
   for (uint i = 0; i < field_count(); ++i) {
     Node* value = field_value(i);
     if (field_is_flat(i)) {
       InlineTypeNode* vt = value->as_InlineType();
-      cnt += vt->add_fields_to_safepoint(worklist, null_markers, sfpt);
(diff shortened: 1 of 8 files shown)


SidneyLann:
@merykitty Hi, valhalla is almost ready for experimental use now, right?

merykitty:
@SidneyLann Valhalla is ready for experimental use, you can either build the project from source (build instructions can be found [here](https://openjdk.org/groups/build/doc/building.html)) or you can grab a prebuilt package [here](https://builds.shipilev.net/). Please give it a try and report to us any issue you find, it would be a great help in the stabilization of Valhalla.

If you want to know whether Valhalla can be released to mainline soon then the answer is we don't know and we are trying our best. I believe an act of trying, reporting issues, and even contributing will help Valhalla to land sooner.

SidneyLann:
Ok. I'll try, but this repo does not allow submit issues. Thank you.

merykitty:
@SidneyLann That's great, you can report an issue by sending an email to [valhalla-dev](https://mail.openjdk.org/mailman/listinfo/valhalla-dev).

TobiHartmann (approved):
> I decide to change this to adding the null marker of a flat field right after all the subfields of it. It makes the shape similar to how we pass nullable flat fields in the ABI, as well as simplifies the implementation since we do not have to keep track of all null markers when traversing the object payload.

Yes, that's much cleaner.  Thanks for the refactoring.

The fix looks good to me, I just added two comments.

TobiHartmann:
Looks like there's a merge conflict now that you pushed https://git.openjdk.org/valhalla/pull/1470.

TobiHartmann (approved):
Thanks for making these adjustments. All tests passed and the fix looks good to me!

merykitty:
Thanks a lot for your speedy reviews!
/integrate

PR #1490, opened 2025-06-20
8359370: [lworld] allow instance fields of identity classes to be readable in the prologue phase

vicente-romero-oracle:
Before this fix only strict fields were readable in the prologue phase. The proposed fix should allow any instance fields of identity classes to be readable in the prologue phase. This implies changes in flow analysis as before we were only tracking final and strict fields. There is also some "cooperation" needed in the code to detect cases when reading a field is not allowed in the prologue phase. For example some methods in Resolve don't have all the needed information at the moment they are dealing with some ASTs and part of the processing needs to be done in Attr

TIA

---------
### Progress

### Issue
 * [JDK-8359370](https://bugs.openjdk.org/browse/JDK-8359370): [lworld] allow instance fields of identity classes to be readable in the prologue phase (**Bug** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/1490/head:pull/1490` \
`$ git checkout pull/1490`

`$ git checkout pull/1490` \
`$ git pull https://git.openjdk.org/valhalla.git pull/1490/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1490`

`$ git pr show -t 1490`

</details>

<a href="https://git.openjdk.org/valhalla/pull/1490.diff">https://git.openjdk.org/valhalla/pull/1490.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/1490#issuecomment-3029446442)
</details>


liach:
Just curious, why is field proxies so late in the stages? Is it because strict fields cannot be used by lowered structures, etc.?

vicente-romero-oracle:
> Just curious, why is field proxies so late in the stages? Is it because strict fields cannot be used by lowered structures, etc.?

sorry I missed your question, well what this phase is doing is kind of similar to what we do in Lower, actually I don't discard folding the code into lower in the future, this is probably one of the reasons it is so close to Lower. We certainly need it after Flow

vicente-romero-oracle:
fixed in: https://github.com/openjdk/valhalla/pull/1523

PR #1494, opened 2025-06-30
8359345: [lworld] C2 crashes with -XX:ForceNonTearable=*

TobiHartmann:
For test method `testFieldLoad1` the graph looks like this after parsing:
Screenshot from 2025-06-30 13-42-44

Now since [JDK-8354068](https://bugs.openjdk.org/browse/JDK-8354068) `MemNode::optimize_simple_memory_chain` rewires the memory input of strict field load `46 LoadL` to before the membar:
Screenshot from 2025-06-30 15-06-13

Later in `PhaseCFG::insert_anti_dependences`, a precedence edge from the membar to the load (now named `15 LoadL`) is added, because the type of the `CastI2N` is `TypeNarrowOop::BOTTOM` and thus `AliasType::is_rewritable` is true (it should be false). The resulting graph is unschedulable:
Screenshot from 2025-06-30 13-37-59

[JDK-8354981](https://bugs.openjdk.org/browse/JDK-8354981) added these `MemBarCPUOrders` around flat, atomic stores and then [JDK-8357474](https://bugs.openjdk.org/browse/JDK-8357474) also added them around loads. The same problem exists for `MemBarCPUOrders` emitted by stores. The memory input of following loads will be wired around them and an anti-dependency will be added.

The fix is to change `CastI2NNode` to a `TypeNode` that carries the exact type. This ways we can also avoid the `CastPPNode` after the decode.

Thanks,
Tobias

---------
### Progress

### Issue
 * [JDK-8359345](https://bugs.openjdk.org/browse/JDK-8359345): [lworld] C2 crashes with -XX:ForceNonTearable=* (**Bug** - P4)

### Contributors
 * Quan Anh Mai `<qamai@openjdk.org>`

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/1494/head:pull/1494` \
`$ git checkout pull/1494`

`$ git checkout pull/1494` \
`$ git pull https://git.openjdk.org/valhalla.git pull/1494/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1494`

`$ git pr show -t 1494`

</details>

<a href="https://git.openjdk.org/valhalla/pull/1494.diff">https://git.openjdk.org/valhalla/pull/1494.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/1494#issuecomment-3022626479)
</details>

### DIFF
--- a/src/hotspot/share/opto/castnode.hpp
+++ b/src/hotspot/share/opto/castnode.hpp
@@ -232,12 +232,14 @@ class CastX2PNode : public Node {
 };
 
 // Cast an integer to a narrow oop
-class CastI2NNode : public Node {
+class CastI2NNode : public TypeNode {
   public:
-  CastI2NNode(Node* ctrl, Node* n) : Node(ctrl, n) { }
+  CastI2NNode(Node* ctrl, Node* n, const Type* t) : TypeNode(t, 2) {
+    init_req(0, ctrl);
+    init_req(1, n);
+  }
   virtual int Opcode() const;
   virtual uint ideal_reg() const { return Op_RegN; }
-  virtual const Type* bottom_type() const { return TypeNarrowOop::BOTTOM; }
 };
 
 //------------------------------CastP2XNode-------------------------------------
--- a/src/hotspot/share/opto/inlinetypenode.cpp
+++ b/src/hotspot/share/opto/inlinetypenode.cpp
@@ -536,9 +536,8 @@ void InlineTypeNode::convert_from_payload(GraphKit* kit, BasicType bt, Node* pay
         if (trust_null_free_oop && field_null_free) {
           val_type = val_type->join_speculative(TypePtr::NOTNULL);
         }
-        value = gvn->transform(new CastI2NNode(kit->control(), value));
+        value = gvn->transform(new CastI2NNode(kit->control(), value, val_type->make_narrowoop()));
         value = gvn->transform(new DecodeNNode(value, val_type->make_narrowoop()));
-        value = gvn->transform(new CastPPNode(kit->control(), value, val_type, ConstraintCastNode::UnconditionalDependency));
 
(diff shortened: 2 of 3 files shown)


merykitty:
I thought final fields should not have anti-dependency? I assume this is where the precedence edge comes from.

merykitty:
Hmmm I think this is not correct. At `Before matching`, `46 LoadL` is a load from `MyValue2.l0`, which has `is_rewritable == false`, but afterward, at `After matching`, `15 loadL` loses this information, why?

TobiHartmann:
Thanks for your comments Quan-Anh, I missed them earlier. Let me double-check!

merykitty:
@TobiHartmann I think I know why. `CastI2N` returns a `TypeNarrowOop::BOTTOM`, which is then casted to the correct type at `DecodeNNode`. However, at matching, we skip through the `DecodeN`. As a result, `15 loadL` is a load from `TypeOopPtr::BOTTOM + 16` which the alias cannot be known to be immutable. I think this mismatch type can happen easily due to dropping `CastPP` at the end of optimization. I suggest clearing `Flag_needs_anti_dependence_check` when we rewire the node at `MemNode::optimize_simple_memory_chain`, or even clearing them at creation if their `adr_type` is non-rewritable.

TobiHartmann:
Good catch! Maybe we should just make the `CastI2N` a `TypeNode` and propagate the right type.

TobiHartmann:
Quick prototype looks promising, I'll give that a try.

TobiHartmann:
I'll pushed this and updated the description. Will run some more testing. Thanks again for looking into this, @merykitty.

TobiHartmann:
Testing is all green. I'm integrating this to get the CI clean. Thanks again!

PR #1497, opened 2025-07-02
JDK-8353182: [lworld] C2: Multiple IR test failures in compiler/gcbarriers/TestZGCBarrierElision.java after JDK-8351569

dafedafe:
This is a very similar issue to [JDK-8343420](https://bugs.openjdk.org/browse/JDK-8343420).

A few `TestZGCBarrierElision` tests fail after [JDK-8351569](https://bugs.openjdk.org/browse/JDK-8351569). These tests check that ZGC barrier elision optimization does not elide necessary barriers and partially do so using `VarHandle::getAndSet` and checking that `GetAndSetP` nodes with barriers are created.
As found out in [JDK-8343420](https://bugs.openjdk.org/browse/JDK-8343420), [JDK-8351569](https://bugs.openjdk.org/browse/JDK-8351569) refactored `Unsafe::getAndSet` (which is used by `VarHandle::getAndSet`) to take value-classes into account, which implies that checks must be introduced to see if the target and argument objects are value-classes. This is done with a new overloaded `getAndSet` method that is now called instead of the old one. This method is not intrinsified (only inlined) and doesn't result in a `GetAndSet` node. It instead relies on the `compareAndSet` method (which gets intrinsified) and results in a `CompareAndSwap` node.

Here as well, as the goal of the test is mainly to check for ZGC barrier presence, the sensible fix for now seems to be to modify the `TestZGCBarrierElision` tests to make them check for `CompareAndSwap` instead of `GetAndSet` nodes.

Tests: Tier 1-3+

### DIFF
--- a/test/hotspot/jtreg/compiler/gcbarriers/TestZGCBarrierElision.java
+++ b/test/hotspot/jtreg/compiler/gcbarriers/TestZGCBarrierElision.java
@@ -187,8 +187,7 @@ class TestZGCCorrectBarrierElision {
     }
 
     @Test
-    // TODO: 8353182
-    //@IR(counts = { IRNode.Z_GET_AND_SET_P_WITH_BARRIER_FLAG, Common.REMAINING, "1" }, phase = CompilePhase.FINAL_CODE)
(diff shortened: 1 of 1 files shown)


dafedafe:
Thanks for your review @TobiHartmann.

PR #1498, opened 2025-07-02
8361082: [lworld] RewriteBytecodesInlineTest fails with SIGSEGV

matias9927:
`RewriteBytecodesInlineTest` fails after # due to a new log message printing the pending exception name even though there may not be a pending exception:
`PENDING_EXCEPTION->klass()->name()->as_C_string()`

This patch refactors the loadable descriptor handling used in `SystemDictionary::load_shared_class` to better illustrate how the loadable descriptors property is handled and it corrects the log messages to be consistent with the messages used in the class file parser. Verified with tier 1-5 tests.

---------
### Progress

### Issue
 * [JDK-8361082](https://bugs.openjdk.org/browse/JDK-8361082): [lworld] RewriteBytecodesInlineTest fails with SIGSEGV (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/1498/head:pull/1498` \
`$ git checkout pull/1498`

`$ git checkout pull/1498` \
`$ git pull https://git.openjdk.org/valhalla.git pull/1498/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1498`

`$ git pr show -t 1498`

</details>

<a href="https://git.openjdk.org/valhalla/pull/1498.diff">https://git.openjdk.org/valhalla/pull/1498.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/1498#issuecomment-3028565663)
</details>

### DIFF
--- a/src/hotspot/share/classfile/classFileParser.cpp
+++ b/src/hotspot/share/classfile/classFileParser.cpp
@@ -6212,7 +6212,7 @@ void ClassFileParser::post_process_parsed_stream(const ClassFileStream* const st
         // Preloading classes for nullable fields that are listed in the LoadableDescriptors attribute
         // Those classes would be required later for the flattening of nullable inline type fields
         TempNewSymbol name = Signature::strip_envelope(sig);
-        if (name != _class_name && is_class_in_loadable_descriptors_attribute(sig)) { //here
+        if (name != _class_name && is_class_in_loadable_descriptors_attribute(sig)) {
           log_info(class, preload)("Preloading class %s during loading of class %s. \
                                    Cause: field type in LoadableDescriptors attribute",
                                    name->as_C_string(), _class_name->as_C_string());
--- a/src/hotspot/share/classfile/classFileParser.cpp
+++ b/src/hotspot/share/classfile/classFileParser.cpp
@@ -6169,13 +6169,21 @@ void ClassFileParser::post_process_parsed_stream(const ClassFileStream* const st
         // Pre-load classes of null-free fields that are candidate for flattening
         TempNewSymbol s = Signature::strip_envelope(sig);
         if (s == _class_name) {
-          THROW_MSG(vmSymbols::java_lang_ClassCircularityError(), err_msg("Class %s cannot have a null-free non-static field of its own type", _class_name->as_C_string()));
+          THROW_MSG(vmSymbols::java_lang_ClassCircularityError(),
+                    err_msg("Class %s cannot have a null-free non-static field of its own type", _class_name->as_C_string()));
         }
-        log_info(class, preload)("Preloading class %s during loading of class %s. Cause: a null-free non-static field is declared with this type", s->as_C_string(), _class_name->as_C_string());
-        Klass* klass = SystemDictionary::resolve_with_circularity_detection_or_fail(_class_name, s, Handle(THREAD, _loader_data->class_loader()), false, THREAD);
+        log_info(class, preload)("Preloading class %s during loading of class %s. \
+                                  Cause: a null-free non-static field is declared with this type",
+                                  s->as_C_string(), _class_name->as_C_string());
+        Klass* klass = SystemDictionary::resolve_with_circularity_detection_or_fail(_class_name, s,
+                                                                                    Handle(THREAD,
+                                                                                    _loader_data->class_loader()),
+                                                                                    false, THREAD);
         if (HAS_PENDING_EXCEPTION) {
-          log_warning(class, preload)("Preloading of class %s during loading of class %s (cause: null-free non-static field) failed: %s",
-                                      s->as_C_string(), _class_name->as_C_string(), PENDING_EXCEPTION->klass()->name()->as_C_string());
+          log_warning(class, preload)("Preloading of class %s during loading of class %s \
+                                      (cause: null-free non-static field) failed: %s",
+                                      s->as_C_string(), _class_name->as_C_string(),
+                                      PENDING_EXCEPTION->klass()->name()->as_C_string());
           return; // Exception is still pending
         }
         assert(klass != nullptr, "Sanity check");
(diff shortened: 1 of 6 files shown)


src/hotspot/share/classfile/systemDictionary.cpp
```diff
@@ -1075,6 +1075,92 @@ bool SystemDictionary::check_shared_class_super_types(InstanceKlass* ik, Handle
   return true;
 }
 
+// Pre-load class referred to in non-static null-free instance field. These fields trigger MANDATORY loading
```
coleenp, line 1078: Add period at the end of the sentence.

src/hotspot/share/classfile/systemDictionary.cpp
```diff
@@ -1075,6 +1075,92 @@ bool SystemDictionary::check_shared_class_super_types(InstanceKlass* ik, Handle
...
+                                name->as_C_string(), ik->name()->as_C_string(),
+                                k->name()->as_C_string());
+    return false;
+  }
+  log_info(class, preload)("Preloading of class %s during loading of shared class %s \
+                           (cause: null-free non-static field) succeeded",
+                           name->as_C_string(), ik->name()->as_C_string());
+
+  assert(real_k != nullptr, "Sanity check");
+  if (real_k->access_flags().is_identity_class()) {
```
coleenp, line 1109: It looks like you could refactor the check for identity and abstract in to InstanceKlass::check_inline_field(TRAPS) and share the code with ClassFileParser. (?)

src/hotspot/share/classfile/systemDictionary.cpp
```diff
@@ -1101,44 +1187,18 @@ InstanceKlass* SystemDictionary::load_shared_class(InstanceKlass* ik,
...
-        TempNewSymbol name = Signature::strip_envelope(sig);
-        log_info(class, preload)("Preloading class %s during loading of shared class %s. Cause: field type in LoadableDescriptors attribute", name->as_C_string(), ik->name()->as_C_string());
-        Klass* real_k = SystemDictionary::resolve_with_circularity_detection_or_fail(ik->name(), name,
-          class_loader, false, CHECK_NULL);
-        Klass* k = ik->get_inline_type_field_klass_or_null(fs.index());
-        if (real_k != k) {
-          // oops, the app has substituted a different version of k!
-          log_warning(class, preload)("Preloading of class %s during loading of shared class %s (cause: field type in LoadableDescriptors attribute) failed : %s",
-                                      name->as_C_string(), ik->name()->as_C_string(), PENDING_EXCEPTION->klass()->name()->as_C_string());
+        bool check = preload_from_null_free_field(ik, class_loader, sig, field_index, CHECK_NULL);
```
coleenp, line 1195: This is a bit confusing that it both has a check for return and CHECK_NULL.  Add a comment that some preloading might fail but not fatally for the shared archive if the class changed in the app.

coleenp (commented):
I have a few comments - maybe some of it can be refactored and shared?

src/hotspot/share/classfile/systemDictionary.cpp
matias9927: This check is being done in Klass so I would need to add it there. I think a better name might be `Klass::check_null_restricted_field(TRAPS)` since these are mandatory conditions for a field to be null restricted. The checks may actually be unnecessary since the `@NullRestricted` annotation can only be used on a concrete value field.

coleenp: It should be InstanceKlass not Klass though since Klass has no fields.  resolve_with_circularity_check does return an InstanceKlass so the return type should be changed to InstanceKlass.  The check_null_restricted_field name seems good though.

src/hotspot/share/oops/instanceKlass.cpp
```diff
@@ -3591,6 +3591,26 @@ bool InstanceKlass::find_inner_classes_attr(int* ooff, int* noff, TRAPS) const {
   return false;
 }
 
+void InstanceKlass::check_null_free_field(InstanceKlass* ik, TRAPS) {
```
coleenp, line 3594: Is ik not the same as 'this'?

src/hotspot/share/classfile/classFileParser.cpp
```diff
@@ -6169,13 +6169,21 @@ void ClassFileParser::post_process_parsed_stream(const ClassFileStream* const st
...
         if (HAS_PENDING_EXCEPTION) {
-          log_warning(class, preload)("Preloading of class %s during loading of class %s (cause: null-free non-static field) failed: %s",
-                                      s->as_C_string(), _class_name->as_C_string(), PENDING_EXCEPTION->klass()->name()->as_C_string());
+          log_warning(class, preload)("Preloading of class %s during loading of class %s \
+                                      (cause: null-free non-static field) failed: %s",
+                                      s->as_C_string(), _class_name->as_C_string(),
+                                      PENDING_EXCEPTION->klass()->name()->as_C_string());
           return; // Exception is still pending
         }
         assert(klass != nullptr, "Sanity check");
```
coleenp, line 6189: Can you call InstanceKlass::check_null_free_field from here too?

src/hotspot/share/oops/instanceKlass.cpp
matias9927: No, `ik` is the class being loaded in `load_shared_class` and `this` is the type of the null restricted field.

src/hotspot/share/classfile/classFileParser.cpp
matias9927: I could but I would need to change the argument of `check_null_free_field` to `Symbol*` since the InstanceKlass isn't available here.

src/hotspot/share/oops/instanceKlass.cpp
```diff
@@ -3591,6 +3591,26 @@ bool InstanceKlass::find_inner_classes_attr(int* ooff, int* noff, TRAPS) const {
   return false;
 }
 
+void InstanceKlass::check_null_free_field(Symbol* container_klass_name, TRAPS) {
```
fparain, line 3594: This method is very misleading, it does nothing related to fields, nor being null-free. And the fact that its name refers to fields but the receiver is the field's type and the container is passed in argument is confusing too, because most methods work the other way around: the receiver is the container, and the arguments tell which fields the query is about.

I'd suggest a static method like that:
can_be_annotated_with_NullRestricted(InstanceKlass* type, Symbol* container_klass_name, TRAPS)

matias9927:
Thanks for the reviews @coleenp and @fparain!
/integrate

PR #1502, opened 2025-07-04
8361166: [lworld] TestG1BarrierGeneration still fails after JDK-8343420

dafedafe:
The `TestG1BarrierGeneration::testArrayStore` and `TestG1BarrierGeneration::testArrayStoreTwice` tests are failing with `-XX:-TieredCompilation` because the compilation is inserting a `CastPP` node for the input of the `StoreP` node (due to missing profiling info) and this is restricting the type of the barrier to `notnull`. The test regex checks for exactly `pre post` (nothing after that) and thus fails.
This change fixes the test by increasing the warmup iterations for the failing tests.

Tests: Tier 1-3+

---------
### Progress

### Issue
 * [JDK-8361166](https://bugs.openjdk.org/browse/JDK-8361166): [lworld] TestG1BarrierGeneration still fails after JDK-8343420 (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/1502/head:pull/1502` \
`$ git checkout pull/1502`

`$ git checkout pull/1502` \
`$ git pull https://git.openjdk.org/valhalla.git pull/1502/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1502`

`$ git pr show -t 1502`

</details>

<a href="https://git.openjdk.org/valhalla/pull/1502.diff">https://git.openjdk.org/valhalla/pull/1502.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/1502#issuecomment-3043688141)
</details>

### DIFF
--- a/test/hotspot/jtreg/compiler/gcbarriers/TestG1BarrierGeneration.java
+++ b/test/hotspot/jtreg/compiler/gcbarriers/TestG1BarrierGeneration.java
@@ -416,16 +416,10 @@ public class TestG1BarrierGeneration {
     }
 
     @Test
-    @IR(applyIfAnd = {"UseCompressedOops", "false", "TieredCompilation", "false"},
-        counts = {IRNode.G1_STORE_P_WITH_BARRIER_FLAG, PRE_AND_POST_NOT_NULL, "1"},
-        phase = CompilePhase.FINAL_CODE)
-    @IR(applyIfAnd = {"UseCompressedOops", "true", "TieredCompilation", "false"},
-        counts = {IRNode.G1_ENCODE_P_AND_STORE_N_WITH_BARRIER_FLAG, PRE_AND_POST_NOT_NULL, "1"},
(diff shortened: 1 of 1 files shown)


chhagedorn:
Thanks for working on this! Have you found out why the `CastPP` is inserted without tiered compilation but not with tiered compilation? Are we missing some profiling? In the latter case, you could also try to increase the warm-up iterations and see if that works instead.

dafedafe:
> Have you found out why the `CastPP` is inserted without tiered compilation but not with tiered compilation? Are we missing some profiling?

Yes, this seems to be the reason.

> In the latter case, you could also try to increase the warm-up iterations and see if that works instead.

Of course! Much nicer solution. Thanks @chhagedorn.

dafedafe:
Thanks for your review @chhagedorn.

PR #1503, opened 2025-07-04
8361250: [lworld] use more IRNode regexes in valhalla IR tests

marc-chevalier:
InlineTypeIRNode is Valhalla-specific and somewhat... antique. I've run into an issue where the old-timey regex for allocation didn't do the job. So, rather than reproducing the fix of [JDK-8314999](https://bugs.openjdk.org/browse/JDK-8314999) again, let's simply use `IRNode`'s regex for Valhalla as well.

The general idea: let's not define node regex specifically for Valhalla, but only for values. I've also introduced regexes for combinations we do all the time: for instance rather than having `ALLOC_OF, ANY_KLASS`, I made a `ALLOC_OF_ANY_KLASS`.

So what now:
- `InlineTypeRegexes` => only value regexes.
- `IRNode` => adapted to allow to define specialisation for `ALLOC_OF`, `ALLOC_ARRAY_OF`, `LOAD_OF_CLASS`, `STORE_OF_CLASS`, `STATIC_CALL_OF_METHOD` for specific classes/functions.
- `InlineTypeIRNode` => define specialisations using the regexes defined in `InlineTypeRegexes`, nothing more!
- `compiler/valhalla/inlinetypes/Test*` => use `IRNode` and `InlineTypeIRNode`.


marc-chevalier:
Address the various comments, found that some tests was passing by accident (failOn on a regex that would never match), improve load/store regexps, fix the tests.

TobiHartmann (approved):
Good catch, I assume you'll upstream the Valhalla independent fix with [JDK-8361492](https://bugs.openjdk.org/browse/JDK-8361492)?

marc-chevalier:
That is the plan, indeed.

marc-chevalier:
/integrate

Thanks @TobiHartmann!

PR #1504, opened 2025-07-07
8361496: [lworld] Treating interfaces as isValue() leads to failed assert

liach:
Reviewed a few sites of isValue() usages that cause immediate trouble with bootstrap and basic tests. Factor out `ValueClass::isValueObjectCompatible` and `isValueObjectInstance` to represent the two states of value classes: field types that can hold value objects, and field types that definitely hold value objects.

I think we still need a more throughout review for all `isValue` usages, such as those in MethodHandle or VarHandle. Preliminary testing hasn't revealed anything yet.

Testing: tier 1-3, seems to have no new failures

### DIFF
--- a/src/java.base/share/classes/java/lang/Class.java
+++ b/src/java.base/share/classes/java/lang/Class.java
@@ -333,18 +333,20 @@ public final class Class<T> implements java.io.Serializable,
                 if (isAnnotation()) {
                     sb.append('@');
                 }
-                if (isValue()) {
-                    sb.append("value ");
-                }
                 if (isInterface()) { // Note: all annotation interfaces are interfaces
                     sb.append("interface");
                 } else {
                     if (isEnum())
                         sb.append("enum");
-                    else if (isRecord())
-                        sb.append("record");
-                    else
-                        sb.append("class");
+                    else {
+                        if (isValue()) {
+                            sb.append("value ");
+                        }
+                        if (isRecord())
+                            sb.append("record");
+                        else
+                            sb.append("class");
+                    }
                 }
                 sb.append(' ');
                 sb.append(getName());
--- a/src/java.base/share/classes/java/lang/Class.java
+++ b/src/java.base/share/classes/java/lang/Class.java
@@ -638,9 +638,7 @@ public final class Class<T> implements java.io.Serializable,
         if (!PreviewFeatures.isEnabled()) {
             return false;
         }
-         if (isPrimitive() || isArray() || isInterface())
-             return false;
-        return ((getModifiers() & Modifier.IDENTITY) == 0);
+        return !isIdentity();
     }
 
     /**
--- a/src/java.base/share/classes/jdk/internal/value/ValueClass.java
+++ b/src/java.base/share/classes/jdk/internal/value/ValueClass.java
@@ -29,14 +29,26 @@ import jdk.internal.access.JavaLangReflectAccess;
 import jdk.internal.access.SharedSecrets;
 import jdk.internal.vm.annotation.IntrinsicCandidate;
 
-import java.lang.reflect.Array;
 import java.lang.reflect.Field;
+import java.lang.reflect.Modifier;
 
 /**
  * Utilities to access package private methods of java.lang.Class and related reflection classes.
  */
-public class ValueClass {
+public final class ValueClass {
     private static final JavaLangReflectAccess JLRA = SharedSecrets.getJavaLangReflectAccess();
+
+    /// {@return whether this field type may store value objects}
+    /// This excludes primitives and includes Object.
+    public static boolean isValueObjectCompatible(Class<?> fieldType) {
+        return !fieldType.isPrimitive() && (!fieldType.isIdentity() || fieldType == Object.class);
+    }
+
+    /// {@return whether an object of this exact class is a value object}
+    /// This excludes abstract value classes and primitives.
+    public static boolean isValueObjectInstance(Class<?> clazz) {
+        return clazz.isValue() && !Modifier.isAbstract(clazz.getModifiers());
+    }
 
     /**
      * {@return {@code true} if the field is NullRestricted}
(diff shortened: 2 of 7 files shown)


src/java.base/share/classes/java/lang/Class.java
```diff
@@ -638,9 +640,7 @@ public boolean isValue() {
         if (!PreviewFeatures.isEnabled()) {
             return false;
         }
-         if (isPrimitive() || isArray() || isInterface())
-             return false;
-        return ((getModifiers() & Modifier.IDENTITY) == 0);
+        return !isIdentity();
```
RogerRiggs, line 643: Copy the lines from PR#1499.
Nothing is value unless preview is enabled.

```suggestion
        return PreviewFeatures.isEnabled() ? !isIdentity() : false;
        
```

liach: This preview switch is retained right above the first deletion in this small piece of diff.

liach:
I have updated the test to test both preview and non-preview behaviors. May I get another review?

test/jdk/valhalla/valuetypes/ValueClassTest.java
```diff
@@ -0,0 +1,82 @@
...
+        assertFalse(isValueObjectCompatible(String.class), "final identity class");
+        assertEquals(PreviewFeatures.isEnabled(), isValueObjectCompatible(Comparable.class), "interface");
+        assertFalse(isValueObjectCompatible(int[].class), "array class");
+        assertFalse(isValueObjectCompatible(Object[].class), "array class");
+        assertFalse(isValueObjectCompatible(Number[].class), "array class");
+        assertFalse(isValueObjectCompatible(Integer[].class), "array class");
+        assertFalse(isValueObjectCompatible(ClassValue[].class), "array class");
+        assertFalse(isValueObjectCompatible(ArrayList[].class), "array class");
+        assertFalse(isValueObjectCompatible(String[].class), "array class");
+        assertFalse(isValueObjectCompatible(Comparable[].class), "array class");
```
RogerRiggs, line 60: If these ever fail, the error messages are not going to be very information  like`expected tru, actual = false Object`;
no mention of preview mode being tested or the method under test.

src/java.base/share/classes/jdk/internal/value/ValueClass.java
```diff
@@ -27,17 +27,30 @@
...
 
+    /// {@return whether this field type may store value objects}
+    /// This excludes primitives and includes Object.
+    public static boolean isValueObjectCompatible(Class<?> fieldType) {
+        return PreviewFeatures.isEnabled() && !fieldType.isPrimitive() && (!fieldType.isIdentity() || fieldType == Object.class);
+    }
+
+    /// {@return whether an object of this exact class is a value object}
+    /// This excludes abstract value classes and primitives.
+    public static boolean isValueObjectInstance(Class<?> clazz) {
```
RogerRiggs, line 50: The method name is a bit misleading, it seems to ask if the parameter is an instance, but it is a class.
Perhaps `isConcreteValueObject()` or `isConcreteValueClass()`.

src/java.base/share/classes/jdk/internal/value/ValueClass.java
```diff
@@ -27,17 +27,30 @@
...
  * Utilities to access package private methods of java.lang.Class and related reflection classes.
  */
-public class ValueClass {
+public final class ValueClass {
     private static final JavaLangReflectAccess JLRA = SharedSecrets.getJavaLangReflectAccess();
 
+    /// {@return whether this field type may store value objects}
+    /// This excludes primitives and includes Object.
+    public static boolean isValueObjectCompatible(Class<?> fieldType) {
+        return PreviewFeatures.isEnabled() && !fieldType.isPrimitive() && (!fieldType.isIdentity() || fieldType == Object.class);
```
RogerRiggs, line 45: The `fieldType == Object.class` would always be true regardless of enable-preview.
That does not seem to be accurate and might be mis-interpreted by a caller.

test/jdk/valhalla/valuetypes/ValueClassTest.java
liach: Junit composes the exception message with the expected, actual values and the string provided, also there is the line number.

liach: I think the exception message is acceptable: I tweaked `isValueObjectCompatible` to erroneousely return `true` for `Integer[].class`, and this is the message:

```
STARTED    ValueClassTest::testIsValueObjectCompatible 'testIsValueObjectCompatible()'
org.opentest4j.AssertionFailedError: array class ==> expected: <false> but was: <true>
        at org.junit.jupiter.api.AssertionFailureBuilder.build(AssertionFailureBuilder.java:151)
        at org.junit.jupiter.api.AssertionFailureBuilder.buildAndThrow(AssertionFailureBuilder.java:132)
        at org.junit.jupiter.api.AssertFalse.failNotFalse(AssertFalse.java:63)
        at org.junit.jupiter.api.AssertFalse.assertFalse(AssertFalse.java:36)
        at org.junit.jupiter.api.Assertions.assertFalse(Assertions.java:239)
        at ValueClassTest.testIsValueObjectCompatible(ValueClassTest.java:56)
        at java.base/java.lang.reflect.Method.invoke(Method.java:565)
        at java.base/java.util.ArrayList.forEach(ArrayList.java:1604)
        at java.base/java.util.ArrayList.forEach(ArrayList.java:1604)
FAILED     ValueClassTest::testIsValueObjectCompatible 'testIsValueObjectCompatible()' [16ms]
```

RogerRiggs (approved):
Looks good thanks for the readability improvements.

RogerRiggs:
Looks good; integrate?

liach:
Thanks for the reminder and the review.

/integrate

PR #1505, opened 2025-07-08
8360653: [lworld] LayoutIterationTest.java fails with -XX:ForceNonTearable=*

liach:
Mark this test as flagless so we will skip it with any layout perturbation.


liach:
Guess people don't bother to review this. I will just integrate.

/integrate

PR #1507, opened 2025-07-10
Remove ensure_objArray and copying code and write a test

coleenp:
This is the diff for removing copy_flatArray_to_objArray and ensure_objArray.  This is relative to lworld but should apply cleanly to your PR Fred.  I tested it against the tier1 tests in the lworld repo (in progress).

### DIFF
--- a/src/hotspot/share/memory/oopFactory.cpp
+++ b/src/hotspot/share/memory/oopFactory.cpp
@@ -147,23 +147,6 @@ flatArrayOop oopFactory::new_flatArray(Klass* k, int length, LayoutKind lk, TRAP
   return oop;
 }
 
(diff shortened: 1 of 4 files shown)


liach:
Note these two paths are only used by `Constructor::newInstance` and `Method::invoke` if system property `-Djdk.reflect.useNativeAccessorOnly=true` is set.

coleenp:
> Note these two paths are only used by Constructor::newInstance and Method::invoke if system property -Djdk.reflect.useNativeAccessorOnly=true is set.

Yes, my test passes at property.

liach (approved):
Looks great

coleenp:
Thanks for reviewing, Fred and Chen.
/integrate

PR #1511, opened 2025-07-17
8357785: [lworld] TestResolvedJavaType fails due to unexpected getInstanceFields order

marc-chevalier:
It's all a matter of order. In short, mainline doesn't sort objects fields by offset, but just take them as they come from the stream. The order can be arbitrary, but in some places, it matters it's the same everywhere. Valhalla is sorting too much for mainline's taste, a merge caused clashes.

But maybe, a bit of archeology:
- Mainline [JDK-8350892: [JVMCI] Align ResolvedJavaType.getInstanceFields with Class.getDeclaredFields](https://bugs.openjdk.org/browse/JDK-8350892): it removed some sorting of fields, keeping it consistent between `reassign_fields_by_klass` and `ciInstanceKlass::compute_nonstatic_fields()`: they both had fields offset-sorted, now they are both stream-sorted. Most of HotSpot agrees with this new order. This change also add a new test checking the order is consistent: `compiler/jvmci/jdk.vm.ci.runtime.test/src/jdk/vm/ci/runtime/test/TestResolvedJavaType.java`.

- Valhalla: [#1429](https://github.com/openjdk/valhalla/pull/1429) brings [JDK-8350892](https://bugs.openjdk.org/browse/JDK-8350892) in Valhalla. This merge is a bit mysterious to me,  how some sorting reappeared in Valhalla-untouched code. For instance, [JDK-8350892](https://bugs.openjdk.org/browse/JDK-8350892) removed the function `
static int sort_field_by_offset(ciField** a, ciField** b)` and
  ```cpp
  // Now sort them by offset, 
  // (In principle, they could mix with superclass fields.)
  fields->sort(sort_field_by_offset);
  ```
  from `int ciInstanceKlass::compute_nonstatic_fields()` (in `ciInstanceKlass.cpp`), but the merge only removes the comment, but leave the `->sort()` and `sort_field_by_offset`. My guess is that it was a tentative fix for some failing tests after merge. Nevertheless `compiler/jvmci/jdk.vm.ci.code.test/src/jdk/vm/ci/code/test/VirtualObjectDebugInfoTest.java` is ProblemListed.

- Valhalla: After this merge, Valhalla continues to assume some ordering, as one can see in [JDK-8355076: [lworld] Incorrect declared_nonstatic_fields computation](https://bugs.openjdk.org/browse/JDK-8355076) that (among other things), makes a new occurrence of `->sort(sort_field_by_offset)`, to be consistent with `InlineKlass::collect_fields`. But now, everything is different! The new test introduced by [JDK-8350892](https://bugs.openjdk.org/browse/JDK-8350892) connects more things together, and enforce the same order in `ciInstanceKlass::compute_nonstatic_fields` and in `Class.getDeclaredFields` (reflexion related, I think. I don't know that so well). Moreover `ciInstanceKlass::compute_nonstatic_fields` has to agree with deoptimization code. In Valhalla, the calling convention has to agree with the layout wrt field ordering. Basically, it makes a big chain of things that needs to agree on the order of fields.

- Valhalla: [JDK-8354366: [lworld] VirtualObjectDebugInfoTest fails after merging jdk-25+16](https://bugs.openjdk.org/browse/JDK-8354366) partially reverts [JDK-8350892: [JVMCI] Align ResolvedJavaType.getInstanceFields with Class.getDeclaredFields](https://bugs.openjdk.org/browse/JDK-8350892) by reintroducing sorting. This allowed to un-ProblemList `compiler/jvmci/jdk.vm.ci.code.test/src/jdk/vm/ci/code/test/VirtualObjectDebugInfoTest.java`, but made `compiler/jvmci/jdk.vm.ci.runtime.test/src/jdk/vm/ci/runtime/test/TestResolvedJavaType.java` fail because now, and that's where I start.

Overall, it's a clash in the order of fields. First question is naturally: does the order really has to agree, or do we put some interfaces, pipes and curtains to adapt between potentially different order? I think we should keep a consistent order, as much as possible. It makes everything simpler, possibly more efficient as we don't have to search if we want to work on two lists... Efforts were made in mainline to keep the order consistent, let's not break that.

Second question is which order. Here, same, I'll invoke mainline that decided to go with stream order. It has also the benefit not to require sorting: get the field stream, ready to use. Lighter, easier, more efficient...

Let's define more precisely the order in which the fields should show up. Given a class `Derived`, that extends the class `Base`.
1. superclass first: for any transitive field `b` of `Base`, and any proper field `d` of `Derived`, `b` should appear before `d`,
2. consistent ordering: the list of (transitive) fields of `Base` must be a subsequence of the list of (transitive) fields of `Derived`. In other words, the order of the fields of `Base` in the list of fields of `Base` is the same as the order of the fields of `Base` in the list of fields of `Derived`.
3. proper field ordering: the proper fields of `Derived`, must appear in the order provided by `JavaFieldStream` (which happens to be the declaration order, but is not guaranteed).

This is enough to define the list of declared fields. It implies in particular that the fields of the super class form a prefix of the whole list of declared fields. It makes possible to isolate the proper fields by removing this prefix (and it's done somewhere). It also is cheap to compute as we can just push fields in the order the stream gives, starting with a recursive call to traverse the supertype's fields.

If we are interested in the list of actual fields, as present in the layout, we simply insert the fields of each flatten field instead of them. The null marker of each field, if it exists, appears immediately after.

Which changes need to be made? Mostly removing sorting
- remove sorting in `ciInstanceKlass::compute_nonstatic_fields_impl`.
- remove member variable `float _sort_offset` from `SigEntry` but add `bool _null_marker` for printing in `nmethod::print_nmethod_labels`.
- remove sorting from `InlineKlass::collect_fields`, but also, change the iterator: use the new `TopDownHierarchicalFieldStreamBase` that will travese the super classes first.
- remove sorting from `get_reassigned_fields`, seems to be a merge accident? The code around comes from [JDK-8350892](https://bugs.openjdk.org/browse/JDK-8350892) without the sort, but [#1429](https://github.com/openjdk/valhalla/pull/1429) introduced sorting.
- revert [JDK-8354366: [lworld] VirtualObjectDebugInfoTest fails after merging jdk-25+16](https://bugs.openjdk.org/browse/JDK-8354366), bringing back the changes of [JDK-8350892: [JVMCI] Align ResolvedJavaType.getInstanceFields with Class.getDeclaredFields](https://bugs.openjdk.org/browse/JDK-8350892).

Bonus:
- Added a printer for `VirtualObjectDebugInfoTest$TestClass` because the test failure message being "this `TestClass` is not equal to this other `TestClass`" is not so helpful.
- Added a flag to print fields collected in `InlineKlass::initialize_calling_convention`. It seems it was deemed useful but missing, and not only by me right now! The flag is called `PrintInlineKlassFields` and takes a comma-separated list of class name patterns to print. Feel free to suggest a better name.
- I've tried to print fields in `ciInstanceKlass::compute_nonstatic_fields` to compare. It works, but this function is called way too many times for each class (not sure why, didn't look into that), so not great to leave a flag for that.

Wow, what a wall of text. But for once, it comes with non-trivial changes.

Thanks,
Marc

---------
### Progress

### Issues
 * [JDK-8357785](https://bugs.openjdk.org/browse/JDK-8357785): [lworld] TestResolvedJavaType fails due to unexpected getInstanceFields order (**Bug** - P4)
 * [JDK-8357186](https://bugs.openjdk.org/browse/JDK-8357186): [lworld] Clean up InlineKlass::collect_fields (**Enhancement** - P4)

### Reviewers
 * [Quan Anh Mai](https://openjdk.org/census#qamai) (@merykitty - Committer) ⚠️ Review applies to [7cb1c3fc](https://git.openjdk.org/valhalla/pull/1511/files/7cb1c3fcb396d8c3ab418d46c721de4caf1f1ffc)
 * [Tobias Hartmann](https://openjdk.org/census#thartmann) (@TobiHartmann - Committer) ⚠️ Review applies to [7cb1c3fc](https://git.openjdk.org/valhalla/pull/1511/files/7cb1c3fcb396d8c3ab418d46c721de4caf1f1ffc)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/1511/head:pull/1511` \
`$ git checkout pull/1511`

`$ git checkout pull/1511` \
`$ git pull https://git.openjdk.org/valhalla.git pull/1511/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1511`

`$ git pr show -t 1511`

</details>

<a href="https://git.openjdk.org/valhalla/pull/1511.diff">https://git.openjdk.org/valhalla/pull/1511.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/1511#issuecomment-3083844125)
</details>

### DIFF
--- a/src/hotspot/share/ci/ciInstanceKlass.cpp
+++ b/src/hotspot/share/ci/ciInstanceKlass.cpp
@@ -471,13 +471,6 @@ ciField* ciInstanceKlass::get_field_by_name(ciSymbol* name, ciSymbol* signature,
   return field;
 }
 
-#if 0
-static int sort_field_by_offset(ciField** a, ciField** b) {
-  return (*a)->offset_in_bytes() - (*b)->offset_in_bytes();
-  // (no worries about 32-bit overflow...)
-}
-#endif
-
 const GrowableArray<ciField*> empty_field_array(0, MemTag::mtCompiler);
 
 void ciInstanceKlass::compute_nonstatic_fields() {
--- a/src/hotspot/share/ci/ciInstanceKlass.cpp
+++ b/src/hotspot/share/ci/ciInstanceKlass.cpp
@@ -471,10 +471,12 @@ ciField* ciInstanceKlass::get_field_by_name(ciSymbol* name, ciSymbol* signature,
   return field;
 }
 
+#if 0
 static int sort_field_by_offset(ciField** a, ciField** b) {
   return (*a)->offset_in_bytes() - (*b)->offset_in_bytes();
   // (no worries about 32-bit overflow...)
 }
+#endif
 
 const GrowableArray<ciField*> empty_field_array(0, MemTag::mtCompiler);
 
@@ -559,7 +561,6 @@ void ciInstanceKlass::compute_nonstatic_fields_impl(const GrowableArray<ciField*
 
     if (fd.is_flat()) {
       // Flat fields are embedded
-      int field_offset = fd.offset();
       Klass* k = get_instanceKlass()->get_inline_type_field_klass(fd.index());
       ciInlineKlass* vk = CURRENT_ENV->get_klass(k)->as_inline_klass();
       // Iterate over fields of the flat inline type and copy them to 'this'
@@ -581,7 +582,6 @@ void ciInstanceKlass::compute_nonstatic_fields_impl(const GrowableArray<ciField*
   if (tmp_declared_fields != nullptr) {
     assert(tmp_declared_fields->length() == declared_field_num, "sanity check failed for class: %s, number of declared fields: %d, expected: %d",
            name()->as_utf8(), tmp_declared_fields->length(), declared_field_num);
-    tmp_declared_fields->sort(sort_field_by_offset);
     _declared_nonstatic_fields = tmp_declared_fields;
   } else {
     _declared_nonstatic_fields = super_declared_fields;
@@ -590,7 +590,6 @@ void ciInstanceKlass::compute_nonstatic_fields_impl(const GrowableArray<ciField*
   if (tmp_fields != nullptr) {
     assert(tmp_fields->length() == field_num, "sanity check failed for class: %s, number of fields: %d, expected: %d",
            name()->as_utf8(), tmp_fields->length(), field_num);
-    tmp_fields->sort(sort_field_by_offset);
     _nonstatic_fields = tmp_fields;
   } else {
     _nonstatic_fields = super_fields;
--- a/src/hotspot/share/ci/ciInstanceKlass.cpp
+++ b/src/hotspot/share/ci/ciInstanceKlass.cpp
@@ -427,8 +427,6 @@ ciField* ciInstanceKlass::get_non_flat_field_by_offset(int field_offset) {
     int field_off = field->offset_in_bytes();
     if (field_off == field_offset) {
       return field;
-    } else if (field_off > field_offset) {
-      break;
     }
   }
   return nullptr;
--- a/src/hotspot/share/code/nmethod.cpp
+++ b/src/hotspot/share/code/nmethod.cpp
@@ -3837,7 +3837,7 @@ void nmethod::print_nmethod_labels(outputStream* stream, address block_begin, bo
     } else {
       bool did_name = false;
       if (is_reference_type(t)) {
-        Symbol* name = (*sig)._symbol;
+        Symbol* name = (*sig)._name;
         name->print_value_on(stream);
         did_name = true;
       }
--- a/src/hotspot/share/code/nmethod.cpp
+++ b/src/hotspot/share/code/nmethod.cpp
@@ -3843,7 +3843,6 @@ void nmethod::print_nmethod_labels(outputStream* stream, address block_begin, bo
       }
       if (!did_name)
         stream->print("%s", type2name(t));
-      // If the entry has a non-default sort_offset, it must be a null marker
       if ((*sig)._null_marker) {
         stream->print(" (null marker)");
       }
--- a/src/hotspot/share/code/nmethod.cpp
+++ b/src/hotspot/share/code/nmethod.cpp
@@ -3844,7 +3844,7 @@ void nmethod::print_nmethod_labels(outputStream* stream, address block_begin, bo
       if (!did_name)
         stream->print("%s", type2name(t));
       // If the entry has a non-default sort_offset, it must be a null marker
-      if ((*sig)._sort_offset != (*sig)._offset) {
+      if ((*sig)._null_marker) {
         stream->print(" (null marker)");
       }
     }
--- a/src/hotspot/share/oops/fieldStreams.hpp
+++ b/src/hotspot/share/oops/fieldStreams.hpp
@@ -193,9 +193,74 @@ class AllFieldStream : public FieldStreamBase {
   AllFieldStream(const InstanceKlass* k):      FieldStreamBase(k->fieldinfo_stream(), k->constants()) {}
 };
 
+template<typename FieldStreamType>
+class HierarchicalFieldStreamBase : public StackObj {
+protected:
+  virtual FieldStreamType& current_stream () = 0;
+  virtual const FieldStreamType& current_stream () const = 0;
+
+public:
+  // bridge functions from FieldStreamBase
+  int index() const {
+    return current_stream().index();
+  }
+
+  AccessFlags access_flags() const {
+    return current_stream().access_flags();
+  }
+
+  FieldInfo::FieldFlags field_flags() const {
+    return current_stream().field_flags();
+  }
+
+  Symbol* name() const {
+    return current_stream().name();
+  }
+
+  Symbol* signature() const {
+    return current_stream().signature();
+  }
+
+  Symbol* generic_signature() const {
+    return current_stream().generic_signature();
+  }
+
+  int offset() const {
+    return current_stream().offset();
+  }
+
+  bool is_contended() const {
+    return current_stream().is_contended();
+  }
+
+  int contended_group() const {
+    return current_stream().contended_group();
+  }
+
+  FieldInfo to_FieldInfo() {
+    return current_stream().to_FieldInfo();
+  }
+
+  fieldDescriptor& field_descriptor() const {
+    return current_stream().field_descriptor();
+  }
+
+  bool is_flat() const {
+    return current_stream().is_flat();
+  }
+
+  bool is_null_free_inline_type() {
+    return current_stream().is_null_free_inline_type();
+  }
+
+  int null_marker_offset() {
+    return current_stream().null_marker_offset();
+  }
+};
+
 // Iterate over fields including the ones declared in supertypes
 template<typename FieldStreamType>
-class HierarchicalFieldStream : public StackObj  {
+class HierarchicalFieldStream : public HierarchicalFieldStreamBase<FieldStreamType>  {
  private:
   const Array<InstanceKlass*>* _interfaces;
   InstanceKlass* _next_klass; // null indicates no more type to visit
@@ -233,6 +298,10 @@ class HierarchicalFieldStream : public StackObj  {
     }
   }
 
+ protected:
+  FieldStreamType& current_stream() override { return _current_stream; }
+  const FieldStreamType& current_stream() const override { return _current_stream; }
+
  public:
   HierarchicalFieldStream(InstanceKlass* klass) :
     _interfaces(klass->transitive_interfaces()),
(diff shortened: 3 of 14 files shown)


src/hotspot/share/ci/ciInstanceKlass.cpp
```diff
@@ -471,11 +471,6 @@ ciField* ciInstanceKlass::get_field_by_name(ciSymbol* name, ciSymbol* signature,
   return field;
 }
 
-static int sort_field_by_offset(ciField** a, ciField** b) {
```
merykitty, line 474: Could you add a comment at `ciInstanceKlass._declared_nonstatic_fields` and `_nonstatic_fields` to specify exactly the order of the fields there? It is a little bit hard to grasp at the first sight, especially with flattened fields.

src/hotspot/share/oops/fieldStreams.hpp
```diff
@@ -193,9 +193,74 @@ class AllFieldStream : public FieldStreamBase {
   AllFieldStream(const InstanceKlass* k):      FieldStreamBase(k->fieldinfo_stream(), k->constants()) {}
 };
 
+template<typename FieldStreamType>
```
merykitty, line 196: It would be good to say what this class does and what the template parameter is for.

src/hotspot/share/oops/fieldStreams.hpp
```diff
@@ -248,63 +317,54 @@ class HierarchicalFieldStream : public StackObj  {
...
-    return _current_stream.offset();
-  }
-
-  bool is_contended() const {
-    return _current_stream.is_contended();
-  }
+/* Iterates on the fields of a class and its super-class top-down (java.lang.Object first)
+ * Doesn't traverse interfaces for now (let's decide how when/if the needs appear).
+ */
+template<typename FieldStreamType>
```
merykitty, line 325: By "doesn't traverse interfaces for now" I assume this traverses both static and non-static fields. Tbh I think traversing static fields seems questionable, or maybe I miss some cases, maybe this stream can discard those, too?

src/hotspot/share/oops/inlineKlass.cpp
```diff
@@ -423,19 +422,14 @@ int InlineKlass::collect_fields(GrowableArray<SigEntry>* sig, float& max_offset,
...
   int offset = base_off + size_helper()*HeapWordSize - (base_off > 0 ? payload_offset() : 0);
   // Null markers are no real fields, add them manually at the end (C2 relies on this) of the flat fields
   if (null_marker_offset != -1) {
-    max_offset += 0.1f; // We add the markers "in-between" because they are no real fields
-    SigEntry::add_entry(sig, T_BOOLEAN, name(), null_marker_offset, max_offset);
+    SigEntry::add_null_marker(sig, name(), null_marker_offset);
     count++;
   }
   SigEntry::add_entry(sig, T_VOID, name(), offset);
-  if (base_off == 0) {
```
merykitty, line 436: Nice, I have a pending issue of making this sort less a minefield but removing it completely is absolutely better. The comment of this function says it sorts the flattened fields in the increasing offset order, please change it, too.

src/hotspot/share/runtime/signature.hpp
```diff
@@ -579,15 +579,16 @@ class SigEntry {
...
   SigEntry()
-    : _bt(T_ILLEGAL), _offset(-1), _sort_offset(-1), _symbol(nullptr) {}
+    : _bt(T_ILLEGAL), _offset(-1), _symbol(nullptr) {}
 
-  SigEntry(BasicType bt, int offset = -1, float sort_offset = -1, Symbol* symbol = nullptr)
-    : _bt(bt), _offset(offset), _sort_offset(sort_offset), _symbol(symbol) {}
+  SigEntry(BasicType bt, int offset, Symbol* symbol, bool null_marker)
+    : _bt(bt), _offset(offset), _symbol(symbol), _null_marker(null_marker) {}
 
+#if 0
```
merykitty, line 591: Why do you wrap it with `#if 0` instead of removing this function?

test/hotspot/jtreg/ProblemList.txt
```diff
@@ -132,7 +132,7 @@ runtime/AccModule/ConstModule.java 8294051 generic-all
 runtime/valhalla/inlinetypes/CircularityTest.java 8349037 generic-all
 runtime/valhalla/inlinetypes/verifier/StrictInstanceFieldsTest.java CODETOOLS-7904031 generic-all
 runtime/valhalla/inlinetypes/verifier/StrictStaticFieldsTest.java CODETOOLS-7904031 generic-all
-compiler/jvmci/jdk.vm.ci.runtime.test/src/jdk/vm/ci/runtime/test/TestResolvedJavaType.java 8357785 generic-all
+runtime/cds/appcds/RewriteBytecodesInlineTest.java 8361082 generic-all
```
merykitty, line 135: This issue is fixed with #1498 , so this seems like a merge conflict, right?

src/hotspot/share/runtime/signature.hpp
marc-chevalier: It's a leftover that survived the clean up, it should have been removed. I was trying to remove some sorting by removing the order function, and see if it's used somewhere legitimate, but this way it was still easily accessible. Anyway, it should have been, and will be, removed.

test/hotspot/jtreg/ProblemList.txt
marc-chevalier: Exactly.

marc-chevalier:
Thanks for looking at it. I think I've addressed the comments. It looks better already: I missed a lot of cleanup (leaving outdated comments and `#if 0` behind, especially), because I was very unsure of what I was doing until it worked.

src/hotspot/share/oops/fieldStreams.hpp
```diff
@@ -193,9 +193,88 @@ class AllFieldStream : public FieldStreamBase {
...
+    return current_stream().null_marker_offset();
+  }
+};
+
+/* Iterate over fields including the ones declared in supertypes.
+ * Derived classes are traversed before base classes, and interfaces
+ * at the end.
+ */
+template<typename FieldStreamType>
+class HierarchicalFieldStream : public HierarchicalFieldStreamBase<FieldStreamType>  {
```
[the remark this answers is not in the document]
marc-chevalier: 1. Yes, good idea. I like that mostly for maintenance. On the other hand, despite my hopes, I never observed really compilers taking advantage of it in the wild (on made up examples, yes, of course).
2. I don't think I can: the public methods of the base class would then be private, and I need them in the public interface of the class. I've tried, and indeed, it doesn't compile.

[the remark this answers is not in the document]
merykitty: I believe they do devirtualize with `final` while non-`final` methods go through the whole virtual call sequence: https://godbolt.org/z/9MM1Kqqx8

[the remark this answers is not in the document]
marc-chevalier: Yes, of course, it can in such cases. But in actual code, I often end up with either an object of the final type (instead of a reference), and then, final doesn't help, or a reference/pointer to a base type, and so I don't save virtual call. In the past,  some big campaign of `final`izing everything were rewarded by no measurable performance improvement. But I still like it, as a hint for the reader not to try to find derived classes (but that the class actually is done implementing the thing it does), a warning that the class wasn't meant to be inherited (so be careful), and still hoping one in a while, it can save some virtual calls.

test/hotspot/jtreg/compiler/jvmci/jdk.vm.ci.code.test/src/jdk/vm/ci/code/test/VirtualObjectDebugInfoTest.java
```diff
@@ -70,6 +70,36 @@ private static class TestClass {
...
+                    .append(intField)
+                    .append("; ")
+                    .append("f: ")
+                    .append(floatField)
+                    .append("; ")
+                    .append("a[")
+                    .append(arrayField.length)
+                    .append("]: ");
+            for (int i = 0; i < arrayField.length; ++i) {
+                if(i != 0) {
```
TobiHartmann, line 89:
```suggestion
                if (i != 0) {
```

src/hotspot/share/runtime/signature.hpp
```diff
@@ -579,44 +579,17 @@ class SigEntry {
  public:
   BasicType _bt;      // Basic type of the argument
   int _offset;        // Offset of the field in its value class holder for scalarized arguments (-1 otherwise). Used for packing and unpacking.
-  float _sort_offset; // Offset used for sorting
-  Symbol* _symbol;    // Symbol for printing
+  Symbol* _name;    // Symbol for printing
+  bool _null_marker;   // Is it a null marker? For printing
```
TobiHartmann, line 583:
```suggestion
  Symbol* _name;      // Symbol for printing
  bool _null_marker;  // Is it a null marker? For printing
```

src/hotspot/share/oops/fieldStreams.hpp
```diff
@@ -248,63 +329,67 @@ class HierarchicalFieldStream : public StackObj  {
...
-
-  Symbol* generic_signature() const {
-    return _current_stream.generic_signature();
-  }
-
-  int offset() const {
-    return _current_stream.offset();
+/* Iterates on the fields of a class and its super-class top-down (java.lang.Object first)
+ * Doesn't traverse interfaces for now, because it's not clear which order would make sense
+ * Let's decide how  when/if the needs appear. Since we are not traversing interfaces, we
```
TobiHartmann, line 336:
```suggestion
 * Let's decide when or if the need arises. Since we are not traversing interfaces, we
```

src/hotspot/share/runtime/signature.hpp
```diff
@@ -579,44 +579,17 @@ class SigEntry {
  public:
   BasicType _bt;      // Basic type of the argument
   int _offset;        // Offset of the field in its value class holder for scalarized arguments (-1 otherwise). Used for packing and unpacking.
-  float _sort_offset; // Offset used for sorting
```
TobiHartmann, line 582: I'm glad this is gone now. Always felt hacky.

src/hotspot/share/runtime/sharedRuntime.cpp
```diff
@@ -2867,8 +2867,8 @@ void CompiledEntrySignature::compute_calling_conventions(bool init) {
             if (bt == T_OBJECT) {
               // Nullable inline type argument, insert InlineTypeNode::NullMarker field right after T_METADATA delimiter
               // Set the sort_offset so that the field is detected as null marker by nmethod::print_nmethod_labels.
```
TobiHartmann, line 2869: This comment needs to be updated. It still mentions the `sort_offset`.

src/hotspot/share/code/nmethod.cpp
```diff
@@ -3837,14 +3837,14 @@ void nmethod::print_nmethod_labels(outputStream* stream, address block_begin, bo
...
       bool did_name = false;
       if (is_reference_type(t)) {
-        Symbol* name = (*sig)._symbol;
+        Symbol* name = (*sig)._name;
         name->print_value_on(stream);
         did_name = true;
       }
       if (!did_name)
         stream->print("%s", type2name(t));
       // If the entry has a non-default sort_offset, it must be a null marker
```
TobiHartmann, line 3846: Comment needs to be updated.

src/hotspot/share/oops/inlineKlass.cpp
```diff
@@ -447,8 +441,21 @@ void InlineKlass::initialize_calling_convention(TRAPS) {
   if (InlineTypeReturnedAsFields || InlineTypePassFieldsAsArgs) {
     ResourceMark rm;
     GrowableArray<SigEntry> sig_vk;
-    float max_offset = 0;
-    int nb_fields = collect_fields(&sig_vk, max_offset);
+    int nb_fields = collect_fields(&sig_vk);
+    if (*PrintInlineKlassFields != '\0') {
+      const char* class_name_str = _name->as_C_string();
+      if (StringUtils::class_list_match(PrintInlineKlassFields, class_name_str)) {
```
TobiHartmann, line 447: Nice! I already added similar printing multiple times when debugging issues in this area. Not sure why I never integrated that.

TobiHartmann (approved):
Thanks for the thorough analysis Marc! I'm glad to see this code being fixed and cleanup up. Looks all good to me, I just found a few minor things.

> Nice, I have a pending issue of making this sort less a minefield but removing it completely is absolutely better.

@merykitty Should we close [JDK-8348547](https://bugs.openjdk.org/browse/JDK-8348547) as duplicate then? If so, I would suggest to integrate the regression tests attached to this bug with this PR. @marc-chevalier you might want to verify that they pass now.

merykitty:
@TobiHartmann Unfortunately, I don't think this PR solves [JDK-8348547](https://bugs.openjdk.org/browse/JDK-8348547). It solves [JDK-8357186](https://bugs.openjdk.org/browse/JDK-8357186), though.

marc-chevalier:
Should I do something to also associate this PR with JDK-8357186 then? I'm thinking of `issue add`, but I mostly mean it as a process question.

src/hotspot/share/runtime/sharedRuntime.cpp
marc-chevalier: I think we can do without the second line. The first says we include a null marker, and the second gives true to the null_marker parameter. Pretty straightforward.

merykitty:
I think `issue add` is appropriate. `ciInstanceKlass::get_non_flat_field_by_offset` assumes increasing offset in `_declared_nonstatic_fields`, please fix it.

TobiHartmann (approved):
> Unfortunately, I don't think this PR solves [JDK-8348547](https://bugs.openjdk.org/browse/JDK-8348547). It solves [JDK-8357186](https://bugs.openjdk.org/browse/JDK-8357186), though.

Ah, makes sense. Thanks for clarifying.

Looks good.

marc-chevalier:
/integrate

Thanks @TobiHartmann & @merykitty

PR #1514, opened 2025-07-22
8363846: [lworld] Make Class.isIdentityClass() non-native

coleenp:
This moves isIdentityClass() implementation to Class.java, and checks within the JVM that is_identity_class() doesn't check the access flags for an ArrayKlass, since AccessFlags aren't initialized in ArrayKlasses.  The AccessFlags should be moved from Klass.hpp to InstanceKlass.cpp in mainline but that's a more complicated change and has several pieces.
Added a test.
Tested with tier1 locally.

---------
### Progress

### Issue
 * [JDK-8363846](https://bugs.openjdk.org/browse/JDK-8363846): [lworld] Make Class.isIdentityClass() non-native (**Enhancement** - P4)

### Reviewers
 * [Roger Riggs](https://openjdk.org/census#rriggs) (@RogerRiggs - Committer)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/1514/head:pull/1514` \
`$ git checkout pull/1514`

`$ git checkout pull/1514` \
`$ git pull https://git.openjdk.org/valhalla.git pull/1514/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1514`

`$ git pr show -t 1514`

</details>

<a href="https://git.openjdk.org/valhalla/pull/1514.diff">https://git.openjdk.org/valhalla/pull/1514.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/1514#issuecomment-3104078637)
</details>

### DIFF
--- a/src/hotspot/share/oops/instanceKlass.cpp
+++ b/src/hotspot/share/oops/instanceKlass.cpp
@@ -3593,7 +3593,7 @@ bool InstanceKlass::find_inner_classes_attr(int* ooff, int* noff, TRAPS) const {
 
 void InstanceKlass::check_can_be_annotated_with_NullRestricted(InstanceKlass* type, Symbol* container_klass_name, TRAPS) {
   assert(type->is_instance_klass(), "Sanity check");
-  if (type->access_flags().is_identity_class()) {
+  if (type->is_identity_class()) {
     ResourceMark rm(THREAD);
     THROW_MSG(vmSymbols::java_lang_IncompatibleClassChangeError(),
               err_msg("Class %s expects class %s to be a value class, but it is an identity class",
--- a/src/java.base/share/classes/java/lang/Class.java
+++ b/src/java.base/share/classes/java/lang/Class.java
@@ -624,7 +624,15 @@ public final class Class<T> implements java.io.Serializable,
      * @since Valhalla
      */
     @PreviewFeature(feature = PreviewFeature.Feature.VALUE_OBJECTS, reflective=true)
-    public native boolean isIdentity();
+    public boolean isIdentity() {
+        if (isPrimitive()) {
+            return false;
+        } else if (PreviewFeatures.isEnabled()) {
+           return isArray() || Modifier.isIdentity(modifiers);
+        } else {
+            return !isInterface();
+        }
+    }
 
     /**
      * {@return {@code true} if this {@code Class} object represents a value
--- a/test/jdk/valhalla/valuetypes/IsIdentityClassTest.java
+++ b/test/jdk/valhalla/valuetypes/IsIdentityClassTest.java
@@ -59,7 +59,7 @@ public class IsIdentityClassTest {
         int imod = Integer.class.getModifiers();
         assertFalseIfPreview(Modifier.isIdentity(imod), "Modifier of Integer should not have IDENTITY set");
         int amod = Integer[].class.getModifiers();
-        assertTrue(Modifier.isIdentity(amod), "Modifier of array of inline types should have IDENTITY set");
+        assertEquals(PreviewFeatures.isPreviewEnabled(), Modifier.isIdentity(amod), "Modifier of array should have IDENTITY set");
     }
 
     @Test
--- a/test/jdk/valhalla/valuetypes/IsIdentityClassTest.java
+++ b/test/jdk/valhalla/valuetypes/IsIdentityClassTest.java
@@ -25,17 +25,14 @@
  * @test
  * @summary Test that IsIdentityClass and modifiers return true for flattened arrays.
  * @library /test/lib
- * @enablePreview
- * @run main/othervm -XX:-UseArrayFlattening -XX:-UseNullableValueFlattening IsIdentityClassTest
- * @run main/othervm -XX:+UseArrayFlattening -XX:+UseNullableValueFlattening IsIdentityClassTest
+ * @modules java.base/jdk.internal.misc
+ *          java.base/jdk.internal.value
+ * @run junit/othervm IsIdentityClassTest
+ * @run junit/othervm --enable-preview -XX:-UseArrayFlattening -XX:-UseNullableValueFlattening IsIdentityClassTest
+ * @run junit/othervm --enable-preview -XX:+UseArrayFlattening -XX:+UseNullableValueFlattening IsIdentityClassTest
  */
 
-/*
- * @test
- * @summary Test that IsIdentityClass and modifiers return true for arrays not in preview.
- * @library /test/lib
- * @run main/othervm IsIdentityClassTest
- */
+import org.junit.jupiter.api.Test;
 
 import java.lang.management.ManagementFactory;
 import java.lang.management.RuntimeMXBean;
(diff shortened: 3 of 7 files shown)


src/java.base/share/classes/java/lang/Class.java
```diff
@@ -624,7 +624,15 @@ public static Class<?> forName(Module module, String name) {
...
     @PreviewFeature(feature = PreviewFeature.Feature.VALUE_OBJECTS, reflective=true)
-    public native boolean isIdentity();
+    public boolean isIdentity() {
+        if (isPrimitive()) {
+            return false;
+        } else if (PreviewFeatures.isEnabled()) {
+           return isArray() || Modifier.isIdentity(modifiers);
+        } else {
+            return !isInterface();
+        }
```
RogerRiggs, line 634: Isn't this going to take a performance hit?

test/jdk/valhalla/valuetypes/IsIdentityClassTest.java
```diff
@@ -0,0 +1,84 @@
...
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+/*
+ * @test
+ * @summary Test that IsIdentityClass and modifiers return true for flattened arrays.
+ * @library /test/lib
+ * @enablePreview
```
RogerRiggs, line 28: Since isIdentity depends on enable-preview, it should be run twice, once with and once without.

src/java.base/share/classes/java/lang/Class.java
coleenp: Yes this is better performance because C2 now doesn't have to intrinsify this native call.

src/hotspot/share/oops/instanceKlass.cpp
```diff
@@ -3593,7 +3593,7 @@ bool InstanceKlass::find_inner_classes_attr(int* ooff, int* noff, TRAPS) const {
 
 void InstanceKlass::check_can_be_annotated_with_NullRestricted(InstanceKlass* type, Symbol* container_klass_name, TRAPS) {
```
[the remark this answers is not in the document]
coleenp: This was part of another change, and we still need this to check if a field can be made an inline type.

test/jdk/valhalla/valuetypes/IsIdentityClassTest.java
coleenp: Added.

src/hotspot/share/oops/instanceKlass.cpp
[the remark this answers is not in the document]
forax: And a strict field can be null

coleenp:
> is_identity_class seems widely used in many areas, such as C2 - do we need to verify those usages?

Good question.  I did check most/attempted all of these places and added the assert in klass.hpp but since AccessFlags are in klass, it is possible to copy them from an ArrayKlass to a ciKlass and fail the is_identity check for an array.  The only way to make this completely safe is to move the access flags, which I started in another repo but it's got pretty hairy since Klass and InstanceKlass are used interchangeably in some of the compiler code that I don't know very well, when I think they really mean InstanceKlass.

src/java.base/share/classes/java/lang/Class.java
RogerRiggs: I'd go ahead and create the isIdentity field and move this code to the constructor.
The modifiers, component type and primitive flag are all available at once.

test/jdk/valhalla/valuetypes/IsIdentityClassTest.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    static void testIsIdentityClass() {
+        Integer[] array0 = new Integer[200];
+        if (UseArrayFlattening) {
+            // NYI assertTrue(ValueClass.isFlatArray(array0));
+        } else {
+            assertFalse(ValueClass.isFlatArray(array0));
+        }
+        assertFalse(Integer.class.isIdentity(), "Integer is not an IDENTITY type");
```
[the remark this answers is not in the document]
RogerRiggs: I'll fixup TEST.properties and the other test in a separate PR.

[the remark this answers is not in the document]
liach: We should be able to use `@enablePreview false` to explicitly turn it off, without touching `TEST.properties`:
https://github.com/openjdk/jtreg/blob/8a906f35aeacc79cc723ed4dd73aa2bb6c7dd8e9/src/share/classes/com/sun/javatest/regtest/config/RegressionTestFinder.java#L644-L650

test/jdk/valhalla/valuetypes/IsIdentityClassTest.java
```diff
@@ -44,48 +41,51 @@
...
+            assertTrue(condition, msg);
+        }
+    }
+
+    @Test
+    void testIsIdentityClass() {
         RuntimeMXBean runtimeMxBean = ManagementFactory.getRuntimeMXBean();
         List<String> arguments = runtimeMxBean.getInputArguments();
-        UseArrayFlattening = !arguments.contains("-XX:-UseArrayFlattening");
+        boolean UseArrayFlattening = !arguments.contains("-XX:-UseArrayFlattening");
```
liach, line 63: Optional comments:

I wonder if it's more usually to obtain flags with serviceability, or explicitly pass -D properties...

Also for these global states, it's better to capture them with a `@Setup static void setup()` method and store in static fields, especially if multiple tests use these states.

test/jdk/valhalla/valuetypes/IsIdentityClassTest.java
```diff
@@ -44,48 +41,51 @@
...
 import jdk.internal.value.ValueClass;
 
 import static jdk.test.lib.Asserts.*;
 
 public class IsIdentityClassTest {
 
-    static boolean UseArrayFlattening = false;
-    static {
+    private static void assertFalseIfPreview(boolean condition, String msg) {
+        if (PreviewFeatures.isEnabled()) {
```
liach, line 52: We can use `assertEquals(!PreviewFeatures.isEnabled(), msg)`

test/jdk/valhalla/valuetypes/IsIdentityClassTest.java
```diff
@@ -0,0 +1,91 @@
...
+        List<String> arguments = runtimeMxBean.getInputArguments();
+        boolean UseArrayFlattening = !arguments.contains("-XX:-UseArrayFlattening");
+        System.out.println("UseArrayFlattening: " + UseArrayFlattening);
+
+        Integer[] array0 = new Integer[200];
+        if (UseArrayFlattening) {
+            // NYI assertTrue(ValueClass.isFlatArray(array0));
+        } else {
+            assertFalse(ValueClass.isFlatArray(array0));
+        }
```
RogerRiggs, line 71: This does not seem to be related to Class.isIdentity(); what is the connection?
Identity or not is not connected to flat or not. Maybe it belongs in a different test.

test/jdk/valhalla/valuetypes/IsIdentityClassTest.java
coleenp: Thats a good suggestion.

test/jdk/valhalla/valuetypes/IsIdentityClassTest.java
coleenp: I copied this out of a hotspot valhalla test.  If I remove the flattening tests, I suppose I won't need this anymore.

test/jdk/valhalla/valuetypes/IsIdentityClassTest.java
coleenp: With Fred's work for array flattening, I was testing that the array would be flattened.  But we do have other tests for this.  I can remove this part of the test.  I was trying to show that even if we've flattened an Integer array, the array itself is an identityClass.

test/jdk/valhalla/valuetypes/IsIdentityClassTest.java
coleenp: There might be a better way to find this out though.  We use this in the Hotspot runtime valhalla tests, but those tests don't use junit, so @things aren't used.  For Hotspot, we're passing -XX flags for now and not using properties.

test/jdk/valhalla/valuetypes/IsIdentityClassTest.java
```diff
@@ -0,0 +1,72 @@
...
+        assertFalseIfPreview(Integer.class.isIdentity(), "Integer is not an IDENTITY type");
+        assertTrue(Integer[].class.isIdentity(), "Arrays of inline types are IDENTITY types");
+    }
+
+    @Test
+    void testModifiers() {
+        int imod = Integer.class.getModifiers();
+        assertFalseIfPreview(Modifier.isIdentity(imod), "Modifier of Integer should not have IDENTITY set");
+        int amod = Integer[].class.getModifiers();
+        assertTrue(Modifier.isIdentity(amod), "Modifier of array of inline types should have IDENTITY set");
```
RogerRiggs, line 62:
```suggestion
        assertEquals(PreviewFeatures.isPreviewEnabled(), Modifier.isIdentity(amod), "Modifier of array should have IDENTITY set");
```
Before Valhalla, Arrays had identity but did not have a modifier bit set.
When --enable-preview, the IDENTITY modifier bit is set.

test/jdk/valhalla/valuetypes/IsIdentityClassTest.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    @Test
+    void testAccessFlags() {
+        Set<AccessFlag> iacc = Integer.class.accessFlags();
+        assertFalseIfPreview(iacc.contains(Modifier.IDENTITY), "Access flags should not contain IDENTITY");
```
RogerRiggs, line 68:
```suggestion
        if (PreviewFeatures.isEnabled()) {
            Set<AccessFlag> iacc = Integer.class.accessFlags();
            assertFalse(iacc.contains(AccessFlag.IDENTITY), "Access flags should not contain IDENTITY");
        }
Without --enable-preview (before Valhalla), there was no IDENTITY accessflag.
```

RogerRiggs (commented):
The test looks good, the conditions are more complicated than is obvious because with --enable-preview new modifier bits are defined with new used.
Thanks for the updates.

RogerRiggs (approved):
Look good, thanks

liach (approved):
The redundant directives in tests are fine.

test/jdk/valhalla/valuetypes/IsIdentityClassTest.java
```diff
@@ -0,0 +1,79 @@
...
+ * questions.
+ */
+
+/*
+ * @test
+ * @summary Test that IsIdentityClass and modifiers return true for arrays that can be flattened.
+ * @library /test/lib
+ * @enablePreview false
+ * @modules java.base/jdk.internal.misc
+ *          java.base/jdk.internal.value
```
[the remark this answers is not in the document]
coleenp: Thanks, I can't merge right now with the latest to test.

coleenp:
Thank you for all your help with this test, Roger and Chen.
/integrate

PR #1520, opened 2025-07-31
8364483: [lworld] Pre-register migrated classes in class loaders

coleenp:
This change preloads and preregisters all classes in the migrated classes lists, in all class loaders.
Tested with tier1 locally and with -XX:+PrintSystemDictionaryAtExit.

### DIFF
--- a/src/hotspot/share/classfile/vmSymbols.cpp
+++ b/src/hotspot/share/classfile/vmSymbols.cpp
@@ -110,6 +110,8 @@ void vmSymbols::initialize() {
 #endif
   }
 
+  initialize_migrated_class_names();
+
 #ifdef ASSERT
   // Check for duplicates:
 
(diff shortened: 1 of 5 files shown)


src/hotspot/share/gc/z/zBarrierSet.inline.hpp
```diff
@@ -35,6 +35,7 @@
 #include "memory/iterator.inline.hpp"
 #include "oops/inlineKlass.inline.hpp"
 #include "utilities/debug.hpp"
+#include "utilities/copy.hpp"
```
coleenp, line 38: This is unrelated but needed to get the product build to build.

coleenp:
I don't love having to enumerate all the classes in more than one place, but I also don't want to add another X macro.

src/hotspot/share/classfile/vmSymbols.cpp
```diff
@@ -296,3 +298,43 @@ vmSymbolID vmSymbols::find_sid(const char* symbol_name) {
...
+  _migrated_class_names[i++] = java_time_OffsetDateTime();
+  _migrated_class_names[i++] = java_time_OffsetTime();
+  _migrated_class_names[i++] = java_time_YearMonth();
+  _migrated_class_names[i++] = java_time_Year();
+  _migrated_class_names[i++] = java_time_Period();
+  _migrated_class_names[i++] = java_time_chrono_ChronoLocalDateImpl();
+  _migrated_class_names[i++] = java_time_chrono_MinguoDate();
+  _migrated_class_names[i++] = java_time_chrono_HijrahDate();
+  _migrated_class_names[i++] = java_time_chrono_JapaneseDate();
+  _migrated_class_names[i++] = java_time_chrono_ThaiBuddhistDate();
```
DanHeidinga, line 339: Should this `assert(i == _migrated_class_names_length);` as a handshake with the size of the set of migrated classes declared in vmSymbols.hpp?

coleenp: That's a good idea.  I didn't know a better C++ way to initialize these.  I'm sure it exists.

coleenp:
Thanks again, Dan.
/integrate

RogerRiggs:
The official list of migrated classes is in `make/modules/java.base/gensrc/GensrcValueClasses.gmk`.
How are all these copies of the list of files kept in sync?

DanHeidinga:
> The official list of migrated classes is in `make/modules/java.base/gensrc/GensrcValueClasses.gmk`.
> How are all these copies of the list of files kept in sync?

This is a temporary measure to make up for classfiles that don't have a `LoadedDescriptors` attribute yet.  Those classes - in user class loaders - won't "know" at link time that the box classes have migrated to be values and therefore won't benefit from the calling convention optimizations.  This PR lets us make those classloaders aware of the changed nature of the boxing classes early and ensures they get the benefits of Valhalla.  The other migrated classes are included because "why not" and so keeping the lists in sync isn't a priority.

PR #1523, opened 2025-08-15
8359370: [lworld] allow instance fields of identity classes to be readable in the prologue phase

vicente-romero-oracle:
Before this fix only strict fields were readable in the prologue phase. The proposed fix should allow any instance fields of identity classes to be readable in the prologue phase. This implies changes in flow analysis as before we were only tracking final and strict fields. There is also some "cooperation" needed in the code to detect cases when reading a field is not allowed in the prologue phase. For example some methods in Resolve don't have all the needed information at the moment they are dealing with some ASTs and part of the processing needs to be done in Attr

TIA

This PR is a remake of https://github.com/openjdk/valhalla/pull/1490

---------
### Progress

### Issue
 * [JDK-8359370](https://bugs.openjdk.org/browse/JDK-8359370): [lworld] allow instance fields of identity classes to be readable in the prologue phase (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/1523/head:pull/1523` \
`$ git checkout pull/1523`

`$ git checkout pull/1523` \
`$ git pull https://git.openjdk.org/valhalla.git pull/1523/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1523`

`$ git pr show -t 1523`

</details>

<a href="https://git.openjdk.org/valhalla/pull/1523.diff">https://git.openjdk.org/valhalla/pull/1523.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/1523#issuecomment-3197949598)
</details>


src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java
```diff
@@ -1252,6 +1254,25 @@ public void visitMethodDef(JCMethodDecl tree) {
 
                 // Attribute method body.
                 attribStat(tree.body, localEnv);
+                if (isConstructor) {
+                    ListBuffer<JCTree> prologueCode = new ListBuffer<>();
+                    for (JCTree stat : tree.body.stats) {
+                        prologueCode.add(stat);
+                        /* gather all the stats in the body until a `super` or `this` constructor invocation is found,
```
mcimadamore, line 1261: I understand that you  wanted to simplify the visitor -- but doing a linear pass on the constructor and creating a new list of statements is also kind of expensive -- maybe when we're done with this change we can see if there's a way to set a flag on the visitor to shortcircuit the analysis after the super call is found.

src/jdk.compiler/share/classes/com/sun/tools/javac/tree/TreeInfo.java
```diff
@@ -964,6 +966,41 @@ public static Name fullName(JCTree tree) {
...
+        for (JCTree node : nodes) {
+            java.util.List<SymAndTree> partialResult = symbolsFor(node);
+            if (!partialResult.isEmpty()) {
+                result.addAll(partialResult);
+            }
+        }
+        return result;
+    }
+
+    public static java.util.List<SymAndTree> symbolsFor(JCTree node) {
```
mcimadamore, line 982: Can you provide some examples on what this method is supposed to do? E.g. example input and output?

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java
```diff
@@ -1263,6 +1284,209 @@ public void visitMethodDef(JCMethodDecl tree) {
...
+        public void visitMethodDef(JCMethodDecl tree) {
+            // ignore any declarative part, mainly to avoid scanning receiver parameters
+            scan(tree.body);
+        }
+
+        void analyzeSymbol(JCTree tree) {
+            tree = TreeInfo.skipParens(tree);
+            Symbol sym = TreeInfo.symbolFor(tree);
+            if (sym != null) {
+                if (!sym.isStatic() && !isMethodArgument(tree)) {
```
mcimadamore, line 1420: if you have a `sym`, in order to understand if something is a method parameter (not argument?) don't you need to check if `sym.owner == MTH` ?

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java
```diff
@@ -1335,6 +1559,11 @@ public void visitVarDef(JCVariableDecl tree) {
                             //fixup local variable type
                             v.type = chk.checkLocalVarType(tree, tree.init.type, tree.name);
                         }
+                        if (v.owner.kind == TYP && !v.isStatic() && v.isStrict()) {
+                            // strict field initializers are inlined in constructor's prologues
+                            CtorPrologueVisitor ctorPrologueVisitor = new CtorPrologueVisitor(initEnv);
```
mcimadamore, line 1564: Nice reuse

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java
```diff
@@ -1263,6 +1284,209 @@ public void visitMethodDef(JCMethodDecl tree) {
...
+        boolean analyzingSelect = false;
+
+        @Override
+        public void visitSelect(JCFieldAccess tree) {
+            // skip if part of a larger select, context is important here
+            if (!analyzingSelect) {
+                boolean previousAnalyzingSelect = analyzingSelect;
+                try {
+                    analyzingSelect = true;
+                    super.visitSelect(tree);
```
mcimadamore, line 1363: Can't we cut recursion here (instead of using `analyzingSelect` ? That's also what the new `TreeInfo.symbolsFor` does. In general it seems like these two visitors are trying to do similar things but are not 100% aligned?

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java
vicente-romero-oracle: this is for cases when we have an argument that is for example of the same type as the current class so like:

class Test {
    String s;
    
    Test(Test t) {
        // the owner of s is Test not MTH so we need to check what is the qualifier for s which at the end is the argument
        // `t` so we ignore it
        String s1 = t.s;
        super();
    }
}

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java
```diff
@@ -1263,6 +1284,209 @@ public void visitMethodDef(JCMethodDecl tree) {
...
+        }
+
+        void analyzeSymbol(JCTree tree) {
+            tree = TreeInfo.skipParens(tree);
+            Symbol sym = TreeInfo.symbolFor(tree);
+            if (sym != null) {
+                if (!sym.isStatic() && !isMethodArgument(tree)) {
+                    if (sym.name == names._this || sym.name == names._super) {
+                        // are we seeing something like `this` or `CurrentClass.this` or `SuperClass.super::foo`?
+                        if (TreeInfo.isExplicitThisReference(
```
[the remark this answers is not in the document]
vicente-romero-oracle: all the code we analyze in this visitor is in the prologue, this is why we pre-select what code we will see

vicente-romero-oracle:
added a new iteration that removes the new visitor at TreeInfo, thanks for the comments so far!

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java
vicente-romero-oracle: I think I like the trade-off, we could try to infer if a constructor invocation corresponds to the class we are interested in. Like for example analyzing the symbol associated to a `super` or `this` invocation. But for erroneous invocations the symbol could be null. So what to do when we find a null symbol? We would have no clues I think.

mcimadamore: let's wait until we address the other comments first. What I had in mind (but can be addressed in a separate PR) was maybe have a general visitor for prologue -- e.g. an helper visitor class that only visits things inside the prologue. Then you can extend that helper visitor here, to do what you need to do.

src/jdk.compiler/share/classes/com/sun/tools/javac/tree/TreeInfo.java
```diff
@@ -1007,6 +1007,8 @@ private static Symbol symbolForImpl(JCTree node) {
             if (node.type != null)
                 return node.type.tsym;
             return null;
+        case TYPECAST:
```
mcimadamore, line 1010: Is this still needed?

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java
mcimadamore: (e.g. we really need to make sure that `analyzeSelect` is not applied too broadly)

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java
```diff
@@ -1263,6 +1284,205 @@ public void visitMethodDef(JCMethodDecl tree) {
...
+                        } else if (rs.isEarlyReference( // now this is a `proper` instance field of the current class
+                                        true,
+                                        localEnv,
+                                        tree.hasTag(SELECT) ? ((JCFieldAccess)tree).selected : null,
+                                        sym
+                                )) {
+                            /* references to fields of identity classes which happen to have initializers are
+                             * not allowed in the prologue
+                             */
+                            if (insideLambdaOrClassDef ||
```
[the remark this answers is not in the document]
vicente-romero-oracle: I think that the idea of restricting the access from lambdas and local classes is that they will capture `this` in order to access the field(s)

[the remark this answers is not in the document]
mcimadamore: Ah! Forgot about that one -- but... with proxy locals that's no longer the case, no?

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java
mcimadamore: I tried to replace this with:

`if (!sym.isStatic() && sym.kind == VAR && sym.owner.kind == TYP) { ... }`

And no tests seem to fail.

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java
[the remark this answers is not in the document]
mcimadamore: (or, do we only do local proxies for strict fields? If so, should we be more uniform here?)

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java
vicente-romero-oracle: yep :(

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java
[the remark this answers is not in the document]
vicente-romero-oracle: no we do local proxies for every field read in the prologue, a getField with a larval `this` will be an error in the VM

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Resolve.java
```diff
@@ -3942,101 +3921,6 @@ private List<Type> pruneInterfaces(Type t) {
         }
         return result.toList();
     }
-    private boolean isReceiverParameter(Env<AttrContext> env, JCFieldAccess tree) {
```
vicente-romero-oracle, line 3945: this method, `isReceiverParameter`, is not used any more, I removed it

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Resolve.java
```diff
@@ -3942,101 +3921,6 @@ private List<Type> pruneInterfaces(Type t) {
...
-     *  - The assignment is not within a lambda, because that would require
-     *    capturing 'this' which is not allowed prior to super().
-     *
-     * <p>
-     * Note, this method doesn't catch all such scenarios, because this method
-     * is invoked for symbol "x" only for "x = 42" but not for "this.x = 42".
-     * We also don't verify that the field has no initializer, which is required.
-     * To catch those cases, we rely on similar logic in Attr.checkAssignable().
-     */
-    private boolean isAllowedEarlyReference(DiagnosticPosition pos, Env<AttrContext> env, VarSymbol v) {
```
vicente-romero-oracle, line 3968: moved this one to Attr

mcimadamore:
> Do we still need the changes in `Attr.checkAssignable` -- e.g. flexible constructor bodies used to check some extra properties in there. Now that we have an LHS variable in the prologue scanner, I wonder if we can unify the checks?

I did an experiment, and adding this:

```
            if (isInLHS && !insideLambdaOrClassDef) {
                // Check instance field assignments that appear in constructor prologues
                if (isEarlyReference(localEnv, tree.hasTag(SELECT) ? ((JCFieldAccess)tree).selected : null, sym)) {
                    // Field may not be inherited from a superclass
                    if (sym.owner != localEnv.enclClass.sym) {
                        log.error(tree, Errors.CantRefBeforeCtorCalled(sym));
                        return;
                    }

                    // Field may not have an initializer
                    if ((sym.flags() & HASINIT) != 0) {
                        log.error(tree, Errors.CantAssignInitializedBeforeCtorCalled(sym));
                        return;
                    }
                }
                return;
            };
 ```
 
 At the start of `analyzeSymbol` seems to work fine. Maybe we want to break `analyzeSymbol` in different parts, one for reads and one for writes.

But one nice consequence of this is that now `isEarlyReference` is only called inside this visitor -- this means we can probably avoid having to pass `isPrologue = true` and, we can also avoid this dance:

```
tree.hasTag(SELECT) ? ((JCFieldAccess)tree).selected : null
```

Which is required at every callsite.

File: src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java (hunk 8 above)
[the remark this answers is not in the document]
Comment (mcimadamore) [reply src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java:1447]: (in your example with static context, my answer is that no, it should _not_ compile. A static context can't access instance fields from an enclosing class)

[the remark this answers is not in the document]
Comment (archiecobbs) [reply src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java:1447]: > (in your example with static context, my answer is that no, it should _not_ compile. A static context can't access instance fields from an enclosing class)

Yep, sorry that was a dumb example - I should have replaced the static context with a lambda.

I'm all for increased flexibility, it's just that it would be nice if that came with a clean mental model.

For example, one possible mental model could be "works like effectively final" - which I think is what you're advocating - but for that we'd have to break the equivalence between `y` and `A.this.y` (probably worth it).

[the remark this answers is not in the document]
Comment (vicente-romero-oracle) [reply src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java:1447]: I guess that we should discuss access from a lambda with Dan, and if we decide the rules should be relaxed then do it but I think as part of a future PR. I'm also all for relaxing restrictions

**Comment by vicente-romero-oracle** [comment]
> > Do we still need the changes in `Attr.checkAssignable` -- e.g. flexible constructor bodies used to check some extra properties in there. Now that we have an LHS variable in the prologue scanner, I wonder if we can unify the checks?
> 
> I did an experiment, and adding this:
> 
> ```
>             if (isInLHS && !insideLambdaOrClassDef) {
>                 // Check instance field assignments that appear in constructor prologues
>                 if (isEarlyReference(localEnv, tree.hasTag(SELECT) ? ((JCFieldAccess)tree).selected : null, sym)) {
>                     // Field may not be inherited from a superclass
>                     if (sym.owner != localEnv.enclClass.sym) {
>                         log.error(tree, Errors.CantRefBeforeCtorCalled(sym));
>                         return;
>                     }
> 
>                     // Field may not have an initializer
>                     if ((sym.flags() & HASINIT) != 0) {
>                         log.error(tree, Errors.CantAssignInitializedBeforeCtorCalled(sym));
>                         return;
>                     }
>                 }
>                 return;
>             };
> ```
> 
> At the start of `analyzeSymbol` seems to work fine. Maybe we want to break `analyzeSymbol` in different parts, one for reads and one for writes.
> 
> But one nice consequence of this is that now `isEarlyReference` is only called inside this visitor -- this means we can probably avoid having to pass `isPrologue = true` and, we can also avoid this dance:
> 
> ```
> tree.hasTag(SELECT) ? ((JCFieldAccess)tree).selected : null
> ```
> 
> Which is required at every callsite.

addressed in last commit, https://github.com/openjdk/valhalla/pull/1523/commits/9014afc5f9cb8b8d3f65f0b9846a27394ef56b51, thanks!

File: src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java [hunk 11]
```diff
@@ -1406,9 +1383,25 @@ public void visitMethodDef(JCMethodDecl tree) {
         }
 
         void analyzeSymbol(JCTree tree) {
-            if (isInLHS && !insideLambdaOrClassDef) return;
-            tree = TreeInfo.skipParens(tree);
             Symbol sym = TreeInfo.symbolFor(tree);
+            if (isInLHS && !insideLambdaOrClassDef) {
```
mcimadamore, line 1387: Question: for stuff that is NOT in LHS -- shouldn't we check that preview features are enabled? E.g. we don't want to allow reads in prologue unless `--enable-preview` is present, right?

vicente-romero-oracle:
> Very nice code consolidation. The new code is more centralized, and there's no more division of labor between Attr and Resolve. I think we should eventually consider something like this for mainline. Kudos!

yes I agree that we should consider something like this for mainline. Thanks for all the guidance / comments here and offline!

PR #1531, opened 2025-08-22
8361352: [lworld] Weird interaction between OSR and value class instances

marc-chevalier:
First, credit to @TobiHartmann for the diagnostic, and a lot of the solution.

# Diagnostic

According to [Strict Field Initialization JEP](https://openjdk.org/jeps/8350458), when a strict field is being initialized, it is not quite immutable, but observally immutable: at first, the field can be only set (during the early larval phase), then it can be only read (late larval or initialized phase), so the last set is the actual value one can ever observe.

The interesting part is that in early larval phase, a field can be subject to some side effects. When applied to a value object, that means that until it reaches the unrestricted state, it is not yet immutable. While being not theoretically necessary, avoiding scalarization and keeping the value object behind a reference is a convenient way to make sure that side effects are correctly applied. This strategy means that we shouldn't scalarized before reaching the unrestricted state. Normally, in C2, finding out what is early larval or not is the job of bytecode parsing, but in OSR compilation, everything about the StartOSR is not parsed, and thus some objects are soundly assumed that they might be larval, when they actually aren't. In the reported example, that leads to drastic performance difference between OSR and non OSR compilation: the second one is able to eliminate allocations since it knows more precisely when the value object can be scalarized.

In the original example:
```java
public value class MyNumber {
    private long d0;
    private MyNumber(long d0) { this.d0 = d0; }
    public MyNumber add(long v) { return new MyNumber(d0 + v); }

    private static void loop() {
        MyNumber dec = new MyNumber(123);
        for (int i = 0; i < 1_000_000_000; ++i) {
            dec = dec.add(i);
        }
    }

        for (int i = 0; i < 10; ++i) {
            loop();
        }
    }
}
```
OSR happens in the loop in `loop`, but here `dec` is not detected to be unrestricted (so immutable, so scalarizable), so the allocation in inlined `add` still needs to happen because we need the buffer for the new `dec`. The first iteration traps at the exit of the loop (unstable if), OSR happens again, followed by a non-OSR compilation, finding correctly that `dec` can be scalarized in the loop, making the third iteration fast.

# Solution

Overall, the solution requires to improve our detection of early larval values. Since we keep parsing as-is, let's do that in `ciTypeFlow`. The logic is quite simple: `ciInlineKlass` can be wrapped in `ciWrapper` to mark when they are known to be non-null. We extend `ciWrapper` to also be able to tell when a value is early larval (which also implies non-null since early larval means that `new` happened).

Then, it's almost a state machine! How is "early larval" introduced:
- `new` bytecode
- receiver of a constructor

How larval is removed:
- call to constructor

Other ways to obtain values are safe (I've experimented and it seems the verifier was keeping me from doing something bad):
- returned values are not early larval
- parameters are not early larval
- receiver of everything but constructor are not early larval

And other guarantees:
- calling twice the constructor is not legal
- merging paths with inconsistent state (early larval vs. unrestricted) is also disallowed

Let's detail how to remove the early larval state on a `ciInlineKlass`. When invoking a method on a receiver, we know that the receiver is not early larval after: either it's a constructor (which initialize the object), or it's a regular method, and the verifier makes sure the receiver was already not early larval before. The tricky part is to remove the larval state on all the copies of the reference: when doing something such as
```
new SomeValueClass
dup
dup
astore 1
invokespecial <init>
```
we still have a copy of the same reference in local 1 and on the stack. All of them now refers to a non-larval value. To take that into account, one must scan all the cells (locals and stack), and remove the early larval wrapper to all the types sharing the same ident  (see `void ciBaseObject::set_ident(uint id)` and `void ciObjectFactory::init_ident_of(ciBaseObject* obj)`). That is enough since early larval values cannot be contained in another object (since we can't make use of them before reaching unrestricted state, see also [Flexible Constructor Bodies](https://openjdk.org/jeps/513)).

Another trick is about the super constructor invocation in a constructor: is that different from a constructor invocation outside of a constructor. Well yes, but actually no:
- outside a constructor, a constructor invocation leaves the object initialized
- in a constructor, a constructor invocation doesn't make the object initialized because the all constructor is not done. But, calling the base class' constructor ensures that `Object.<init>` is being called, climbing up the hierarchy. Once `Object`'s constructor has been reached, the object is late larval.

In both case, after a constructor invocation, the object is unrestricted (late larval or initialized), and in the case of a value object, immutable, so scalarizable, which is what matters here.

The case of an abstract value class needs a bit of hack. An abstract value class is an `InstanceKlass` and a `ciInstanceKlass`, so we should take care of not doing the aforementioned logic only for `ciInlineKlass`, but also for `ciInstanceKlass` that are not identity classes.

# Tests

Testing is not quite easy. First, there is a microbenchmark.

Running
```
make test TEST="micro:org.openjdk.bench.valhalla.loops.osr.LarvalDetectionAboveOSR" CONF_NAME=linux-x64
```
before the fix would give
```
Iteration   1: 1874.498 ms/op
Iteration   2: 1662.409 ms/op
Iteration   3: 204.573 ms/op
Iteration   4: 201.177 ms/op
Iteration   5: 201.526 ms/op
Iteration   6: 200.718 ms/op
Iteration   7: 198.127 ms/op
Iteration   8: 200.535 ms/op
Iteration   9: 201.079 ms/op
Iteration  10: 203.341 ms/op
```
and now
```
Iteration   1: 212.941 ms/op
Iteration   2: 225.214 ms/op
Iteration   3: 204.319 ms/op
Iteration   4: 199.884 ms/op
Iteration   5: 201.875 ms/op
Iteration   6: 202.331 ms/op
Iteration   7: 200.829 ms/op
Iteration   8: 200.794 ms/op
Iteration   9: 200.428 ms/op
Iteration  10: 203.278 ms/op
```
That's better! For this microbench, we need to do everything that we shouldn't usually do: no warmup, single measure... Indeed, as soon as we do more than OSR, the effect disappear.

Otherwise, it's nice to see that the allocation is not there anymore in OSR. That could be an IR framework test, but alas, IR framework doesn't support OSR so easily. The test is thus running a subprocess and parsing the output quite basically. This should be ported to an IR test once the IR framework supports OSR.

Thanks,
Marc

---------
### Progress

### Issue
 * [JDK-8361352](https://bugs.openjdk.org/browse/JDK-8361352): [lworld] Weird interaction between OSR and value class instances (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/1531/head:pull/1531` \
`$ git checkout pull/1531`

`$ git checkout pull/1531` \
`$ git pull https://git.openjdk.org/valhalla.git pull/1531/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1531`

`$ git pr show -t 1531`

</details>

<a href="https://git.openjdk.org/valhalla/pull/1531.diff">https://git.openjdk.org/valhalla/pull/1531.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/1531#issuecomment-3214533928)
</details>

### DIFF
--- a/src/hotspot/share/ci/ciType.cpp
+++ b/src/hotspot/share/ci/ciType.cpp
@@ -148,7 +148,7 @@ ciWrapper::ciWrapper(ciType* type, int properties)
   assert(!type->is_wrapper(), "Thou shall not double wrap!");
   assert(type->is_inlinetype()
              // An abstract value type is an instance_klass
-             || (type->is_instance_klass() && type->as_instance_klass()->flags().is_abstract() && !type->as_instance_klass()->flags().is_identity())
+             || (type->is_instance_klass() && !type->as_instance_klass()->flags().is_identity())
              // An unloaded inline type is an instance_klass (see ciEnv::get_klass_by_name_impl())
              || (type->is_instance_klass() && !type->is_loaded()),
          "should only be used for inline types");
--- a/src/hotspot/share/ci/ciType.cpp
+++ b/src/hotspot/share/ci/ciType.cpp
@@ -140,3 +140,16 @@ void ciReturnAddress::print_impl(outputStream* st) {
 ciReturnAddress* ciReturnAddress::make(int bci) {
   GUARDED_VM_ENTRY(return CURRENT_ENV->get_return_address(bci);)
 }
+
+ciWrapper::ciWrapper(ciType* type, int properties)
+    : ciType(type->basic_type()),
+      _type(type),
+      _properties(properties) {
+  assert(!type->is_wrapper(), "Thou shall not double wrap!");
+  assert(type->is_inlinetype()
+             // An abstract value type is an instance_klass
+             || (type->is_instance_klass() && type->as_instance_klass()->flags().is_abstract() && !type->as_instance_klass()->flags().is_identity())
+             // An unloaded inline type is an instance_klass (see ciEnv::get_klass_by_name_impl())
+             || (type->is_instance_klass() && !type->is_loaded()),
+         "should only be used for inline types");
+}
--- a/test/hotspot/jtreg/compiler/valhalla/inlinetypes/LarvalDetectionAboveOSR.java
+++ b/test/hotspot/jtreg/compiler/valhalla/inlinetypes/LarvalDetectionAboveOSR.java
@@ -27,7 +27,7 @@
  * @summary In OSR compilation, we must correctly determine the initialization
  * state of value objects coming from above the OSR start, and not consider
  * everything as potentially early larval. Value objects that are known to be
- * unrestricted (late larval of fully initialized) are immutable, and can be
+ * unrestricted (late larval or fully initialized) are immutable, and can be
  * scalarized.
  * @library /test/jdk/java/lang/invoke/common
  * @enablePreview
--- a/test/hotspot/jtreg/compiler/valhalla/inlinetypes/LarvalDetectionAboveOSR.java
+++ b/test/hotspot/jtreg/compiler/valhalla/inlinetypes/LarvalDetectionAboveOSR.java
@@ -1,0 +1,280 @@
+/*
+ * Copyright (c) 2025, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+/**
+ * @test
+ * @bug 8361352
+ * @summary In OSR compilation, value objects coming from above the OSR start
+ * must be correctly found to be early larval so we know they are
+ * immutable, and they can be scalarized.
+ * @library /test/jdk/java/lang/invoke/common
+ * @enablePreview
+ * @run main compiler.valhalla.inlinetypes.LarvalDetectionAboveOSR
+ */
+
+package compiler.valhalla.inlinetypes;
+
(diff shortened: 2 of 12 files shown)


test/micro/org/openjdk/bench/valhalla/loops/osr/LarvalDetectionAboveOSR.java
```diff
@@ -0,0 +1,39 @@
+package org.openjdk.bench.valhalla.loops.osr;
```
TobiHartmann, line 1: Copyright header is missing.

test/hotspot/jtreg/compiler/valhalla/inlinetypes/LarvalDetectionAboveOSR.java
```diff
@@ -0,0 +1,281 @@
...
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+/**
+ * @test
+ * @bug 8361352
+ * @summary In OSR compilation, value objects coming from above the OSR start
+ * must be correctly found to be early larval so we know they are
+ * immutable, and they can be scalarized.
```
TobiHartmann, line 29: I suggest to rephrase this: "to be early larval so we know they are immutable" is confusing because early larvals are **mutable**.

src/hotspot/share/ci/ciType.cpp
```diff
@@ -140,3 +140,16 @@ void ciReturnAddress::print_impl(outputStream* st) {
...
 }
+
+ciWrapper::ciWrapper(ciType* type, int properties)
+    : ciType(type->basic_type()),
+      _type(type),
+      _properties(properties) {
+  assert(!type->is_wrapper(), "Thou shall not double wrap!");
+  assert(type->is_inlinetype()
+             // An abstract value type is an instance_klass
+             || (type->is_instance_klass() && type->as_instance_klass()->flags().is_abstract() && !type->as_instance_klass()->flags().is_identity())
```
TobiHartmann, line 151: Just wondering, is the `type->as_instance_klass()->flags().is_abstract()` condition even needed? Same in `ciTypeFlow::get_start_state()`.

TobiHartmann (approved):
Thanks for working on this Marc. Analysis and fix look good. I just added a few minor comments.

src/hotspot/share/ci/ciType.cpp
marc-chevalier: Probably not. An instance class without identity must be an abstract value class. At least, I don't see another way... But just to be sure, I'm running some tests.

TobiHartmann: Right, I think so too. Thanks for checking!

marc-chevalier: I've realized that since `ciInlineKlass` is a class derived from `ciInstanceKlass`, `type->is_instance_klass()` is also true for `ciInlineKlass`, which messed up my experiment but I guess it's reasonable.

TobiHartmann (approved):

marc-chevalier:
After the successful testing with asserting that the abstract was useless (implied by instanceKlass without identity), I've push the fix. The condition is actually slightly relaxed as it also catches final value classes, and not only abstract value classes, but anyway, we want both.

TobiHartmann (approved):
Thanks for verifying. Looks good to me. Ship it! :)

marc-chevalier:
/integrate

Thanks!

PR #1533, opened 2025-08-25
8364095: [lworld] AccessFlags location and release need futher work for Valhalla (IDENTITY and SUPER)

liach:


test/langtools/tools/javac/classfiles/attributes/innerclasses/InnerAnnotationsInInnerAnnotationTest.java
```diff
@@ -26,7 +26,6 @@
  * @bug 8042251
  * @summary Testing InnerClasses_attribute of inner annotations in inner annotation.
  * @library /tools/lib /tools/javac/lib ../lib
- * @enablePreview
```
coleenp, line 29: I think you should update the copyrights in these files.

liach: Good point. [Updated the copyright script](https://github.com/openjdk/jdk/pull/26998) to work for valhalla and done.

src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/ClassWriter.java
```diff
@@ -1758,10 +1763,13 @@ public void writeClassFile(OutputStream out, ClassSymbol c)
         acount += writeExtraAttributes(c);
 
         poolbuf.appendInt(JAVA_MAGIC);
+        boolean markedPreview;
         if (preview.isEnabled() && preview.usesPreview(c.sourcefile)) {
             poolbuf.appendChar(ClassFile.PREVIEW_MINOR_VERSION);
+            markedPreview = true;
         } else {
             poolbuf.appendChar(target.minorVersion);
+            markedPreview = target.minorVersion == ClassFile.PREVIEW_MINOR_VERSION;
```
mcimadamore, line 1772: I don't think `target.minorVersion` can ever be preview?

src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/ClassReader.java
```diff
@@ -3345,7 +3345,8 @@ long adjustClassFlags(ClassSymbol c, long flags) {
             flags &= ~ACC_MODULE;
             flags |= MODULE;
         }
-        if (((flags & ACC_IDENTITY) != 0 && !isMigratedValueClass(flags)) || (majorVersion < V67.major && (flags & INTERFACE) == 0)) {
+        if (((flags & ACC_IDENTITY) != 0 && !isMigratedValueClass(flags))
```
mcimadamore, line 3348: do we still need the `isMigratedValueClass` ? Aren't migrated value classes treated the same as ordinary value classes thanks to the work we did to load differerent classfiles based on version?

src/java.base/share/classes/java/lang/invoke/MethodHandleProxies.java
```diff
@@ -345,7 +345,7 @@ private static byte[] createTemplate(ClassLoader loader, ClassDesc proxyDesc, Cl
                         ClassLoaders.platformClassLoader() : loader)))
                         .build(proxyDesc, clb -> {
             clb.withSuperclass(CD_Object)
-               .withFlags((PreviewFeatures.isEnabled() ? ACC_IDENTITY  : 0) | ACC_FINAL | ACC_SYNTHETIC)
+               .withFlags(ACC_SUPER | ACC_FINAL | ACC_SYNTHETIC)
```
mcimadamore, line 348: I presume this change is because the classfile API will be able to filter out ACC_SUPER based on the target version?

src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/ClassReader.java
liach: I only adjusted this because this is incorrectly marking Java 25 classes as value classes in InnerClasses. This call didn't affect my patch, should I investigate it now?

src/java.base/share/classes/java/lang/reflect/AccessFlag.java
```diff
@@ -465,7 +458,7 @@ public enum Location {
          * @see Modifier#interfaceModifiers()
          * @jvms 4.1 The {@code ClassFile} Structure
          */
-        CLASS(ACC_PUBLIC | ACC_FINAL | (PreviewFeatures.isEnabled() ? ACC_IDENTITY : ACC_SUPER) |
+        CLASS(ACC_PUBLIC | ACC_FINAL | ACC_IDENTITY |
```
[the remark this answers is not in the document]
liach: This is just the bit masks, where ACC_SUPER is equivalent to ACC_IDENTITY. So 0x20 is always there in classes, but newly introduced to inner classes, so yes.

src/java.base/share/classes/java/lang/invoke/MethodHandleProxies.java
liach: No, because the old code is wrong - the flags are no-ops if the class file version is not latest release's preview (by default it is latest release.0)

src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/ClassReader.java
mcimadamore: It's ok to investigate in a separate PR

src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/ClassWriter.java
mcimadamore: IIRC, `forcePreview` makes use of _any_ feature behave as if they were preview features, meaning you will get classfile version pollution. But it doesn't alter Target. So yes, the code you suggest should be correct.

liach:
Thanks for the reviews! This will fix javap and facilitate further development, and there's no new failure, so I will go ahead and integrate.

/integrate

PR #1534, opened 2025-08-27
8366200: Modify existing use of jimage fields to make space for preview mode flags

david-beaumont:


src/jdk.jlink/share/classes/jdk/tools/jlink/internal/ImageResourcesTree.java
```diff
@@ -117,36 +130,51 @@ static final class PackageReference {
...
 
         PackageNode(String name, Node parent) {
             super(name, parent);
         }
 
+        int refCount() {
+            return references.size();
+        }
+
+        Stream<PackageReference> sortedRefs() {
```
RogerRiggs, line 161: Method name: "Refs" is pretty generic, in a class named ResourceTree.
How about sortedPackageRefs().

RogerRiggs (approved):
looks ok.

src/jdk.jlink/share/classes/jdk/tools/jlink/internal/ImageResourcesTree.java
david-beaumont: I mean it's in a class called `PackageNode` really (and isn't public), but I take your point.

PR #1536, opened 2025-08-27
8366214: [lworld] Use Objects.equals in HashMap and ConcurrentHashmap

RogerRiggs:
As observed in [JDK-8366043](https://bugs.openjdk.org/browse/JDK-8366043) [lworld] (LIFE = Legacy Idiom For Equality) causes performance regressions.
Updating HashMap and ConcurrentHashMap to use `java.util.Objects.equals` will make it easier to measure performance of options that remove or modify the use of `==`

Replace constructs like:
```
-                        ((k = e.key) == key || (key != null && key.equals(k))))
with:
+                        Objects.equals(key, k))
```

The changes in ConcurrentHashMap are a bit different due to the use of null as a sentinel.

The order of arguments to the .equals methods must remain the same to ensure compatibility.

---------
### Progress

### Issue
 * [JDK-8366214](https://bugs.openjdk.org/browse/JDK-8366214): [lworld] Use Objects.equals in HashMap and ConcurrentHashmap (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/1536/head:pull/1536` \
`$ git checkout pull/1536`

`$ git checkout pull/1536` \
`$ git pull https://git.openjdk.org/valhalla.git pull/1536/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1536`

`$ git pr show -t 1536`

</details>

<a href="https://git.openjdk.org/valhalla/pull/1536.diff">https://git.openjdk.org/valhalla/pull/1536.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/1536#issuecomment-3229537179)
</details>

### DIFF
--- a/src/java.base/share/classes/java/util/HashMap.java
+++ b/src/java.base/share/classes/java/util/HashMap.java
@@ -576,14 +576,14 @@ public class HashMap<K,V> extends AbstractMap<K,V>
         if ((tab = table) != null && (n = tab.length) > 0 &&
             (first = tab[(n - 1) & (hash = hash(key))]) != null) {
             if (first.hash == hash && // always check first node
-                ((k = first.key) == key || (key != null && key.equals(k))))
+                    Objects.equals(key, first.key))
                 return first;
             if ((e = first.next) != null) {
                 if (first instanceof TreeNode)
                     return ((TreeNode<K,V>)first).getTreeNode(hash, key);
                 do {
                     if (e.hash == hash &&
-                        ((k = e.key) == key || (key != null && key.equals(k))))
+                            Objects.equals(key, e.key))
                         return e;
                 } while ((e = e.next) != null);
             }
@@ -639,7 +639,7 @@ public class HashMap<K,V> extends AbstractMap<K,V>
         else {
             Node<K,V> e; K k;
             if (p.hash == hash &&
-                ((k = p.key) == key || (key != null && key.equals(k))))
+                    Objects.equals(key, p.key))
                 e = p;
             else if (p instanceof TreeNode)
                 e = ((TreeNode<K,V>)p).putTreeVal(this, tab, hash, key, value);
@@ -652,7 +652,7 @@ public class HashMap<K,V> extends AbstractMap<K,V>
                         break;
                     }
                     if (e.hash == hash &&
-                        ((k = e.key) == key || (key != null && key.equals(k))))
+                            Objects.equals(key, e.key))
                         break;
                     p = e;
                 }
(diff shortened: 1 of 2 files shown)


src/java.base/share/classes/java/util/concurrent/ConcurrentHashMap.java
```diff
@@ -665,8 +666,8 @@ public final boolean equals(Object o) {
             return ((o instanceof Map.Entry) &&
                     (k = (e = (Map.Entry<?,?>)o).getKey()) != null &&
                     (v = e.getValue()) != null &&
-                    (k == key || k.equals(key)) &&
-                    (v == (u = val) || v.equals(u)));
+                    (Objects.equals(k, key)) &&
```
liach, line 669: `k.equals(key)` should be sufficient here.

src/java.base/share/classes/java/util/HashMap.java
```diff
@@ -2024,7 +2023,7 @@ final TreeNode<K,V> find(int h, Object k, Class<?> kc) {
                     p = pl;
                 else if (ph < h)
                     p = pr;
-                else if ((pk = p.key) == k || (k != null && k.equals(pk)))
+                else if (Objects.equals(k, (pk = p.key)))
```
liach, line 2026: Stylistic: don't need parentheses around `pk = p.key`

src/java.base/share/classes/java/util/concurrent/ConcurrentHashMap.java
```diff
@@ -678,7 +679,7 @@ Node<K,V> find(int h, Object k) {
                 do {
                     K ek;
                     if (e.hash == h &&
-                        ((ek = e.key) == k || (ek != null && k.equals(ek))))
+                        (ek = e.key) != null && Objects.equals(k, ek))
```
liach, line 682: Looking at the code in HashMap, I think you can just use `Objects.equals(k, ek = e.key)`, or just `k.equals(ek = e.key)` because `Object::equals` is supposed to return false on null.

src/java.base/share/classes/java/util/HashMap.java
RogerRiggs: Stylistically, it calls attention to the assignment in the middle of an expression.

src/java.base/share/classes/java/util/concurrent/ConcurrentHashMap.java
liach: I think your current logic differs when `e.key == null` and `k == null` - previously it evaluates to true, now this test fails.

liach: Indeed, forgot this file is CHM; thought this one was the regular HM.

PR #1538, opened 2025-08-29
Fix issues in javac combo tests

mcimadamore:
A [change](https://bugs.openjdk.org/browse/JDK-8334294) made last year added `@enablePreview` to some tests using the combinatorial javac testing framework.
I believe the issue was that back then we had no way to load different version of a class (e.g. `Optional`) based on whether we were running on preview vs. non-preview so. Since the framework uses `Optional` it was treated as having an hard dependency on a preview feature, and compilation failed.
To fix this, we added `@enablePreview` in the tests that depend on that framework -- this meant the framework was now compiled with `--enable-preview` which solved the issue.

But this workaround is no longer needed today. And, worse, as we keep getting new tests from mainline, such tests will fail if they depend on the combinatorial framework, as they will depend on a "preview" classfile. So this means that to keep test working we will have to add `@enablePreview` to an increasing number of tests. This doesn't scale and since we no longer need it, it seems better to just just drop the `@enablePreview` directives that were added, effectively reverting the code to what is there in mainline/master.

### DIFF
--- a/test/langtools/tools/javac/ConditionalExpressionResolvePending.java
+++ b/test/langtools/tools/javac/ConditionalExpressionResolvePending.java
@@ -25,7 +25,6 @@
  * @test
  * @bug 8234899
  * @summary Verify behavior w.r.t. preview feature API errors and warnings
- * @enablePreview
  * @library /tools/lib /tools/javac/lib
  * @modules
  *      java.base/jdk.internal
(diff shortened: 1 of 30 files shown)


vicente-romero-oracle (approved):
looks good, thanks for fixing this

PR #1542, opened 2025-09-03
8366802: [lworld] jdk/javadoc/doccheck/checks/jdkCheckLinks.java fails since jdk-26+26

liach:
The missing link points to the preview note on `Attributes::loadableDescriptors`. Simply marking this API as preview fixes it.

Also I fixed the since version from 23 to Valhalla.

---------
### Progress
- [x] Change must not contain extraneous whitespace

### Issue
 * [JDK-8366802](https://bugs.openjdk.org/browse/JDK-8366802): [lworld] jdk/javadoc/doccheck/checks/jdkCheckLinks.java fails since jdk-26+26 (**Bug** - P4)

### Reviewers
 * [David Simms](https://openjdk.org/census#dsimms) (@MrSimms - Committer)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/1542/head:pull/1542` \
`$ git checkout pull/1542`

Update a local copy of the PR: \
`$ git checkout pull/1542` \
`$ git pull https://git.openjdk.org/valhalla.git pull/1542/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 1542`

View PR using the GUI difftool: \
`$ git pr show -t 1542`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
<a href="https://git.openjdk.org/valhalla/pull/1542.diff">https://git.openjdk.org/valhalla/pull/1542.diff</a>

</details>
<details><summary>Using Webrev</summary>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/1542#issuecomment-3249460292)
</details>

PR #1439, opened 2025-04-21
8355227: [lworld] Clarify preview API status for some valhalla APIs

liach:
Per [JEP 12](https://openjdk.org/jeps/12):

> In order for frameworks to deliver early support for a preview language feature, frameworks should utilize the reflective preview APIs that are co-developed with it. This will allow frameworks to keep up when programs are written to use the permanent version of the language feature.

According to these rules, the `Objects` methods and the `IdentityException` class themselves must be reflective preview APIs: programs compiled without preview feature dependencies that are running with preview features enabled should be able to access these APIs, as at run-time, they may receive value objects and `IdentityException` thrown by the JVM.

The only remaining essential preview APIs are the constructors of `IdentityException`: non-preview programs have no reason to construct them.

In addition, fixed a few JEP feature reference: notably, add reference to [Strict Fields JEP](https://openjdk.org/jeps/8350458).

Also, made ClassFile API new models reflective preview APIs - class file processors wish to process structures from preview class files even if the current runtime is not in preview. This is already happening to javac and javap.

---------
### Progress
- [x] Change must not contain extraneous whitespace

### Issue
 * [JDK-8355227](https://bugs.openjdk.org/browse/JDK-8355227): [lworld] Clarify preview API status for some valhalla APIs (**Bug** - P4)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/1439/head:pull/1439` \
`$ git checkout pull/1439`

Update a local copy of the PR: \
`$ git checkout pull/1439` \
`$ git pull https://git.openjdk.org/valhalla.git pull/1439/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 1439`

View PR using the GUI difftool: \
`$ git pr show -t 1439`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
<a href="https://git.openjdk.org/valhalla/pull/1439.diff">https://git.openjdk.org/valhalla/pull/1439.diff</a>

</details>
<details><summary>Using Webrev</summary>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/1439#issuecomment-2819535696)
</details>

PR #1543, opened 2025-09-03
LoadableDescriptorsAttrTest2 uses preview API Attributes.loadableDescriptors

liach:
Fixes new javac failures caused by #1542 when Attributes.loadableDescriptors is now preview. (It should probably be a reflective preview API; that will be addressed in #1439)


liach:
Test passes on CI against existing builds and locally, integrating.

/integrate

PR #1544, opened 2025-09-04
8366440: [lworld] Two tests broken with JDK-8364483

coleenp:
I added some logging and fixed the tests so that one doesn't rely on a migrated class in java.base to throw a VerifyError (will make a parallel fix to this test in mainline).  The other, I changed the logging message that it is looking for. More tests in this area will be added.
Tested locally.

### DIFF
--- a/src/hotspot/share/classfile/classFileParser.cpp
+++ b/src/hotspot/share/classfile/classFileParser.cpp
@@ -6218,7 +6218,7 @@ void ClassFileParser::post_process_parsed_stream(const ClassFileStream* const st
                                           "(cause: field type in LoadableDescriptors attribute) but loaded class is not a value class",
                                           name->as_C_string(), _class_name->as_C_string());
             }
-            } else {
+          } else {
             log_warning(class, preload)("Preloading of class %s during loading of class %s "
                                         "(cause: field type in LoadableDescriptors attribute) failed : %s",
                                         name->as_C_string(), _class_name->as_C_string(),
@@ -6229,11 +6229,16 @@ void ClassFileParser::post_process_parsed_stream(const ClassFileStream* const st
             CLEAR_PENDING_EXCEPTION;
           }
         } else {
-          // Just poking the system dictionary to see if the class has already be loaded
+          // Just poking the system dictionary to see if the class has already be loaded. Looking for migrated classes
+          // used when --enable-preview when jdk isn't compiled with --enable-preview so doesn't include LoadableDescriptors.
+          // This is temporary.
           oop loader = loader_data()->class_loader();
           InstanceKlass* klass = SystemDictionary::find_instance_klass(THREAD, name, Handle(THREAD, loader));
           if (klass != nullptr && klass->is_inline_klass()) {
             _inline_layout_info_array->adr_at(fieldinfo.index())->set_klass(InlineKlass::cast(klass));
+            log_info(class, preload)("Preloading of class %s during loading of class %s "
+                                     "(cause: field type not in LoadableDescriptors attribute) succeeded",
+                                     name->as_C_string(), _class_name->as_C_string());
           }
         }
       }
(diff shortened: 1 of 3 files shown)


src/hotspot/share/classfile/classFileParser.cpp
```diff
@@ -6229,11 +6229,16 @@ void ClassFileParser::post_process_parsed_stream(const ClassFileStream* const st
             CLEAR_PENDING_EXCEPTION;
           }
         } else {
-          // Just poking the system dictionary to see if the class has already be loaded
+          // Just poking the system dictionary to see if the class has already be loaded. Looking for migrated classes
+          // used when --enable-preview when jdk isn't compiled with --enable-preview so doesn't include LoadableDescriptors.
+          // This is temporary.
```
Arraying, line 6234: I was talking to Simms earlier and we think we should have a classfile check (i.e. ensure we are running whatever major + preview we are on at the moment), and then disable the whole loadabledescriptor functionality if the version doesn't match up. When we do this, we should probably consider that here as well. What do you think?

src/hotspot/share/classfile/classFileParser.cpp
```diff
@@ -6229,11 +6229,16 @@ void ClassFileParser::post_process_parsed_stream(const ClassFileStream* const st
...
           oop loader = loader_data()->class_loader();
           InstanceKlass* klass = SystemDictionary::find_instance_klass(THREAD, name, Handle(THREAD, loader));
           if (klass != nullptr && klass->is_inline_klass()) {
             _inline_layout_info_array->adr_at(fieldinfo.index())->set_klass(InlineKlass::cast(klass));
+            log_info(class, preload)("Preloading of class %s during loading of class %s "
+                                     "(cause: field type not in LoadableDescriptors attribute) succeeded",
```
Arraying, line 6240: "field type not in LoadableDescriptors" yet "succeeded" seems contradictory. Correct me if I'm wrong, but I think successful preloading implies that this class was (potentially transitively) in a LoadableDescriptors attribute of some class being loaded.

I'm not sure I follow why we suddenly introduce a log message here which is never checked in the tests. Could you elaborate?

src/hotspot/share/classfile/classFileParser.cpp
Arraying: Just to clarify, this PR is fine as-is but do you agree that we should include this part of code if that classfile checking manifests itself?

coleenp: It's under EnablePreview, like a lot of the code.  I'm hoping EnablePreview might check for classfile version?  Although it's a good question, maybe we do this speculative lookup only if classfile version is not preview.

This code unconditionally loads a bunch of classes in the bootclass loader, but doesn't add them to other class loaders as initiated classes unless EnablePreview is on.  I was thinking about how to load them only for EnablePreview as a separate change.

src/hotspot/share/classfile/classFileParser.cpp
coleenp: I couldn't figure out how to write this test (even though we do reach this code), or rather I didn't want to delay this change while figuring out how to write a test because it was breaking the CI.

Even though this doesn't technically preload the class, I wanted the log message to look like the other preloading messages, so kept "Preloading".  And I wanted to keep "field type".   How about:

Preloading of class %s during loading of class $s (cause: field type where LoadableDescriptors attribute is missing) succeeded

Although that's really long.  Suggestions welcome!

coleenp: I reworded the logging message so hopefully it makes more sense.

Arraying (approved):
I don't see the new logging message but since we discussed this extensively in the meeting I'm very happy to approve nonetheless. Let me know if you would like me to look at anything else, otherwise thank you for fixing these tests!!

coleenp:
Thanks Paul.  The path with the log message needs a test, but I want to integrate this to fix the noise in the testing.  Thank you for reviewing and your comments.
/integrate

PR #1548, opened 2025-09-04
8366913: [lworld] Add ACC_STRICT_INIT to ClassFile

liach:
@arraying discovered that java.lang.classfile.ClassFile is missing a constant for ACC_STRICT_INIT. This was due to previous iterations of strict fields proposal calling this flag simply ACC_STRICT like strictfp. The usages of ACC_STRICT and Modifier.STRICT has been reviewed, and ones that actually refer to STRICT_INIT (which is also not a source modifier) are migrated.


liach:
No new failure in tier 1-3, integrating.

/integrate

PR #1549, opened 2025-09-05
8365978: [lworld] C2: assert(vk->maybe_flat_in_array()) when using compareAndSetFlatValue with -XX:-UseArrayFlattening

marc-chevalier:
`Unsafe::compareAndSetFlatValue` calls `Unsafe::compareAndSetFlatValueAsBytes` which then calls `Unsafe::putFlatValue` on a flat array created by `Unsafe::newSpecialArray`.

https://github.com/openjdk/valhalla/blob/858be30119bd5bc37a69d16042523f53bea71a36/src/java.base/share/classes/jdk/internal/misc/Unsafe.java#L2870-L2875

`putFlatValue` can be intrinsified in

https://github.com/openjdk/valhalla/blob/858be30119bd5bc37a69d16042523f53bea71a36/src/hotspot/share/opto/library_call.cpp#L2727

that calls `cast_to_flat_array`

https://github.com/openjdk/valhalla/blob/858be30119bd5bc37a69d16042523f53bea71a36/src/hotspot/share/opto/library_call.cpp#L2831

which fails the assert

https://github.com/openjdk/valhalla/blob/858be30119bd5bc37a69d16042523f53bea71a36/src/hotspot/share/opto/graphKit.cpp#L1875-L1876

because of

https://github.com/openjdk/valhalla/blob/858be30119bd5bc37a69d16042523f53bea71a36/src/hotspot/share/oops/inlineKlass.cpp#L287-L290

Of course, if array flattening is disabled, one can't have flat arrays. And indeed, `Unsafe::newSpecialArray` will raise if called. So at runtime, the call to `Unsafe::compareAndSetFlatValue` should simply raise. But when compiled the assert is hit, crashing the VM, because it cannot tell that the code is dead, but can check that if it's not the array is not flat. That doesn't seem reasonable to me. I propose to insert a trap instead. Even in the case where this code wouldn't be dead (like if `Unsafe::newSpecialArray` is just undefined behavior instead of throwing), crashing the compiler doesn't seem like a good option.

The situation is actually more surprising than it seems: using `Unsafe::compareAndSetFlatValue` on a flat field with `-XX:-UseArrayFlattening` sounds reasonable, but doesn't actually work since the implementation will (attempt to) create flat arrays under the hood. It is not clear to me whether it's actually desirable, as `Unsafe::compareAndSetFlatValue` introduce some coupling of field and array flattening, but on the other hand, it's an unsafe API, so it's not that crazy to require more constrains to use.

Thanks,
Marc

---------
### Progress

### Issue
 * [JDK-8365978](https://bugs.openjdk.org/browse/JDK-8365978): [lworld] C2: assert(vk-&gt;maybe_flat_in_array()) when using compareAndSetFlatValue with -XX:-UseArrayFlattening (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/1549/head:pull/1549` \
`$ git checkout pull/1549`

`$ git checkout pull/1549` \
`$ git pull https://git.openjdk.org/valhalla.git pull/1549/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1549`

`$ git pr show -t 1549`

</details>

<a href="https://git.openjdk.org/valhalla/pull/1549.diff">https://git.openjdk.org/valhalla/pull/1549.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/1549#issuecomment-3258078724)
</details>

### DIFF
--- a/src/hotspot/share/opto/library_call.cpp
+++ b/src/hotspot/share/opto/library_call.cpp
@@ -2825,11 +2825,7 @@ bool LibraryCallKit::inline_unsafe_flat_access(bool is_store, AccessKind kind) {
       }
       ptr = basic_plus_adr(base, ConvL2X(offset));
     } else {
-      if (!UseArrayFlattening) {
-        uncommon_trap(Deoptimization::Reason_intrinsic,
-                      Deoptimization::Action_none);
-        return true;
-      } else {
+      if (UseArrayFlattening) {
         // Flat array must have an exact type
         bool is_null_free = layout != LayoutKind::NULLABLE_ATOMIC_FLAT;
         bool is_atomic = layout != LayoutKind::NON_ATOMIC_FLAT;
@@ -2841,6 +2837,10 @@ bool LibraryCallKit::inline_unsafe_flat_access(bool is_store, AccessKind kind) {
         if (ptr_type->field_offset().get() != 0) {
           ptr = _gvn.transform(new CastPPNode(control(), ptr, ptr_type->with_field_offset(0), ConstraintCastNode::StrongDependency));
         }
+      } else {
+        uncommon_trap(Deoptimization::Reason_intrinsic,
+                      Deoptimization::Action_none);
+        return true;
       }
     }
   } else {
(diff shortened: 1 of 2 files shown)


TobiHartmann (approved):
Thanks for the detailed analysis Marc. I don't like the coupling between flat arrays and fields in the compareAndSet implementation but that's not something that the JIT can fix :) Looks good to me!

marc-chevalier:
/integrate

Thanks @TobiHartmann

PR #1551, opened 2025-09-06
8366841 objects equals no acmp

RogerRiggs:
Modify java.util.Objects.equals to avoid acmp (==) when --enable-preview.

There is a simple bug fix in the SimpleValueGraphs test.


liach (approved):
Looks reasonable.

PR #1555, opened 2025-09-09
8352737: [lworld] Fix TestIntrinsics after VarHandle support cleanup

marc-chevalier:
This remove bailout in 8 cases in `test/hotspot/jtreg/compiler/valhalla/inlinetypes/TestIntrinsics.java`. This use a new small type that is tiny enough to be safely flattened in any case.

This change also makes sure that both branches in `test_*` are taken by asserting the value of `UseArrayFlattening`/`UseFieldFlattening`. One can check that there is at least one scenario for each case. Just add `Asserts.assertTrue(false);` if you have a doubt. Since each case run in a separate VM, I couldn't find an easy way to count that across scenarii, each branch is taken at least once.

There are still some bailout since `compareAndSetFlatValue` (and exchange one too) require array flattening. See https://github.com/openjdk/valhalla/pull/1549 & [JDK-8365978](https://bugs.openjdk.org/browse/JDK-8365978).

Thanks,
Marc


marc-chevalier:
/integrate

Thanks @TobiHartmann!

PR #1559, opened 2025-09-10
8367263: [lworld] fix compiler/valhalla/inlinetypes/TestIntrinsics.java after current merges of JDK-8352737 and JDK-8366705

marc-chevalier:
```
TEST64_LAYOUT = U.arrayLayout(TEST64_ARRAY.getClass());
```
is not correct anymore since [JDK-8366705](https://bugs.openjdk.org/browse/JDK-8366705) as the layout is no longer a class property, but an instance property (if I understood right)

Weirdly enough, code like
 ```
U.arrayBaseOffset(TEST64_ARRAY.getClass());
U.arrayIndexScale(TEST64_ARRAY.getClass());
```
got also their `.getClass()` removed in [JDK-8366705](https://bugs.openjdk.org/browse/JDK-8366705):
https://github.com/openjdk/valhalla/blob/0817750d434463a606523ac03b5461802f0328e1/test/hotspot/jtreg/compiler/valhalla/inlinetypes/TestIntrinsics.java#L650-L654
but there the compiler wasn't yelling at me. Is it overloaded and it works with instance or class, and the instance overload will just query the class. Not sure... I'm still removing it!

Thanks,
Marc


marc-chevalier:
/integrate

Thanks @MrSimms & @TobiHartmann!

PR #1592, opened 2025-09-16
8349037: [lworld] runtime/valhalla/inlinetypes/CircularityTest.java crashes with SIGSEGV

fparain:
runtime/valhalla/inlinetypes/CircularityTest.java was written to test the behavior of the JVM with respect of the initialization of classes of default values when the JVM was in charge of initializing null-restricted fields.
JEP 401 went into a different direction, the concept of default value has been removed, and the concept of strict fields has been defined to force user code to reliably initialized all strict fields before their first read.
Strict fields have their own tests ensuring static fields initialization rules are correctly enforced.
This test is then obsolete and should be removed from the repository.


fparain:
Thank your for the review Lois.

PR #1595, opened 2025-09-17
8367698: [lworld] New lint category for code that would not be allowed in the prologue

vicente-romero-oracle:
This PR introduces a new Xlint category: `initialization`. When enabled it will indicate if some code in an identity class couldn't be placed in the prologue phase. This applies to field initializers and constructors with no explicit `super` invocation. So for example for code like:

```
class Test {
    int i = 0;

    Test() {
        this.i = 1;
    }
}
```

a warning will be issued as if there were a super invocation at the end of the constructor, the compiler would issue an error.

### DIFF
--- a/src/jdk.compiler/share/classes/com/sun/tools/javac/code/Lint.java
+++ b/src/jdk.compiler/share/classes/com/sun/tools/javac/code/Lint.java
@@ -279,6 +279,12 @@ public class Lint {
         INCUBATING("incubating", false),
 
         /**
+         * Warn about code in identity classes that wouldn't be allowed in early
+         * construction due to a this dependency.
+         */
+        INITIALIZATION("initialization"),
+
+        /**
           * Warn about compiler possible lossy conversions.
           */
         LOSSY_CONVERSIONS("lossy-conversions"),
--- a/src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java
+++ b/src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java
@@ -1189,6 +1189,7 @@ public class Attr extends JCTree.Visitor {
                 // Add an implicit super() call unless an explicit call to
                 // super(...) or this(...) is given
                 // or we are compiling class java.lang.Object.
+                boolean addedSuperInIdentityClass = false;
                 if (isConstructor && owner.type != syms.objectType) {
                     if (!TreeInfo.hasAnyConstructorCall(tree)) {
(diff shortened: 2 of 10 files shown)


src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java
```diff
@@ -1246,11 +1248,13 @@ public void visitMethodDef(JCMethodDecl tree) {
...
                                 TreeInfo.isConstructorCall(mi)) {
-                            break;
+                            if (!addedSuperInIdentityClass || !allowValueClasses) {
+                                break;
+                            }
                         }
                     }
                     if (!prologueCode.isEmpty()) {
-                        CtorPrologueVisitor ctorPrologueVisitor = new CtorPrologueVisitor(localEnv);
+                        CtorPrologueVisitor ctorPrologueVisitor = new CtorPrologueVisitor(localEnv, addedSuperInIdentityClass && allowValueClasses);
```
mcimadamore, line 1257: I suppose I'd rather not run the visitor if the lint warning is disabled?

mcimadamore: Although, it might be tricky to see whether the warning is enabled or not here. The new design is to issue all warnings all the time, on the basis that typically it's too early to filter here,

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Lint.java
```diff
@@ -278,6 +278,12 @@ public enum LintCategory {
          */
         INCUBATING("incubating", false),
 
+        /**
+         * Warn about code in identity classes that wouldn't be allowed in early
+         * construction due to a this dependency.
+         */
+        INITIALIZATION("initialization"),
```
mcimadamore, line 285: IMHO this should contain the word `strict` somewhere.

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java
mcimadamore: scratch that -- you need the check always. I just don't get the `allowValueClass` filter here, but I guess we want to make the warnings conditional on --enable-preview, which is ok

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Lint.java
vicente-romero-oracle: sorry, do you mean the category or the comment?

vicente-romero-oracle:
thanks for the review

PR #1599, opened 2025-09-18
8368020: Remove preloading of null-restricted static fields classes

fparain:
Remove preloading of of null-restricted static fields classes. This code was needed when such fields were initialized with default values. In the current model, null-restricted fields are always strict and are initialized accordingly to the strict fields model.

Tested on Mach5, tiers 1 to 3.

Fred


coleenp (approved):
Thank you! This looks really nice.

fparain:
Thanks Coleen for the review.

PR #1602, opened 2025-09-19
8367415: Fully encapsulate array allocation in oopFactory

coleenp:
This makes oopFactory and deoptimization the only friends of private ArrayKlass::allocate_instance.  This is similar to the PR https://github.com/openjdk/jdk/pull/27372 in mainline with some extra handling for new_refArray (rather than calling RefArrayKlass::allocate_instance() directly.
Tested with tier1-4.


src/hotspot/share/oops/objArrayKlass.hpp
```diff
@@ -89,9 +92,7 @@ class ObjArrayKlass : public ArrayKlass {
   static ObjArrayKlass* allocate_objArray_klass(ClassLoaderData* loader_data,
                                                 int n, Klass* element_klass, TRAPS);
 
-  static ArrayDescription array_layout_selection(Klass* element, ArrayProperties properties);
-
-  virtual objArrayOop allocate_instance(int length, ArrayProperties props, TRAPS);
+ public:
```
fparain, line 95: Is this needed? There's already one at line 60.

coleenp: It's not.  I moved the methods to be where they are in mainline.  Thanks for noticing this.

coleenp:
Thanks Fred.
/integrate

PR #1610, opened 2025-09-22
8368099: [lworld] Crash in frame::sender_for_compiled_frame while processing JFR sampling request

pchilano:
Please review this small fix. The trick to create a synthetic compiled frame if processing a sample request at a poll on return doesn’t work for frames that need stack repair. The stack walking code will read the frame size from the stack, but the memory has already been overwritten by the SafepointBlob frame. See comments in JBS issue for full analysis.

I tested the patch along the fix for 8368002 and verified that Jetty.java doesn’t crash anymore.

Thanks,
Patricio


pchilano:
Thanks for the review Markus.
/integrate

liach:
Thanks for this fix!

/sponsor

PR #1621, opened 2025-09-23
8368467: [lworld] Add new flag generation for jimage to support preview mode

david-beaumont:
Adds support for writing preview related flags into jimage files.

Preview mode is complex. It's not nearly as simple as "does something in /modules/xxx/... have an entry in /modules/xxx/META-INF/preview/...".

Specific issues include:

Supporting preview-only resources without forcing a double lookup on everything.
Changing the set of entries in /packages/xxx directories to account for preview only packages in some modules.
Minimising the work done during image reader initialization to only need to process the small number of preview resources (rather than scanning the whole file to look for them).
The new flags added by this code address these issues, but calculating them correctly with only minor adjustments to the existing code was not feasible, it just became a lot larger and very complex.

To address this, a new type (ModuleReference) is introduced to track and then merge information about packages seen in each module. This allows a much simpler inner loop for processing resource paths when building the node tree, combined with a subsequent merging stage to produce the final package information for each module.

Not that since ModuleReference is needed during jimage reading, that class is already present in the previous PR on which this is based, but it starts to be used to calculate the module flags in this PR.

This PR can also adds the ImageReader unit tests for preview mode, which rely on being able to generate jimage files with preview mode flags in.

Compare and review this against https://github.com/openjdk/valhalla/pull/1619.

---------
### Progress

### Integration blocker
&nbsp;⚠️ Dependency #1619 must be integrated first

### Issue
 * [JDK-8368467](https://bugs.openjdk.org/browse/JDK-8368467): [lworld] Add new flag generation for jimage to support preview mode (**Sub-task** - P3)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/1621/head:pull/1621` \
`$ git checkout pull/1621`

`$ git checkout pull/1621` \
`$ git pull https://git.openjdk.org/valhalla.git pull/1621/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1621`

`$ git pr show -t 1621`

</details>

<a href="https://git.openjdk.org/valhalla/pull/1621.diff">https://git.openjdk.org/valhalla/pull/1621.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/1621#issuecomment-3325286883)
</details>


test/jdk/jdk/internal/jimage/ImageLocationTest.java
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    @Test
+    public void getFlags_packageFlags() {
+        assertThrows(
+                IllegalArgumentException.class,
+                () -> ImageLocation.getFlags("/packages/pkgname", p -> true));
+    }
+
+    @Test
```
david-beaumont: The `getPackageFlags` tests check that module references get merged correctly to create the parent directory flags we expect. While adding individual resources you build up a *list* of references in each package, one for each resource (directly or indirectly) in that package.
This is a key reason for how the new code can simplify things around the outer "add each resource in turn" loop, by just collecting the references and merging them later to get the right result.

PR #1632, opened 2025-09-25
8348972: [lworld] C1/C2: Update Valhalla with UseCompactObjectHeaders for oop->klass load/stores

chhagedorn:
This patch fixes various issues in C1 and C2 (11 bugs in total) when running with `-XX:+UseCompactObjectHeaders`.

Most of the bugs could either be traced back to:
- Unconditionally assuming old layout with klass pointer following immediately the mark word.
- Small mistakes when merging JEP 450.
- Using the wrong basic type `T_OBJECT` for flat arrays instead of `T_FLAT_ELEMENT`. In Valhalla, we require that the elements of an array are aligned:https://github.com/openjdk/valhalla/blob/04efe5c66b70b9dca1a5c538c9a04e7ad93c107a/src/hotspot/share/oops/arrayOop.hpp#L94-L97
https://github.com/openjdk/valhalla/blob/04efe5c66b70b9dca1a5c538c9a04e7ad93c107a/src/hotspot/share/oops/arrayOop.hpp#L57-L61
When we now wrongly use `T_OBJECT`, it is not a problem with default flags (using compressed klass pointers and no COH): An array has a 12 byte header + 4 byte length which is 16 bytes in total - no alignment required. But with COH, we now have a header of 8 bytes + 4 byte length. We now require alignment but when we pass in `T_OBJECT`, we miss the alignment which leads to crashes down the line. I went through all the uses of `base_offset_in_bytes()` and fixed more such issues.

To review: I made separate commits for each fixed bug, each message summarizing the found issue. To better understand individual changes of the patch, it might help to look at the commits separately.

**Testing:**
- Apart from one bug, I could trigger them with existing tests. So, I have not added new tests for these issues separately.
- One bug only manifested with a new test (`test127a`) which I added with this patch.
- tier1-5 + compiler stress:
  - No additional flags, i.e. (mostly) without COH: Looks good and only known failures
  - With explicitly setting COH:
    - A lot of CDS related failures on which @matias9927 is working ([JDK-8367959](https://bugs.openjdk.org/browse/JDK-8367959))
    - Test timeout with `TestIntrinsics`: Unrelated to COH but only triggering with COH ([JDK-8368274](https://bugs.openjdk.org/browse/JDK-8368274)).
    - Otherwise, looks good with only known failures.

Thanks,
Christian

---------
### Progress

### Issue
 * [JDK-8348972](https://bugs.openjdk.org/browse/JDK-8348972): [lworld] C1/C2: Update Valhalla with UseCompactObjectHeaders for oop-&gt;klass load/stores (**Sub-task** - P3)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/1632/head:pull/1632` \
`$ git checkout pull/1632`

`$ git checkout pull/1632` \
`$ git pull https://git.openjdk.org/valhalla.git pull/1632/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1632`

`$ git pr show -t 1632`

</details>

<a href="https://git.openjdk.org/valhalla/pull/1632.diff">https://git.openjdk.org/valhalla/pull/1632.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/1632#issuecomment-3333397342)
</details>

### DIFF
--- a/src/hotspot/cpu/aarch64/c1_MacroAssembler_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/c1_MacroAssembler_aarch64.cpp
@@ -185,14 +185,20 @@ void C1_MacroAssembler::initialize_header(Register obj, Register klass, Register
   assert_different_registers(obj, klass, len);
 
   if (UseCompactObjectHeaders || EnableValhalla) {
+    // COH: Markword contains class pointer which is only known at runtime.
+    // Valhalla: Could have value class which has a different prototype header to a normal object.
+    // In both cases, we need to fetch dynamically.
     ldr(t1, Address(klass, Klass::prototype_header_offset()));
     str(t1, Address(obj, oopDesc::mark_offset_in_bytes()));
   } else {
+    // Otherwise: Can use the statically computed prototype header which is the same for every object.
     mov(t1, checked_cast<int32_t>(markWord::prototype().value()));
     str(t1, Address(obj, oopDesc::mark_offset_in_bytes()));
   }
 
   if (!UseCompactObjectHeaders) {
+    // COH: Markword already contains class pointer. Nothing else to do.
+    // Otherwise: Fetch klass pointer following the markword
     if (UseCompressedClassPointers) { // Take care not to kill klass
       encode_klass_not_null(t1, klass);
       strw(t1, Address(obj, oopDesc::klass_offset_in_bytes()));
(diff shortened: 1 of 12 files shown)


chhagedorn:
Thanks Tobias!

PR #1647, opened 2025-09-30
8367767: [lworld] Interpreter adherence to new memory model rules for strict instance fields

fparain:
Create new method entry for the interpreter to add a memory barrier when entering the j.l.Object constructor.


fparain:
Thanks for the idea and review Dan.

PR #1656, opened 2025-10-03
8367785: [lworld] Compiler adherence to new memory model rules for strict instance fields

TobiHartmann:
The new memory model rules for strict instance fields (see [JDK-8366372](https://bugs.openjdk.org/browse/JDK-8366372)) require a memory barrier at the beginning of `java.lang.Object::<init>`. The JITs will try to omit it if the receiver type is statically known (potentially via CHA + a dependency) and the class does not contain any strict fields.

Potentially, there's some more tricks that we can play. Most importantly, I think we can omit barriers for final fields at the end of the constructors if the field is also strict. I filed [JDK-8369166](https://bugs.openjdk.org/browse/JDK-8369166) for this.

I added a test that already caught [JDK-8369044](https://bugs.openjdk.org/browse/JDK-8369044) and will fail immediately without the JIT changes (on AArch64).

Thanks,
Tobias

---------
### Progress

### Issue
 * [JDK-8367785](https://bugs.openjdk.org/browse/JDK-8367785): [lworld] Compiler adherence to new memory model rules for strict instance fields (**Sub-task** - P2)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/1656/head:pull/1656` \
`$ git checkout pull/1656`

`$ git checkout pull/1656` \
`$ git pull https://git.openjdk.org/valhalla.git pull/1656/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1656`

`$ git pr show -t 1656`

</details>

<a href="https://git.openjdk.org/valhalla/pull/1656.diff">https://git.openjdk.org/valhalla/pull/1656.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/1656#issuecomment-3370307083)
</details>


src/hotspot/share/c1/c1_GraphBuilder.cpp
```diff
@@ -4299,6 +4299,34 @@ bool GraphBuilder::try_inline_full(ciMethod* callee, bool holder_known, bool ign
   caller_state->truncate_stack(args_base);
   assert(callee_state->stack_size() == 0, "callee stack must be empty");
 
+  // Check if we need a membar at the beginning of the java.lang.Object
+  // constructor to satisfy the memory model for strict fields.
+  if (EnableValhalla && method()->intrinsic_id() == vmIntrinsics::_Object_init) {
```
liach, line 4304:
```suggestion
  if (EnableValhalla && callee->intrinsic_id() == vmIntrinsics::_Object_init) {
```
Fits better in the context of this function.

src/hotspot/share/c1/c1_LIRGenerator.cpp
```diff
@@ -3014,6 +3014,12 @@ void LIRGenerator::do_Base(Base* x) {
     java_index += type2size[t];
   }
 
+  // Check if we need a membar at the beginning of the java.lang.Object
+  // constructor to satisfy the memory model for strict fields.
+  if (EnableValhalla && method()->intrinsic_id() == vmIntrinsics::_Object_init) {
```
liach, line 3019: Should we perform checks to try skip this fence here?

src/hotspot/share/c1/c1_GraphBuilder.cpp
TobiHartmann: I would rather leave the code as is because we enter the callee scope just above. Also, code further below also uses `method()`.

test/hotspot/jtreg/compiler/valhalla/inlinetypes/TestStrictFieldBarriers.java
```diff
@@ -0,0 +1,227 @@
...
+        Thread reader = new Thread(() -> {
+            for (int i = 0; i < 100_000; ++i) {
+                // We don't check individual fields here because the checks need to be
+                // as fast a possible to increase the likelyhood of a race condition.
+                int res = sharedA1.x & sharedB1.x & sharedC1_1.x & sharedC1_1.y & sharedC1_2.x & ((C1)sharedC1_2).y &
+                          sharedA2.x & sharedB2.x & sharedC2_1.x & sharedC2_1.y & sharedC2_2.x & ((C2)sharedC2_2).y &
+                          sharedD1_1.x & ((D1)sharedD1_2).x & sharedD2_1.x & ((D2)sharedD2_2).x;
+                if (res != 1) {
+                    System.err.println("Incorrect field value observed!");
+                    System.exit(1);
```
chhagedorn, line 204: Is there any particular reason why you don't throw a `RuntimeException` here instead?

chhagedorn (approved):
Looks good to me and nice test!

test/hotspot/jtreg/compiler/valhalla/inlinetypes/TestStrictFieldBarriers.java
TobiHartmann: Yes, it's the best way (I think) to make sure all threads immediately exit if one thread observes an error. Throwing an exception will not stop the other threads.

TobiHartmann:
Thanks for the review Christian!

test/hotspot/jtreg/compiler/valhalla/inlinetypes/TestStrictFieldBarriers.java
chhagedorn: Got it, thanks for the explanation!

PR #1664, opened 2025-10-06
8368795: [lworld] javac rejects assignments to components of array fields with initializers in the prologue

vicente-romero-oracle:
still a work in progress


src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java
```diff
@@ -1356,6 +1356,20 @@ public void visitIdent(JCIdent tree) {
             analyzeSymbol(tree);
         }
 
+        boolean isIndexed = false;
+
+        @Override
+        public void visitIndexed(JCArrayAccess tree) {
+            boolean previewIsIndexed = isIndexed;
```
liach, line 1363: Do you mean `previousIsIndexed`?

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java
```diff
@@ -1434,7 +1448,9 @@ void analyzeSymbol(JCTree tree) {
                     }
                     // Field may not have an initializer
                     if ((sym.flags() & HASINIT) != 0) {
-                        reportPrologueError(tree, sym, true);
+                        if (!sym.type.hasTag(ARRAY) || !isIndexed) {
```
liach, line 1451: Does this work for complex ones like `(this.a = stuff)[b] = 5`?

Maybe we can try updating `isInLHS` to point to a JCTree and all the tests to be something like `tree == isInLHS` so this just won't run for array indexed trees?

vicente-romero-oracle:
test

vicente-romero-oracle:
tier1-3 tests are passing

PR #1971, opened 2026-01-26
8376346: [lworld] Basic Shenandoah support

shipilev:
Noticed that Shenandoah tests are failing in current lworld. This should fix most of the issues ahead of mainline integration. There are still plenty of TODO items we would need to resolve, but this at very least gives us a baseline to improve on.

Additional testing: 
 -  [x] Linux x86_64 server fastdebug, `hotspot_gc_shenandoah`
 -  [x] Linux x86_64 server fastdebug, `hotspot_valhalla jdk_valhalla`
 -  [x] Linux x86_64 server fastdebug, `hotspot_valhalla jdk_valhalla` with `-XX:+UseShenandoahGC`
 -  [x] Linux x86_64 server fastdebug, `hotspot_valhalla jdk_valhalla` with `-XX:+UseShenandoahGC -XX:ShenandoahGCHeuristics=aggressive`

---------

---------
### Progress

### Issue
 * [JDK-8376346](https://bugs.openjdk.org/browse/JDK-8376346): [lworld] Basic Shenandoah support (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/1971/head:pull/1971` \
`$ git checkout pull/1971`

`$ git checkout pull/1971` \
`$ git pull https://git.openjdk.org/valhalla.git pull/1971/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1971`

`$ git pr show -t 1971`

</details>

<a href="https://git.openjdk.org/valhalla/pull/1971.diff">https://git.openjdk.org/valhalla/pull/1971.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/1971#issuecomment-4659028875)
</details>


shipilev:
I would appreciate reviews so this thing could land. @liach, who should I ping?

liach:
@lfoltan Can you help take a look? Lois is coordinating the delivery of JEP 401.

src/hotspot/share/opto/inlinetypenode.cpp
```diff
@@ -1378,6 +1378,11 @@ InlineTypeNode* InlineTypeNode::make_from_flat_impl(GraphKit* kit, ciInlineKlass
   if (vk->is_naturally_atomic(null_free)) {
     do_atomic = false;
   }
+  if (do_atomic && UseShenandoahGC && vk->contains_oops()) {
```
merykitty: Can you instead reject flattening if the member type contains an oop?

shipilev: Maybe, I would really like to mark the paths where we need more work, like this. It follows the similar precedents in current Valhalla codebase, I think.

merykitty: You can add `assert` here so that future support will notify the lack of support from C2 here.

shipilev: Maybe? I do not have time for this this week, but I can try next week. What exactly are you having in mind? Putting down some feature flags when Shenandoah is enabled?

merykitty: In `fieldLayoutBuilder.cpp`, we have this:

    if (UseNullableNonAtomicValueFlattening) {
      _nullable_non_atomic_layout_size_in_bytes = new_raw_size;
      _null_marker_offset = null_marker_offset;
      _null_free_non_atomic_layout_alignment = _payload_alignment;
    }
    if (UseNullableValueFlattening) {
      // For the nullable atomic layout, the size mut be compatible with the platform capabilities
      int nullable_atomic_size = round_up_power_of_2(new_raw_size);
      if (nullable_atomic_size <= (int)MAX_ATOMIC_OP_SIZE) {
        _nullable_atomic_layout_size_in_bytes = nullable_atomic_size;
        _null_marker_offset = null_marker_offset;
      }
    }

I wonder if it is a good idea to insert a little bit of code here to disable these when `UseShenandoahGC` is `true` and the layout has oop. What do you think @fparain?

shipilev: I now pushed a simpler kludge: like ZGC, Shenandoah can "rely" on the implementation detail that the embedded oops are "only" the problem when we have compressed oops. Otherwise the atomic payloads have at most one oop, which are apparently handled okay. This is still a kludge, but it errs to make things slower instead of losing atomicity.

merykitty (approved):
Thanks, this workaround is also fine.

shipilev:
AFAICS, GHA failures are known failures from mainline and should disappear after the next merge.

Thanks for reviews, please take it in when you see fit.

/integrate

PR #2180, opened 2026-02-27
8377101: [lworld] Automatically generate value class sources based on source annotations

david-beaumont:
Adds a new AnnotationProcessor to read the @MigratedValueClass on prospective value classes and generate the equivalent source file with the 'value' keyword at each annotated class declaration.

This adds new PROCESSOR_PATH variable to the compilation macro and use it for annotation processing.
This also requires moving the plugin discovery path to the new variable (was in classpath before) because the addition of a '--processor-path' flag disables using the classpath for plugin discovery.

For now this is all limited to `java.base` but that's expandable later easily. There was an issue just enabling it for all modules however (esp. `jdk.incubator` which might be special).

While there is currently no specific unit test for this code, if it were to fail to generate the correct value classes, a lot of other downstream tests would fail.

---------

---------
### Progress

### Issue
 * [JDK-8377101](https://bugs.openjdk.org/browse/JDK-8377101): [lworld] Automatically generate value class sources based on source annotations (**Enhancement** - P4)

### Reviewers
 * [Chen Liang](https://openjdk.org/census#liach) (@liach - Committer) ⚠️ Review applies to [6b9608a8](https://git.openjdk.org/valhalla/pull/2180/files/6b9608a87664976d64128e92b45126684f4e6999)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2180/head:pull/2180` \
`$ git checkout pull/2180`

`$ git checkout pull/2180` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2180/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2180`

`$ git pr show -t 2180`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2180.diff">https://git.openjdk.org/valhalla/pull/2180.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2180#issuecomment-3972489865)
</details>


david-beaumont:
An example of output is, given the `Integer.java` source file:

**src/java.base/share/classes/java/lang/Integer.java**
```
@jdk.internal.MigratedValueClass
@jdk.internal.ValueBased
public final class Integer extends Number
```

**build/linux-x64/support/gensrc-valueclasses/java.base/java/lang/Integer.java**
```
@jdk.internal.MigratedValueClass
@jdk.internal.ValueBased
public final value class Integer extends Number
```

make/modules/java.base/Gensrc.gmk
```diff
@@ -37,7 +37,6 @@ include gensrc/GensrcMisc.gmk
 include gensrc/GensrcModuleLoaderMap.gmk
 include gensrc/GensrcRegex.gmk
 include gensrc/GensrcScopedMemoryAccess.gmk
-include gensrc/GensrcValueClasses.gmk
```
liach: Let's remove this gmk file too.

david-beaumont:
Now I see there are tests in `test/jdk/valhalla/valuetypes`, I'm wondering if the annotation processor should (at least temporarily) be in:

`make/jdk/src/classes/build/tools/valhalla/valuetypes`
rather than:
`make/jdk/src/classes/build/tools/valueclasses`

liach:
I think `valueclasses` directory is better. The test directory names I think come from earlier iterations of valhalla that we have not bothered to update yet.

david-beaumont:
Hmm, since I'm now adding a test in `test/jdk/valhalla/valuetypes` for it, I think I will move it to aid text-searching and discovery. I mean when this is all integrated the 'valhalla' bit should go away, so it'll just be:
`test/jdk/valuetypes` and `make/jdk/src/classes/build/tools/valuetypes`

RogerRiggs:
At some point, the test directory/tests specific to valhalla should be merged into the functional areas, but that may be only after they are out of preview-mode.

make/jdk/src/classes/build/tools/valueclasses/GenValueClasses.java
```diff
@@ -0,0 +1,286 @@
...
+    public synchronized void init(ProcessingEnvironment processingEnv) {
+        super.init(processingEnv);
+        this.processingEnv = processingEnv;
+        String outDir = this.processingEnv.getOptions().get(OUTDIR_OPTION_KEY);
+        if (outDir == null) {
+            throw new IllegalStateException(
+                    "Must specify -A" + OUTDIR_OPTION_KEY + "=<output-directory-path>"
+                            + " for annotation processor: " + GenValueClasses.class.getName());
+        }
+        this.outDir = Path.of(outDir.replace('/', File.separatorChar));
```
RogerRiggs: Separator replacement is not necessary, windows supports both.

make/jdk/src/classes/build/tools/valueclasses/GenValueClasses.java
RogerRiggs, line 1: OpenJDK has been moving away from `getXXXX` style names to just XXX, consider doing that here in new code.

david-beaumont:
Note: If you are testing this, you *must* `make clean` first, or else it will fail with a warning about the AP option not being recognised.

david-beaumont:
This is now tested and passes.

Obviously there's a question of how to access the annotation processor *for* testing "legitimately", but I think the tests are valuable enough to want to keep in one way or another.

liach (approved):
Impressive work! For the injection logic, what if we have something like:
```
@MigratedValueClass
record Test() {}
```
where would value be injected?

make/jdk/src/classes/build/tools/valhalla/valuetypes/GenValueClasses.java
```diff
@@ -0,0 +1,289 @@
...
+ * for compilation as preview classes. Note that more than one class in a given
+ * source file may be annotated.
+ *
+ * <p>Class re-writing is achieved by injecting the "value" keyword in front of
+ * class declarations for all annotated elements in the original source file.
+ *
+ * <p>Note that there are two annotations in use for value classes, but since
+ * we must generate sources for abstract classes, we only process one of them.
+ * <ul>
+ *     <li>{@code @jdk.internal.ValueBased} appears on concrete value classes.
```
jddarcy: I would have expected to the processor to process (and likely claim) both annotations. Can you say more here?

make/jdk/src/classes/build/tools/valhalla/valuetypes/GenValueClasses.java
```diff
@@ -0,0 +1,289 @@
...
+public final class GenValueClasses extends AbstractProcessor {
+    // Matches preprocessor option flag in CompileJavaModules.gmk.
+    private static final String OUTDIR_OPTION_KEY = "valueclasses.outdir";
+
+    private ProcessingEnvironment processingEnv = null;
+    private Path outDir = null;
+    private Trees trees = null;
+
+    @Override
+    public synchronized void init(ProcessingEnvironment processingEnv) {
```
jddarcy: Was there a reason using the Filer API was ruled out?

david-beaumont (commented):
- Addressed some comments.
- Test is no longer relying on hard assumptions about test directories in relation to build directories.
- Still need to investigate Messager API.

david-beaumont:
I added some use of the Messager API, but not everywhere since some exceptions really shouldn't happen (e.g. failing to read files properly). @jddarcy please let me know if it looks idiomatic now.

There's currently a weird unreliability wrt the actual builds (incremental building sometimes seems to not find the annotation processor, so I think there's a dependency missing that would trigger it being re-built (or something).

david-beaumont:
The build issue I'm seeing is:
* `make clean jdk-image` : Everything works
* Then `make jdk-image` : Everything works
* Make text edit to `GenValueClasses.java` and `make jdk-image` : Build breaks

But I don't know enough about the JDK build setup to know if this is expected/normal (since it's non-hermetic in a lot of ways and having to `make clean` is fairly normal.

make/jdk/src/classes/build/tools/valhalla/valuetypes/GenValueClasses.java
jddarcy: > After talking with Jan, this seemed the better option. What would the Filer API have simplified here?

After some further consideration, I could give better guidance if I understood more about the overall architecture about how the annotation processor was being used in the build process and how the Valhalla build would be sequenced with this change.

I have not attempted to research how the Valhalla build differs from a mainline build, but I'm generally aware there is some kind of post-processing of classes in java.base to make them value classes in the Valhalla images output.

I think two basic approaches to running an annotation processor to aid the post processing would be:

1) As an integrated part of the "base build", have the annotation processor run and generate the new source files in the chosen directory. (Is these were completely new files and new not meant to replace existing files in the output, the compilation of the new files could be handled as part of the built-in annotation processing compilation machinery).

2) As a post-pass after the base build, scan over the source (or class files) in question and find the annotated classes to have replacements be generated.

For something closer to 1) the Filer has a few facilities to do extra checking and some infrastructure for concepts like the SOURCE_OUTPUT location.

make/CompileJavaModules.gmk
```diff
@@ -154,6 +173,7 @@ ifneq ($(COMPILER), bootjdk)
         MODULE := $(MODULE), \
         SRC := $(wildcard $(MODULE_PREVIEW_SRC_DIRS)), \
         INCLUDES := $(JDK_USER_DEFINED_FILTER), \
+        EXTRA_DEPS := $(VALUETYPE_GENSRC_PROCESSOR_DEP), \
```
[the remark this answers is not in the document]
david-beaumont: Yeah, I wasn't completely clear if DEPENDS or EXTRA_DEPS was the right thing. It wasn't clear that one was internal.

The "depends" thing is interesting to me, because there's a conceptual distinction between "the code I'm compiling depends on this library" and "the action I'm performing should be re-performed if this thing was rebuilt".

Because the output from the compiler is the same, it's the output of the AP that differs when the processor is recompiled. I don't think this distinction is actually pulled out anywhere, so I guessed at "EXTRA_DEPS" as the right thing because "DEPENDS" feels like library dependencies etc.

But they do appear to be used in exactly the same places, so this distinction was always tenuous.

I'll make it use DEPENDS unless we find something better.

make/jdk/src/classes/build/tools/valhalla/valuetypes/GenValueClasses.java
RogerRiggs: @lahodaj We started with an explicit list of classes and a somewhat fragile makefile. 
Two places to update.   An extra pass over the source with fgrep probably could have made the list of classes more resiliant but an annotation processor and/or a javac plugin were suggested as possibilities.

david-beaumont:
Latest push is basically functional with two important caveats:
1. Fix for JDK-8378740 needs to be in and merged into lworld before the Xlint directives can be removed.
2. I have not been able to get it to recompile the module sources when the annotation processor itself changes (despite using basically the same approach as the depend plugin in terms of dependency management).

(1) is just a blocker which will go away, but I think I need help with (2).

david-beaumont:
So as regards the internal variable thing, I should make an "actual-processor-path" variable of some kind and modify that to pass through? I'll have a look.

david-beaumont:
I think this is now working, and correctly incrementally rebuilds the preview classes when the tool source changes.
I added a new "augmented" variable a'la classpath.

The only thing to wait for now is the Xlint stuff I think.

make/CompileJavaModules.gmk
```diff
@@ -68,6 +69,22 @@ MODULESOURCEPATH := $(call GetModuleSrcPath)
 # Add imported modules to the modulepath
 MODULEPATH := $(call PathList, $(IMPORT_MODULES_CLASSES))
 
+################################################################################
+# Setup preprocessor flags
+# The output directory must be present in GENERATED_PREVIEW_SUBDIRS in Modules.gmk.
+# Temporarily restrict this to java.base, but it can be expanded later.
+# TODO: Remove Xlint directives below once the fix in JDK-8378740 is merged into lworld.
```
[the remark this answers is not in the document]
david-beaumont: I can remove the "removal" directive, but not the "processor" one, since that's "correctly" reporting unused annotations. I'm in discussion with Jan to see if we can do better.

liach (approved):

make/jdk/src/classes/build/tools/valhalla/valuetypes/GenValueClasses.java
```diff
@@ -0,0 +1,289 @@
...
+ *     <li>{@code @jdk.internal.ValueBased} appears on concrete value classes.
+ *     <li>{@code @jdk.internal.MigratedValueClass} appears on concrete and
+ *         abstract value classes.
+ * </ul>
+ */
+@SupportedAnnotationTypes({"jdk.internal.MigratedValueClass"})
+@SupportedOptions("valueclasses.outdir")
+public final class GenValueClasses extends AbstractProcessor {
+    // Matches preprocessor option flag in CompileJavaModules.gmk.
+    private static final String OUTDIR_OPTION_KEY = "valueclasses.outdir";
```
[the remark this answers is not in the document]
ExE-Boss: I wish that it was allowed to refer to private fields of the target class from top‑level class annotations.

(This is already possible for inner classes due to nestmateness)

make/jdk/src/classes/build/tools/valhalla/valuetypes/GenValueClasses.java
```diff
@@ -0,0 +1,297 @@
...
+ *         abstract value classes.
+ * </ul>
+ */
+@SupportedAnnotationTypes(GenValueClasses.MIGRATED_VALUE_CLASS_ANNOTATION)
+@SupportedOptions("valueclasses.outdir")
+public final class GenValueClasses extends AbstractProcessor {
+    static final String MIGRATED_VALUE_CLASS_ANNOTATION = "jdk.internal.MigratedValueClass";
+
+    // Matches preprocessor option flag in CompileJavaModules.gmk.
+    private static final String OUTDIR_OPTION_KEY = "valueclasses.outdir";
```
ExE-Boss:
```suggestion
@SupportedAnnotationTypes(GenValueClasses.MIGRATED_VALUE_CLASS_ANNOTATION)
@SupportedOptions(GenValueClasses.OUTDIR_OPTION_KEY)
public final class GenValueClasses extends AbstractProcessor {
    static final String MIGRATED_VALUE_CLASS_ANNOTATION = "jdk.internal.MigratedValueClass";

    // Matches preprocessor option flag in CompileJavaModules.gmk.
    static final String OUTDIR_OPTION_KEY = "valueclasses.outdir";
```

make/common/JavaCompilation.gmk
```diff
@@ -302,13 +302,18 @@ define SetupJavaCompilationBody
     # including the compilation output on the classpath, so that incremental
     # compilations in unnamed module can refer to other classes from the same
     # source root, which are not being recompiled in this compilation:
-    $1_AUGMENTED_CLASSPATH += $$(BUILDTOOLS_OUTPUTDIR)/depend $$($1_BIN)
+    $1_AUGMENTED_CLASSPATH += $$($1_BIN)
+    $1_PROCESSOR_PATH += $$(BUILDTOOLS_OUTPUTDIR)/depend
```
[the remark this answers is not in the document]
david-beaumont: I think this is now using only "good practice" in terms of the Makefile changes. Please take another look.

liach (approved):
annotation consumption changes lgtm

make/jdk/src/classes/build/tools/valhalla/valuetypes/GenValueClasses.java
```diff
@@ -0,0 +1,300 @@
...
+ */
+@SupportedAnnotationTypes("*" /* JDK-8382613: GenValueClasses.MIGRATED_VALUE_CLASS_ANNOTATION */)
+@SupportedOptions(GenValueClasses.OUTDIR_OPTION_KEY)
+public final class GenValueClasses extends AbstractProcessor {
+    static final String MIGRATED_VALUE_CLASS_ANNOTATION = "jdk.internal.MigratedValueClass";
+
+    // Matches preprocessor option flag in CompileJavaModules.gmk.
+    static final String OUTDIR_OPTION_KEY = "valueclasses.outdir";
+
+    private ProcessingEnvironment processingEnv = null;
```
jddarcy: There is a protected processingEnv field define in the superclass you can use instead.

make/jdk/src/classes/build/tools/valhalla/valuetypes/GenValueClasses.java
```diff
@@ -0,0 +1,300 @@
...
+ * we must generate sources for abstract classes, we only process one of them.
+ * <ul>
+ *     <li>{@code @jdk.internal.ValueBased} appears on concrete value classes.
+ *     <li>{@code @jdk.internal.MigratedValueClass} appears on concrete and
+ *         abstract value classes.
+ * </ul>
+ *
+ * @implNote This annotation currently accepts and consumes <em>all</em>
+ * annotations. This is a temporary situation to avoid unwanted warnings about
+ * unrelated annotations. See JDK-8382613 for more details.
```
jddarcy, line 83: Note as a convention, we don't typically include bug numbers in comments in this fashion.

make/jdk/src/classes/build/tools/valhalla/valuetypes/GenValueClasses.java
```diff
@@ -0,0 +1,300 @@
...
+    }
+
+    @Override
+    public boolean process(Set<? extends TypeElement> annotations, RoundEnvironment env) {
+        // Don't do anything if there was an error in init().
+        if (outDir != null) {
+            // We don't have direct access to MigratedValueClass classes here.
+            Optional<? extends TypeElement> valueClassAnnotation =
+                    getAnnotation(annotations, MIGRATED_VALUE_CLASS_ANNOTATION);
+            if (valueClassAnnotation.isPresent()) {
```
jddarcy, line 126: I would expect a call to RoundEnvironment.getElementsAnnotatedWith() to be used in this method. Is there a reason for avoiding that?

If you need to operate on the _files_ containing the annotated types, you could use Elements.getFileObjectOf to create a set of them.

make/jdk/src/classes/build/tools/valhalla/valuetypes/GenValueClasses.java
david-beaumont: Yeah, I know. It's temporary though. I'll change it if you want, but I see not downside to giving a proper reference for the comment (unless you know of a better way, such as the bug URL).

make/jdk/src/classes/build/tools/valhalla/valuetypes/GenValueClasses.java
david-beaumont: I'm not super familiar with APs, so it's probably just that. Let me have a look...

make/jdk/src/classes/build/tools/valhalla/valuetypes/GenValueClasses.java
david-beaumont: I didn't spot it was visible. Ta.

make/jdk/src/classes/build/tools/valhalla/valuetypes/GenValueClasses.java
david-beaumont: Hmm, it is used inside the method I call slightly further down (~L145).

PR #2311, opened 2026-04-08
8381574: [lworld] Merge of JDK-8365501 broke the scalarized calling convention for abstract methods

TobiHartmann:
The merge of https://github.com/openjdk/jdk/commit/444a8fa14e8ab016b8aae018054c5dc1eb843fee into Valhalla broke the scalarized calling convention for abstract methods. Unfortunately, we didn't have any tests that caught this so this went unnoticed for a while until @rwestrel wondered why it's so tricky to come up with a test case for [JDK-8381563](https://bugs.openjdk.org/browse/JDK-8381563) :slightly_smiling_face: 

The problem is that no adapters are generated anymore for abstract methods because they are not needed. However, with the scalarized calling convention, we use the adapter to keep track of the scalarized signature. I re-enabled adapter generation if the scalarized calling convention is enabled - overhead should be benign because these are cached/shared after all.

I also added tests to catch similar breakages in the future, refactored some surrounding code and added asserts.

Thanks,
Tobias

---------

---------
### Progress

### Issue
 * [JDK-8381574](https://bugs.openjdk.org/browse/JDK-8381574): [lworld] Merge of JDK-8365501 broke the scalarized calling convention for abstract methods (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2311/head:pull/2311` \
`$ git checkout pull/2311`

`$ git checkout pull/2311` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2311/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2311`

`$ git pr show -t 2311`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2311.diff">https://git.openjdk.org/valhalla/pull/2311.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2311#issuecomment-4206884464)
</details>

### DIFF
--- a/src/hotspot/share/oops/method.cpp
+++ b/src/hotspot/share/oops/method.cpp
@@ -169,11 +169,17 @@ address Method::get_c2i_entry() {
 }
 
 address Method::get_c2i_inline_entry() {
+  if (is_abstract()) {
+    return SharedRuntime::get_handle_wrong_method_abstract_stub();
+  }
   assert(adapter() != nullptr, "must have");
   return adapter()->get_c2i_inline_entry();
 }
 
 address Method::get_c2i_inline_ro_entry() {
+  if (is_abstract()) {
+    return SharedRuntime::get_handle_wrong_method_abstract_stub();
+  }
   assert(adapter() != nullptr, "must have");
   return adapter()->get_c2i_inline_ro_entry();
 }
@@ -1342,20 +1348,12 @@ void Method::link_method(const methodHandle& h_method, TRAPS) {
   // called from the vtable.  We need adapters on such methods that get loaded
   // later.  Ditto for mega-morphic itable calls.  If this proves to be a
   // problem we'll make these lazily later.
-  if (is_abstract()) {
-    address wrong_method_abstract = SharedRuntime::get_handle_wrong_method_abstract_stub();
-    h_method->_from_compiled_entry = wrong_method_abstract;
-    h_method->_from_compiled_inline_entry = wrong_method_abstract;
-    h_method->_from_compiled_inline_ro_entry = wrong_method_abstract;
-  } else if (_adapter == nullptr) {
-    (void) make_adapters(h_method, CHECK);
-#ifndef ZERO
-    assert(adapter()->is_linked(), "Adapter must have been linked");
-#endif
-    h_method->_from_compiled_entry = adapter()->get_c2i_entry();
-    h_method->_from_compiled_inline_entry = adapter()->get_c2i_inline_entry();
-    h_method->_from_compiled_inline_ro_entry = adapter()->get_c2i_inline_ro_entry();
+  if (_adapter == nullptr && (!h_method->is_abstract() || InlineTypePassFieldsAsArgs)) {
+    make_adapters(h_method, CHECK);
   }
+  h_method->_from_compiled_entry = h_method->get_c2i_entry();
+  h_method->_from_compiled_inline_entry = h_method->get_c2i_inline_entry();
+  h_method->_from_compiled_inline_ro_entry = h_method->get_c2i_inline_ro_entry();
 
   // ONLY USE the h_method now as make_adapter may have blocked
 
@@ -1374,8 +1372,8 @@ void Method::link_method(const methodHandle& h_method, TRAPS) {
   }
 }
 
-address Method::make_adapters(const methodHandle& mh, TRAPS) {
-  assert(!mh->is_abstract(), "abstract methods do not have adapters");
+void Method::make_adapters(const methodHandle& mh, TRAPS) {
+  assert(!mh->is_abstract() || InlineTypePassFieldsAsArgs, "abstract methods do not have adapters");
   PerfTraceTime timer(ClassLoader::perf_method_adapters_time());
 
   // Adapters for compiled code are made eagerly here.  They are fairly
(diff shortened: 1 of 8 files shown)


marc-chevalier (approved):
It looks fine to me. I don't get the thing about `#ifndef ZERO` but it was there before, just moved, so I think I don't need to understand too much.

TobiHartmann:
Thanks for the review Marc!

> I don't get the thing about #ifndef ZERO but it was there before, just moved, so I think I don't need to understand too much.

Yeah, me neither :smile:

chhagedorn (approved):
Nice catch! Looks good to me, too.

TobiHartmann:
My new test triggered another issue (good job, test!):

```
#
#  Internal Error (/oracle/valhalla2/open/src/hotspot/share/opto/graphKit.cpp:2046), pid=589180, tid=589199
#  assert(C->inlining_incrementally() || C->strength_reduction()) failed: must be during devirtualization of calls

Current CompileTask:
C2:4520 1113    b  4       compiler.valhalla.inlinetypes.TestCallingConvention::test62 (6 bytes)

Stack: [0x00007fa3bba60000,0x00007fa3bbb60000],  sp=0x00007fa3bbb5c0f0,  free space=1008k
V  [libjvm.so+0x10b7418]  GraphKit::set_arguments_for_java_call(CallJavaNode*, bool)+0xc58  (graphKit.cpp:2046)
V  [libjvm.so+0xa33369]  DirectCallGenerator::generate(JVMState*)+0x2f9  (callGenerator.cpp:185)
V  [libjvm.so+0xe1b690]  Parse::do_call()+0x340  (doCall.cpp:703)
V  [libjvm.so+0x198f1d8]  Parse::do_one_bytecode()+0x458  (parse2.cpp:3620)
V  [libjvm.so+0x1977a9c]  Parse::do_one_block()+0x24c  (parse1.cpp:1732)
V  [libjvm.so+0x1978ee0]  Parse::do_all_blocks()+0x130  (parse1.cpp:746)
V  [libjvm.so+0x197c684]  Parse::Parse(JVMState*, ciMethod*, float)+0xac4  (parse1.cpp:650)
V  [libjvm.so+0xa2ae55]  ParseGenerator::generate(JVMState*)+0x135  (callGenerator.cpp:99)
V  [libjvm.so+0xc3cf8a]  Compile::Compile(ciEnv*, ciMethod*, int, Options, DirectiveSet*)+0x186a  (compile.cpp:831)
V  [libjvm.so+0xa282c8]  C2Compiler::compile_method(ciEnv*, ciMethod*, int, bool, DirectiveSet*)+0x4b8  (c2compiler.cpp:150)
V  [libjvm.so+0xc4d08f]  CompileBroker::invoke_compiler_on_method(CompileTask*)+0x6ff  (compileBroker.cpp:2345)
V  [libjvm.so+0xc4e8f0]  CompileBroker::compiler_thread_loop()+0x530  (compileBroker.cpp:1989)
V  [libjvm.so+0x122922b]  JavaThread::thread_main_inner()+0x13b  (javaThread.cpp:773)
V  [libjvm.so+0x1d06ca6]  Thread::call_run()+0xb6  (thread.cpp:243)
V  [libjvm.so+0x1923388]  thread_native_entry(Thread*)+0x118  (os_linux.cpp:931)
```

This assert was added with [JDK-8372824](https://bugs.openjdk.org/browse/JDK-8372824) and is too strong.

`CallGenerator::for_guarded_call` creates a `PredictedCallGenerator` that will emit the `subtype_check_receiver` but since we run with `AlwaysIncrementalInline`, it's stored in `LateInlineCallGenerator::_inline_cg` for later and no type check is emitted in the graph yet. `LateInlineCallGenerator::generate` then delegates to `DirectCallGenerator::generate` which creates a direct call to be late-inlined. When processing the arguments for that call, we assert because the receiver is non-scalarized. That's expected because the abstract receiver object was not casted to a concrete type yet (but the receiver type that we get from the signature of the target method is concrete). That casting will only happen when late-inlining happens. FTR, these receiver type checks were introduced by [JDK-8268405](https://bugs.openjdk.org/browse/JDK-8268405).

I weakened the assert, updated the comment and added a targeted regression test for this particular case.

TobiHartmann:
Testing revealed yet another issue. I'm investigating and will update the PR later.

TobiHartmann:
I finally found the second issue. Now that the calling convention is re-enabled for abstract methods, more tests in `TestC2CCalls` test what they are supposed to and in combination with `-XX:+StressCallingConvention` they triggered an issue with how mismatched methods are handled. I think it's a regression from [JDK-8369185](https://bugs.openjdk.org/browse/JDK-8369185) that was integrated while the calling convention for abstract methods was already broken and this therefore went unnoticed.

The problem is that code in `SharedRuntime::resolve_helper` always looks at the resolved method to check for a mismatch but if the virtual call is optimized, i.e. converted to a static call, C2 will use the calling convention of the selected method which might not be mismatched. The fix is straight forward and we don't need a regression test because this reliably triggers in our testing with `TestC2CCalls.java`.

marc-chevalier (approved):
I'm convinced.

TobiHartmann:
Thank you Marc! :slightly_smiling_face:

chhagedorn (approved):
Nice tests I'd say! Update looks good.

TobiHartmann:
Thanks Christian!

PR #2312, opened 2026-04-08
8381854: [lworld] Add acmp maps test to the FieldLayoutAnalyzer

fparain:
Add acmp_maps verification to the FieldLayoutAnalyzer.
The FieldLayoutAnalyzer recomputes the memory segments used by fields from the available layout descriptions, and compare the results to the acmp_maps computed by the JVM.
The verification is automatically performed for all tests calling the `check()` method of the FieldLayoutAnalyzer (like the ValueRandomLayoutTest.java).

Tested repeatedly with ValueRandomLayoutTest.java on Mach5.

Thanks,

Fred

---------
### Progress

### Issue
 * [JDK-8381854](https://bugs.openjdk.org/browse/JDK-8381854): [lworld] Add acmp maps test to the FieldLayoutAnalyzer (**Bug** - P4)

### Reviewers
 * [Paul Hübner](https://openjdk.org/census#phubner) (@Arraying - Committer) ⚠️ Review applies to [1ac3b7e7](https://git.openjdk.org/valhalla/pull/2312/files/1ac3b7e7a14e6196db861cf1398c7488df39cb59)
 * [Matias Saavedra Silva](https://openjdk.org/census#matsaave) (@matias9927 - Committer)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2312/head:pull/2312` \
`$ git checkout pull/2312`

`$ git checkout pull/2312` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2312/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2312`

`$ git pr show -t 2312`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2312.diff">https://git.openjdk.org/valhalla/pull/2312.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2312#issuecomment-4207454596)
</details>

### DIFF
--- a/test/hotspot/jtreg/runtime/valhalla/inlinetypes/field_layout/FieldLayoutAnalyzer.java
+++ b/test/hotspot/jtreg/runtime/valhalla/inlinetypes/field_layout/FieldLayoutAnalyzer.java
@@ -849,7 +849,7 @@ public class FieldLayoutAnalyzer {
     checkAcmpMap();
   }
 
-  void postPrecessing() {
+  void postProcessing() {
     // Abstract value class don't have a first field offset in the log, and this information is needed
     // in the verification of acmp maps. The log doesn't contain the information needed to identify
     // abstract value classes, so the value is computed for all classes with an uninitialized fieldFieldOffset field
@@ -889,6 +889,6 @@ public class FieldLayoutAnalyzer {
       System.out.println("Error while processing line: " + lo.getCurrentLine());
       throw t;
     }
-    postPrecessing();
+    postProcessing();
   }
  }
--- a/test/hotspot/jtreg/runtime/valhalla/inlinetypes/field_layout/FieldLayoutAnalyzer.java
+++ b/test/hotspot/jtreg/runtime/valhalla/inlinetypes/field_layout/FieldLayoutAnalyzer.java
@@ -27,10 +27,7 @@ import java.io.PrintStream;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
-
-
 import jdk.test.lib.Asserts;
-
 
 public class FieldLayoutAnalyzer {
 
--- a/test/hotspot/jtreg/runtime/valhalla/inlinetypes/field_layout/FieldLayoutAnalyzer.java
+++ b/test/hotspot/jtreg/runtime/valhalla/inlinetypes/field_layout/FieldLayoutAnalyzer.java
@@ -25,15 +25,12 @@ package runtime.valhalla.inlinetypes.field_layout;
 
 import java.io.PrintStream;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 
-import javax.management.RuntimeErrorException;
 
 import jdk.test.lib.Asserts;
-import jdk.test.lib.process.OutputAnalyzer;
-import jdk.test.lib.process.ProcessTools;
+
 
 public class FieldLayoutAnalyzer {
 
@@ -108,6 +105,8 @@ public class FieldLayoutAnalyzer {
     }
   }
 
+  static public record AcmpMapSegment(int offset, int size) {}
+
   static public record FieldBlock (int offset,
                             BlockType type,
                             int size,
@@ -136,6 +135,7 @@ public class FieldLayoutAnalyzer {
     }
 
     boolean isFlat() { return type == BlockType.FLAT; } // Warning: always return false for inherited fields, even flat ones
+    boolean isRegular() { return type == BlockType.REGULAR; }
     boolean hasNullMarker() { return layoutKind.hasNullMarker(); }
 
     static FieldBlock parseField(String line) {
@@ -204,10 +204,14 @@ public class FieldLayoutAnalyzer {
     String[] lines;
     ArrayList<FieldBlock> staticFields;
     ArrayList<FieldBlock> nonStaticFields;
+    ArrayList<AcmpMapSegment> nonOopAcmpMap;
+    ArrayList<Integer> oopAcmpMap;
 
     private ClassLayout() {
       staticFields = new ArrayList<FieldBlock>();
       nonStaticFields = new ArrayList<FieldBlock>();
+      nonOopAcmpMap = new ArrayList<AcmpMapSegment>();
+      oopAcmpMap = new ArrayList<Integer>();
     }
 
     boolean hasNullFreeNonAtomicLayout() { return nullFreeNonAtomicLayoutSize != -1; }
@@ -375,9 +379,25 @@ public class FieldLayoutAnalyzer {
         } else {
           cl.nullMarkerOffset = -1;
         }
+        // Non-oop acmp map <offset,size>: <12,1> <16,4> ...
         Asserts.assertTrue(lo.getCurrentLine().startsWith("Non-oop acmp map <offset,size>"), lo.getCurrentLine());
+        String[] acmp_nonoop_line = lo.getCurrentLine().split("\\s+");
+        Asserts.assertTrue(acmp_nonoop_line.length >= 4, "Wrong format for non-oop acmp map line: " + lo.getCurrentLine());
+        for (int i = 4; i < acmp_nonoop_line.length; i++) {
+          String[] offset_size = acmp_nonoop_line[i].split(",");
+          Asserts.assertTrue(offset_size.length == 2, "Wrong format for acmp map entry: " + acmp_nonoop_line[i]);
+          int offset = Integer.parseInt(offset_size[0].substring(1)); // skipping '<'
+          int size = Integer.parseInt(offset_size[1].substring(0, offset_size[1].length() - 1)); // skipping '>'
+          cl.nonOopAcmpMap.add(new AcmpMapSegment(offset, size));
+        }
         lo.moveToNextLine();
+        // oop acmp map: 12 16 ...
         Asserts.assertTrue(lo.getCurrentLine().startsWith("oop acmp map"), lo.getCurrentLine());
+        String[] acmp_oop_line = lo.getCurrentLine().split("\\s+");
+        Asserts.assertTrue(acmp_oop_line.length >= 3, "Wrong format for oop acmp map line: " + lo.getCurrentLine());
+        for (int i = 3; i < acmp_oop_line.length; i++) {
+          cl.oopAcmpMap.add(Integer.parseInt(acmp_oop_line[i]));
+        }
         lo.moveToNextLine();
       } else {
         cl.isValue = false;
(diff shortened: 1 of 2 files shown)


test/hotspot/jtreg/runtime/valhalla/inlinetypes/field_layout/FieldLayoutAnalyzer.java
```diff
@@ -745,6 +765,79 @@ void checkBufferedLayout() {
...
   }
 
+  void addFlatFieldToBitMap(ClassLayout cl, boolean[] bitmap, int offset, LayoutKind layoutKind) {
+    if (cl.superName != null) {
+      ClassLayout superLayout = getClassLayout(cl.superName);
+      addClassToBitMap(superLayout, bitmap, offset);
+    }
+    int fieldOffset = offset - cl.firstFieldOffset;
+    for (FieldBlock fb : cl.nonStaticFields) {
+      if (fb.isRegular() && fb.name().compareTo("\".empty\"") != 0) {
```
Arraying: Nit: use `.equals` on the String.

test/hotspot/jtreg/runtime/valhalla/inlinetypes/field_layout/FieldLayoutAnalyzer.java
```diff
@@ -745,6 +765,79 @@ void checkBufferedLayout() {
...
+    if (layout.superName != null) {
+      ClassLayout superLayout = getClassLayout(layout.superName);
+      addClassToBitMap(superLayout, bitmap, baseOffset);
+    }
+    if (baseOffset != 0) {
+      baseOffset = baseOffset - layout.firstFieldOffset;
+    }
+    for (FieldBlock block : layout.nonStaticFields) {
+      if (block.isRegular() && block.name().compareTo("\".empty\"") != 0) {
+        for (int i =  baseOffset + block.offset(); i <  baseOffset +block.offset() + block.size(); i++) {
```
Arraying: Nit: double space after =

test/hotspot/jtreg/runtime/valhalla/inlinetypes/field_layout/FieldLayoutAnalyzer.java
```diff
@@ -375,9 +379,25 @@ static ClassLayout parseClassLayout(LogOutput lo) {
         } else {
           cl.nullMarkerOffset = -1;
         }
+        // Non-oop acmp map <offset,size>: <12,1> <16,4> ...
         Asserts.assertTrue(lo.getCurrentLine().startsWith("Non-oop acmp map <offset,size>"), lo.getCurrentLine());
+        String[] acmp_nonoop_line = lo.getCurrentLine().split("\\s+");
+        Asserts.assertTrue(acmp_nonoop_line.length >= 4, "Wrong format for non-oop acmp map line: " + lo.getCurrentLine());
+        for (int i = 4; i < acmp_nonoop_line.length; i++) {
```
Arraying: Maybe more a matter of personal style, but I think this may too sensitive to changes in HotSpot.  I would suggest to do the following:
```java
String searchingFor = "Non-oop acmp map <offset,size>";
// assert current line contains searchingFor
String toParse = lo.getCurrentLine().substring(searchingFor.length);
String[] acmpNoopLine = toParse.split("\\s+");
// assert there is at least one
for (String tuple : acmpNoopLine) // ...
```
The rationale being if I update this message, change indentation, etc. I don't want to spend too much time changing the index numbers everywhere in this test.

test/hotspot/jtreg/runtime/valhalla/inlinetypes/field_layout/FieldLayoutAnalyzer.java
```diff
@@ -375,9 +379,25 @@ static ClassLayout parseClassLayout(LogOutput lo) {
         } else {
           cl.nullMarkerOffset = -1;
         }
+        // Non-oop acmp map <offset,size>: <12,1> <16,4> ...
         Asserts.assertTrue(lo.getCurrentLine().startsWith("Non-oop acmp map <offset,size>"), lo.getCurrentLine());
+        String[] acmp_nonoop_line = lo.getCurrentLine().split("\\s+");
```
Arraying: Nit: lowerCamelCase in Java.

Arraying: May be worth doing a `.trim()` before splitting by whitespace, it's good practice so we don't end up with any surprises, especially since our output has trailing spaces at the moment.

test/hotspot/jtreg/runtime/valhalla/inlinetypes/field_layout/FieldLayoutAnalyzer.java
```diff
@@ -375,9 +379,25 @@ static ClassLayout parseClassLayout(LogOutput lo) {
...
+          int offset = Integer.parseInt(offset_size[0].substring(1)); // skipping '<'
+          int size = Integer.parseInt(offset_size[1].substring(0, offset_size[1].length() - 1)); // skipping '>'
+          cl.nonOopAcmpMap.add(new AcmpMapSegment(offset, size));
+        }
         lo.moveToNextLine();
+        // oop acmp map: 12 16 ...
         Asserts.assertTrue(lo.getCurrentLine().startsWith("oop acmp map"), lo.getCurrentLine());
+        String[] acmp_oop_line = lo.getCurrentLine().split("\\s+");
+        Asserts.assertTrue(acmp_oop_line.length >= 3, "Wrong format for oop acmp map line: " + lo.getCurrentLine());
+        for (int i = 3; i < acmp_oop_line.length; i++) {
```
Arraying: Same comments as with the non-OOP map apply here.

test/hotspot/jtreg/runtime/valhalla/inlinetypes/field_layout/FieldLayoutAnalyzer.java
Arraying: Nit: prefer `.equals` rather than comparison.

test/hotspot/jtreg/runtime/valhalla/inlinetypes/field_layout/FieldLayoutAnalyzer.java
```diff
@@ -745,6 +765,79 @@ void checkBufferedLayout() {
...
+        }
+      } else if (fb.isFlat()) {
+        ClassLayout fcl = getClassLayout(fb.fieldClass);
+        addFlatFieldToBitMap(fcl, bitmap, fieldOffset + fb.offset(), fb.layoutKind);
+      }
+    }
+    if (layoutKind.hasNullMarker()) {
+      Asserts.assertTrue(cl.hasNullMarker());
+      int nullMarkerOffset = fieldOffset + cl.nullMarkerOffset;
+      Asserts.assertFalse(bitmap[nullMarkerOffset], "Overlap in null marker at offset " + nullMarkerOffset + " of class " + cl.name);
```
Arraying, line 785: If we are in a nullable layout, concrete, and have no fields, the null marker is repurposed as `.empty` IIRC. In such a case, what does it get printed as? My concern is that we will see the null marker as `.empty` above, mark it, and then fail this assertion.

test/hotspot/jtreg/runtime/valhalla/inlinetypes/field_layout/FieldLayoutAnalyzer.java
```diff
@@ -745,6 +765,79 @@ void checkBufferedLayout() {
...
+        addFlatFieldToBitMap(cl, bitmap, baseOffset + block.offset(), block.layoutKind);
+      }
+    }
+  }
+
+  void checkAcmpMap() {
+    for (ClassLayout layout : layouts) {
+      if (!layout.isValue) continue;
+      System.out.println("Checking acmp map of class " + layout.name);
+      boolean[] acmpBitMap = new boolean[layout.instanceSize];
```
Arraying, line 815: If I understand the format properly, it could be a nice addition to verify that the leftover `false` regions in the bitmap correspond to the padding `BlockType`.

test/hotspot/jtreg/runtime/valhalla/inlinetypes/field_layout/FieldLayoutAnalyzer.java
```diff
@@ -753,6 +846,26 @@ public void check() {
...
+    // Abstract value class don't have a first field offset in the log, and this information is needed
+    // in the verification of acmp maps. The log doesn't contain the information needed to identify
+    // abstract value classes, so the value is computed for all classes with an uninitialized fieldFieldOffset field
+    for (ClassLayout layout : layouts) {
+      if (layout.firstFieldOffset == 0) {
+        int firstOffset = 0;
+        for (FieldBlock block : layout.nonStaticFields) {
+          if (block.isRegular() || block.isFlat() || block.type() == BlockType.INHERITED) {
+            if (block.offset() < firstOffset || firstOffset == 0) {
+              firstOffset = block.offset();
```
Arraying, line 859: Nit: break loop.

Arraying (commented):
Thanks for this change. This is a valuable thing to have in our test, and I think the logic of using a bitmap to check for overlaps makes a lot of sense. I've left some code feedback and potential future ideas.

test/hotspot/jtreg/runtime/valhalla/inlinetypes/field_layout/FieldLayoutAnalyzer.java
fparain: The synthetic `.empty` field is filtered out when constructing the bitmap from fields information (lines 775 and 802), so its slot is available for the null-marker.

test/hotspot/jtreg/runtime/valhalla/inlinetypes/field_layout/FieldLayoutAnalyzer.java
fparain: `false` regions could be padding, or empty, or reserved, in fact all block types that are not real basic types or the null marker.

test/hotspot/jtreg/runtime/valhalla/inlinetypes/field_layout/FieldLayoutAnalyzer.java
fparain: The expression used to split the line takes care of removing consecutive whitespaces and trailing whitespaces.

test/hotspot/jtreg/runtime/valhalla/inlinetypes/field_layout/FieldLayoutAnalyzer.java
fparain: I don't like the format of the log nor the way it is parsed (I have to blame myself for that).
The whole system could be reworked to have a log format that is both readable and easily parsable, combined with a real parser in the FieldLayoutAnalyzer. But this is beyond the scope of this PR.

test/hotspot/jtreg/runtime/valhalla/inlinetypes/field_layout/FieldLayoutAnalyzer.java
Arraying: Good point. 

Right now, we are testing that what we get from the acmp map is correct (i.e., fields do not overlap). I think it would be valuable to test (perhaps a follow-up) from the opposite angle as well: is everything that's _not_ in the acmp map accounted for? Are there any unexpected leftover areas?

For instance, let's say that we introduce a bug which forgets to include a field in the acmp map. We end up with a bitmap of e.g. length 5, and the (faulty) acmp map says there are fields in indices 1, 2, and 4. After the test is run, the bitmap looks as follows:
```
[F | T | T | F | T]
```
We know that index 1 corresponds to padding, so we mark that as T as well. We have no empty, reserved, etc. blocks. The final bitmap looks as follows:
```
[T | T | T | F | T]
```
At this point, the bitmap _should_ only be truthy. If we add an assertion, that will fail. It turns out, index 3 corresponds to the region that got skipped during the (faulty) acmp map creation. By looking at what's left, we ensure nothing is left out by accident.

Does any of that make sense? 😅

fparain: I don't really understand your point. The acmp_map must contain all the segments that are relevant for a substitutability test, and only the segments that are relevant, so padding areas must not be in the map. The new test reconstructs a map of the relevant segments using the field information, and then compare this reconstructed map with the VM generated map. At line 834, the test checks that at each index, the two maps are consistent, they must agree that either it is a relevant area or it is not. If the VM generated acmp_map is missing a field area, this field area will show up in the reconstructed map, and the test will detect the mismatch.

Arraying: > so padding areas must not be in the map.

Not in the acmp map, but I was under the impression that `FieldLayoutAnalyzer` has this information. What I'm suggesting combines existing generic layout tests with the new acmp bitmap test.

> this field area will show up in the reconstructed map

That's a good point (or maybe my example was just bad). 

To summarize, I'm suggesting something along the lines of:
* Get the field layout of value class `X`. Save all of the `FieldBlock`s as `blocks`. 
* Create a bitmap that spans the whole field layout region (I think we already do this right now).
* Get the acmp maps, set the bitmap indices to `true` and ensure no collisions (we are doing this already).
* Iterate through every block `b` in `blocks`:
    * If we haven't seen `b`'s type yet in the acmp maps (e.g., we already marked the null marker, so skip that:
        * Ensure the corresponding bitmap index is `false` and then set to `true`.
* Iterate through the bitmap at every index `i`:
    * Ensure that the value at index `i` is `true`. If not, we've "lost" something in some part of our bookkeeping.
    
I hope that explains the idea a bit better. Thoughts?

Arraying (approved):
Thanks!

test/hotspot/jtreg/runtime/valhalla/inlinetypes/field_layout/FieldLayoutAnalyzer.java
```diff
@@ -753,6 +846,27 @@ public void check() {
     checkSubClasses();
     checkNullMarkers();
     checkBufferedLayout();
+    checkAcmpMap();
+  }
+
+  void postPrecessing() {
```
matias9927: I think this is meant to be `void postProcessing()`

test/hotspot/jtreg/runtime/valhalla/inlinetypes/field_layout/FieldLayoutAnalyzer.java
```diff
@@ -25,15 +25,12 @@
...
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 
-import javax.management.RuntimeErrorException;
 
 import jdk.test.lib.Asserts;
-import jdk.test.lib.process.OutputAnalyzer;
-import jdk.test.lib.process.ProcessTools;
+
```
matias9927, line 31: The whitespace with the imports should be cleaned up

matias9927 (changes requested):
Change looks good! Just some cosmetic comments

test/hotspot/jtreg/runtime/valhalla/inlinetypes/field_layout/FieldLayoutAnalyzer.java
fparain: As we discussed off-line, this is a different kind of verification unrelated to acmp_maps. This should be addressed in a different PR.

matias9927 (approved):
LGTM!

fparain:
Thanks @Arraying @matias9927 for the reviews.
/integrate

PR #2314, opened 2026-04-09
8379694: [lworld] Add value class test coverage to java.util.Arrays

bwhuang-us:

**Tests Updated**

**TEST.groups**
Add java/util/Arrays to valhalla_adopted, bringing the Arrays test directory into the Valhalla-adopted test group.

**AsList.java**
Add value-oriented object-array datasets: an @AsValueClass record, boxed Integer, LocalDate, and Optional. This extends the existing iterator contract test to ensure Arrays.asList() behaves the same for value-capable elements, including assertSame on iteration.

**ArrayObjectMethods.java**
Add an @AsValueClass record and extend the random object/nested-object generators so value objects and value-object arrays participate in the existing toString, deepToString, hashCode, deepHashCode, and deepEquals coverage. This broadens mixed-content object-array coverage rather than adding a separate value-only test path.

**ArraysEqCmpTest.java**
Add an @AsValueClass Point type and a corresponding ArrayType<ValuePoints> so the existing Arrays.equals, Arrays.compare, and Arrays.mismatch matrix now runs on value-object arrays too. The object-array same-element tests were also generalized to use the ArrayType abstraction instead of assuming Integer[].

**Big.java**
Add a large-array Point[] path using an @AsValueClass record. This extends the existing huge-array binarySearch and range-sort coverage to value-object arrays.

**CopyMethods.java**
Add Point, LocalDate, and Optional to the reflective copy/copyOfRange test matrix, plus the coercion and cloner wiring needed for those types. This extends the existing generic copy-method framework to value-capable reference types.

**Correct.java**
Add an @AsValueClass Point type plus default-sort and comparator-sort subrange tests for Point[], using the existing “compare against reference sort” approach.

**Fill.java**
Add an @AsValueClass array target and verify existing exception behavior when filling it with the wrong type or with an out-of-bounds range.

**HashCode.java**
Add an @AsValueClass Point test that checks Arrays.hashCode(Point[]) is equal for equal value-object arrays and stable across repeated calls.

**SetAllTest.java**
Add Point[] coverage for Arrays.setAll and Arrays.parallelSetAll, including normal generation cases and the existing null-argument exception checks.

**Sorting.java**
Add an @AsValueClass Point sort path to the main sorting test suite. This extends the existing object-sort coverage to value objects for both natural ordering and comparator ordering.

**StreamAndSpliterator.java**
Extend the existing null-array and bad-range exception tests to additional reference-array types: Integer[], Long[], Optional[], and LocalDate[]. This broadens object-array coverage for Arrays.stream, Arrays.spliterator, and Spliterators.spliterator.

**largeMemory/ParallelPrefix.java**
Add an @AsValueClass VInt type and extend the existing parallelPrefix framework to cover value-object arrays, including normal prefix computation plus NPE/IAE/AIOOBE checks.

**Tests With No Change**

**FloatDoubleOrder.java**
No change -- primitive-only float/double ordering test.

**SortingIntBenchmarkTestJMH.java, SortingLongBenchmarkTestJMH.java**
No change -- primitive-array JMH benchmarks.

**SortingNearlySortedPrimitive.java**
No change -- primitive-only nearly-sorted coverage.

**TimSortStackSize.java, TimSortStackSize2.java**
No change -- TimSort stack-shape regression tests over existing Integer[] inputs.

---------
### Progress

### Issue
 * [JDK-8379694](https://bugs.openjdk.org/browse/JDK-8379694): [lworld] Add value class test coverage to java.util.Arrays (**Task** - P4)

### Reviewers
 * [Chen Liang](https://openjdk.org/census#liach) (@liach - Committer) ⚠️ Review applies to [2795a743](https://git.openjdk.org/valhalla/pull/2314/files/2795a743e406fb2470ad03000b5ac3797b963656)
 * [Leonid Mesnik](https://openjdk.org/census#lmesnik) (@lmesnik - Committer)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2314/head:pull/2314` \
`$ git checkout pull/2314`

`$ git checkout pull/2314` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2314/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2314`

`$ git pr show -t 2314`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2314.diff">https://git.openjdk.org/valhalla/pull/2314.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2314#issuecomment-4226993188)
</details>

### DIFF
--- a/test/jdk/java/util/Arrays/Correct.java
+++ b/test/jdk/java/util/Arrays/Correct.java
@@ -26,10 +26,13 @@
  * @bug 4726380 8037097
  * @summary Check that different sorts give equivalent results.
  * @key randomness
+ * @library /test/lib
  * @run junit Correct
  */
 
 import java.util.*;
+
+import jdk.test.lib.valueclass.AsValueClass;
 
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
@@ -44,6 +47,26 @@ public class Correct {
     static final int ITERATIONS = 1000;
     static final int TEST_SIZE = 1000;
 
+    @AsValueClass
+    record Point(int x, int y) implements Comparable<Point> {
+        public int compareTo(Point o) {
+            int c = Integer.compare(x, o.x);
+            return c != 0 ? c : Integer.compare(y, o.y);
+        }
+    }
+
+    @Test
+    public void testDefaultSortPoint() {
+        for (int i=0; i<ITERATIONS; i++) {
+            int size = rnd.nextInt(TEST_SIZE) + 1;
+            Point[] array1 = getPointArray(size);
+            Point[] array2 = Arrays.copyOf(array1, array1.length);
+            Arrays.sort(array1, array1.length/3, array1.length/2);
+            stupidSort(array2, array2.length/3, array2.length/2);
+            Assertions.assertArrayEquals(array2, array1, "Arrays did not match. size=" + size);
+        }
+    }
+
     @Test
     public void testDefaultSort() {
         for (int i=0; i<ITERATIONS; i++) {
@@ -54,6 +77,27 @@ public class Correct {
             stupidSort(array2, array2.length/3, array2.length/2);
             Assertions.assertArrayEquals(array2, array1, "Arrays did not match. size=" + size);
         }
+    }
+
+    @ParameterizedTest
+    @MethodSource("pointComparators")
+    public void testComparatorSortPoint(Comparator<Point> comparator) {
+        for (int i=0; i<ITERATIONS; i++) {
+            int size = rnd.nextInt(TEST_SIZE) + 1;
+            Point[] array1 = getPointArray(size);
+            Point[] array2 = Arrays.copyOf(array1, array1.length);
+            Arrays.sort(array1, array1.length/3, array1.length/2, comparator);
+            stupidSort(array2, array2.length/3, array2.length/2, comparator);
+            Assertions.assertArrayEquals(array2, array1, "Arrays did not match. size=" + size);
+        }
+    }
+
+    static Point[] getPointArray(int size) {
+        Point[] blah = new Point[size];
+        for (int x=0; x<size; x++) {
+            blah[x] = new Point(rnd.nextInt(), rnd.nextInt());
+        }
+        return blah;
     }
 
     @ParameterizedTest
@@ -75,6 +119,44 @@ public class Correct {
             blah[x] = new Integer(rnd.nextInt());
         }
         return blah;
+    }
+
+    static void stupidSort(Point[] a1, int from, int to) {
+        if (from > to - 1)
+            return;
+
+        for (int x=from; x<to; x++) {
+            Point lowest = a1[x];
+            int lowestIndex = x;
+            for (int y=x + 1; y<to; y++) {
+                if (a1[y].compareTo(lowest) < 0) {
+                    lowest = a1[y];
+                    lowestIndex = y;
+                }
+            }
+            if (lowestIndex != x) {
+                swap(a1, x, lowestIndex);
+            }
+        }
+    }
+
+    static void stupidSort(Point[] a1, int from, int to, Comparator<Point> comparator) {
+        if (from > to - 1)
+            return;
+
+        for (int x=from; x<to; x++) {
+            Point lowest = a1[x];
+            int lowestIndex = x;
+            for (int y=x + 1; y<to; y++) {
+                if (comparator.compare(a1[y], lowest) < 0) {
+                    lowest = a1[y];
+                    lowestIndex = y;
+                }
+            }
+            if (lowestIndex != x) {
+                swap(a1, x, lowestIndex);
+            }
+        }
     }
 
     static void stupidSort(Integer[] a1, int from, int to) {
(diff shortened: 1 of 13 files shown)


test/jdk/java/util/Arrays/Correct.java
```diff
@@ -77,6 +121,44 @@ static Integer[] getIntegerArray(int size) {
...
+                    lowestIndex = y;
+                }
+            }
+            if (lowestIndex != x) {
+                swap(a1, x, lowestIndex);
+            }
+        }
+    }
+
     static void stupidSort(Integer[] a1, int from, int to) {
```
liach: I recommend changing the static types of Integer[] to Comparable[] instead of introducing identical copies of stupid sort. Same for the comparator version, but the comparator version can take Object[]

test/jdk/java/util/Arrays/HashCode.java
```diff
@@ -167,8 +173,32 @@ public static void main(String[] args) throws Exception {
...
+    private static void testValueClassArrayHashCode() {
+        Point[] a = {new Point(1, 2), new Point(3, 4)};
+        Point[] b = {new Point(1, 2), new Point(3, 4)}; // distinct instances, same fields
+        int hashA = Arrays.hashCode(a);
+        int hashB = Arrays.hashCode(b);
+        if (hashA != hashB) {
+            throw new RuntimeException("testValueClassArrayHashCode: expected equal hash codes " +
+                "for equal value class arrays, got " + hashA + " and " + hashB);
+        }
+        // Consistency with repeated calls (no identity hash interference)
```
liach: There's nothing relevant to identity hash here, dead comment?

test/jdk/java/util/Arrays/HashCode.java
```diff
@@ -24,14 +24,20 @@
...
  */
 
 import java.lang.reflect.Method;
 import java.util.Arrays;
+import jdk.test.lib.valueclass.AsValueClass;
 
 public class HashCode {
+
+    @AsValueClass
+    record Point(int x, int y) {}
```
liach: You might consider make its hashCode just return `Objects.identityHashCode` if preview is enabled for better coverage?
Like:
```
if (Point.class.isValue()) {
    return Objects.identityHashCode(this);
}
```

test/jdk/java/util/Arrays/SetAllTest.java
```diff
@@ -164,6 +186,18 @@ public void testSetAllDouble(String name, int size, IntToDoubleFunction generato
         assertDoubleArrayEquals(result, expected, 0.05, "setAll(double[], IntToDoubleFunction) case " + name + " failed.");
     }
 
+    @ParameterizedTest
+    @MethodSource("pointTests")
+    public void testSetAllPoint(String name, int size, IntFunction<Point> generator, Point[] expected) {
+        Point[] result = new Point[size];
+        Arrays.setAll(result, generator);
+        Assertions.assertArrayEquals(expected, result, "setAll(Point[], IntFunction<Point>) case " + name + " failed.");
```
liach, line 195: The message is redundant because junit already does it but since this is a copycat guess it's fine.

test/jdk/java/util/Arrays/SetAllTest.java
```diff
@@ -279,4 +313,34 @@ public void testDoubleSetNulls() {
...
     }
+
+    @Test
+    public void testPointSetNulls() {
+        Point[] ar = new Point[2];
+
+        try {
+            Arrays.setAll((Point[]) null, (IntFunction<Point>) i -> new Point(i, i));
+            fail("Arrays.setAll(null, foo) should throw NPE");
+        } catch (NullPointerException npe) {
```
liach: Let's do `assertThrows(NullPointerException.class, () -> ...)` instead.

test/jdk/java/util/Arrays/HashCode.java
bwhuang-us: I will change the Point record to a class, as value classes fall back to Object.hashCode() which is computed via System.identityHashCode(), per JEP 401.

test/jdk/java/util/Arrays/HashCode.java
bwhuang-us: This comment is for the repeated call on hashCode() below. It doesn't specifically demonstrate anything about identity hash behavior. I will reword it to "Repeated calls should produce the same array hash code"

test/jdk/java/util/Arrays/SetAllTest.java
bwhuang-us: Are you suggesting to completely remove the assertion message or keep the "setAll" and "parallelSetAll" cases to indicate the operations?

liach: You can keep `setAll`/`parallelSetAll`. Even though they can be deduced from line numbers, that is not quite straightforward.

JUnit when reporting the failure already includes the `name` argument passed for `ParameterizedTest`, or you can make the `@ParameterizedTest(name = "{0}")` so it substitutes the name arg in its parameterized name.

liach (approved):
Lgtm, the error message beautification is trivial.

test/jdk/java/util/Arrays/SetAllTest.java
bwhuang-us: That makes sense. I tried using JUnit's parameterized name to improve the signal a bit, which avoids the noisy lamdba toString().
`@ParameterizedTest(name = "{0}, size={1}, data={3}")`
This gives output like: 
SetAllTest::testSetAllDouble '[3] ten, 10, [0.0, 1.1, 2.2, ...]'

lmesnik (approved):

PR #2325, opened 2026-04-16
8380525: [lworld] C2: Various issues with abstract value classes leading to bad casts with is_inlinetype()

chhagedorn:
This patch fixes various issues found with abstract value classes where the assumption was that we have a concrete value class but we actually could have an abstract value class as well.

The main issue found is that a concrete value class is represented with an `InlineKlass` while an abstract value class is represented with an `InstanceKlass`. Therefore, we cannot cast the latter to `InlineKlass`. While working on this, I checked all `is_inlinetype()` uses and found more issues. I also found issues around the `Arrays.copyOf/copyOfRange` intrinsic but left it as follow-up work and filed [JDK-8382226](https://bugs.openjdk.org/browse/JDK-8382226) for it. 

Here is a list of fixes included in this patch, together with corresponding tests:
1. `GraphKit::get_layout_helper()`: Wrong old assumption that nullable implies not-flat.
2. Checking whether we have a load/store from/to a field of an empty value class:
a. C1: `GraphBuilder::access_field()`, for static and instance field.
b. C2: `Parse::do_get_xxx()`, for loading
c. C2: `Parse::do_set_xxx()`, for storing
3. `arraycopy` instrinsic: Not bailing out when src or dest is a flat abstract value class array - in this case we don't know the layout. We currently hit an assert due to an invalid cast, assuming concrete value classes when an array is flat.

Additional changes:
- Added some more asserts
- Some small code cleanups

Thanks,
Christian

---------

---------
### Progress

### Issue
 * [JDK-8380525](https://bugs.openjdk.org/browse/JDK-8380525): [lworld] C2: Various issues with abstract value classes leading to bad casts with is_inlinetype() (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2325/head:pull/2325` \
`$ git checkout pull/2325`

`$ git checkout pull/2325` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2325/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2325`

`$ git pr show -t 2325`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2325.diff">https://git.openjdk.org/valhalla/pull/2325.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2325#issuecomment-4259550821)
</details>

### DIFF
--- a/src/hotspot/share/c1/c1_GraphBuilder.cpp
+++ b/src/hotspot/share/c1/c1_GraphBuilder.cpp
@@ -1899,7 +1899,7 @@ void GraphBuilder::access_field(Bytecodes::Code code) {
         null_check(val);
 
         ciType* field_type = field->type();
-        if (field_type->is_loaded() && field_type->is_inlinetype() && field->type()->as_inline_klass()->is_empty() &&
+        if (field_type->is_loaded() && field_type->is_inlinetype() && field_type->as_inline_klass()->is_empty() &&
             (!method()->is_class_initializer() || field->is_flat())) {
           // Storing to a field of an empty, null-free inline type that is already initialized. Ignore.
           break;
@@ -2095,7 +2095,7 @@ void GraphBuilder::access_field(Bytecodes::Code code) {
 
       ciType* field_type = field->type();
       if (field->is_null_free() && field_type->is_loaded() && field_type->is_inlinetype() &&
-          field->type()->as_inline_klass()->is_empty() && (!method()->is_object_constructor() || field->is_flat())) {
+          field_type->as_inline_klass()->is_empty() && (!method()->is_object_constructor() || field->is_flat())) {
         // Storing to a field of an empty, null-free inline type that is already initialized. Ignore.
         null_check(obj);
         null_check(val);
@@ -2111,7 +2111,7 @@ void GraphBuilder::access_field(Bytecodes::Code code) {
       } else {
         // Flat field
         assert(!needs_patching, "Can't patch flat inline type field access");
-        ciInlineKlass* inline_klass = field->type()->as_inline_klass();
+        ciInlineKlass* inline_klass = field_type->as_inline_klass();
         if (field->is_atomic()) {
           if (field->is_null_free()) {
             null_check(val);
--- a/src/hotspot/share/c1/c1_GraphBuilder.cpp
+++ b/src/hotspot/share/c1/c1_GraphBuilder.cpp
@@ -1825,8 +1825,8 @@ void GraphBuilder::access_field(Bytecodes::Code code) {
   bool will_link;
   ciField* field = stream()->get_field(will_link);
   ciInstanceKlass* holder = field->holder();
-  BasicType field_type = field->type()->basic_type();
-  ValueType* type = as_ValueType(field_type);
+  BasicType field_basic_type = field->type()->basic_type();
+  ValueType* type = as_ValueType(field_basic_type);
 
   // call will_link again to determine if the field is valid.
   const bool needs_patching = !holder->is_loaded() ||
@@ -1891,16 +1891,19 @@ void GraphBuilder::access_field(Bytecodes::Code code) {
       if (state_before == nullptr) {
         state_before = copy_state_for_exception();
       }
-      if (field_type == T_BOOLEAN) {
+      if (field_basic_type == T_BOOLEAN) {
         Value mask = append(new Constant(new IntConstant(1)));
         val = append(new LogicOp(Bytecodes::_iand, val, mask));
       }
       if (field->is_null_free()) {
         null_check(val);
-      }
-      if (field->is_null_free() && field->type()->is_loaded() && field->type()->as_inline_klass()->is_empty() && (!method()->is_class_initializer() || field->is_flat())) {
-        // Storing to a field of an empty, null-free inline type that is already initialized. Ignore.
-        break;
+
+        ciType* field_type = field->type();
+        if (field_type->is_loaded() && field_type->is_inlinetype() && field->type()->as_inline_klass()->is_empty() &&
+            (!method()->is_class_initializer() || field->is_flat())) {
+          // Storing to a field of an empty, null-free inline type that is already initialized. Ignore.
+          break;
+        }
       }
       append(new StoreField(append(obj), offset, field, val, true, state_before, needs_patching));
       break;
@@ -1963,7 +1966,7 @@ void GraphBuilder::access_field(Bytecodes::Code code) {
             assert(replacement->is_linked() || !replacement->can_be_linked(), "should already by linked");
             // Writing an (integer) value to a boolean, byte, char or short field includes an implicit narrowing
             // conversion. Emit an explicit conversion here to get the correct field value after the write.
-            switch (field_type) {
+            switch (field_basic_type) {
             case T_BOOLEAN:
             case T_BYTE:
               replacement = append(new Convert(Bytecodes::_i2b, replacement, type));
@@ -2085,12 +2088,14 @@ void GraphBuilder::access_field(Bytecodes::Code code) {
       if (state_before == nullptr) {
         state_before = copy_state_for_exception();
       }
-      if (field_type == T_BOOLEAN) {
+      if (field_basic_type == T_BOOLEAN) {
         Value mask = append(new Constant(new IntConstant(1)));
         val = append(new LogicOp(Bytecodes::_iand, val, mask));
       }
 
-      if (field->is_null_free() && field->type()->is_loaded() && field->type()->as_inline_klass()->is_empty() && (!method()->is_object_constructor() || field->is_flat())) {
+      ciType* field_type = field->type();
+      if (field->is_null_free() && field_type->is_loaded() && field_type->is_inlinetype() &&
+          field->type()->as_inline_klass()->is_empty() && (!method()->is_object_constructor() || field->is_flat())) {
         // Storing to a field of an empty, null-free inline type that is already initialized. Ignore.
         null_check(obj);
         null_check(val);
--- a/src/hotspot/share/opto/inlinetypenode.cpp
+++ b/src/hotspot/share/opto/inlinetypenode.cpp
@@ -595,9 +595,7 @@ void InlineTypeNode::store(GraphKit* kit, Node* base, Node* ptr, bool immutable_
     Node* field_ptr = kit->basic_plus_adr(base, ptr, field_off);
     if (field->is_flat()) {
       // Recursively store the flat inline type field
-      ciInlineKlass* fvk = ft->as_inline_klass();
       bool atomic = field->is_atomic();
-
       field_val->as_InlineType()->store_flat(kit, base, field_ptr, atomic, immutable_memory, field_null_free, decorators);
     } else {
       // Store field value to memory
--- a/src/hotspot/share/opto/macroArrayCopy.cpp
+++ b/src/hotspot/share/opto/macroArrayCopy.cpp
@@ -1550,8 +1550,6 @@ void PhaseMacroExpand::expand_arraycopy_node(ArrayCopyNode *ac) {
 
   RegionNode* slow_region = new RegionNode(1);
   transform_later(slow_region);
-  const bool both_flat = top_src->is_flat() && top_dest->is_flat();
-  const bool both_exact = top_src->klass_is_exact() && top_dest->klass_is_exact();
   const bool same_nullness = top_src->is_null_free() == top_dest->is_null_free();
 
   if (!ac->is_arraycopy_validated()) {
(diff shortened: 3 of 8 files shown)


src/hotspot/share/c1/c1_GraphBuilder.cpp
```diff
@@ -1891,16 +1891,19 @@ void GraphBuilder::access_field(Bytecodes::Code code) {
...
       }
-      if (field_type == T_BOOLEAN) {
+      if (field_basic_type == T_BOOLEAN) {
         Value mask = append(new Constant(new IntConstant(1)));
         val = append(new LogicOp(Bytecodes::_iand, val, mask));
       }
       if (field->is_null_free()) {
         null_check(val);
-      }
-      if (field->is_null_free() && field->type()->is_loaded() && field->type()->as_inline_klass()->is_empty() && (!method()->is_class_initializer() || field->is_flat())) {
```
chhagedorn, line 1901: Also merged together under `is_null_free()`

src/hotspot/share/opto/inlinetypenode.cpp
```diff
@@ -595,9 +595,7 @@ void InlineTypeNode::store(GraphKit* kit, Node* base, Node* ptr, bool immutable_
     Node* field_ptr = kit->basic_plus_adr(base, ptr, field_off);
     if (field->is_flat()) {
       // Recursively store the flat inline type field
-      ciInlineKlass* fvk = ft->as_inline_klass();
       bool atomic = field->is_atomic();
-
```
chhagedorn, line 600: Unused

src/hotspot/share/opto/macroArrayCopy.cpp
```diff
@@ -1541,6 +1550,7 @@ void PhaseMacroExpand::expand_arraycopy_node(ArrayCopyNode *ac) {
 
   RegionNode* slow_region = new RegionNode(1);
   transform_later(slow_region);
+  const bool same_nullness = top_src->is_null_free() == top_dest->is_null_free();
```
chhagedorn, line 1561: Just some code sharing and simplification with code below

src/hotspot/share/opto/macroArrayCopy.cpp
```diff
@@ -1491,19 +1492,27 @@ void PhaseMacroExpand::expand_arraycopy_node(ArrayCopyNode *ac) {
...
   // Figure out the size and type of the elements we will be copying.
   //
-  // We have no stub to copy flat inline type arrays with oop
-  // fields if we need to emit write barriers.
   //
+  // Go to the slow path but avoid the native method wrapper to JVM_ArrayCopy when:
+  // 1) The component types are not the same or void.
+  // 2) src and dest do not have the same flatness.
+  // 3) src or dest is a flat abstract value class array (we don't know the exact layout and cannot use the stub).
+  // 4) dest is a flat value class array with oop fields and requires to emit write barriers (we don't have such a stub).
```
merykitty: This should be `src or dest` similar to other comments, because loading an oop may require load barrier, too.

merykitty: Besides, should it be that `src` and `dest` have the same `layout_kind`, can we arraycopy a nullable array to a null-free array, or an atomic to a non-atomic array?

src/hotspot/share/opto/parse3.cpp
```diff
@@ -154,16 +154,19 @@ void Parse::do_get_xxx(Node* obj, ciField* field) {
...
+      // Loading from a field of an empty inline type. Just return the default instance.
+      ld = InlineTypeNode::make_all_zero(_gvn, vk);
+    } else if (field->is_flat()) {
+      // Loading from a flat inline type field.
+      bool is_immutable = field->is_final() && field->is_strict();
+      bool atomic = field->is_atomic();
+      ld = InlineTypeNode::make_from_flat(this, vk, obj, adr, atomic, is_immutable, field->is_null_free(), IN_HEAP | MO_UNORDERED);
+    }
+  }
+  if (ld == nullptr) {
```
merykitty, line 169: Probably add an assert that `field` is not flat here.

src/hotspot/share/opto/parse3.cpp
```diff
@@ -273,22 +276,29 @@ void Parse::do_put_xxx(Node* obj, ciField* field, bool is_field) {
...
+      inc_sp(1);
+      bool is_immutable = field->is_final() && field->is_strict();
+      bool atomic = field->is_atomic();
+      val->as_InlineType()->store_flat(this, obj, adr, atomic, is_immutable, field->is_null_free(), IN_HEAP | MO_UNORDERED);
+      dec_sp(1);
+      do_store = false;
+    }
+  }
+  if (do_store) {
     // Store the value.
```
merykitty, line 303: Similarly, assert that `field` is not flat.

TobiHartmann (approved):
The abstract value classes keep biting us - thanks for thoroughly checking all occurrences of  `is_inlinetype()`! The fixes look good to me (minus what Quan Anh already mentioned :slightly_smiling_face: ). Nice tests.

src/hotspot/share/c1/c1_GraphBuilder.cpp
```diff
@@ -1891,16 +1891,19 @@ void GraphBuilder::access_field(Bytecodes::Code code) {
...
-        // Storing to a field of an empty, null-free inline type that is already initialized. Ignore.
-        break;
+
+        ciType* field_type = field->type();
+        if (field_type->is_loaded() && field_type->is_inlinetype() && field->type()->as_inline_klass()->is_empty() &&
```
dean-long:
```suggestion
        if (field_type->is_loaded() && field_type->is_inlinetype() && field_type->as_inline_klass()->is_empty() &&
```

src/hotspot/share/c1/c1_GraphBuilder.cpp
```diff
@@ -2085,12 +2088,14 @@ void GraphBuilder::access_field(Bytecodes::Code code) {
...
-      if (field_type == T_BOOLEAN) {
+      if (field_basic_type == T_BOOLEAN) {
         Value mask = append(new Constant(new IntConstant(1)));
         val = append(new LogicOp(Bytecodes::_iand, val, mask));
       }
 
-      if (field->is_null_free() && field->type()->is_loaded() && field->type()->as_inline_klass()->is_empty() && (!method()->is_object_constructor() || field->is_flat())) {
+      ciType* field_type = field->type();
+      if (field->is_null_free() && field_type->is_loaded() && field_type->is_inlinetype() &&
+          field->type()->as_inline_klass()->is_empty() && (!method()->is_object_constructor() || field->is_flat())) {
```
dean-long:
```suggestion
          field_type->as_inline_klass()->is_empty() && (!method()->is_object_constructor() || field->is_flat())) {
```

src/hotspot/share/opto/macroArrayCopy.cpp
```diff
@@ -1491,19 +1492,27 @@ void PhaseMacroExpand::expand_arraycopy_node(ArrayCopyNode *ac) {
...
-       bs->array_copy_requires_gc_barriers(alloc != nullptr, T_OBJECT, false, false, BarrierSetC2::Optimization))) {
-    // The component types are not the same or are not recognized.  Punt.
-    // (But, avoid the native method wrapper to JVM_ArrayCopy.)
-    {
-      Node* mem = ac->in(TypeFunc::Memory);
-      merge_mem = generate_slow_arraycopy(ac, &ctrl, mem, &io, TypePtr::BOTTOM, src, src_offset, dest, dest_offset, length, false);
-    }
+  bool go_to_slow_path = src_elem != dest_elem || top_src->is_flat() != top_dest->is_flat() || dest_elem == T_VOID;
+  if ((top_src->is_flat() && !top_src->elem()->is_inlinetypeptr()) ||
+      (top_dest->is_flat() && !top_dest->elem()->is_inlinetypeptr())) {
```
dean-long: Doesn't an `is_flat` array imply concrete elements here?  I didn't think we could create a flat array with abstract elements.  For example, ObjArrayKlass::array_layout_selection checks element->is_abstract().  FlatArrayKlass asserts element_klass->is_inline_klass().  And FlatArrayKlass::element_klass() returns InlineKlass*.

src/hotspot/share/opto/macroArrayCopy.cpp
```diff
@@ -1491,19 +1492,27 @@ void PhaseMacroExpand::expand_arraycopy_node(ArrayCopyNode *ac) {
...
+    go_to_slow_path = true;
+  } else if (top_dest->is_flat() &&
+             bs->array_copy_requires_gc_barriers(alloc != nullptr, T_OBJECT, false, false, BarrierSetC2::Optimization) &&
+             top_dest->elem()->inline_klass()->contains_oops()) {
```
dean-long, line 1514: && short-circuits and I see you swapped the order.  We should do the cheaper operation first.  Is that array_copy_requires_gc_barriers() or elem()->inline_klass()->contains_oops()?

src/hotspot/share/opto/macroArrayCopy.cpp
dean-long, line 1561: Can we move it down closer to its use?

dean-long: Oh I see that it is used both inside and outside if the following `if`.

src/hotspot/share/opto/macroArrayCopy.cpp
```diff
@@ -1591,8 +1602,8 @@ void PhaseMacroExpand::expand_arraycopy_node(ArrayCopyNode *ac) {
   // This is where the memory effects are placed:
   const TypePtr* adr_type = nullptr;
   Node* dest_length = (alloc != nullptr) ? alloc->in(AllocateNode::ALength) : nullptr;
-  if (top_src->is_flat() && top_dest->is_flat() &&
-      top_src->is_null_free() == top_dest->is_null_free()) {
+  assert(top_src->is_flat() == top_dest->is_flat(), "must have bailed out before");
+  if (top_src->is_flat() && same_nullness) {
```
dean-long: This is definitely simpler, but it takes me a couple seconds to see this is the opposite of `!top_src->is_flat() || !same_nullness`.  Since we are guaranteed to need this expression at least once, how about doing something like:
```
const bool same_nullness ...
const bool flat_and_same_nullness = top_src->is_flat() && same_nullness;
```
once, then we could have one `if (!flat_and_same_nullness)` and one `if (flat_and_same_nullness)`.

src/hotspot/share/opto/macroArrayCopy.cpp
chhagedorn: I think `array_copy_requires_gc_barriers()` is cheaper since `contain_oops()` also involves a guarded VM entry.

src/hotspot/share/opto/macroArrayCopy.cpp
chhagedorn: That's a good idea! Updated.

src/hotspot/share/opto/macroArrayCopy.cpp
chhagedorn: At first sight, it looks like it must be implying it. It's true that we cannot directly create a flat abstract value class array. But we could have this:
https://github.com/openjdk/valhalla/blob/8f06d5e939e11431a275ea0d5c171ffe9b3d67b1/test/hotspot/jtreg/compiler/valhalla/inlinetypes/TestArrays.java#L3925-L3932
 
We create a flat array of a concrete value class in each branch and then merge them together. The type system propagates the flatness information and we end up with a flat abstract value class array. The elements are now abstract and represented by an `InstanceKlass` and not an `InlineKlass` anymore. We don't know the exact layout of the elements because it could either be `BadCastV1` or `BadCastV2` or something else if there are even more sub classes of `BadCastA` (the type system only sees `BadCastA`).

src/hotspot/share/opto/macroArrayCopy.cpp
chhagedorn: > This should be `src or dest` similar to other comments, because loading an oop may require load barrier, too.

Good catch, updated. Also refactored the code slightly and added more asserts.

> Besides, should it be that src and dest have the same layout_kind, can we arraycopy a nullable array to a null-free array, or an atomic to a non-atomic array?

That's a good point and I think you're definitely right about atomicity - this does not seem to be handled. In this patch, I only address the abstract value class case and added some comments and small clean-ups to keep the scope smaller. There is already an umbrella bug to fix the remaining issues: [JDK-8251971](https://bugs.openjdk.org/browse/JDK-8251971).

For matching nullness, we handle it for the non-validated cases further down:
https://github.com/openjdk/valhalla/blob/2f943e10e4cab414515b52371bec38f19eda81fe/src/hotspot/share/opto/macroArrayCopy.cpp#L1595-L1599

and for validated cases, we handle it here:
https://github.com/openjdk/valhalla/blob/8f9586645baae882b8a60e3111d00c600cdf1089/src/hotspot/share/opto/library_call.cpp#L6647-L6672

So, I think nullness should be currently handled. I'm not sure if we have enough tests to cover all of this, though.

chhagedorn:
Thanks @TobiHartmann, @merykitty and @dean-long for your reviews!

I've pushed an update and addressed your comments.

merykitty (approved):
Thanks a lot, looks good to me.

dean-long (approved):
Looks good.

TobiHartmann (approved):
Still good.

chhagedorn:
Thanks for the reviews @TobiHartmann, @merykitty, and @dean-long!

/integrate

PR #2326, opened 2026-04-16
8384885: [lworld] Implementation update for value class serialization

liach:
For preview of the serialization implementation cleanup.

See the mainline diff version:
https://gist.github.com/a3440f2b9da5111fe20315e59ed17380

---------

---------
### Progress

### Issue
 * [JDK-8384885](https://bugs.openjdk.org/browse/JDK-8384885): [lworld] Implementation update for value class serialization (**Bug** - P2)

### Reviewers
 * [Alan Bateman](https://openjdk.org/census#alanb) (@AlanBateman - no project role) ⚠️ Review applies to [5b26ade8](https://git.openjdk.org/valhalla/pull/2326/files/5b26ade8f8f68b90fa3848c592ef5172395f2d73)
 * [Dan Smith](https://openjdk.org/census#dlsmith) (@dansmithcode - Committer) ⚠️ Review applies to [fc83d143](https://git.openjdk.org/valhalla/pull/2326/files/fc83d14331aaa50d522722e20c1a3e13d6e1839e)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2326/head:pull/2326` \
`$ git checkout pull/2326`

`$ git checkout pull/2326` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2326/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2326`

`$ git pr show -t 2326`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2326.diff">https://git.openjdk.org/valhalla/pull/2326.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2326#issuecomment-4479763689)
</details>


src/java.base/share/classes/java/io/ObjectInputStream.java
```diff
@@ -2305,13 +2240,13 @@ private Object readExternalObject(ObjectStreamClass desc, boolean unshared)
...
+    private Object readRecordOrValue(ObjectStreamClass desc, boolean isRecord) throws IOException {
+        ObjectStreamClass.ClassDataSlot[] slots = desc.getClassDataLayout();
+        if (slots.length != 1) {
             // skip any superclass stream field values
-            for (int i = 0; i < slots.size()-1; i++) {
-                if (slots.get(i).hasData) {
-                    new FieldValues(slots.get(i).desc, true);
+            for (int i = 0; i < slots.length-1; i++) {
+                if (slots[i].hasData) {
+                    new FieldValues(slots[i].desc, true);
```
viktorklang-ora, line 2250: What's the performance impact of allocating the intermediate array?

liach: The intermediate array is allocated in the mainline ObjectInputStream. I reverted some code changes unrelated to the JEP 401 effort. And this is a package-private method that returns a trusted array.

src/java.base/share/classes/java/io/ObjectInputStream.java
```diff
@@ -2305,13 +2240,13 @@ private Object readExternalObject(ObjectStreamClass desc, boolean unshared)
      * Null is returned from readRecord and later the exception is thrown at
      * the exit of {@link #readObject(Class)}.
      */
-    private Object readRecord(ObjectStreamClass desc, boolean unshared) throws IOException {
-        List<ClassDataSlot> slots = desc.getClassDataLayout();
-        if (slots.size() != 1) {
+    private Object readRecordOrValue(ObjectStreamClass desc, boolean isRecord) throws IOException {
```
viktorklang-ora: My preference would be to introduce a readValue instead of passing around a boolean to indicate whether record or value.

src/java.base/share/classes/java/io/ObjectStreamClass.java
```diff
@@ -260,6 +193,8 @@ InvalidClassException newInvalidClassException() {
     /** session-cache of record deserialization constructor
      * (in de-serialized OSC only), or null */
     private MethodHandle deserializationCtr;
+    /** value deserialization method identified by {@link DeserializeConstructor}.*/
+    private Executable deserializeConstructor;
```
viktorklang-ora: These two fields seems confusingly similarly named, possible to differentiate the names a bit for clarity?

src/java.base/share/classes/java/io/ObjectStreamClass.java
```diff
@@ -188,9 +123,7 @@ protected Map<FieldReflectorKey, FieldReflector> computeValue(Class<?> type) {
     /** true if represents record type */
     private boolean isRecord;
     /** true if represents a value class */
-    private boolean isValue;
-    /** The DeserializationMode for this class. */
-    private DeserializationMode factoryMode;
+    private boolean isValueOrStrictInit;
```
viktorklang-ora: The comment reads one thing but the name of the field tells a slightly different story. If we need to differentiate "value or strict init" in the variable name, then it would make sense to explain in the comment.

src/java.base/share/classes/java/io/ObjectStreamClass.java
viktorklang-ora: Also, can a type have both a `deserializationCtr` AND a `deserializeConstructor` or are they mutually exclusive?

src/java.base/share/classes/java/io/ObjectStreamClass.java
```diff
@@ -932,17 +840,14 @@ boolean isSerializable() {
...
         requireInitialized();
-        return isValue;
+        return isValueOrStrictInit;
     }
 
-    /**
-     * {@return the factory mode for deserialization}
-     */
-    DeserializationMode factoryMode() {
+    boolean isDeserializeConstructor() {
```
viktorklang-ora: This method name reads a bit odd. Is it "useDeserializeConstructor", "hasDeserializeConstructor", "isDeserializeConstructor" or would there exist a better name to reflect its purpose?

src/java.base/share/classes/java/io/ObjectStreamClass.java
```diff
@@ -967,15 +872,13 @@ boolean hasWriteObjectData() {
...
      * non-externalizable and its first non-serializable superclass defines an
-     * accessible no-arg constructor, or if the class is a value class with a @DeserializeConstructor
-     * constructor or static factory.
-     * Otherwise, returns false.
+     * accessible no-arg constructor.  Otherwise, returns false.
      */
     boolean isInstantiable() {
         requireInitialized();
-        return (cons != null || (isValue() && canonicalCtr != null));
+        return (cons != null);
```
viktorklang-ora, line 890:
```suggestion
        return cons != null;
```

src/java.base/share/classes/java/io/ObjectStreamClass.java
```diff
@@ -1430,45 +1337,50 @@ private ObjectStreamClass getVariantFor(Class<?> cl)
      * @return a MethodHandle, null if none found
```
viktorklang-ora: This method no longer returns a MethodHandle, so the javadoc needs updating to reflect that

src/java.base/share/classes/java/io/ObjectStreamClass.java
```diff
@@ -1430,45 +1337,50 @@ private ObjectStreamClass getVariantFor(Class<?> cl)
...
         for (ObjectStreamField field : fields) {
-            int argIndex = field.getArgIndex();
-            final Class<?> paramtype = params.get(argIndex);
-            if (!field.getType().equals(paramtype)) {
+            var i = map.get(field.getName());
+            if (i == null) {
+                return false; // Name mismatch
+            }
+            if (!field.getType().equals(params[i])) {
                 return false;
```
viktorklang-ora:
```suggestion
                return false; // Type mismatch
```

src/java.base/share/classes/java/io/ObjectStreamClass.java
```diff
@@ -1739,9 +1653,9 @@ private static ObjectStreamField[] getDefaultSerialFields(Class<?> cl) {
         ArrayList<ObjectStreamField> list = new ArrayList<>();
         int mask = Modifier.STATIC | Modifier.TRANSIENT;
 
-        for (int i = 0, argIndex = 0; i < clFields.length; i++) {
+        for (int i = 0; i < clFields.length; i++) {
             if ((clFields[i].getModifiers() & mask) == 0) {
-                list.add(new ObjectStreamField(clFields[i], false, true, argIndex++));
+                list.add(new ObjectStreamField(clFields[i], false, true));
             }
         }
```
viktorklang-ora, line 1676: Since we aren't using the index beyond looking at the "current" array index value, we might as well remove that complication:

```suggestion
        for (var clField : clFields.length) {
if ((clField.getModifiers() & mask) == 0) {
                list.add(new ObjectStreamField(clField, false, true));
}
}
```

test/jdk/java/io/Serializable/valueObjects/SimpleValueGraphs.java
```diff
@@ -364,7 +364,7 @@ public String toString(int depth) {
     void testExternalizableNotSer() {
         var obj = new ValueExt();
         var ex = Assertions.assertThrows(InvalidClassException.class, () -> serialize(obj));
-        Assertions.assertEquals("SimpleValueGraphs$ValueExt; Externalizable not valid for value class", ex.getMessage());
+        Assertions.assertEquals("SimpleValueGraphs$ValueExt; cannot serialize value class", ex.getMessage());
```
viktorklang-ora, line 366: Shouldn't this error message say "externalize" rather than "serialize" (based off of the test name)

test/jdk/java/io/Serializable/valueObjects/SimpleValueGraphs.java
```diff
@@ -373,7 +373,7 @@ void testExternalizableNotDeser() throws IOException {
         byte[] bytes = serialize(obj);
         byte[] newBytes = patchBytes(bytes, "IdentExt", "ValueExt");
         var ex = Assertions.assertThrows(InvalidClassException.class, () -> deserialize(newBytes));
-        Assertions.assertTrue(ex.getMessage().contains("Externalizable not valid for value class"));
+        Assertions.assertTrue(ex.getMessage().contains("cannot serialize value class"));
```
viktorklang-ora, line 375: same as https://github.com/openjdk/valhalla/pull/2326/changes#r3355827173

test/jdk/java/io/Serializable/valueObjects/ValueSerializationTest.java
```diff
@@ -103,6 +106,24 @@ public NonSerializablePoint(int x, int y) {
         }
     }
 
+    public static class NonSerializableStrictPoint implements Serializable {
+        static {
+            for (var f : NonSerializableStrictPoint.class.getDeclaredFields()) {
+                assertTrue(f.isStrictInit(), f.getName());
+            }
```
viktorklang-ora, line 113: I think it would be good to also validate that we are indeed seeing the fields (i.e. assert that we see "x" and "y")

src/java.base/share/classes/java/io/ObjectStreamClass.java
liach: I named this `deserializeConstructor` because this is for `@DeserializeConstructor`. Maybe we can rename both the annotation and the field "deserialization factory" or some other better name instead?

src/java.base/share/classes/java/io/ObjectStreamClass.java
liach: Restoration to mainline: https://github.com/openjdk/jdk/blob/ee53e7569e211d74648b8d9ccf3f039888938f2c/src/java.base/share/classes/java/io/ObjectStreamClass.java#L847

viktorklang-ora:
I've had a look over this PR, but I think it would be good to have an expert like @stuart-marks look at it as well.

test/jdk/java/io/Serializable/valueObjects/SimpleValueGraphs.java
liach: I don't think we need distinct serialize/externalize messages for Externalizable: this distinction was arbitrary from an earlier implementation. The true distinction is migrated wrappers versus all other non-writeReplace Serializable value classes.

src/java.base/share/classes/java/io/ObjectStreamClass.java
liach: `i == null` is not straightforward for "name mismatch" case. I should make that comment more specific, like "Name not accounted by the factory" and "Name match, type mismatch"

src/java.base/share/classes/java/io/ObjectStreamClass.java
liach: This is how it is in mainline, I don't think I should introduce noisy diff to the mainline review PR.

viktorklang-ora: Makes sense. I wonder if it would be easier (for future PRs in this space) to do the mainline-alignment as a separate, initial, PR, then do a follow-up PR with the specific changes.

test/jdk/java/io/Serializable/valueObjects/ValueSerializationTest.java
liach: That would be an assertion on the `getDeclaredFields` facility. I am more concerned about this annotation processing because the `@driver` clause is more likely to be accidentally omitted or have typos in test directives and becomes hard to debug. There is no such problem with `Class.getDeclaredFields` so I don't think field omissions is a problem.

viktorklang-ora: I think there's a semantic difference between asserting "any fields returned must be strict" and "two fields are to be returned, named "x" and "y" and they must both be strict".

liach: There is, but I don't think this argues for complicating such a sanity check. You can always request extra coverage for every failure path of every test you write; it is an endless rabbit hole, and I argue a misused `@driver` has no compile-time or validation from other tests that it deserves some sanity test; meanwhile `getDeclaredFields` is well-tested and much less error-prone, so that's where I put the stopping line.

src/java.base/share/classes/java/io/ObjectStreamClass.java
liach: I don't think I can - lworld is well tested so it's almost impossible to split the mainline alignment out without causing some test failures (

src/java.base/share/classes/java/io/ObjectStreamClass.java
liach: See the updated comments in the latest push.

src/java.base/share/classes/java/io/ObjectStreamClass.java
liach: Right. I renamed this to `requiresDeserializeFactory` to better convey the intentions.

src/java.base/share/classes/java/io/ObjectStreamClass.java
liach: I renamed this to `deserializeFactory` and the annotation to `@DeserializeFactory`.

src/java.base/share/classes/java/io/ObjectInputStream.java
liach: Good point. The boolean is actually accessible as `desc.isRecord()`. I have now renamed this class to `FactorySupport` and split the record and factory paths, leaving the method handle building code in common.

src/java.base/share/classes/java/io/ObjectStreamClass.java
liach: Renamed to `hasDeserializeConstructor`, pairs up with the new `requiresDeserializeConstructor`.

test/jdk/java/io/Serializable/valueObjects/SimpleValueGraphs.java
liach: In the latest push, I removed the dead misleading message (which the old test tested against) from `desc.isExternalizable() && !desc.isProxy()` branch in `ObjectOutputStream` and added an assert indicating where an exception should have been through.

src/java.base/share/classes/jdk/internal/value/DeserializeFactory.java
```diff
@@ -39,7 +39,7 @@
  */
 @Retention(RetentionPolicy.RUNTIME)
 @Target(value={CONSTRUCTOR, METHOD})
-public @interface DeserializeConstructor {
+public @interface DeserializeFactory {
```
viktorklang-ora: "Factory" tends to mean "static method", so if we need something which is applicable for both constructors and factory methods, I think we need to have a name change. "Deserializer"?

liach: Good idea, I love simple names!

viktorklang-ora:
@liach I've reviewed this PR, and have no further comments at this time. It would be beneficial to have another Core Libraries reviewer chime in here as well, maybe @AlanBateman might be able to give a second opinion.

src/java.base/share/classes/java/io/ObjectStreamClass.java
```diff
@@ -1422,54 +1338,66 @@ private ObjectStreamClass getVariantFor(Class<?> cl)
...
-        var params = mh.type().parameterList();
-        if (params.size() != fields.length) {
-            return false;    // Mismatch in count of fields and parameters
+    private static boolean isDeserializer(Executable exec,
+                                          ObjectStreamField[] fields) {
+        if (exec.getParameterCount() != fields.length) {
+            return false;
+        }
+
+        var deserializer = exec.getDeclaredAnnotation(Deserializer.class);
```
AlanBateman, line 1376: @liach There's a bit too much "var" usage in isDeserializer so it's hard to review. It is of course okay when the type is on the RHS but when there is nothing to hint at the type it makes it hard to read.

AlanBateman:
> @liach I've reviewed this PR, and have no further comments at this time. It would be beneficial to have another Core Libraries reviewer chime in here as well, maybe @AlanBateman might be able to give a second opinion.

Okay, it removes a lot of the code so the resulting changes for serialiation will be much small, good.

src/java.base/share/classes/jdk/internal/value/ValueClass.java
```diff
@@ -139,4 +142,31 @@ public static Object[] copyOfRangeSpecialArray(Object[] array, int from, int to)
      */
     @IntrinsicCandidate
     public static native boolean isAtomicArray(Object[] array);
+
+    // This class also serves as a lazy holder of its singleton instance
+    private static final class StrictInstanceFieldClassValue extends ClassValue<Boolean> {
```
AlanBateman, line 147: Good use of CV.

src/java.base/share/classes/java/io/ObjectStreamClass.java
```diff
@@ -1422,54 +1338,66 @@ private ObjectStreamClass getVariantFor(Class<?> cl)
...
-        MethodHandles.Lookup lookup = MethodHandles.lookup();
-        MethodHandle mh = Stream.concat(
+     * @return an Executable, null if none found
+     */
+    private static Executable findDeserializer(Class<?> clazz,
+                                               ObjectStreamField[] fields) {
+        return Stream.concat(
                 Arrays.stream(clazz.getDeclaredMethods()).filter(m -> Modifier.isStatic(m.getModifiers())),
                 Arrays.stream(clazz.getDeclaredConstructors()))
-                .filter(m -> m.isAnnotationPresent(DeserializeConstructor.class))
```
AlanBateman, line 1440: If it turns out we only use this on the wrapper constructors then I would be tempted to leave the static factory support for the code going to main line. I note your comment in the mainline PR about having flexibility but keeping unused code would be good, if possible.

AlanBateman (approved):
I did a pass over this and I think it okay. (Some of it just reverting code back to what it is in main line so it will reduce the overall patch).

src/java.base/share/classes/java/io/ObjectStreamClass.java
liach: I currently have cases actively testing static factories in test/jdk/java/io/Serializable/valueObjects/SerializedObjectCombo.java. I think this support has very low overhead and in the long run, we still might want to move back to using `valueOf` if we can eliminate regressions.

dlsmith-dev (approved):
I can't speak for the relative merits of the prior vs. proposed implementation, but appreciate the push for minimizing the diff. I'm a Committer here to make Skara happy :-)

liach:
Thanks for the reviews. As in #2350, I will integrate this to get more sufficient testing, and this also removed some bad uses of things like `@MigratedValueClass` that blocks other changes. Sanity ran Serializable tests after a quick merge.

/integrate

PR #2329, opened 2026-04-17
8382027: [lworld] if_acmp nullchecks can produce false positives

Arraying:
Hi all,

This fixes implementation bugs in the interpreter and C1 for `if_acmp<cc>` bytecodes.

The substitutability check that we perform to determine equality of value objects requires that certain conditions have been checked beforehand. Namely, that it's not the same object, neither is null, and both are value objects with the same klass.

At the point of the null-check, it is already clear that the oops are unequal; hence the nullcheck checks `lhs == 0 xor rhs == 0`. Both the AArch64 interpreter and x86 C1 code falsely ANDs the two operands. This can yield false positives if two oops have no set-bit positions in common (e.g., `1010` and `0101`). The operands need to be compared separately; the interpreter for x86 and C1 for AArch64 already does this.

Note that this fix is covered by the many test cases failing in `compiler/valhalla/inlinetypes/`.

Testing: tiers 1-4 on Linux (x64, AArch64), macOS (x64, AArch64), Windows (x64).

---------

---------
### Progress

### Issue
 * [JDK-8382027](https://bugs.openjdk.org/browse/JDK-8382027): [lworld] if_acmp nullchecks can produce false positives (**Bug** - P4)

### Reviewers
 * [Stefan Karlsson](https://openjdk.org/census#stefank) (@stefank - Committer)

### Contributors
 * Axel Boldt-Christmas `<aboldtch@openjdk.org>`
 * Joel Sikström `<jsikstro@openjdk.org>`
 * Stefan Karlsson `<stefank@openjdk.org>`

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2329/head:pull/2329` \
`$ git checkout pull/2329`

`$ git checkout pull/2329` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2329/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2329`

`$ git pr show -t 2329`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2329.diff">https://git.openjdk.org/valhalla/pull/2329.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2329#issuecomment-4267092314)
</details>

### DIFF
--- a/src/hotspot/cpu/aarch64/templateTable_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/templateTable_aarch64.cpp
@@ -2029,14 +2029,15 @@ void TemplateTable::if_acmp(Condition cc) {
   __ mov(is_inline_type_mask, markWord::inline_type_pattern);
 
   if (Arguments::is_valhalla_enabled()) {
+    // The substitutability test is only necessary if r1 and r0 are not the same...
     __ cmp(r1, r0);
     __ br(Assembler::EQ, (cc == equal) ? taken : not_taken);
 
-    // might be substitutable, test if either r0 or r1 is null
-    __ andr(r2, r0, r1);
-    __ cbz(r2, (cc == equal) ? not_taken : taken);
+    // ... neither are null...
+    __ cbz(r1, (cc == equal) ? not_taken : taken);
+    __ cbz(r0, (cc == equal) ? not_taken : taken);
 
-    // and both are values ?
+    // ...and both are values...
     __ ldr(r2, Address(r1, oopDesc::mark_offset_in_bytes()));
     __ andr(r2, r2, is_inline_type_mask);
     __ ldr(r4, Address(r0, oopDesc::mark_offset_in_bytes()));
@@ -2045,7 +2046,7 @@ void TemplateTable::if_acmp(Condition cc) {
     __ cmp(r2,  is_inline_type_mask);
     __ br(Assembler::NE, (cc == equal) ? not_taken : taken);
 
-    // same value klass ?
+    // ...with the same value klass
     __ load_metadata(r2, r1);
     __ load_metadata(r4, r0);
     __ cmp(r2, r4);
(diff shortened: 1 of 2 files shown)


xmas92 (commented):
lgtm. 

I think this went unnoticed for so long (seems to have been in the code for 6 years) is because it is rare for GC to not have a heap address layout where where some subset of bits are always `1`.

This only happens with `-UseCompressedOops`, (or with some compressed modes depending on the specific base selection). With ZGC it never happened as we always have a disjoint heap base.

Arraying:
> lgtm.
> 
> I think this went unnoticed for so long (seems to have been in the code for 6 years) is because it is rare for GC to not have a heap address layout where where some subset of bits are always `1`.
> 
> This only happens with `-UseCompressedOops`, (or with some compressed modes depending on the specific base selection). With ZGC it never happened as we always have a disjoint heap base.

Thanks! I've been able to reproduce wrong behaviour with `-XX:+UseCompressedOops` yielding base 0, shift 3 that appears to be fixed after this patch.

stefank (approved):

TobiHartmann:
Just a fly-by comment: This issue needs a more descriptive title.

TobiHartmann (approved):
C1 fix looks good to me. Thanks!

Arraying:
> Just a fly-by comment: This issue needs a more descriptive title.

I've updated the title to reflect the issue is with `if_acmp`.

fparain (approved):
LGTM
Thank you for the investigation and the fix.

xmas92 (approved):

merykitty:
Shouldn't the title be about "false negatives"? We mistakenly assume that `x == 0 || y == 0` is equivalent to `(x & y) == 0`. However, there are cases when `x != 0` and `y != 0` but `(x & y) == 0`. This means that we mistakenly take the path `x == 0 || y == 0`, which leads to `acmp` mistakenly returning `false`.

Edit: My mistake, just ignore :)

Arraying:
Thanks for the reviews everyone.
/integrate

PR #2332, opened 2026-04-17
8381866: [lworld] C2: SIGSEGV in SafePointNode::disconnect_from_root

benoitmaillard:
This PR prevents a segmentation fault during `Compile::process_inline_types` caused by a missing check for failure.

This was originally spotted on our CI when running `applications/ctw/modules/java_desktop.java` with `StressBailout`, and is reproducible with replay compilation.

Near the end of `Compile::Optimize`, we have:
https://github.com/openjdk/valhalla/blob/f29dba868499db3b1369efee8ea22365efa36849/src/hotspot/share/opto/compile.cpp#L3238-L3243

In `Compile::process_late_inline_calls_no_inline`, we have a `if (failing())  return;` check. When using `StressBailout`, this might trigger a compilation bailout.

Before valhalla this was fine because there is a failure check a few lines after the call to `process_late_inline_calls_no_inline`. Now we also have `process_inline_types(igvn)`, and this can go very wrong if `C->root()` is `nullptr`. Therefore we need a new `if failing()` check before processing the inline types.

I have checked other usages of `process_inline_types` and it seems there is no cause for concern there.

I don't think it's worth it to try to produce a regression test here as it would rely on setting a stress seed and would therefore be very sensitive.

### Testing
- [ ] tier1-3, plus some internal testing

Thank you for reviewing!

---------
### Progress

### Issue
 * [JDK-8381866](https://bugs.openjdk.org/browse/JDK-8381866): [lworld] C2: SIGSEGV in SafePointNode::disconnect_from_root (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2332/head:pull/2332` \
`$ git checkout pull/2332`

`$ git checkout pull/2332` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2332/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2332`

`$ git pr show -t 2332`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2332.diff">https://git.openjdk.org/valhalla/pull/2332.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2332#issuecomment-4269077143)
</details>

### DIFF
--- a/src/hotspot/share/opto/compile.cpp
+++ b/src/hotspot/share/opto/compile.cpp
@@ -3239,6 +3239,9 @@ void Compile::Optimize() {
     // More opportunities to optimize virtual and MH calls.
     // Though it's maybe too late to perform inlining, strength-reducing them to direct calls is still an option.
     process_late_inline_calls_no_inline(igvn);
+    if (failing()) {
+      return;
+    }
     process_inline_types(igvn);
   }
 


marc-chevalier:
That looks nice, but why is it so rare? Shouldn't it happen a bit all the time with this flag?

benoitmaillard:
I think there are a lot of places where we call `Compile::failing` (there are 31 references in `Compile::Optimize`), and since we randomly fail at each call, that's quite the sequence of events. The probability is _quite low_ @marc-chevalier.

Edit: for this specific case, `failing` is called 21745 times. Given that we have `-XX:StressBailoutMean=100000`, this gives us a probability of failure at this specific callsite of roughly `1/10000`.

marc-chevalier:
So, if I trick the code to make it not fail elsewhere, but here with high probability, it should happen very often, on many tests?

merykitty (approved):
Thanks for the fix, it looks good to me.

TobiHartmann (approved):
Looks good to me too. Thanks for fixing!

benoitmaillard:
@marc-chevalier yes, most likely.

Thank for reviewing @marc-chevalier @TobiHartmann @merykitty!

/integrate

PR #2333, opened 2026-04-17
8374475: [lworld] C2: assert(_gvn.type(l)->higher_equal(type)) failed: must constrain OSR typestate

merykitty:
Hi,

The reason for the failure is that, we separately compute the instance type `type`, which is used in the assert, and the klass type `klass_type`, which is used for the check cast, so there is a chance during the computation of `type`, the class has only 1 single concrete subclass, and before the computation of `klass_type`, another subtype is loaded, leading to `klass_type` being the supertype of `type`, which triggers the assert. Computing these 2 in the opposite order likely does not work, either, because we can have concurrent class unloading.

As a result, in this PR, I reworked `check_interpreter_type` to receive a `ciType`, this also helps simplify the implementation, as previously, we needed to obtain the null restriction and larval-ness of the value indirectly, while they are readily available in the `ciType` obtained from `ciTypeFlow`. Trying to keep the assert for every case is cumbersome, and verifying the type right after the check cast seems unnecessary, so I relaxed the assert to only verify the correct kind of value.

Testing:

Please take a look and leave your review, thanks a lot.

---------

---------
### Progress

### Issue
 * [JDK-8374475](https://bugs.openjdk.org/browse/JDK-8374475): [lworld] C2: assert(_gvn.type(l)-&gt;higher_equal(type)) failed: must constrain OSR typestate (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2333/head:pull/2333` \
`$ git checkout pull/2333`

`$ git checkout pull/2333` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2333/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2333`

`$ git pr show -t 2333`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2333.diff">https://git.openjdk.org/valhalla/pull/2333.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2333#issuecomment-4269809406)
</details>

### DIFF
--- a/src/hotspot/share/opto/parse.hpp
+++ b/src/hotspot/share/opto/parse.hpp
@@ -443,7 +443,7 @@ class Parse : public GraphKit {
 
   // OSR helpers
   Node* fetch_interpreter_state(int index, const Type* type, Node* local_addrs);
-  Node* check_interpreter_type(Node* l, const Type* type, const TypeKlassPtr* klass_type, SafePointNode* &bad_type_exit, bool is_larval);
+  Node* check_interpreter_type(Node* l, ciType* ci_type, SafePointNode* &bad_type_exit);
   void  load_interpreter_state(Node* osr_buf);
 
   // Functions for managing basic blocks:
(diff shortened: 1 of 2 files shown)


TobiHartmann (approved):
Looks good to me, nice refactoring!

merykitty:
@TobiHartmann Thanks a lot for your review!

/integrate

PR #2339, opened 2026-04-20
8284443: [lworld] Follow-up work from support for scalarization of nullable inline types in the calling convention

TobiHartmann:
This patch addresses various follow-up TODOs from [JDK-8278390](https://bugs.openjdk.org/browse/JDK-8278390).

I added PR comments to explain the changes in more detail.

Thanks,
Tobias

---------

---------
### Progress

### Issue
 * [JDK-8284443](https://bugs.openjdk.org/browse/JDK-8284443): [lworld] Follow-up work from support for scalarization of nullable inline types in the calling convention (**Enhancement** - P3)

### Reviewers
 * [Quan Anh Mai](https://openjdk.org/census#qamai) (@merykitty - Committer) ⚠️ Review applies to [2cb66e28](https://git.openjdk.org/valhalla/pull/2339/files/2cb66e285d4d94b2b55e9d127b7d63fdfa170e2d)
 * [Christian Hagedorn](https://openjdk.org/census#chagedorn) (@chhagedorn - Committer) ⚠️ Review applies to [2cb66e28](https://git.openjdk.org/valhalla/pull/2339/files/2cb66e285d4d94b2b55e9d127b7d63fdfa170e2d)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2339/head:pull/2339` \
`$ git checkout pull/2339`

`$ git checkout pull/2339` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2339/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2339`

`$ git pr show -t 2339`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2339.diff">https://git.openjdk.org/valhalla/pull/2339.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2339#issuecomment-4281551128)
</details>


src/hotspot/cpu/aarch64/aarch64.ad
```diff
@@ -2203,8 +2203,9 @@ void MachVEPNode::emit(C2_MacroAssembler *masm, PhaseRegAlloc* ra_) const
   if (!_verified) {
     __ ic_check(1);
   } else {
-    // TODO 8284443 Avoid creation of temporary frame
```
TobiHartmann, line 2206: I don't think we can easily avoid this, unfortunately. But I added a comment that explains why it's needed. See [JDK-8293994](https://bugs.openjdk.org/browse/JDK-8293994) for background.

src/hotspot/cpu/aarch64/c1_MacroAssembler_aarch64.cpp
```diff
@@ -314,7 +314,6 @@ int C1_MacroAssembler::scalarized_entry(const CompiledEntrySignature* ces, int f
   // C1 code is not hot enough to micro optimize the nmethod entry barrier with an out-of-line stub
   bs->nmethod_entry_barrier(this, nullptr /* slow_path */, nullptr /* continuation */, nullptr /* guard */);
 
-  // FIXME -- call runtime only if we cannot in-line allocate all the incoming inline type args.
```
TobiHartmann, line 317: Not something we want to implement at this point.

src/hotspot/cpu/aarch64/frame_aarch64.inline.hpp
```diff
@@ -463,7 +463,6 @@ inline frame frame::sender_for_compiled_frame(RegisterMap* map) const {
     nmethod* nm = _cb->as_nmethod_or_null();
     if (nm != nullptr && nm->is_compiled_by_c1() && nm->method()->has_scalarized_args() &&
         pc() < nm->verified_inline_entry_point()) {
-      // TODO 8284443 Can't we do that by not passing 'dont_gc_arguments' in case 'StubId::c1_buffer_inline_args_id' in 'Runtime1::generate_code_for'?
```
TobiHartmann, line 466: No because we need this for the nmethod at method entry only, not for the stub itself.

src/hotspot/cpu/aarch64/interp_masm_aarch64.cpp
```diff
@@ -734,19 +734,6 @@ void InterpreterMacroAssembler::remove_activation(TosState state,
...
-      Label skip_stress;
-      ldr(rscratch1, Address(rfp, frame::interpreter_frame_method_offset * wordSize));
-      ldrw(rscratch1, Address(rscratch1, Method::flags_offset()));
-      tstw(rscratch1, MethodFlags::has_scalarized_return_flag());
-      br(Assembler::EQ, skip_stress);
-      load_klass(r0, r0);
-      orr(r0, r0, 1);
-      bind(skip_stress);
-    }
-#endif
```
TobiHartmann, line 749: The problem is that `MethodFlags::has_scalarized_return_flag` is not accurate because it's set at method link time when the return type might not be loaded yet. After this patch, it's also not used anywhere else but for this stress path, so I decided to simply remove it. The same stress path exists in C1 and C2 which is sufficient for stress testing.

src/hotspot/cpu/x86/c1_MacroAssembler_x86.cpp
```diff
@@ -306,7 +306,6 @@ int C1_MacroAssembler::scalarized_entry(const CompiledEntrySignature* ces, int f
   // C1 code is not hot enough to micro optimize the nmethod entry barrier with an out-of-line stub
   bs->nmethod_entry_barrier(this, nullptr /* slow_path */, nullptr /* continuation */);
 
-  // FIXME -- call runtime only if we cannot in-line allocate all the incoming inline type args.
```
TobiHartmann, line 309: Not something we want to implement at this point.

src/hotspot/cpu/x86/frame_x86.inline.hpp
```diff
@@ -440,7 +440,6 @@ inline frame frame::sender_for_compiled_frame(RegisterMap* map) const {
     nmethod* nm = _cb->as_nmethod_or_null();
     if (nm != nullptr && nm->is_compiled_by_c1() && nm->method()->has_scalarized_args() &&
         pc() < nm->verified_inline_entry_point()) {
-      // TODO 8284443 Can't we do that by not passing 'dont_gc_arguments' in case 'StubId::c1_buffer_inline_args_id' in 'Runtime1::generate_code_for'?
```
TobiHartmann, line 443: No because we need this for the nmethod at method entry only, not for the stub itself.

src/hotspot/cpu/x86/interp_masm_x86.cpp
```diff
@@ -1085,19 +1085,6 @@ void InterpreterMacroAssembler::remove_activation(TosState state,
...
-    if (StressCallingConvention && false) {
-      Label skip_stress;
-      movptr(rscratch1, Address(rbp, frame::interpreter_frame_method_offset * wordSize));
-      movl(rscratch1, Address(rscratch1, Method::flags_offset()));
-      testl(rcx, MethodFlags::has_scalarized_return_flag());
-      jcc(Assembler::zero, skip_stress);
-      load_klass(rax, rax, rscratch1);
-      orptr(rax, 1);
-      bind(skip_stress);
-    }
```
TobiHartmann, line 1100: The problem is that MethodFlags::has_scalarized_return_flag is not accurate because it's set at method link time when the return type might not be loaded yet. After this patch, it's also not used anywhere else but for this stress path, so I decided to simply remove it. The same stress path exists in C1 and C2 which is sufficient for stress testing.

src/hotspot/cpu/x86/x86.ad
```diff
@@ -2605,8 +2605,9 @@ void MachVEPNode::emit(C2_MacroAssembler* masm, PhaseRegAlloc* ra_) const
   if (!_verified) {
     __ ic_check(1);
   } else {
-    // TODO 8284443 Avoid creation of temporary frame
```
TobiHartmann, line 2608: I don't think we can easily avoid this, unfortunately. But I added a comment that explains why it's needed. See [JDK-8293994](https://bugs.openjdk.org/browse/JDK-8293994) for background.

src/hotspot/share/asm/macroAssembler_common.cpp
```diff
@@ -90,8 +90,8 @@ MacroAssembler::RegState* MacroAssembler::init_reg_state(VMRegPair* regs, int nu
 #ifdef COMPILER2
 int MacroAssembler::unpack_inline_args(Compile* C, bool receiver_only) {
   assert(C->has_scalarized_args(), "inline type argument scalarization is disabled");
-  Method* method = C->method()->get_Method();
-  const GrowableArray<SigEntry>* sig = method->adapter()->get_sig_cc();
+  ciMethod* method = C->method();
```
TobiHartmann, line 93: Since this is called from the compiler threads, all accesses should go through the CI.

src/hotspot/share/c1/c1_Compilation.cpp
```diff
@@ -597,12 +596,6 @@ Compilation::Compilation(AbstractCompiler* compiler, ciEnv* env, ciMethod* metho
 #endif
   CompilationMemoryStatisticMark cmsm(directive);
 
-  {
-    ResetNoHandleMark rnhm;
-    // TODO 8284443 Should only be computed once
-    _compiled_entry_signature.compute_calling_conventions(false);
```
TobiHartmann, line 603: I move this to where it's needed but caching doesn't seem feasible.

src/hotspot/share/oops/inlineKlass.cpp
```diff
@@ -194,7 +194,6 @@ int InlineKlass::collect_fields(GrowableArray<SigEntry>* sig, int base_off, int
     assert(!fs.access_flags().is_static(), "TopDownHierarchicalNonStaticFieldStreamBase should not let static fields pass.");
     int offset = base_off + fs.offset() - (base_off > 0 ? payload_offset() : 0);
     InstanceKlass* field_holder = fs.field_descriptor().field_holder();
-    // TODO 8284443 Use different heuristic to decide what should be scalarized in the calling convention
```
TobiHartmann, line 197: I think the heuristic is good as-is.

src/hotspot/share/oops/method.cpp
```diff
@@ -1336,9 +1335,6 @@ void Method::link_method(const methodHandle& h_method, TRAPS) {
       SharedRuntime::native_method_throw_unsatisfied_link_error_entry(),
       !native_bind_event_is_interesting);
   }
-  if (InlineTypeReturnedAsFields && returns_inline_type() && !has_scalarized_return()) {
-    set_has_scalarized_return();
```
TobiHartmann, line 1340: Removed because it's not accurate. See comments above.

src/hotspot/share/opto/callGenerator.cpp
```diff
@@ -1068,29 +1068,6 @@ JVMState* PredictedCallGenerator::generate(JVMState* jvms) {
...
   }
 
-  // Allocate inline types if they are merged with objects (similar to Parse::merge_common())
-  uint tos = kit.jvms()->stkoff() + kit.sp();
-  uint limit = slow_map->req();
-  for (uint i = TypeFunc::Parms; i < limit; i++) {
-    Node* m = kit.map()->in(i);
-    Node* n = slow_map->in(i);
-    const Type* t = gvn.type(m)->meet_speculative(gvn.type(n));
-    // TODO 8284443 still needed?
```
TobiHartmann, line 1092: Turns out the entire block is not needed anymore.

src/hotspot/share/opto/chaitin.cpp
```diff
@@ -2463,7 +2463,27 @@ void PhaseChaitin::dump_frame() const {
   OptoReg::Name return_addr = _matcher.return_addr();
 
   reg = OptoReg::add(reg, -1);
+
+  // Special fixed slots
```
TobiHartmann, line 2467: Added improved printing of the special fixed stack slots. "stack increment" and "null marker" is Valhalla specific. "original deopt pc" is not.

src/hotspot/share/opto/inlinetypenode.cpp
```diff
@@ -1141,9 +1140,6 @@ InlineTypeNode* InlineTypeNode::make_from_oop_impl(GraphKit* kit, Node* oop, ciI
     vt->set_null_marker(gvn);
     Node* payload_ptr = kit->basic_plus_adr(oop, vk->payload_offset());
     vt->load(kit, oop, payload_ptr, true, true, IN_HEAP | MO_UNORDERED, visited);
-// TODO 8284443
-//    assert(!null_free || vt->as_InlineType()->is_all_zero(&gvn) || init_ctl != kit->control() || !gvn.type(oop)->is_inlinetypeptr() || oop->is_Con() || oop->Opcode() == Op_InlineType ||
-//           AllocateNode::Ideal_allocation(oop, &gvn) != nullptr || vt->as_InlineType()->is_loaded(&gvn) == oop, "inline type should be loaded");
   }
   assert(vt->is_allocated(&gvn), "inline type should be allocated");
```
TobiHartmann, line 1148: This assert is just way too brittle. I fixed it multiple times before and I don't think it's useful. Let's just remove it.

src/hotspot/share/opto/inlinetypenode.cpp
```diff
@@ -1387,7 +1383,12 @@ Node* InlineTypeNode::tagged_klass(ciInlineKlass* vk, PhaseGVN& gvn) {
 
 void InlineTypeNode::pass_fields(GraphKit* kit, Node* n, uint& base_input, bool in, bool null_free, bool root) {
   if (root) {
-    n->init_req(base_input++, get_oop());
+    if (is_allocated(&kit->gvn())) {
+      // Keep the information that 'this' is buffered
+      n->init_req(base_input++, this);
+    } else {
+      n->init_req(base_input++, get_oop());
+    }
```
TobiHartmann, line 1389: Filed [JDK-8382527](https://bugs.openjdk.org/browse/JDK-8382527) as a follow-up on this.

src/hotspot/share/opto/inlinetypenode.cpp
```diff
@@ -1417,10 +1418,6 @@ void InlineTypeNode::pass_fields(GraphKit* kit, Node* n, uint& base_input, bool
       }
     }
   }
-  // The last argument is used to pass the null marker to compiled code and not required here.
-  if (!null_free && !in) {
-    n->init_req(base_input++, kit->top());
-  }
 }
```
TobiHartmann, line 1425: I removed the requirement to return this arg, see changes to `src/hotspot/share/opto/type.cpp`.

src/hotspot/share/opto/graphKit.cpp
```diff
@@ -2030,7 +2030,6 @@ void GraphKit::set_arguments_for_java_call(CallJavaNode* call) {
     uint arg_idx = i - TypeFunc::Parms;
     Node* arg = argument(arg_idx);
     const Type* t = domain->field_at(i);
-    // TODO 8284443 A static call to a mismatched method should still be scalarized
```
TobiHartmann, line 2033: Not worth the complexity because this would also need changes in `TypeFunc::make` where the "is virtual call" information is messy to propagate. Let's leave it for now.

src/hotspot/share/opto/type.cpp
```diff
@@ -7325,31 +7330,32 @@ const TypeTuple* osr_domain() {
...
   }
-  const TypeTuple* domain_sig = is_osr_compilation ? osr_domain() : TypeTuple::make_domain(method, ignore_interfaces, false);
-  const TypeTuple* domain_cc = has_scalar_args ? TypeTuple::make_domain(method, ignore_interfaces, true) : domain_sig;
   ciSignature* sig = method->signature();
   bool has_scalar_ret = !method->is_native() && sig->return_type()->is_inlinetype() && sig->return_type()->as_inline_klass()->can_be_returned_as_fields();
-  const TypeTuple* range_sig = TypeTuple::make_range(sig, ignore_interfaces, false);
-  const TypeTuple* range_cc = has_scalar_ret ? TypeTuple::make_range(sig, ignore_interfaces, true) : range_sig;
-  tf = TypeFunc::make(domain_sig, domain_cc, range_sig, range_cc);
-  if (!is_osr_compilation) {
+  // Don't cache on scalarized return because the range depends on 'is_call'
```
TobiHartmann, line 7354: We need to distinguish between the return signature of a method at calls vs. at compilation of that method because at calls we return an additional `InlineTypeNode::NullMarker` field.

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
TobiHartmann, line 7493: See [JDK-8382417](https://bugs.openjdk.org/browse/JDK-8382417).

src/hotspot/cpu/x86/macroAssembler_x86.cpp
TobiHartmann, line 6464: See [JDK-8382417](https://bugs.openjdk.org/browse/JDK-8382417).

src/hotspot/share/runtime/sharedRuntime.cpp
```diff
@@ -2904,8 +2899,8 @@ void CompiledEntrySignature::compute_calling_conventions(bool init) {
     int arg_num = 0;
     if (!_method->is_static()) {
       // We shouldn't scalarize 'this' in a value class constructor
-      if (holder->is_inline_klass() && InlineKlass::cast(holder)->can_be_passed_as_fields() && !_method->is_object_constructor() &&
-          (init || _method->is_scalarized_arg(arg_num))) {
+      if (InlineTypePassFieldsAsArgs && holder->is_inline_klass() && InlineKlass::cast(holder)->can_be_passed_as_fields() &&
```
merykitty: Do you think it is a good idea to reduce the sites in which these flag checks are necessary? In this particular instance, principally, I think `InlineKlass::can_be_passed_as_fields` should return `false` if `InlineTypePassFieldsAsArgs` is `false`, removing the need to explicitly check the flag here.

src/hotspot/share/runtime/sharedRuntime.cpp
```diff
@@ -2920,7 +2915,7 @@ void CompiledEntrySignature::compute_calling_conventions(bool init) {
     }
     for (SignatureStream ss(_method->signature()); !ss.at_return_type(); ss.next()) {
       BasicType bt = ss.type();
-      if (bt == T_OBJECT) {
+      if (InlineTypePassFieldsAsArgs && bt == T_OBJECT) {
```
merykitty, line 2923: Similarly, this can also be removed.

src/hotspot/share/opto/type.cpp
```diff
@@ -7325,31 +7330,32 @@ const TypeTuple* osr_domain() {
 }
 
 //------------------------------make-------------------------------------------
-const TypeFunc* TypeFunc::make(ciMethod* method, bool is_osr_compilation) {
+const TypeFunc* TypeFunc::make(ciMethod* method, bool is_call, bool is_osr_compilation) {
```
merykitty, line 7339: It is better to have a description for the meaning of `is_call`.

src/hotspot/share/opto/subnode.cpp
```diff
@@ -1148,13 +1148,14 @@ static inline Node* isa_const_java_mirror(PhaseGVN* phase, Node* n, bool& might_
...
+  Node* uncast_in1 = in(1)->uncast();
+  Node* uncast_in2 = in(2)->uncast();
+  if (uncast_in1->is_InlineType() && phase->type(uncast_in2)->is_zero_type()) {
     // Null checking a scalarized but nullable inline type. Check the null marker
     // input instead of the oop input to avoid keeping buffer allocations alive.
-    return new CmpINode(in(1)->as_InlineType()->get_null_marker(), phase->intcon(0));
+    return new CmpINode(uncast_in1->as_InlineType()->get_null_marker(), phase->intcon(0));
   }
-  if (in(1)->is_InlineType() || in(2)->is_InlineType()) {
+  if (uncast_in1->is_InlineType() || uncast_in2->is_InlineType()) {
```
merykitty, line 1158: This looks pretty suspicious tbh. What if we want to fast-path the substitutability check in C2, and compare 2 objects by pointer before looking at each of their fields?

src/hotspot/share/ci/ciInlineKlass.hpp
```diff
@@ -65,6 +65,7 @@ class ciInlineKlass : public ciInstanceKlass {
   bool can_be_passed_as_fields() const;
```
merykitty, line 69: I think a description of how these values are passed in and out, or a pointer to an existing description is appreciated here,

merykitty (commented):
A first pass through the `opto` and `ci` changes from me.

src/hotspot/share/ci/ciInlineKlass.hpp
TobiHartmann: I added a comment here and another more detailed one to `InlineKlass::initialize_calling_convention`. What do you think?

src/hotspot/share/opto/subnode.cpp
TobiHartmann: I agree. Unfortunately, it's needed to ensure that the substitutability test is optimized after being inlined - together with the code in `CmpLNode::Ideal`. Maybe we can remove/improve it with [JDK-8381826](https://bugs.openjdk.org/browse/JDK-8381826)?

src/hotspot/share/opto/type.cpp
TobiHartmann: I added a comment explaining the purpose of this method and its arguments.

src/hotspot/share/runtime/sharedRuntime.cpp
TobiHartmann: Actually, `InlineKlass::can_be_passed_as_fields` already returns false for `!InlineTypePassFieldsAsArgs` so this particular check is not needed. I reverted it (though it might actually nice for consistency with below code - but I don't have a strong opinion on this).

src/hotspot/share/runtime/sharedRuntime.cpp
TobiHartmann: I added this to address a comment that @fparain had when reading this code:
> [...] a point that was bothering me was that `InlineTypePassFieldsAsArgs` is tested after the call to ss.as_inline_klass(), this is not obvious when reading the code above but the test is performed in vk->can_be_passed_as_fields() (this is the only thing this method tests) which is called after having poke the system dictionary."

The expensive part here is `ss.as_inline_klass()`, so we definitely want to guard this by `InlineTypePassFieldsAsArgs` but since that method is also called from other places when `!InlineTypePassFieldsAsArgs && InlineTypeReturnedAsFields`, we can't easily move that check into the method. Also, as Fred mentions, it's easier to read when the check is in the caller.

What do you think?

TobiHartmann:
Thanks for the review @merykitty! I replied to all your comments.

src/hotspot/share/oops/inlineKlass.cpp
```diff
@@ -220,6 +219,16 @@ int InlineKlass::collect_fields(GrowableArray<SigEntry>* sig, int base_off, int
...
+//
+// For arguments, an inline type can be passed in scalarized form instead of as a single
+// oop: the calling convention uses an optional buffer oop together with a null marker,
+// followed by the field values, assigned to the normal argument registers and stack slots.
+// See CompiledEntrySignature::compute_calling_conventions.
+//
+// For returns, an inline type is returned in scalarized form via multiple return registers:
+// the first word is a tri-state value (null, tagged InlineKlass*, or oop) and the remaining
+// registers carry the field values.
 void InlineKlass::initialize_calling_convention(TRAPS) {
```
merykitty, line 232: Thanks for the explanation, I see that below we talk about a null marker slot at scalarized return value, is it the null marker of the return value, or is it something else? If it is the null marker of the return value, it seems the tri-state of the oop is unnecessary, as we are passing the values similar to how we pass them as an argument.

src/hotspot/share/opto/type.cpp
merykitty: Thanks, I understand it now.

src/hotspot/share/runtime/sharedRuntime.cpp
merykitty: I see, in that case keeping `InlineTypePassFieldsAsArgs` in the caller is more preferable then.

merykitty (approved):
Thanks for your answers, the PR looks good to me.

src/hotspot/share/oops/inlineKlass.cpp
TobiHartmann: Unfortunately, it's a bit more complicated than that. The actual call will **not** return the null marker as a separate field but use the tri-state inf the first register to signal null. However, in C2 IR, we model the null marker as a separate projection and therefore need to add it to the signature. The "translation" from the tri state to the separate field is done here:
https://github.com/openjdk/valhalla/blob/8b2263f38bdc38b7d219d9eb813ceea1311077b2/src/hotspot/cpu/x86/x86.ad#L4624-L4626

merykitty: I see, thanks for your clarification.

src/hotspot/share/ci/ciMethod.cpp
```diff
@@ -1616,8 +1616,7 @@ bool ciMethod::is_scalarized_buffer_arg(int idx) const {
 }
 
 bool ciMethod::has_scalarized_args() const {
-  VM_ENTRY_MARK;
-  return get_Method()->has_scalarized_args();
+  GUARDED_VM_ENTRY(return get_Method()->has_scalarized_args();)
```
chhagedorn: You could do the same for `inline_arg_length()` above.

src/hotspot/share/opto/type.cpp
```diff
@@ -2210,13 +2210,15 @@ static void collect_inline_fields(ciInlineKlass* vk, const Type** field_array, u
 
 //------------------------------make-------------------------------------------
 // Make a TypeTuple from the range of a method signature
-const TypeTuple *TypeTuple::make_range(ciSignature* sig, InterfaceHandling interface_handling, bool ret_vt_fields) {
+const TypeTuple *TypeTuple::make_range(ciSignature* sig, InterfaceHandling interface_handling, bool ret_vt_fields, bool is_call) {
```
chhagedorn:
```suggestion
const TypeTuple* TypeTuple::make_range(ciSignature* sig, InterfaceHandling interface_handling, bool ret_vt_fields, bool is_call) {
```

src/hotspot/share/opto/type.cpp
```diff
@@ -7309,8 +7313,9 @@ const Type* TypeAryKlassPtr::base_element_type(int& dims) const {
 
 //------------------------------make-------------------------------------------
 const TypeFunc *TypeFunc::make(const TypeTuple *domain_sig, const TypeTuple* domain_cc,
-                               const TypeTuple *range_sig, const TypeTuple *range_cc) {
-  return (TypeFunc*)(new TypeFunc(domain_sig, domain_cc, range_sig, range_cc))->hashcons();
+                               const TypeTuple *range_sig, const TypeTuple *range_cc,
```
chhagedorn:
```suggestion
                               const TypeTuple* range_sig, const TypeTuple* range_cc,
```

src/hotspot/share/opto/type.hpp
```diff
@@ -2346,8 +2346,8 @@ class TypeNarrowKlass : public TypeNarrowPtr {
 //------------------------------TypeFunc---------------------------------------
 // Class of Array Types
 class TypeFunc : public Type {
-  TypeFunc(const TypeTuple *domain_sig, const TypeTuple *domain_cc, const TypeTuple *range_sig, const TypeTuple *range_cc)
-    : Type(Function), _domain_sig(domain_sig), _domain_cc(domain_cc), _range_sig(range_sig), _range_cc(range_cc) {}
+  TypeFunc(const TypeTuple *domain_sig, const TypeTuple *domain_cc, const TypeTuple *range_sig, const TypeTuple *range_cc, bool scalarized_return)
```
chhagedorn:
```suggestion
  TypeFunc(const TypeTuple *domain_sig, const TypeTuple* domain_cc, const TypeTuple* range_sig, const TypeTuple* range_cc, bool scalarized_return)
```

chhagedorn (approved):
A few minor things, otherwise, it looks good to me, too.

TobiHartmann:
Thanks for the reviews Quan Anh and Christian!

PR #2344, opened 2026-04-20
8382560: [lworld] Unsafe access to klass name in SystemDictionary::try_preload_from_loadable_descriptors

jsikstro:
Hello,

When looking at the usages of `InstanceKlass::get_inline_type_field_klass_or_null` I saw that `SystemDictionary::try_preload_from_loadable_descriptors` uses the returned value even though it may be null, which can result in a SIGSGV. This change aligns well with what is done just above in `SystemDictionary::preload_from_null_free_field`, using the local variable `name` instead of getting the name from the klass.

Given the right conditions this will result in a crash, but it seems rare as it requires "the stars to be aligned just right" for it to happen.

I'm not sure how much testing testing there is for classes which have inlined and non-inlined fields in them in connection to the code being changed. Such classes would result in an inline_layout_info_array having null-entires, since it is a sparse array.

Testing:
* Oracle's tier1-3
* Testing with a local reproducer that this does not crash with the changes

---------

---------
### Progress

### Issue
 * [JDK-8382560](https://bugs.openjdk.org/browse/JDK-8382560): [lworld] Unsafe access to klass name in SystemDictionary::try_preload_from_loadable_descriptors (**Bug** - P4)

### Reviewers
 * [Matias Saavedra Silva](https://openjdk.org/census#matsaave) (@matias9927 - Committer) ⚠️ Review applies to [22f68cb2](https://git.openjdk.org/valhalla/pull/2344/files/22f68cb2d24ae2aab057093e9223a4cc98a4adfc)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2344/head:pull/2344` \
`$ git checkout pull/2344`

`$ git checkout pull/2344` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2344/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2344`

`$ git pr show -t 2344`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2344.diff">https://git.openjdk.org/valhalla/pull/2344.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2344#issuecomment-4287824004)
</details>

### DIFF
--- a/src/hotspot/share/classfile/systemDictionary.cpp
+++ b/src/hotspot/share/classfile/systemDictionary.cpp
@@ -1164,7 +1164,7 @@ void SystemDictionary::try_preload_from_loadable_descriptors(InstanceKlass* ik, 
                                   "(cause: field type in LoadableDescriptors attribute) failed : "
                                   "app substituted a different version of %s",
                                   name->as_C_string(), ik->name()->as_C_string(),
-                                  k->name()->as_C_string());
+                                  name->as_C_string());
       return;
     } else if (real_k != nullptr) {
       log_info(class, preload)("Preloading of class %s during loading of shared class %s "
--- a/src/hotspot/share/classfile/systemDictionary.cpp
+++ b/src/hotspot/share/classfile/systemDictionary.cpp
@@ -1128,9 +1128,8 @@ bool SystemDictionary::preload_from_null_free_field(InstanceKlass* ik, Handle cl
     // oops, the app has substituted a different version of k! Does not fail fatally
     log_info(class, preload)("Preloading of class %s during loading of shared class %s "
                                 "(cause: null-free non-static field) failed : "
-                                "app substituted a different version of %s",
-                                name->as_C_string(), ik->name()->as_C_string(),
-                                name->as_C_string());
+                                "app substituted a different version",
+                                name->as_C_string(), ik->name()->as_C_string());
     return false;
   }
   log_info(class, preload)("Preloading of class %s during loading of shared class %s "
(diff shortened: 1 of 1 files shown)


matias9927 (approved):
Fix looks good, thanks!

I don't think this issue will need a regression test.

src/hotspot/share/classfile/systemDictionary.cpp
```diff
@@ -1164,7 +1164,7 @@ void SystemDictionary::try_preload_from_loadable_descriptors(InstanceKlass* ik,
                                   "(cause: field type in LoadableDescriptors attribute) failed : "
                                   "app substituted a different version of %s",
                                   name->as_C_string(), ik->name()->as_C_string(),
-                                  k->name()->as_C_string());
+                                  name->as_C_string());
```
xmas92: What is the difference between `name` and `k->name` here. It sounds like `name` originates from a signature. Not sure what the value is in printing `name` twice.

I'd expect something more along the lines of the following, even if we expect that `name` and `k->name` are the same symbol, it would leave breadcrumbs incase they ever diverged. 

```suggestion
                                  k != nullptr ? k->name->as_C_string() : "[name unavailable]");
```

jsikstro: Yes, name is from a signature
```
```

I don't think `name` and `k->name()` should ever diverge since both the signature and the klass from `get_inline_type_field_klass_or_null` are from the same "field_index" in the caller. If anything, I'd expect a stronger invariant, maybe with an assert, that they are the same if k is non-null.

I agree that the print could do without printing the name twice, but I don't have anything against the current version.

xmas92: Alright. I do not think I understand exactly what it means that we fail with preloading these classes. 

I think in both this code patch "app substituted a different version of %s" -> "app substituted a different version" is probably a better logging. But not relevant for this fix patch.

xmas92 (approved):

src/hotspot/share/classfile/systemDictionary.cpp
fparain: +1 on removal of the second printing of the class name.

jsikstro: I removed the second printing of the class name and adjusted the indentation to match the first line. I also adjusted the other very similar print in `SystemDictionary::preload_from_null_free_field`.

xmas92 (approved):

fparain (approved):
LGTM

jsikstro:
Thank you for the reviews everyone! I reran testing locally, which looks good.
/integrate

PR #2349, opened 2026-04-21
8381408: [lworld] Cleanup hotspot/cpu code

coleenp:
The valhalla code had refactored object allocation in TemplateTable::_new because it was called for buffering reads from flat fields.  This buffering code was moved to runtime in JDK-8375719.  Rather than refactoring OpenJDK to match this, I chose to move the object allocation code back to where it was, with the UseCompactObjectHeaders modifications.

Tested with tier1-3.

---------

---------
### Progress

### Issue
 * [JDK-8381408](https://bugs.openjdk.org/browse/JDK-8381408): [lworld] Cleanup hotspot/cpu code (**Enhancement** - P3)

### Reviewers
 * [Dean Long](https://openjdk.org/census#dlong) (@dean-long - no project role) ⚠️ Review applies to [72258d62](https://git.openjdk.org/valhalla/pull/2349/files/72258d628e18a5ae4ba8ae4497ca873f3b7d5a28)

### Contributors
 * Paul Hübner `<phubner@openjdk.org>`

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2349/head:pull/2349` \
`$ git checkout pull/2349`

`$ git checkout pull/2349` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2349/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2349`

`$ git pr show -t 2349`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2349.diff">https://git.openjdk.org/valhalla/pull/2349.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2349#issuecomment-4290804334)
</details>

### DIFF
--- a/src/hotspot/cpu/aarch64/c1_LIRGenerator_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/c1_LIRGenerator_aarch64.cpp
@@ -1147,7 +1147,6 @@ void LIRGenerator::do_NewInstance(NewInstance* x) {
   __ move(reg, result);
 }
 
-
 void LIRGenerator::do_NewTypeArray(NewTypeArray* x) {
   CodeEmitInfo* info = nullptr;
   if (x->state_before() != nullptr && x->state_before()->force_reexecute()) {
--- a/src/hotspot/cpu/aarch64/c1_LIRGenerator_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/c1_LIRGenerator_aarch64.cpp
@@ -1147,6 +1147,7 @@ void LIRGenerator::do_NewInstance(NewInstance* x) {
   __ move(reg, result);
 }
 
+
 void LIRGenerator::do_NewTypeArray(NewTypeArray* x) {
   CodeEmitInfo* info = nullptr;
   if (x->state_before() != nullptr && x->state_before()->force_reexecute()) {
--- a/src/hotspot/cpu/aarch64/aarch64.ad
+++ b/src/hotspot/cpu/aarch64/aarch64.ad
@@ -6886,7 +6886,7 @@ instruct loadConP(iRegPNoSp dst, immP con)
 
   ins_cost(INSN_COST * 4);
   format %{
-    "mov  $dst, $con\t# ptr"
+    "mov  $dst, $con\t# ptr\n\t"
   %}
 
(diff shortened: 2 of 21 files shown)


dean-long:
The JBS issue says this is about white space, but I guess it is really about making merges with mainline JDK easier.  If that's the case, I'm not sure this is a net win.  Why not just leave the refactoring if it makes _new more readable?

coleenp:
Yes, it is to make the merge to mainline easier.  I could have gone the other direction (and I have a patch that does that too), but the refactoring only applies to x86 and aarch64 and if we really wanted it to be consistent, we'd do the other cpus.  Moving it back makes it consistent.  Also, in-place in TemplateTable::_new didn't have as many register passing and stomping in another place, which is a problem with methods in the template interpreter.

The other thing I didn't like about how the refactoring ended up was that there was a version in InterpreterMacroAssembler and then another one in MacroAssembler, since the latter did not want to post DTRACE events at one time.  So the way it was makes you go looking in too many places.

We could refactor it someday.

coleenp:
Additionally, the refactoring changed some instructions at least on aarch64 that someone would have to review if we kept it.

dean-long (approved):
This looks good.  Sorry for the delay.  For some reason I thought the moved code was new in Valhalla, but in fact it is mostly the same code as mainline JDK.  There seems to be a disagreement between aarch64 and x86 on whether the loop should run forwards or backwards, but mainline has the same issue.

src/hotspot/cpu/aarch64/c1_LIRGenerator_aarch64.cpp
```diff
@@ -1147,6 +1147,7 @@ void LIRGenerator::do_NewInstance(NewInstance* x) {
   __ move(reg, result);
 }
 
+
```
jsikstro: This doesn't seem to be here in mainline.

coleenp: fixed.

coleenp:
Thanks for reviewing, Dean and Joel.
/integrate

jsikstro (approved):
👍

coleenp:
Thanks again Dean and Joel.
/integrate

PR #2350, opened 2026-04-21
8382556: [lworld] newConstructorForSerialization should not bypass strict init

liach:
ReflectionFactory.newConstructorForSerializatoin might bypass strict initialization. We should make sure the skipped classes in the constructor chain do not declare strict instance fields.

This updates the specification of jdk.unsupported's ReflectionFactory to reflect this behavioral update.

---------

---------
### Progress

### Issue
 * [JDK-8382556](https://bugs.openjdk.org/browse/JDK-8382556): [lworld] newConstructorForSerialization should not bypass strict init (**Bug** - P3)

### Reviewers
 * [Shaojin Wen](https://openjdk.org/census#swen) (@wenshao - no project role)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2350/head:pull/2350` \
`$ git checkout pull/2350`

`$ git checkout pull/2350` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2350/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2350`

`$ git pr show -t 2350`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2350.diff">https://git.openjdk.org/valhalla/pull/2350.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2350#issuecomment-4290769844)
</details>


src/java.base/share/classes/jdk/internal/reflect/MethodHandleAccessorFactory.java
```diff
@@ -117,11 +118,13 @@ static ConstructorAccessorImpl newConstructorAccessor(Constructor<?> ctor) {
      * @param decl the class to instantiate
      * @param ctor the constructor to call
      * @return an accessible constructor
+     * @throws UnsupportedOperationException if the constructor is not from
+     *         a superclass, if the instantiated class is a value class, or if
+     *         any superclass between the declaring class of the constructor and
```
dansmithcode, line 123: The "any superclass" rule should apply to value classes too.

(Here's how I described it in the serialization spec: "Then, in superclass-to-subclass order, starting with the first superclass that implements Serializable: if the class is a value class, or if it declares an instance field with the ACC_STRICT_INIT field modifier (see [JVMS 4.5](https://cr.openjdk.org/~dlsmith/jep401/jep401-20260518/specs/strict-fields-jvms.html#jvms-4.5)), an InvalidClassException is thrown.")

src/java.base/share/classes/jdk/internal/reflect/MethodHandleAccessorFactory.java
```diff
@@ -135,17 +138,24 @@ static ConstructorAccessorImpl newSerializableConstructorAccessor(Class<?> decl,
         }
     }
 
-    private static boolean constructorInSuperclass(Class<?> decl, Constructor<?> ctor) {
+    private static void checkConstructorInSuperclass(Class<?> decl, Constructor<?> ctor) {
```
dansmithcode: I'm not sure about the terminology we're using here, but I'm finding it hard to interpret this method name, suggest renaming somehow. We're not really checking the _constructor_ so much as checking all the classes on the path to the constructor.

src/java.base/share/classes/jdk/internal/reflect/MethodHandleAccessorFactory.java
liach: Do you mean to reject any superclass that is an abstract value class here?

dansmithcode: Yes. Any value class, abstract or not, with fields or not, must have its instances initialized using a real constructor.

liach: I thought this might break java.lang.Number, but after checking this shouldn't hurt. Will make this change.

dansmithcode: Oh, that's interesting. What happens to BigInteger under this rule (a Serializable identity class that extends a Serializable abstract value class)? How is it deserialized?

Maybe we're going to have to be more permissive with stateless abstract value classes...

liach: So here the first non-serializable superclass that can execute normal construction is `java.lang.Object`. Serialization spec for now says `BigInteger` is allocated and `Object`'s no-arg constructor is called, and the fields are injected. So yes, here `java.lang.Number` initialization in the middle must be skipped and that doesn't work.

It worked with `java.lang.Integer` and other wrapper classes because they use their deserialization factory instead of this allocate, call super constructor, and fill procedure.

dansmithcode: Okay, sounds like we agree we need to adjust the spec.

I will work on that change, but the expected rule is what it says here in the javadoc: the instantiated class may not be a value class, and no field declared by a Serializable class may be `STRICT_INIT`.

dansmithcode: Filed bug: [JDK-8385170](https://bugs.openjdk.org/browse/JDK-8385170)

src/java.base/share/classes/jdk/internal/value/ValueClass.java
```diff
@@ -138,4 +141,32 @@ public static Object[] copyOfRangeSpecialArray(Object[] array, int from, int to)
...
+        private static final StrictInstanceFieldClassValue INSTANCE = new StrictInstanceFieldClassValue();
+
+        private StrictInstanceFieldClassValue() {}
+
+        @Override
+        protected Boolean computeValue(Class<?> type) {
+            if ((type.getModifiers() & (Modifier.ABSTRACT | Modifier.FINAL)) == (Modifier.ABSTRACT | Modifier.FINAL)) {
+                // Not class or interface
+                return false;
+            }
```
viktorklang-ora: Might be better to put this check in `public static boolean hasStrictInstanceField(Class<?> cl)` so the cheap check doesn't need to populate the ClassValue at all.

viktorklang-ora: (We can keep the check in `computeValue` as well)

src/java.base/share/classes/jdk/internal/value/ValueClass.java
```diff
@@ -139,4 +142,44 @@ public static Object[] copyOfRangeSpecialArray(Object[] array, int from, int to)
...
+                    return true;
+                }
+            }
+            return false;
+        }
+    }
+
+    /// Returns whether a class or interface declares strict instance fields.
+    /// This does not include inherited instance fields.
+    public static boolean hasStrictInstanceField(Class<?> cl) {
```
viktorklang-ora, line 169: Is this intended as a long-term solution?

viktorklang-ora:
I've had a look over this PR, but I think it would be good to have an expert like @stuart-marks look at it as well.

src/java.base/share/classes/jdk/internal/value/ValueClass.java
liach: Yes, since this is only meaningfully used by legacy serialization, I think this ugly approach is acceptable.

src/jdk.unsupported/share/classes/sun/reflect/ReflectionFactory.java
```diff
@@ -68,6 +68,11 @@ public static ReflectionFactory getReflectionFactory() {
      * @param cl the class to instantiate
      * @param constructorToCall the constructor to call
      * @return an accessible constructor
+     * @throws UnsupportedOperationException if the constructor to call is
+     *         not suitable for the class to instantiate, such as if the
+     *         constructor is not of the given class or one of its supertype,
```
viktorklang-ora:
```suggestion
     *         constructor is not of the given class or one of its supertypes,
```

src/jdk.unsupported/share/classes/sun/reflect/ReflectionFactory.java
```diff
@@ -79,10 +84,13 @@ public Constructor<?> newConstructorForSerialization(Class<?> cl,
     /**
      * Returns an accessible no-arg constructor for a class.
      * The no-arg constructor is found searching the class and its supertypes.
+     * The no-arg constructor would ensure all strictly-initialized fields
```
viktorklang-ora:
```suggestion
    * The no-arg constructor is found searching the class and its supertypes,
    * and must ensure that all strictly-initialized fields
```

src/jdk.unsupported/share/classes/sun/reflect/ReflectionFactory.java
```diff
@@ -79,10 +84,13 @@ public Constructor<?> newConstructorForSerialization(Class<?> cl,
...
+     * declared in the class and its supertypes are properly initialized.
      *
      * @param cl the class to instantiate
-     * @return a no-arg constructor for the class or {@code null} if
-     *     the class or supertypes do not have a suitable no-arg constructor
+     * @return a no-arg constructor for the class or {@code null} if the
```
viktorklang-ora:
```suggestion
     * @return a no-arg constructor for the class, or {@code null} if the
```

viktorklang-ora:
@liach I've reviewed this PR, and have no further comments at this time. It would be beneficial to have another Core Libraries reviewer chime in here as well, maybe @AlanBateman might be able to give a second opinion.

src/jdk.unsupported/share/classes/sun/reflect/ReflectionFactory.java
```diff
@@ -68,6 +68,11 @@ public static ReflectionFactory getReflectionFactory() {
      * @param cl the class to instantiate
      * @param constructorToCall the constructor to call
      * @return an accessible constructor
+     * @throws UnsupportedOperationException if the constructor to call is
+     *         not suitable for the class to instantiate, such as if the
+     *         constructor is not of the given class or one of its supertypes,
+     *         or this constructor would skip initialization of
+     *         strictly-initialized fields
      */
     public Constructor<?> newConstructorForSerialization(Class<?> cl,
```
AlanBateman, line 77: The 1-arg newConstructorForSerialization returns null, the 2-arg newConstructorForSerialization may return null or throw UOE. Is that existing behavior (before value classes and strict-init fields), just checking.

(although this is the jdk.unsupported module, the ReflectionFactory "API" is there for 3rd party serialization libraries to continue to work until legacy serialization goes away)

liach: That is the existing behavior on mainline, from `jdk.internal.reflect.ReflectionFactory::generateConstructor`. (The 2-arg version never returns null, that might be a misunderstanding from your part)

The 1-arg version on mainline automatically finds a suitable no-arg constructor from the first non-serializable superclass of the specified class, returns null the specified class is not serializable, no constructor is suitable, but it ensures the found constructor never fails with a UOE once passed to `generateConstructor`.

The 2-arg version just has a shortcut to avoid generation if `constructorToCall` is declared in the specified class. It delegates everything to `generateConstructor`, which is a method that either returns successfully or throws UOE in `MethodHandleAccessorFactory::newSerializableConstructorAccessor`.

Thus, when I updated these APIs, I made the 1-arg version return null for a value class or a strict field calss, and the 2-arg version throw UOE.

dlsmith-dev (approved):
Looks good

wenshao (approved):
No issues found. LGTM! ✅ _— qwen3.7-max via Qwen Code /review_

liach:
Thanks for the reviews so far. I will integrate this to lworld so it can be tested on CI regularly.

/integrate

src/jdk.unsupported/share/classes/sun/reflect/ReflectionFactory.java
AlanBateman: Okay, we can discuss this again when the changes make it to the PR for main line.

src/jdk.unsupported/share/classes/sun/reflect/ReflectionFactory.java
```diff
@@ -83,12 +83,12 @@ public Constructor<?> newConstructorForSerialization(Class<?> cl,
 
     /**
      * Returns an accessible no-arg constructor for a class.
-     * The no-arg constructor is found searching the class and its supertypes.
-     * The no-arg constructor would ensure all strictly-initialized fields
+    * The no-arg constructor is found searching the class and its supertypes,
+    * and must ensure that all strictly-initialized fields
```
ExE-Boss, line 87: Wrong indentation:
```suggestion
     * The no-arg constructor is found searching the class and its supertypes,
     * and must ensure that all strictly-initialized fields
```

ExE-Boss (changes requested):

PR #2351, opened 2026-04-22
8380927: [lworld] C2: Missed Value() optimization for InlineType

benoitmaillard:
This PR prevents a missed optimization caused by a breach of idempotence in `InlineTypeNode::Value` when handling contradicting method return profiles.

## Conflicting return profiles

Suppose we have the following java code, and that we have record profile data for both callsites (`call(...)` and `callWrapper(...)`):

```java
static MyValue callWrapper(...) {
    return call(...); // callsite #1
}

static MyValue test() {
    return callWrapper(...); // callsite #2
}
```

It could be that `test` is not the only place where we call `callWrapper(...)`, and that the other call sites result in wildy different executions of callsite `#1` with wildly different return values at that callsite.

Suppose that callsite `#1` always returns `null`, except when `callWrapper` is called from callsite `#2` where it throws an exception. In `ciMethod::return_profiled_type`, we end up with the following `ProfilePtrKind` when compiling `test`:
- For callsite `#1`: `ProfileAlwaysNull`, because we have only observed `null` (the rest of the time an exception was thrown, so we didn't see anything)
- For callsite `#2`: `ProfileNeverNull`, because we have never observed a return value and this is basically the default

Please look at the test from this PR for a more complete example.

These two profiles contradict each other. If we do incremental inlining (probably not limited to it, but easier to reproduce), we can easily end up in a situation where a speculative type derived from the return of the callsite is `NotNull` before inlining the outer callsite, and then moves to `ptr:null` after inlining. This can confuse `Value` implementations that rely on `filter_speculative`. This is handled gracefully for `ConstraintCastNode`, but not for `InlineTypeNode`.

## Missed optimization

In `InlineTypeNode::Value`, we have:
https://github.com/openjdk/valhalla/blob/e391f76e7db834f50c1a0af466daf95a65a79f23/src/hotspot/share/opto/inlinetypenode.cpp#L1619

Suppose we have the following types (only the speculative part is shown):
```
oop: instptr:compiler/valhalla/inlinetypes/TestPollutedCallsiteProfileInlineType$MyValue:NotNull:exact *,iid=bot (flat in array)
_type: instptr:compiler/valhalla/inlinetypes/TestPollutedCallsiteProfileInlineType$MyValue:NotNull:exact *,iid=bot (flat in array)
```

After this step we have `t = oop = _type`.

The oop input suddenly changes types because of incremental inlining, as explained in the previous section. It goes from `NotNull` to `ptr:null` because the type from the polluted profile propagates.

Then at the next IGVN pass, we have:
```
oop: ptr:null
_type: instptr:compiler/valhalla/inlinetypes/TestPollutedCallsiteProfileInlineType$MyValue:NotNull:exact *,iid=bot (flat in array)
```

Intuitively, the resulting type should be `TOP`. But `TypePtr::cleanup_speculative` does not like it, and we end up with `t.speculatve() = nullptr`:

https://github.com/openjdk/valhalla/blob/e391f76e7db834f50c1a0af466daf95a65a79f23/src/hotspot/share/opto/type.cpp#L2966-L2979

We end up with no speculative type, which means `speculative = non_speculative`. Then in verification we do another `Value` pass:

```
oop: ptr:null
_type: nullptr
```

This gives us `ptr:null` for the resulting speculative type. This breaks the idempotence of `Value`, and is reported as missed optimization.

## Fix

I think ideally we would like to be able to have `TOP` as a valid speculative type (this is the most accurate it could get), but this would clearly break a lot of things (I have tried). I think for now we should have the same behavior as in `ConstraintCastNode::Value`: detect that the speculative type was suppressed by `cleanup_speculative`, and call `filter_speculative` again when it is the case. This is roughly the same as [JDK-8371716](https://bugs.openjdk.org/browse/JDK-8371716).

### Testing

This issue was initially as a failure in `test59` of `TestCallingConvention.java` with `-XX:-TieredCompilation -XX:VerifyIterativeGVN=1110`. I think this specific test exhibits some quite interesting behavior, but is quite hard to debug (it's not the first time it is failing). For this reason I really wanted to reduce this to a less opaque test case, and I think this test pattern (especially the profiling part) could be useful in the future as well.

- [x] tier1-3, plus some internal testing

Thank you for reviewing!

---------

---------
### Progress

### Issue
 * [JDK-8380927](https://bugs.openjdk.org/browse/JDK-8380927): [lworld] C2: Missed Value() optimization for InlineType (**Bug** - P5)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2351/head:pull/2351` \
`$ git checkout pull/2351`

`$ git checkout pull/2351` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2351/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2351`

`$ git pr show -t 2351`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2351.diff">https://git.openjdk.org/valhalla/pull/2351.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2351#issuecomment-4296168013)
</details>

### DIFF
--- a/src/hotspot/share/opto/inlinetypenode.cpp
+++ b/src/hotspot/share/opto/inlinetypenode.cpp
@@ -1617,6 +1617,12 @@ const Type* InlineTypeNode::Value(PhaseGVN* phase) const {
     return Type::TOP;
   }
   const Type* t = toop->filter_speculative(_type);
+  // Because of contradicting type profiling, we can end up with top as speculative type,
+  // which would then get removed by cleanup_speculative. In this case we have to run filter_speculative
+  // again, otherwise we would break the idempotence of Value
+  if (t->speculative() == nullptr && toop->speculative() != nullptr) {
+    t = toop->filter_speculative(t);
+  }
   if (t->singleton()) {
     // Don't replace InlineType by a constant
     t = _type;
--- a/test/hotspot/jtreg/compiler/valhalla/inlinetypes/TestCallingConvention.java
+++ b/test/hotspot/jtreg/compiler/valhalla/inlinetypes/TestCallingConvention.java
@@ -1490,21 +1490,20 @@ public class TestCallingConvention {
     }
 
     // Method handle with a scalarized return that will always throw an exception
-    // TODO JDK-8380927 this should be re-introduced once fixed
-    // @Test
-    // public static MyValue2 test59(MyValue2 val) throws Throwable {
-    //     return (MyValue2)test59_mh.invokeExact(val);
-    // }
+    @Test
+    public static MyValue2 test59(MyValue2 val) throws Throwable {
+        return (MyValue2)test59_mh.invokeExact(val);
+    }
 
-    // @Run(test = "test59")
-    // @Warmup(10000) // Trigger compilation of LambdaForm method
-    // public void test59_verifier() throws Throwable {
-    //     try {
-    //         test59(null);
-    //     } catch (ArithmeticException e) {
-    //         // Expected
-    //     }
-    // }
+    @Run(test = "test59")
+    @Warmup(10000) // Trigger compilation of LambdaForm method
+    public void test59_verifier() throws Throwable {
+        try {
+            test59(null);
+        } catch (ArithmeticException e) {
+            // Expected
+        }
+    }
 
     static interface Interface60 {
         public void m(MyValue2 val);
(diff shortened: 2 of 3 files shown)


merykitty (approved):
Thanks, the fix makes sense to me.

TobiHartmann (approved):
Looks good to me too. Thanks for fixing this.

benoitmaillard:
Thank you for the reviews @TobiHartmann @merykitty!

/integrate

PR #2353, opened 2026-04-22
8381472: [lworld] Fast path for acmp for migrated classes

marc-chevalier:
`acmp` for value object might be expensive: it may need substitutability test, which is done through a Java call. That is the last resort when everything failed, that is: we are in the presence of two value objects of the same class, not statically known or speculated (profile absent or polluted). While a lot of the acmp logic is trying to improve acmp at compile-time, this change proposes a runtime check to take a fast path.

The idea is that if the object is small enough, we can read all the fields at once with a single 8-byte load, we exclude the non-interesting bits with a mask, and we can simply compare the masked value from both objects.

The first step happens at class loading. When the class is loaded, we check whether all the fields can be read at once, from which offset, and the mask to filter out useless bits. Useless bits may have two sources:
- padding between fields
- header: this happens because we cannot necessarily start reading at `payload_offset`, the payload might be smaller than a long, and we could overread. On the other hand, the header is always long enough (never less than 8 bytes), so we may read from somewhere in the header to make sure we don't go further than the payload. This piece of header must be filtered out.

This strategy may not apply if the object is too big, or if it contains an oop. Loading oop without precaution, concurrently with the GC is not a good idea. This limitation is rather harmless: this fast path is mostly made not to pay the price of a complicated `acmp` for magically migrated classes (such as `Integer`), which are small and don't contain oops.

There is an interesting corner case: the mask can be 0 for an empty value class. We use a negative offset to signal that the fast path doesn't apply, since it is not a valid value for the offset: we shouldn't load before the object.

The second step is to prepare the runtime check. In `do_acmp`, when we used to prepare the Java call, now we have a parallel fast path. The condition uses the fact that we already have the class node available (used for checking operands have the same types). We can get the fast acmp offset from the class, check if it looks valid (that is non-negative), and if so, take the fast path that will load the mask, do the two loads and masking, and compare the given values. If the class of the operand is known precisely at parsing, we don't emit the fast path since the call will be nicely intrinsified.

The third step is to prevent the fast path to interfere with the faster path. When the class is not known at parsing-time, but after some optimizations it becomes precise, we already emitted the fast path for acmp while we now can intrinsify the substitutability test. That means, not only we will still pay the price of the fast path test, but also, we might not even take the intrinsified substitutability test (if the fast path applies). Thus, comes a hack. It's not very nice, but it's not too bad. When we are about to intrinsify, we check around if we can find the fast path test. This test is rather specific because nothing else needs the address of the fast acmp offset. If we find such a test, we simply replace it a constant, not to take the fast path. This works because C2 is unable to fold the load of the fast acmp offset/mask, even when the class is a constant, and that the check was carefully crafted to be already in ideal form. Since the check is quite precise, we don't risk to mess with an unrelated `If`. It is also not too bad if this test is not eliminated: we might end up taking the fast path instead of the faster path.

This is not very nice, and not a definitive version. A way forward is probably through some macro node that gather all the resources needed to generate the acmp IR, and that can pick when is the best moment.

But... does it work? It was successfully tested on `tier1,tier2,tier3,tier4,hs-comp-stress,valhalla-comp-stress`, and also with `+UseCompactObjectHeaders` (because one can be worried about loads).

But how well does it perform? In average, some benchmarks show improvements, and other are not affected. I've also wrote a microbench:

with fast path:
```
Mode  Cnt  Score   Error  Units
FastPath.big_mix                      avgt   15  1.487 ± 0.039  ns/op
FastPath.custom_multi_field           avgt   15  1.172 ± 0.021  ns/op
FastPath.custom_single_field          avgt   15  1.217 ± 0.034  ns/op
FastPath.empty_objects                avgt   15  0.632 ± 0.013  ns/op
FastPath.empty_objects_heterogeneous  avgt   15  0.896 ± 0.037  ns/op
FastPath.heterogeneous_eq             avgt   15  1.217 ± 0.040  ns/op
FastPath.heterogeneous_eq_neq         avgt   15  1.309 ± 0.039  ns/op
FastPath.heterogeneous_neq            avgt   15  1.191 ± 0.026  ns/op
FastPath.homogeneous_eq               avgt   15  0.739 ± 0.021  ns/op
FastPath.homogeneous_eq_neq           avgt   15  0.872 ± 0.021  ns/op
FastPath.homogeneous_neq              avgt   15  0.867 ± 0.014  ns/op
FastPath.identity_objects             avgt   15  0.462 ± 0.011  ns/op
FastPath.numeric_classes              avgt   15  1.207 ± 0.034  ns/op
FastPath.too_big                      avgt   15  4.733 ± 0.131  ns/op
FastPath.with_oops                    avgt   15  2.668 ± 0.082  ns/op
```
without:
```
Benchmark                             Mode  Cnt  Score   Error  Units
FastPath.big_mix                      avgt   15  2.274 ± 0.076  ns/op
FastPath.custom_multi_field           avgt   15  4.722 ± 0.112  ns/op
FastPath.custom_single_field          avgt   15  3.928 ± 0.118  ns/op
FastPath.empty_objects                avgt   15  0.630 ± 0.012  ns/op
FastPath.empty_objects_heterogeneous  avgt   15  1.651 ± 0.052  ns/op
FastPath.heterogeneous_eq             avgt   15  6.952 ± 0.163  ns/op
FastPath.heterogeneous_eq_neq         avgt   15  5.557 ± 0.179  ns/op
FastPath.heterogeneous_neq            avgt   15  4.298 ± 0.150  ns/op
FastPath.homogeneous_eq               avgt   15  0.752 ± 0.019  ns/op
FastPath.homogeneous_eq_neq           avgt   15  0.869 ± 0.016  ns/op
FastPath.homogeneous_neq              avgt   15  0.874 ± 0.019  ns/op
FastPath.identity_objects             avgt   15  0.463 ± 0.014  ns/op
FastPath.numeric_classes              avgt   15  3.751 ± 0.050  ns/op
FastPath.too_big                      avgt   15  4.704 ± 0.148  ns/op
FastPath.with_oops                    avgt   15  2.455 ± 0.097  ns/op
```

We can see a few cases. For instance:
- `too_big` or `with_oops` are cases where the fast path doesn't apply. So, on top of taking always the slow path, we pay at runtime the cost of the test.
- all the `homogeneous_*`, here profiling does wonders: the type is speculated, and we are taking the intrinsified test. The fast path is not generated (or removed).
- `heterogeneous_*`, here profiling is polluted, but all the classes can use the fast path, the win is quite huge
- `big_mix` has everything (identity class, value classes with and without fast path and null), the fast path is generated, but not always taken. The win is still there.

The worst case of this change is when the `acmp` involves value objects that are never candidate to the fast path. If it becomes problematic, I think we could consider profiling that, in the same way we profile the occurrences of `null` or of identity classes. If we have a profile that shows that none of the classes seen before would be eligible for the fast path, we could skip it. If the profile shows that fast path is possible, it wouldn't need a lot of fast path-compatible value objects to make it overall worth it. I'm not convinced this level of details make sense at the moment: most of the value classes are the automatically migrated onces and most of them are fast path-compatible.

Thanks,
Marc

---------

---------
### Progress

### Issue
 * [JDK-8381472](https://bugs.openjdk.org/browse/JDK-8381472): [lworld] Fast path for acmp for migrated classes (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2353/head:pull/2353` \
`$ git checkout pull/2353`

`$ git checkout pull/2353` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2353/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2353`

`$ git pr show -t 2353`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2353.diff">https://git.openjdk.org/valhalla/pull/2353.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2353#issuecomment-4296621572)
</details>


src/hotspot/share/code/nmethod.cpp
```diff
@@ -3492,11 +3461,11 @@ void nmethod::print_on_impl(outputStream* st) const {
                                              p2i(oops_begin()),
                                              p2i(oops_end()),
                                              oops_size());
-  if (mutable_data_size() > 0) st->print_cr(" mutable data [" INTPTR_FORMAT "," INTPTR_FORMAT "] = %d",
+  if (mutable_data_size () > 0) st->print_cr(" mutable data   [" INTPTR_FORMAT "," INTPTR_FORMAT "] = %d",
```
marc-chevalier, line 3495: Not actually related, but I was annoyed at the misaligned output.

src/hotspot/share/oops/instanceKlass.cpp
```diff
@@ -4155,10 +4155,10 @@ void InstanceKlass::print_on(outputStream* st) const {
   st->print(BULLET"local interfaces:  "); local_interfaces()->print_value_on(st);      st->cr();
   st->print(BULLET"trans. interfaces: "); transitive_interfaces()->print_value_on(st); st->cr();
 
-  st->print(BULLET"secondary supers: "); secondary_supers()->print_value_on(st); st->cr();
+  st->print(BULLET"secondary supers:  "); secondary_supers()->print_value_on(st); st->cr();
```
marc-chevalier, line 4158: More alignment issues.

test/hotspot/jtreg/compiler/valhalla/inlinetypes/TestAcmpFastPath.java
```diff
@@ -0,0 +1,409 @@
...
+        return a == b;
+    }
+    @Test
+    @IR(phase = {CompilePhase.AFTER_PARSING}, counts = {AND_L, "3", STATIC_CALL_OF_METHOD, "isSubstitutable.*", "1"})
+    @IR(phase = {CompilePhase.PRINT_IDEAL}, counts = {AND_L, "3", STATIC_CALL_OF_METHOD, "isSubstitutable.*", "1"})
+    boolean neq_object(Object a, Object b) {
+        return a != b;
+    }
+
+    // No acm fast path: the type is precise, and the call will be intrinsified
```
TobiHartmann: Should be `acmp, a few more occurrences below and in the other test.

src/hotspot/share/oops/inlineKlass.hpp
```diff
@@ -99,6 +99,19 @@ class InlineKlass: public InstanceKlass {
     int _null_marker_offset;       // expressed as an offset from the beginning of the object for a heap buffered value
                                    // payload_offset must be subtracted to get the offset from the beginning of the payload
 
+    /* When we can't intrinsify the substitutability check, we can still avoid the call to isSubstitutable at runtime if the value object is small enough.
+     * If all the fields are contained at once in a single long, we can load such a long on both operand, use a bitwise mask to remove the extra bits
```
TobiHartmann:
```suggestion
     * If all the fields are contained at once in a single long, we can load such a long from both operands, use a bitwise mask to remove the extra bits
```

src/hotspot/share/opto/parse2.cpp
```diff
@@ -2373,6 +2430,81 @@ void Parse::do_acmp(BoolTest::mask btest, Node* left, Node* right) {
...
+
+  if (cmp_i->in(1)->Opcode() != Op_LoadI) return nullptr;
+  LoadNode* load_offset = cmp_i->in(1)->as_Load();
+  if (!cmp_i->in(2)->is_ConI()) return nullptr;
+  ConINode* zero_i = cmp_i->in(2)->as_ConI();
+  assert(zero_i->type()->is_int() != nullptr, "");
+  if (!zero_i->type()->is_int()->is_con(0)) return nullptr;
+
+  assert(load_offset->in(2) != nullptr, "");
+  if(!load_offset->in(2)->is_AddP()) return nullptr;
```
TobiHartmann:
```suggestion
  if (!load_offset->in(2)->is_AddP()) return nullptr;
```

src/hotspot/share/opto/parse2.cpp
```diff
@@ -2373,6 +2430,81 @@ void Parse::do_acmp(BoolTest::mask btest, Node* left, Node* right) {
...
+  assert(offset_addr_add->in(AddPNode::Base) != nullptr, "");
+  assert(offset_addr_add->in(AddPNode::Address) != nullptr, "");
+  assert(offset_addr_add->in(AddPNode::Offset) != nullptr, "");
+  if (!offset_addr_add->in(AddPNode::Base)->is_top()) return nullptr;
+  if (offset_addr_add->in(AddPNode::Address)->Opcode() != Op_LoadP) return nullptr;
+  LoadNode* load_members = offset_addr_add->in(AddPNode::Address)->as_Load();
+  if (!is_con_offset(offset_addr_add->in(AddPNode::Offset), InlineKlass::fast_acmp_offset_offset())) return nullptr;
+
+  assert(load_members->in(2) != nullptr, "");
+  if(!load_members->in(2)->is_AddP()) return nullptr;
```
TobiHartmann:
```suggestion
  if (!load_members->in(2)->is_AddP()) return nullptr;
```

TobiHartmann (approved):
Great work Marc. This looks good to me but someone from the runtime team should also take a look. Maybe @fparain?

src/hotspot/share/oops/inlineKlass.hpp
```diff
@@ -191,6 +204,12 @@ class InlineKlass: public InstanceKlass {
   void set_null_marker_offset(int offset)                     { members()._null_marker_offset = offset; }
   int null_marker_offset_in_payload() const                   { return null_marker_offset() - payload_offset(); }
 
+  int64_t fast_acmp_offset() const                              { return members()._fast_acmp_offset; }
```
TobiHartmann: Indentation is off here.

test/hotspot/jtreg/compiler/valhalla/inlinetypes/TestAcmpFastPath.java
marc-chevalier: Thanks to the power of copy-paste, I can do the same trivial typo an unbounded amount of times.

src/hotspot/share/oops/inlineKlass.hpp
marc-chevalier: Is it a coincidence that this is 2 characters too long, and there is exactly 2 more characters in `fast_acmp_offset` than in `fast_acmp_mask`? Once again, copy-paste-jutsu.

src/hotspot/share/classfile/classFileParser.cpp
```diff
@@ -5355,6 +5357,84 @@ void ClassFileParser::create_acmp_maps(InstanceKlass* ik, TRAPS) {
...
+    mask <<= leading_zeroes;
+    assert(count_leading_zeros(mask) == 0, "fast acmp mask can be moved further!");
+    int offset = _layout_info->_payload_offset - leading_zeroes / BitsPerByte;
+    assert(offset >= 0, "fast acmp path shouldn't load before the object");
+    vk->set_fast_acmp_offset(offset);
+    vk->set_fast_acmp_mask(mask);
+  }
+
+#else
+  /* Leaving the mask and offset to default values (that is just return;") is a correct and easy way to implement this for other endianness, but it will
```
coleenp: Drive by comments: The hotspot coding standard is to use // comments not /* */ ones.  Can you change these and all the ones below?

src/hotspot/share/classfile/classFileParser.cpp
```diff
@@ -5355,6 +5357,84 @@ void ClassFileParser::create_acmp_maps(InstanceKlass* ik, TRAPS) {
...
+   * have a runtime cost in do_acmp, without the benefit. It is better, and probably easy with access to such an architecture, to adapt the logic above
+   * for big-endian architectures, but filling the mask from the other end. I prefer not to do it blindly.
+   */
+  Unimplemented()
+#endif
+}
```
coleenp, line 5430: I think this and the method above this aren't reallly parsing any classfiles.  They should be moved to be members of InstanceKlass instead.

src/hotspot/share/classfile/classFileParser.cpp
```diff
@@ -5355,6 +5357,84 @@ void ClassFileParser::create_acmp_maps(InstanceKlass* ik, TRAPS) {
...
+
```
coleenp, line 5431: This is very complicated and needs more commentary in the source file about what all this compression and casting is about.
If it only yields a small performance increase, should we really have this?

coleenp (commented):
I have a couple of comments about this code.

src/hotspot/share/classfile/classFileParser.cpp
```diff
@@ -5355,6 +5357,84 @@ void ClassFileParser::create_acmp_maps(InstanceKlass* ik, TRAPS) {
...
+void ClassFileParser::set_fast_acmp_members(InlineKlass* vk) const {
+#ifdef VM_LITTLE_ENDIAN
+  GrowableArray<InlineKlassAtOffset> worklist;
+
+  auto make_mask_piece = [](int start, int size)-> int64_t { return right_n_bits<int64_t>(size * BitsPerByte) << (start * BitsPerByte); };
+  int64_t mask = 0;
+
+  worklist.push(InlineKlassAtOffset{0, vk});
+  while (!worklist.is_empty()) {
+    InlineKlassAtOffset inline_klass_at_offset = worklist.pop();
```
fparain: Why is this code constructing the mask from the field meta-data instead of re-using the acmp_maps?

marc-chevalier:
@fparain Yes, that's better of course. At first, I thought I'd need more details, but actually no. Now, `set_fast_acmp_members` is somewhat similar to `create_acmp_maps`.

This new version should be equivalent to the old one: I ran some tests with both in, and assert they give the same results. It seems good. I've reduced the endianness-specific bits.

There is now a flag to disable this fast path, as requested.

marc-chevalier:
As for performance, the microbench shows that the cost of the test is small compared to the potential win, so a small fraction of fast path should be enough for being beneficial. But there are also real benchmarks:
- Renaissance-ChiSquare: the gap between mainline and valhalla with enable-preview seems to drop from 30-45% to ~20%
- Renaissance-DecTree: the gap reduction is more modest, but still seems observable, from ~21% to ~18%
- Renaissance-NaiveBayes: this one doesn't seem to be improved (not negatively either)

I think the outcome is worth doing this fast path, especially now that the computation of the mask is much simpler.

TobiHartmann (approved):
Nice simplification by using the `_nonoop_acmp_map`. The new version looks good to me and the benchmark results show that this is worth doing.

src/hotspot/share/classfile/classFileParser.cpp
```diff
@@ -5355,6 +5355,60 @@ void ClassFileParser::create_acmp_maps(InstanceKlass* ik, TRAPS) {
...
+  for (int i = 0; i < _layout_info->_nonoop_acmp_map->length(); i++) {
+    int field_start = _layout_info->_nonoop_acmp_map->at(i)._offset - _layout_info->_payload_offset;
+    int field_size = _layout_info->_nonoop_acmp_map->at(i)._size;
+    int field_end = field_start + field_size - 1;
+    if (field_end >= BytesPerLong) {  // Too far! Can't fit in an 8-byte load, fast path will not be taken
+      return;
+    }
+    int64_t mask_piece = make_mask_piece(field_start, field_size);
+    mask |= mask_piece;
+  }
```
fparain, line 5395: A small description of the algorithm used to compute the mask would be welcome (may be with a little example). Some details, like the fact it is filled starting from the least significant bit is not obvious at first, and explains the need for the shift below.
I've checked it on paper with a few examples, but review would have been faster with a description.

src/hotspot/share/classfile/classFileParser.cpp
```diff
@@ -5355,6 +5355,60 @@ void ClassFileParser::create_acmp_maps(InstanceKlass* ik, TRAPS) {
...
+  auto make_mask_piece = [](int start, int size)-> int64_t { return right_n_bits<int64_t>(size * BitsPerByte) << (start * BitsPerByte); };
+#else
+  // Leaving the mask and offset to default values (that is just "return;") is a correct and easy way to implement this for other endianness, but it will
+  // have a runtime cost in do_acmp, without the benefit. It is better, and probably easy with access to such an architecture, to adapt the logic above
+  // for big-endian architectures, but filling the mask from the other end. I prefer not to do it blindly.
+  Unimplemented()
+#endif
+
+  for (int i = 0; i < _layout_info->_nonoop_acmp_map->length(); i++) {
+    int field_start = _layout_info->_nonoop_acmp_map->at(i)._offset - _layout_info->_payload_offset;
```
fparain: The entries in the acmp_maps are not fields, they are just pieces of contiguous memory used by one or several fields. An entry can cover multiple fields, or a field can be split over several entries. Renaming the local variables in this method avoiding the term "field" would prevent confusion.

src/hotspot/share/classfile/classFileParser.cpp
```diff
@@ -5355,6 +5355,60 @@ void ClassFileParser::create_acmp_maps(InstanceKlass* ik, TRAPS) {
...
+    // accept null argument.
+    vk->set_fast_acmp_offset(0);
+    vk->set_fast_acmp_mask(mask);
+  } else {
+#ifdef VM_LITTLE_ENDIAN
+    int leading_zeroes = static_cast<int>(count_leading_zeros(mask));
+    assert(leading_zeroes % BitsPerByte == 0, "we should mask full bytes");
+    mask <<= leading_zeroes;
+    assert(count_leading_zeros(mask) == 0, "fast acmp mask can be moved further!");
+    int offset = _layout_info->_payload_offset - leading_zeroes / BitsPerByte;
```
fparain, line 5420: The fast_acmp_offset is not guaranteed to be aligned in any particular way (not on word boundary for instance). Is it going to be OK for code using it?

marc-chevalier: Yes, the access is known to be mismatch and unaligned. It is already tested by anything whose payload is too short to keep the whole long busy. For instance, `Integer`.

fparain (approved):
LGTM.
Thank you for taking my comments into consideration.

TobiHartmann (approved):
Still looks good to me too. Thanks Marc!

marc-chevalier:
/integrate

Thanks all for review!

PR #2356, opened 2026-04-22
8382814: [lworld] C2: crash with late inlining of call with circular value argument

rwestrel:
When doing late inlining, `GraphKit::access_load_at()` may be called
to populate an `InlineTypeNode`. When it executes, it, sometimes,
causes the `map` used by the `GraphKit` instance to change. When that
happens, `CallGenerator::do_late_inline_helper()` continues execution
with an outdated `map` which causes issues during parsing.

---------

---------
### Progress

### Issue
 * [JDK-8382814](https://bugs.openjdk.org/browse/JDK-8382814): [lworld] C2: crash with late inlining of call with circular value argument (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2356/head:pull/2356` \
`$ git checkout pull/2356`

`$ git checkout pull/2356` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2356/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2356`

`$ git pr show -t 2356`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2356.diff">https://git.openjdk.org/valhalla/pull/2356.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2356#issuecomment-4296899287)
</details>

### DIFF
--- a/src/hotspot/share/opto/callGenerator.cpp
+++ b/src/hotspot/share/opto/callGenerator.cpp
@@ -701,6 +701,8 @@ void CallGenerator::do_late_inline_helper() {
         // field of the inline type. Build InlineTypeNodes from the inline type arguments.
         GraphKit arg_kit(jvms, &gvn);
         Node* vt = InlineTypeNode::make_from_multi(&arg_kit, call, t->inline_klass(), j, /* in= */ true, /* null_free= */ !t->maybe_null());
+        // GraphKit::access_load_at() may be called from InlineTypeNode::make_from_multi() and it may change the map
+        // that arg_kit uses.
         map = arg_kit.map();
         map->set_control(arg_kit.control());
         map->set_argument(jvms, i1, vt);
--- a/src/hotspot/share/opto/callGenerator.cpp
+++ b/src/hotspot/share/opto/callGenerator.cpp
@@ -701,6 +701,7 @@ void CallGenerator::do_late_inline_helper() {
         // field of the inline type. Build InlineTypeNodes from the inline type arguments.
         GraphKit arg_kit(jvms, &gvn);
         Node* vt = InlineTypeNode::make_from_multi(&arg_kit, call, t->inline_klass(), j, /* in= */ true, /* null_free= */ !t->maybe_null());
+        map = arg_kit.map();
         map->set_control(arg_kit.control());
         map->set_argument(jvms, i1, vt);
       } else {
--- a/test/hotspot/jtreg/compiler/valhalla/inlinetypes/TestLateInliningIncorrectArgs.java
+++ b/test/hotspot/jtreg/compiler/valhalla/inlinetypes/TestLateInliningIncorrectArgs.java
@@ -25,7 +25,7 @@
  * @test
  * @enablePreview
  * @requires (os.simpleArch == "x64" | os.simpleArch == "aarch64")
- * @run main/othervm -XX:-BackgroundCompilation -XX:+AlwaysIncrementalInline ${test.main.class}
+ * @run main/othervm -XX:-BackgroundCompilation -XX:+IgnoreUnrecognizedVMOptions -XX:+AlwaysIncrementalInline ${test.main.class}
  */
 
 package compiler.valhalla.inlinetypes;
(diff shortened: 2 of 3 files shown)


src/hotspot/share/opto/callGenerator.cpp
```diff
@@ -715,6 +701,9 @@ void CallGenerator::do_late_inline_helper() {
         // field of the inline type. Build InlineTypeNodes from the inline type arguments.
         GraphKit arg_kit(jvms, &gvn);
         Node* vt = InlineTypeNode::make_from_multi(&arg_kit, call, t->inline_klass(), j, /* in= */ true, /* null_free= */ !t->maybe_null());
+        // GraphKit::access_load_at() may be called from InlineTypeNode::make_from_multi() and it may change the map
+        // that arg_kit uses.
+        map = arg_kit.map();
```
merykitty, line 706: Should we have a single `GraphKit` that is used to collect all arguments? What if we have 2 arguments that affect the map of `arg_kit`, it seems that it will cause the `map` to diverge, am I right?

rwestrel: Something like the test I just pushed? It seems to work ok.
`GraphKit::SavedState::SavedState()` and `GraphKit::SavedState::~SavedState()` is where the new map originates from and AFAICT, `SavedState` wires things up so it doesn't matter if we have multiple changes in map.

TobiHartmann: Right, I think it's fine.

TobiHartmann (approved):
Looks good to me.

src/hotspot/share/opto/callGenerator.cpp
dean-long, line 706: We also do something similar in Compile::build_start_state(), so does it also need to be changed?
If the map changes, which map should we call disconnect_inputs on at line 721?  Only the new map, the old map, or both?

rwestrel: > We also do something similar in Compile::build_start_state(), so does it also need to be changed?

I don't see how things can go wrong there. I added an assert anyway.

> If the map changes, which map should we call disconnect_inputs on at line 721? Only the new map, the old map, or both?

Good question. I'm not sure line 721 is required. Late inlining, in general, leaves plenty of dead partially connected nodes behind that are collected when `Compile::inline_incrementally_cleanup()` is called.

test/hotspot/jtreg/compiler/valhalla/inlinetypes/TestLateInliningIncorrectArgs.java
```diff
@@ -0,0 +1,91 @@
...
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+/**
+ * @test
+ * @enablePreview
+ * @requires (os.simpleArch == "x64" | os.simpleArch == "aarch64")
+ * @run main/othervm -XX:-BackgroundCompilation -XX:+AlwaysIncrementalInline ${test.main.class}
```
TobiHartmann: `Error: VM option 'AlwaysIncrementalInline' is develop and is available only in debug version of VM.`

rwestrel: fixed in new commit

TobiHartmann (approved):
Looks good to me!

rwestrel:
@merykitty @dean-long any objection that I integrate this PR?

merykitty (approved):
Looks good to me, then.

dean-long:
If you don't mind, I'd like to take a closer look at GraphKit::SavedState.  It seems like it should be possible keep the same map when we restore.

dean-long (approved):

dean-long:
OK, I think GraphKit::SavedState could be rewritten to keep the same map, but I think it would be a much bigger and riskier change.

rwestrel:
@TobiHartmann @merykitty @dean-long thanks for the reviews

PR #2357, opened 2026-04-22
8381531: [lworld] Startup performance regression with Valhalla caused by Method::link_method

matias9927:
The method `Method::link_method()` has a notable impact on overhead and results in notable performance regressions from mainline. The source of this performance difference seems to be the aggressive calculation and recalculation of signature calling conventions for methods. This patch introduces the possibility for methods to inherit adapters from super methods so long as its calling conventions are compatible with the calling conventions introduced by local interfaces. 

This design reports a x-y% improvment.

Verified with tier 1-5 tests.

---------

---------
### Progress
- [ ] Change must be properly reviewed (1 review required, with at least 1 [Committer](https://openjdk.org/bylaws#committer))

### Issue
 * [JDK-8381531](https://bugs.openjdk.org/browse/JDK-8381531): [lworld] Startup performance regression with Valhalla caused by Method::link_method (**Bug** - P3) ⚠️ Issue is not open.

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2357/head:pull/2357` \
`$ git checkout pull/2357`

`$ git checkout pull/2357` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2357/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2357`

`$ git pr show -t 2357`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2357.diff">https://git.openjdk.org/valhalla/pull/2357.diff</a>

</details>


src/hotspot/share/runtime/sharedRuntime.cpp
```diff
@@ -3221,6 +3271,43 @@ AdapterHandlerEntry* AdapterHandlerLibrary::get_adapter(const methodHandle& meth
   ResourceMark rm;
   bool new_entry = false;
 
+  // Inherit adapter from super method as long as it is compatible with adapters from local interfaces
```
DanHeidinga, line 3274: Can we wrap this in a `if (!method->is_static()) {...}` block?

PR #2361, opened 2026-04-23
8382941: [lworld] Incorrect handling of naturally atomic values

fparain:
Fixes in the logic to determine if a value class is naturally atomic or not. The previous implementation incorrectly considered values classes with a single non-empty nullable field as being naturally atomic.
Also includes improvements to the InstanceKlass::is_naturally_atomic() to prevent bad usages.

Tested locally.
More testing in progress on Mach5 (tiers 1-4).

---------

---------
### Progress

### Issue
 * [JDK-8382941](https://bugs.openjdk.org/browse/JDK-8382941): [lworld] Incorrect handling of naturally atomic values (**Bug** - P3)

### Reviewers
 * [Stefan Karlsson](https://openjdk.org/census#stefank) (@stefank - Committer) ⚠️ Review applies to [0301f5d2](https://git.openjdk.org/valhalla/pull/2361/files/0301f5d20505b209994c38cd488cb92fda5df1f1)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2361/head:pull/2361` \
`$ git checkout pull/2361`

`$ git checkout pull/2361` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2361/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2361`

`$ git pr show -t 2361`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2361.diff">https://git.openjdk.org/valhalla/pull/2361.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2361#issuecomment-4307367070)
</details>

### DIFF
--- a/src/hotspot/share/classfile/fieldLayoutBuilder.cpp
+++ b/src/hotspot/share/classfile/fieldLayoutBuilder.cpp
@@ -79,7 +79,7 @@ static LayoutKind field_layout_selection(FieldInfo field_info, Array<InlineLayou
   if (field_info.field_flags().is_null_free_inline_type()) {
     assert(field_info.access_flags().is_strict(), "null-free fields must be strict");
     if (vk->must_be_atomic() || AlwaysAtomicAccesses) {
-      if (vk->is_naturally_atomic() && vk->has_null_free_non_atomic_layout()) return LayoutKind::NULL_FREE_NON_ATOMIC_FLAT;
+      if (vk->is_naturally_atomic(true) && vk->has_null_free_non_atomic_layout()) return LayoutKind::NULL_FREE_NON_ATOMIC_FLAT;
       return (vk->has_null_free_atomic_layout() && can_use_atomic_flat) ? LayoutKind::NULL_FREE_ATOMIC_FLAT : LayoutKind::REFERENCE;
     } else {
       return vk->has_null_free_non_atomic_layout() ? LayoutKind::NULL_FREE_NON_ATOMIC_FLAT : LayoutKind::REFERENCE;
@@ -976,7 +976,7 @@ void FieldLayoutBuilder::inline_class_field_sorting() {
         assert(_inline_layout_info_array->adr_at(field_index)->klass() != nullptr, "Klass must have been set");
         _has_inlined_fields = true;
         InlineKlass* vk = _inline_layout_info_array->adr_at(field_index)->klass();
-        if (!vk->is_naturally_atomic()) _has_non_naturally_atomic_fields = true;
+        if (!vk->is_naturally_atomic(true)) _has_non_naturally_atomic_fields = true;
         group->add_flat_field(idx, vk, lk);
         _inline_layout_info_array->adr_at(field_index)->set_kind(lk);
         _nonstatic_oopmap_count += vk->nonstatic_oop_map_count();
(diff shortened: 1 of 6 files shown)


src/hotspot/share/oops/instanceKlass.cpp
```diff
@@ -179,6 +179,13 @@ void InlineLayoutInfo::print_on(outputStream* st) const {
   st->print("_null_marker_offset: %d", _null_marker_offset);
 }
 
+bool InstanceKlass::is_naturally_atomic(bool null_free) const {
+  assert(!is_identity_class(), "Doesn't have sense for an identity class");
+  if (!_misc_flags.is_naturally_atomic()) return false;
+  if (null_free) return true;
+  return InlineKlass::cast(this)->is_empty_inline_type();
+}
```
stefank, line 201: FWIW, I think the following version of the code is easier to understand:
```suggestion
bool InstanceKlass::is_naturally_atomic(bool null_free) const {
  assert(!is_identity_class(), "Doesn't have sense for an identity class");
  if (null_free) {
    // No extra null-marker - just check the layout of the fields
    _misc_flags.is_naturally_atomic()
  } else {
    // Requires a null-marker - can't have any other fields
    return InlineKlass::cast(this)->is_empty_inline_type();
  }
}
```

stefank (approved):
Looks good.

Me and Axel has been brainstorming about if we could figure out a name that could help us get rid of this new boolean and at the same time clarify the code and help prevent bugs like the one that this PR is fixing. The question whether something is naturally atomic depends on more than the explicit fields of a value, but rather what a specific value payload where the fields are residing contains a null-marker or not. I would like to come up with a name to describe the set of explicit fields described in the value excluding any potential null-marker. I think that could help clarify the code for the readers. We haven't managed to come up with a name like that yet, so let's defer that to the future.

I find the style used to implement the new is_naturally_atomic quite terse and hard-to-read, so I've given a suggestion that I think is faster to understand. It's OK to skip it if you don't like it.

Arraying (approved):
The changes look good, thank you! I agree with Stefan's suggestion on making it more readable. I would prefer to have a test case where we exercise that an empty value is naturally atomic, so we cover both branches of `InstanceKlass::is_naturally_atomic` explicitly. I'm not sure if that is covered by other tests already, or if it could be added to `test_9`.

src/hotspot/share/classfile/fieldLayoutBuilder.cpp
```diff
@@ -79,7 +79,7 @@ static LayoutKind field_layout_selection(FieldInfo field_info, Array<InlineLayou
   if (field_info.field_flags().is_null_free_inline_type()) {
     assert(field_info.access_flags().is_strict(), "null-free fields must be strict");
     if (vk->must_be_atomic() || AlwaysAtomicAccesses) {
-      if (vk->is_naturally_atomic() && vk->has_null_free_non_atomic_layout()) return LayoutKind::NULL_FREE_NON_ATOMIC_FLAT;
+      if (vk->is_naturally_atomic(true) && vk->has_null_free_non_atomic_layout()) return LayoutKind::NULL_FREE_NON_ATOMIC_FLAT;
```
jsikstro: It's a bit hard to know what the true here means from first glance. Maybe we can add comments to the plain boolean calls?

```suggestion
      if (vk->is_naturally_atomic(true /* null_free */) && vk->has_null_free_non_atomic_layout()) return LayoutKind::NULL_FREE_NON_ATOMIC_FLAT;
```

src/hotspot/share/oops/instanceKlass.cpp
fparain: The method body has been updated according to your suggestion. A comment has also been added above the method to explain the meaning of being naturally atomic. Hope it helps.

src/hotspot/share/classfile/fieldLayoutBuilder.cpp
fparain: Fixed.

fparain:
> I would prefer to have a test case where we exercise that an empty value is naturally atomic, so we cover both branches of `InstanceKlass::is_naturally_atomic` explicitly. I'm not sure if that is covered by other tests already, or if it could be added to `test_9`.

A test with empty values has been added.

Arraying (approved):
LGTM. Thanks for the updates.

fparain:
Thanks @stefank @Arraying @jsikstro for the reviews.
/integrate

PR #2363, opened 2026-04-24
8382704: [lworld] GetLocalObject/GetLocalInstance for THIS value objects during construction

sspitsyn:
This enhancement is to a add JVMTI debugging support for value object construction.
The fix includes:
 - Minor `GetLocalObject` and `GetLocalInstance` spec clarifications
 - JVMTI functions `GetLocalObject` and `GetLocalInstance` now return a snapshot of the `THIS` object if it is requested during a value object construction.
 - This update adds new test with a needed test coverage: `test/hotspot/jtreg/serviceability/jvmti/valhalla/GetCtorLocal`.
 - Some corrections in the test `test/hotspot/jtreg/serviceability/jvmti/valhalla/GetSetLocal` for unification with the test above.

Testing:
  - Tested locally with new test `GetCtorLocal`
  - Tested with mach5 tiers 1-6

---------

---------
### Progress

### Issue
 * [JDK-8382704](https://bugs.openjdk.org/browse/JDK-8382704): [lworld] GetLocalObject/GetLocalInstance for THIS value objects during construction (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2363/head:pull/2363` \
`$ git checkout pull/2363`

`$ git checkout pull/2363` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2363/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2363`

`$ git pr show -t 2363`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2363.diff">https://git.openjdk.org/valhalla/pull/2363.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2363#issuecomment-4310443986)
</details>

### DIFF
--- a/src/hotspot/share/prims/jvmti.xml
+++ b/src/hotspot/share/prims/jvmti.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <?xml-stylesheet type="text/xsl" href="jvmti.xsl"?>
 <!--
- Copyright (c) 2002, 2025, Oracle and/or its affiliates. All rights reserved.
+ Copyright (c) 2002, 2026, Oracle and/or its affiliates. All rights reserved.
  DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
 
  This code is free software; you can redistribute it and/or modify it
@@ -5889,7 +5889,7 @@ class C2 extends C1 implements I2 {
             <description>
               On return, points to the variable's value.
               If the frame's method is a value object constructor and
-              the requested local is the "<code>this</code>" object then the
+              the requested local is the "<code>this</code>" object. the
               value_ptr is set to a snapshot of the "<code>this</code>" object.
             </description>
         </param>
@@ -5944,7 +5944,7 @@ class C2 extends C1 implements I2 {
           <outptr><jobject/></outptr>
             <description>
               On return, points to the variable's value.
-              If the frame's method is a value object constructor then the 
+              If the frame's method is a value object constructor, the 
               value_ptr is set to a snapshot of the "<code>this</code>" object.
             </description>
         </param>
--- a/src/hotspot/share/prims/jvmti.xml
+++ b/src/hotspot/share/prims/jvmti.xml
@@ -5889,7 +5889,7 @@ class C2 extends C1 implements I2 {
             <description>
               On return, points to the variable's value.
               If the frame's method is a value object constructor and
-              the requested local is the "<code>this</code>" object. the
+              the requested local is the "<code>this</code>" object, the
               value_ptr is set to a snapshot of the "<code>this</code>" object.
             </description>
         </param>
--- a/src/hotspot/share/prims/jvmti.xml
+++ b/src/hotspot/share/prims/jvmti.xml
@@ -5888,6 +5888,9 @@ class C2 extends C1 implements I2 {
           <outptr><jobject/></outptr>
             <description>
               On return, points to the variable's value.
+              If the frame's method is a value object constructor and
+              the requested local is the "<code>this</code>" object then the
+              value_ptr is set to a snapshot of the "<code>this</code>" object.
             </description>
         </param>
       </parameters>
@@ -5941,6 +5944,8 @@ class C2 extends C1 implements I2 {
           <outptr><jobject/></outptr>
             <description>
               On return, points to the variable's value.
+              If the frame's method is a value object constructor then the 
+              value_ptr is set to a snapshot of the "<code>this</code>" object.
             </description>
         </param>
       </parameters>
--- a/src/hotspot/share/prims/jvmtiImpl.cpp
+++ b/src/hotspot/share/prims/jvmtiImpl.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2003, 2025, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2003, 2026, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
--- a/src/hotspot/share/prims/jvmtiImpl.cpp
+++ b/src/hotspot/share/prims/jvmtiImpl.cpp
@@ -48,7 +48,7 @@
 #include "runtime/interfaceSupport.inline.hpp"
 #include "runtime/javaCalls.hpp"
 #include "runtime/javaThread.hpp"
-#include "runtime/jniHandles.hpp"
+#include "runtime/jniHandles.inline.hpp"
 #include "runtime/os.hpp"
 #include "runtime/serviceThread.hpp"
 #include "runtime/signature.hpp"
@@ -327,6 +327,7 @@ VM_BaseGetOrSetLocal::VM_BaseGetOrSetLocal(JavaThread* calling_thread, jint dept
   , _jvf(nullptr)
   , _set(set)
   , _self(self)
+  , _need_clone(false)
   , _result(JVMTI_ERROR_NONE)
 {
 }
@@ -476,6 +477,30 @@ bool VM_BaseGetOrSetLocal::check_slot_type_no_lvt(javaVFrame* jvf) {
   return true;
 }
 
+void VM_BaseGetOrSetLocal::check_and_clone_this_value_object() {
+  oop obj = JNIHandles::resolve(_value.l);
+  HandleMark hm(_calling_thread);
+  Handle obj_h(_calling_thread, obj);
+
+  assert(_type == T_OBJECT, "sanity check");
+  assert(obj != nullptr, "expected non-null oop");
+  assert(obj_h()->is_inline(), "expected inline oop");
+  assert(_index == 0, "expected slot 0 for THIS object");
+
+  InlineKlass* klass = InlineKlass::cast(obj_h()->klass());
+  inlineOop obj_copy = klass->allocate_instance(_calling_thread);
+  if (obj_copy == nullptr) {
+    _result = JVMTI_ERROR_OUT_OF_MEMORY;
+  } else {
+    inlineOop thisObj = inlineOop(obj_h());
+    // copy object payload into the object snapshot
+    BufferedValuePayload src(thisObj);
+    BufferedValuePayload dst(obj_copy, klass);
+    src.copy_to(dst);
+  }
+  _value.l = JNIHandles::make_local(_calling_thread, obj_copy);
+}
+
 static bool can_be_deoptimized(vframe* vf) {
   return (vf->is_compiled_frame() && vf->fr().can_be_deoptimized());
 }
@@ -610,12 +635,25 @@ void VM_BaseGetOrSetLocal::doit() {
           // Wrap the oop to be returned in a local JNI handle since
           // oops_do() no longer applies after doit() is finished.
           oop obj = locals->obj_at(_index)();
+
+          if (Arguments::is_valhalla_enabled()) {
+            bool is_ctor = _jvf->method()->is_object_constructor();
+            if (is_ctor && _index == 0 && obj != nullptr && obj->is_inline()) {
+              _need_clone = true; // need to allocate an object snapshot in doit_epilogue
+            }
+          }
           _value.l = JNIHandles::make_local(_calling_thread, obj);
           break;
         }
         default: ShouldNotReachHere();
       }
     }
+  }
+}
+
+void VM_BaseGetOrSetLocal::doit_epilogue() {
+  if (_need_clone) {
+    check_and_clone_this_value_object();
   }
 }
 
--- a/src/hotspot/share/prims/jvmtiImpl.cpp
+++ b/src/hotspot/share/prims/jvmtiImpl.cpp
@@ -500,7 +500,8 @@ void VM_BaseGetOrSetLocal::check_and_clone_this_value_object() {
 
     // Must ensure the content of the buffered value is visible
     // before publishing the buffered value oop
-    OrderAccess::storestore();  }
+    OrderAccess::storestore();
+  }
   _value.l = JNIHandles::make_local(_calling_thread, obj_copy);
 }
 
--- a/src/hotspot/share/prims/jvmtiImpl.cpp
+++ b/src/hotspot/share/prims/jvmtiImpl.cpp
@@ -497,7 +497,10 @@ void VM_BaseGetOrSetLocal::check_and_clone_this_value_object() {
     BufferedValuePayload src(thisObj);
     BufferedValuePayload dst(obj_copy, klass);
     src.copy_to(dst);
-  }
+
+    // Must ensure the content of the buffered value is visible
+    // before publishing the buffered value oop
+    OrderAccess::storestore();  }
   _value.l = JNIHandles::make_local(_calling_thread, obj_copy);
 }
 
--- a/test/hotspot/jtreg/serviceability/jvmti/valhalla/GetCtorLocal/ValueGetCtorLocal.java
+++ b/test/hotspot/jtreg/serviceability/jvmti/valhalla/GetCtorLocal/ValueGetCtorLocal.java
@@ -53,13 +53,9 @@ public class ValueGetCtorLocal {
 
         public ValueHolder(int v) {
             f1 = new ValueClass(v, v + 100);
-            if (!nTestCtorThis(Thread.currentThread())) {
-                throw new RuntimeException("Failed: error in nTestCtorThis #1");
-            }
+            testCtorThis(Thread.currentThread());
             f2 = new ValueClass(v + 1, v + 200);
-            if (!nTestCtorThis(Thread.currentThread())) {
-                throw new RuntimeException("Failed: error in nTestCtorThis #2");
-            }
+            testCtorThis(Thread.currentThread());
         }
     }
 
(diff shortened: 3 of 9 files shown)


src/hotspot/share/prims/jvmtiImpl.cpp
```diff
@@ -610,6 +635,13 @@ void VM_BaseGetOrSetLocal::doit() {
           // Wrap the oop to be returned in a local JNI handle since
           // oops_do() no longer applies after doit() is finished.
           oop obj = locals->obj_at(_index)();
+
+          if (Arguments::is_valhalla_enabled()) {
+            bool is_ctor = _jvf->method()->is_object_constructor();
+            if (is_ctor && _index == 0 && obj != nullptr && obj->is_inline()) {
```
dean-long, line 645: I thought local slots can be recycled, so checking for local 0 is not reliable.  I would think it is necessary to check the LocalVariableTable to find the bci range for the "this" variable.

sspitsyn: Thank you for the comment. The `LocalVariableTable` is not always available. The implementation has many checks in place. Please, see `check_slot_type_lvt()` and `check_slot_type_no_lvt()`.

src/hotspot/share/prims/jvmti.xml
```diff
@@ -5888,6 +5888,9 @@ class C2 extends C1 implements I2 {
           <outptr><jobject/></outptr>
             <description>
               On return, points to the variable's value.
+              If the frame's method is a value object constructor and
+              the requested local is the "<code>this</code>" object then the
+              value_ptr is set to a snapshot of the "<code>this</code>" object.
```
plummercj, line 5893:
```suggestion
              If the frame's method is a value object constructor and
              the requested local is the "<code>this</code>" object, the
              value_ptr is set to a snapshot of the "<code>this</code>" object.
```

src/hotspot/share/prims/jvmti.xml
```diff
@@ -5941,6 +5944,8 @@ class C2 extends C1 implements I2 {
           <outptr><jobject/></outptr>
             <description>
               On return, points to the variable's value.
+              If the frame's method is a value object constructor then the 
+              value_ptr is set to a snapshot of the "<code>this</code>" object.
```
plummercj:
```suggestion
              If the frame's method is a value object constructor, the 
              value_ptr is set to a snapshot of the "<code>this</code>" object.
```

src/hotspot/share/prims/jvmtiImpl.cpp
dean-long: OK.  I also noticed that the implementation of GetLocalInstance also uses local 0.

dean-long (approved):

test/hotspot/jtreg/serviceability/jvmti/valhalla/GetCtorLocal/ValueGetCtorLocal.java
```diff
@@ -0,0 +1,72 @@
...
+            }
+        }
+    }
+
+    public static void main(String[] args) throws Exception {
+        ValueClass testObj1 = new ValueClass(7, 8);
+        ValueHolder testObj2 = new ValueHolder(9);
+    }
+
+    private static native boolean nTestCtorThis(Thread thread);
```
plummercj: What does the "n" in "nTestCtorThis" mean?

test/hotspot/jtreg/serviceability/jvmti/valhalla/GetCtorLocal/libValueGetCtorLocal.cpp
```diff
@@ -0,0 +1,124 @@
...
+  LOG("nTestCtorThis: obj0: %p obj_this: %p objects are equal: %d\n",
+      (void*)obj0, (void*)obj_this, result);
+
+  if (!result) {
+    fatal(jni, "Failed: obj0 != obj_this\n");
+  }
+  if (cached_this == nullptr) { // first call to nTestCtorThis
+    cached_this = jni->NewGlobalRef(obj_this);
+  } else {
+    // cached_this must be a snapshot that not mutate with the ctor changes
```
plummercj:
```suggestion
    // cached_this must be a snapshot that does not mutate when the ctor changes field values
```

test/hotspot/jtreg/serviceability/jvmti/valhalla/GetSetLocal/libValueGetSetLocal.cpp
```diff
@@ -32,7 +32,7 @@ extern "C" {
 #endif
 
 static jvmtiEnv *jvmti = nullptr;
-
+static jobject cached_this = nullptr;
```
plummercj: I don't think this is used.

test/lib/jdk/test/lib/jvmti/jvmti_common.hpp
```diff
@@ -279,6 +279,15 @@ get_frame_count(jvmtiEnv *jvmti, JNIEnv* jni, jthread thread) {
   return frame_count;
 }
 
+static jmethodID
+get_frame_method(jvmtiEnv *jvmti, JNIEnv* jni, jthread thread, jint depth) {
+  jmethodID method;
+  jlocation loc;
+  jvmtiError err = jvmti->GetFrameLocation(thread, depth, &method, &loc);
+  check_jvmti_status(jni, err, "notifyFramePop: Failed in JVMTI GetFrameLocation");
```
plummercj: The error message needs updating.

plummercj (commented):
Copyrights need updating.

plummercj (commented):
I think you still need to update com/sun/jdi/valhalla/CtorDebuggingTest.java. I'm seeing the following two asserts fail:

            // and later we get the same oop (although it's content is changing).
            assertEquals(breakpointHandler.thisAtBreakpoint(bkpt1), breakpointHandler.thisAtBreakpoint(bkpt2));
            assertEquals(breakpointHandler.thisAtBreakpoint(bkpt2), breakpointHandler.thisAtBreakpoint(bkpt3));

I think after your changes the assumption now is that they should not be equal. Also this assert fails:

            // Now thisAt2 has the same content as v3.
            assertEquals(thisAt2, v3);

I think the expectation now is also that they should not be equal.

test/hotspot/jtreg/serviceability/jvmti/valhalla/GetCtorLocal/ValueGetCtorLocal.java
sspitsyn: Thank you for catching this. I think it means native. I'll get rid of this prefix.

src/hotspot/share/prims/jvmti.xml
sspitsyn: Thanks, taken.

src/hotspot/share/prims/jvmti.xml
sspitsyn: Thanks, taken.

test/hotspot/jtreg/serviceability/jvmti/valhalla/GetSetLocal/libValueGetSetLocal.cpp
sspitsyn: Thanks, removed.

test/lib/jdk/test/lib/jvmti/jvmti_common.hpp
sspitsyn: Updated, thanks.

test/hotspot/jtreg/serviceability/jvmti/valhalla/GetCtorLocal/libValueGetCtorLocal.cpp
sspitsyn: Thanks. Fixed now.

sspitsyn:
> I think you still need to update com/sun/jdi/valhalla/CtorDebuggingTest.java. I'm seeing the following two asserts fail:

Good catch. It seems that I've prepared these changes but lost on my way.

sspitsyn:
I believe, I've made all changes suggested by reviewers. Please, let me know if I've missed anything.

test/hotspot/jtreg/serviceability/jvmti/valhalla/GetCtorLocal/libValueGetCtorLocal.cpp
```diff
@@ -0,0 +1,90 @@
...
+      (void*)obj0, (void*)obj_this, is_equal);
+
+  if (!is_equal) {
+    fatal(jni, "Failed: obj0 != obj_this\n");
+  }
+  if (cached_this == nullptr) { // first call to testCtorThis
+    cached_this = jni->NewGlobalRef(obj_this);
+  } else {
+    // cached_this must be a snapshot that idoes not mutate
+    // when the ctor changes field values
```
plummercj:
```suggestion
    // cached_this must be a snapshot that does not mutate
    // when the ctor changes field values
```

test/jdk/com/sun/jdi/valhalla/CtorDebuggingTest.java
```diff
@@ -396,10 +396,10 @@ protected void runTests() throws Exception {
         // Analyze gathered data depending on the testcase.
         if (v1 == null && v2 == null && v3 == null) {
             // No other references.
-            // ObjectID is generated at the 1st breakpoint (reference to heap object being constructed),
-            // and later we get the same oop (although it's content is changing).
-            assertEquals(breakpointHandler.thisAtBreakpoint(bkpt1), breakpointHandler.thisAtBreakpoint(bkpt2));
-            assertEquals(breakpointHandler.thisAtBreakpoint(bkpt2), breakpointHandler.thisAtBreakpoint(bkpt3));
+            // ObjectID is generated at the 1st breakpoint (reference to a snapshot of heap object being constructed),
+            // and later we get snapshots of the updated oop as it's content is changing.
```
plummercj:
```suggestion
            // and later we get snapshots of the updated oop as its content is changing.
```

src/hotspot/share/prims/jvmtiImpl.cpp
```diff
@@ -476,6 +477,30 @@ bool VM_BaseGetOrSetLocal::check_slot_type_no_lvt(javaVFrame* jvf) {
...
+  inlineOop obj_copy = klass->allocate_instance(_calling_thread);
+  if (obj_copy == nullptr) {
+    _result = JVMTI_ERROR_OUT_OF_MEMORY;
+  } else {
+    inlineOop thisObj = inlineOop(obj_h());
+    // copy object payload into the object snapshot
+    BufferedValuePayload src(thisObj);
+    BufferedValuePayload dst(obj_copy, klass);
+    src.copy_to(dst);
+  }
```
plummercj: When you combine the semantics of klass->allocate_instance() and copying with BufferedValuePayload, are you guaranteed to have a fully initialized object? Also, does BufferedValuePayload do deep cloning, and if yes, is that necessary? I think it would be good to have someone from the GC team to look at this specific code to make sure there are no GC concerns with what you are doing.

test/hotspot/jtreg/serviceability/jvmti/valhalla/GetCtorLocal/libValueGetCtorLocal.cpp
sspitsyn: Thanks. Fixed now.

test/jdk/com/sun/jdi/valhalla/CtorDebuggingTest.java
sspitsyn: Fixed now, thanks!

src/hotspot/share/prims/jvmtiImpl.cpp
sspitsyn: Thanks for the questions.

> When you combine the semantics of klass->allocate_instance() and copying with BufferedValuePayload, are you guaranteed to have a fully initialized object?

What do you mean by this? The purpose is to create a snapshot even though the object has not been fully initialized yet. In fact, it can be non-initialized at all. Is it a concern about a need for a memory order action like:
```
    // Must ensure the content of the buffered value is visible
    // before publishing the buffered value oop
    OrderAccess::storestore();
```
Please, see `FlatValuePayload::read(TRAPS)`.
I guess, we can check with @dean-long if it is really needed.

sspitsyn: Update: I've added a call to the `OrderAccess::storestore();`.

sspitsyn: > Also, does BufferedValuePayload do deep cloning, and if yes, is that necessary? 

I don't think it is necessary. I'm not sure there is a reason for that. We create a value object snapshot during initialization. It is needed because the value object can mutate during its construction, and so, the `ObjectReference` we create can cause some trouble. It is not clear if it is allowed for value objects to have mutable references to other objects in their fields. But if it is, then it has to be okay and right thing to do to avoid their snashotting.

src/hotspot/share/prims/jvmtiImpl.cpp
```diff
@@ -476,6 +477,33 @@ bool VM_BaseGetOrSetLocal::check_slot_type_no_lvt(javaVFrame* jvf) {
...
+  } else {
+    inlineOop thisObj = inlineOop(obj_h());
+    // copy object payload into the object snapshot
+    BufferedValuePayload src(thisObj);
+    BufferedValuePayload dst(obj_copy, klass);
+    src.copy_to(dst);
+
+    // Must ensure the content of the buffered value is visible
+    // before publishing the buffered value oop
+    OrderAccess::storestore();  }
```
dean-long: A storestore() seems necessary before publishing the jobject to other threads, if copy_to() is used.  But why can't we do this instead:
```
oop obj_copy = src.read(_calling_thread);
if (obj_copy == nullptr) {
  _result = JVMTI_ERROR_OUT_OF_MEMORY;
} else {
  _value.l = JNIHandles::make_local(_calling_thread, obj_copy);
}
```

src/hotspot/share/prims/jvmtiImpl.cpp
plummercj: Regarding initialization, does allocate_instance() return a fully initialized object (all nulls and 0's). I assume it does.

Regarding BufferedValuePayload doing deep cloning, it is likely best if it does not, so I just wanted to confirm that this it does not. Also, does it honor GC memory barriers?

src/hotspot/share/prims/jvmtiImpl.cpp
plummercj:
```suggestion
    OrderAccess::storestore();
  }
```

sspitsyn: Addressed, thanks.

sspitsyn: Thanks!

>  But why can't we do this instead:

The `src` is an object of the `BufferedValuePayload` class which does not have the `read(TRAPS)` function.
I was thinking to introduce such function but have a doubt there is a strong need for that. Also, it is not urgent and this code is not my area of expertise.

src/hotspot/share/prims/jvmtiImpl.cpp
sspitsyn: > Also, does it honor GC memory barriers?

There is no need in an extra GC barrier after `src.copy_to(dst);`. The implementation of the BufferedValuePayload::copy_to()` which makes a call to the `HeapAccess<>::value_copy(src, dst)` for a buffered layout which should have all needed barriers in place.

plummercj: Ok. thanks for looking into it.

plummercj (approved):

src/hotspot/share/prims/jvmti.xml
```diff
@@ -5888,6 +5888,9 @@ class C2 extends C1 implements I2 {
           <outptr><jobject/></outptr>
             <description>
               On return, points to the variable's value.
+              If the frame's method is a value object constructor and
+              the requested local is the "<code>this</code>" object, the
+              value_ptr is set to a snapshot of the "<code>this</code>" object.
```
plummercj, line 5893: One thing I just realized is that all our spec updates are going to have to clarify that they are for when in preview mode. I'm not too sure what the best approach and wording for this is. I think it is ok for you to push this PR as-is and we can take care of the spec cleanup later.

dean-long (approved):

src/hotspot/share/prims/jvmti.xml
sspitsyn: Agreed. It is not clear yet how to make a spec clarification just for preview release. Yes, we can take care of this later.

dean-long, line 5893: Would it be clearer to say "the value_ptr is set to a copy of the this object containing a snapshot of the current contents"?

plummercj: I agree, and I was actually planning on using language like this in the JDI and JDWP specs. I was thinking "current state", but "current contents" works also. In any case, we don't need to resolve the spec wording now. Just agree on the concept and the implementation.

sspitsyn: Thank you for the suggestion. Yes, we can polish it later.

dholmes-ora: If the spec update only applies to valhalla/preview then you may be able to use the changes in the JNI spec to guide the wording.

plummercj: JNI AllocObject uses "if preview features are enabled" in a couple of places. That's probably a good example.

PR #2370, opened 2026-04-27
8381268: [lworld] C2: assert(adr->is_top() || C->get_alias_index(gvn.type(adr)->is_ptr(), true) == C->get_alias_index(adr_type, true)) failed: adr and adr_type must agree

chhagedorn:
### Failing Test
The assert added in mainline to verify that `adr_type` and `adr`'s type match fails in Valhalla when facing a load from a cloned flat array (i.e. called `clone()`) in `TestArrays::test28()`:
https://github.com/openjdk/valhalla/blob/cfda30f3735276318c291718ca1f3daaa0e124c8/test/hotspot/jtreg/compiler/valhalla/inlinetypes/TestArrays.java#L808-L813

This triggers when trying to add array elements to safepoints.

### Mismatched `AddP` Type vs. Calculated `adr_type`
In `PhaseMacroExpand::make_arraycopy_load()`, we have the following type for the `base` of the flat array `src`:
```
//  gvn.type(base) with _offset = 0, _field_offset = bottom:
aryptr:flat:instptr:.../MyValue2 (...):NotNull:exact *,iid=bot[int:10] (...):NotNull:exact:flat(+bot):null free
```
We then try to add the flat array load for `src.x` which is at `offset = 32` (16 bytes for the array object header + another 16 bytes to get to `x` at offset 16 (`src.v` is first at offset 0)). Here we get a mismatch between the computed `adr_type` and the type stored for `adr` (i.e. an `AddP`) which is later used when creating the `LoadNode`:
```
// adr_type with _offset = 32, _field_offset = bottom:
aryptr:flat:instptr:.../MyValue2 (...):NotNull:exact *,iid=bot[int:10] (...):NotNull:exact:flat(+bot):null free[2]

// gvn.type(adr) with _offset = 16, _field_offset = 16:
aryptr:flat:instptr:.../MyValue2 (...):NotNull:exact *,iid=bot[int:10] (...):NotNull:exact:flat(+16):null free[0]
```
The types have different `_field_offset` values and thus end up on a different alias class which triggers the assert. 

### Reason for Mismatch
The problem is that in `PhaseMacroExpand::make_arraycopy_load()`, we simply call `add_offset()`, regardless of whether it's a flat array or not:
https://github.com/openjdk/valhalla/blob/cfda30f3735276318c291718ca1f3daaa0e124c8/src/hotspot/share/opto/macro.cpp#L323-L326

This ignores `_field_offset` and leaves it at `bottom`. But when calling `AddPNode::Value()`, we call `TypeAryPtr::add_field_offset_and_offset()` which also considers `_field_offset` and properly sets it to 16 and correspondingly `_offset` to 16.

### Fixing the Mismatch
A straight forward fix is to use `add_field_offset_and_offset()` instead of `add_offset()` when having an array pointer. This works to address the observing assertion failure. This is what's suggested in this patch. I also applied some small clean-ups.

### Another Lurking Bug
In `add_field_offset_and_offset()`, we have special code to find out what the `_field_offset` is for a flat array pointer:
https://github.com/openjdk/valhalla/blob/583bdbc45a0c044eb0e26436983e9f416f806ed6/src/hotspot/share/opto/type.cpp#L5799-L5809

This assumes that `_field_offset` is implicitly 0 even when it's set to `bottom` which could actually mean any field offset. This most likely seems to hold in all the cases we currently know of but is not correct in the general case. We should fix that as well. But when restricting this to `_field_offset != bottom`, we hit other assertions. This requires more investigation.

To move forward with this patch and enable more testing again, I suggest to fix this second issue separately, either with [JDK-8358079](https://bugs.openjdk.org/browse/JDK-8358079) or a new issue.

Thanks,
Christian

---------

---------
### Progress

### Issue
 * [JDK-8381268](https://bugs.openjdk.org/browse/JDK-8381268): [lworld] C2: assert(adr-&gt;is_top() || C-&gt;get_alias_index(gvn.type(adr)-&gt;is_ptr(), true) == C-&gt;get_alias_index(adr_type, true)) failed: adr and adr_type must agree (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2370/head:pull/2370` \
`$ git checkout pull/2370`

`$ git checkout pull/2370` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2370/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2370`

`$ git pr show -t 2370`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2370.diff">https://git.openjdk.org/valhalla/pull/2370.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2370#issuecomment-4326722259)
</details>

### DIFF
--- a/src/hotspot/share/opto/type.cpp
+++ b/src/hotspot/share/opto/type.cpp
@@ -5783,29 +5783,32 @@ const TypeAryPtr* TypeAryPtr::with_field_offset(int offset) const {
 }
 
 const TypePtr* TypeAryPtr::add_field_offset_and_offset(intptr_t offset) const {
+  if (!is_flat() || !klass_is_exact() || offset == OffsetBot || offset == OffsetTop) {
+    return add_offset(offset);
+  }
+
+  // Handle flat concrete value class array with known 'offset' which could refer to an actual field in the flat storage.
   int adj = 0;
-  if (is_flat() && klass_is_exact() && offset != Type::OffsetBot && offset != Type::OffsetTop) {
-    if (_offset.get() != OffsetBot && _offset.get() != OffsetTop) {
-      adj = _offset.get();
-      offset += _offset.get();
+  if (_offset != Offset::bottom && _offset != Offset::top) {
+    adj = _offset.get();
+    offset += _offset.get();
+  }
+  uint header = arrayOopDesc::base_offset_in_bytes(T_FLAT_ELEMENT);
+  if (_field_offset != Offset::bottom && _field_offset != Offset::top) {
+    offset += _field_offset.get();
+    if (_offset == Offset::bottom || _offset == Offset::top) {
+      offset += header;
     }
-    uint header = arrayOopDesc::base_offset_in_bytes(T_FLAT_ELEMENT);
-    if (_field_offset.get() != OffsetBot && _field_offset.get() != OffsetTop) {
-      offset += _field_offset.get();
-      if (_offset.get() == OffsetBot || _offset.get() == OffsetTop) {
-        offset += header;
-      }
-    }
-    if (elem()->make_oopptr()->is_inlinetypeptr() && (offset >= (intptr_t)header || offset < 0)) {
-      // Try to get the field of the inline type array element we are pointing to
-      ciInlineKlass* vk = elem()->inline_klass();
-      int shift = flat_log_elem_size();
-      int mask = (1 << shift) - 1;
-      intptr_t field_offset = ((offset - header) & mask);
-      ciField* field = vk->get_field_by_offset(field_offset + vk->payload_offset(), false);
-      if (field != nullptr || field_offset == vk->null_marker_offset_in_payload()) {
-        return with_field_offset(field_offset)->add_offset(offset - field_offset - adj);
-      }
+  }
+  if (elem()->make_oopptr()->is_inlinetypeptr() && (offset >= (intptr_t)header || offset < 0)) {
+    // Try to get the field of the inline type array element we are pointing to
+    ciInlineKlass* vk = elem()->inline_klass();
+    int shift = flat_log_elem_size();
+    int mask = (1 << shift) - 1;
+    int field_offset = static_cast<int>((offset - header) & mask);
+    ciField* field = vk->get_field_by_offset(field_offset + vk->payload_offset(), false);
+    if (field != nullptr || field_offset == vk->null_marker_offset_in_payload()) {
+      return with_field_offset(field_offset)->add_offset(offset - field_offset - adj);
     }
   }
   return add_offset(offset - adj);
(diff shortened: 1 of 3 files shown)


src/hotspot/share/opto/type.cpp
```diff
@@ -5783,29 +5783,32 @@ const TypeAryPtr* TypeAryPtr::with_field_offset(int offset) const {
...
-      if (field != nullptr || field_offset == vk->null_marker_offset_in_payload()) {
-        return with_field_offset(field_offset)->add_offset(offset - field_offset - adj);
-      }
+  }
+  if (elem()->make_oopptr()->is_inlinetypeptr() && (offset >= (intptr_t)header || offset < 0)) {
+    // Try to get the field of the inline type array element we are pointing to
+    ciInlineKlass* vk = elem()->inline_klass();
+    int shift = flat_log_elem_size();
+    int mask = (1 << shift) - 1;
+    int field_offset = static_cast<int>((offset - header) & mask);
```
chhagedorn, line 5808: Added `static_cast` here (IDE warning about narrowing type).

src/hotspot/share/opto/type.cpp
```diff
@@ -5783,29 +5783,32 @@ const TypeAryPtr* TypeAryPtr::with_field_offset(int offset) const {
 }
 
 const TypePtr* TypeAryPtr::add_field_offset_and_offset(intptr_t offset) const {
+  if (!is_flat() || !klass_is_exact() || offset == OffsetBot || offset == OffsetTop) {
+    return add_offset(offset);
+  }
```
chhagedorn, line 5788: I added a quick bailout here for non-flat and abstract value class arrays to have fewer nested levels below.

src/hotspot/share/opto/type.cpp
```diff
@@ -5783,29 +5783,32 @@ const TypeAryPtr* TypeAryPtr::with_field_offset(int offset) const {
 }
 
 const TypePtr* TypeAryPtr::add_field_offset_and_offset(intptr_t offset) const {
+  if (!is_flat() || !klass_is_exact() || offset == OffsetBot || offset == OffsetTop) {
```
chhagedorn, line 5786: Removed unneeded `Type::` prefix.

src/hotspot/share/opto/type.cpp
```diff
@@ -5783,29 +5783,32 @@ const TypeAryPtr* TypeAryPtr::with_field_offset(int offset) const {
...
-      adj = _offset.get();
-      offset += _offset.get();
-    }
-    uint header = arrayOopDesc::base_offset_in_bytes(T_FLAT_ELEMENT);
-    if (_field_offset.get() != OffsetBot && _field_offset.get() != OffsetTop) {
-      offset += _field_offset.get();
-      if (_offset.get() == OffsetBot || _offset.get() == OffsetTop) {
-        offset += header;
-      }
+  if (_offset != Offset::bottom && _offset != Offset::top) {
```
chhagedorn, line 5792: Simplified to use `operator!=()` and `operator==()` instead of calling `get()`.

TobiHartmann (approved):
That looks good to me. Thanks for fixing!

> To move forward with this patch and enable more testing again, I suggest to fix this second issue separately, either with [JDK-8358079](https://bugs.openjdk.org/browse/JDK-8358079) or a new issue.

Let's adjust JDK-8358079 to track this.

chhagedorn:
Thanks for your review Tobias!

>> To move forward with this patch and enable more testing again, I suggest to fix this second issue separately, either with [JDK-8358079](https://bugs.openjdk.org/browse/JDK-8358079) or a new issue.
>
> Let's adjust JDK-8358079 to track this.

Let's do that. I will update it accordingly.

/integrate

PR #2371, opened 2026-04-27
8383415: [lworld] Flat contended oop field might not be included in OopMap

jsikstro:
Hello,

Embdedded flat contended fields might not be included in the class' OopMap if there are no other oop fields in the holder class.

Flat fields are handled separately in fieldLayoutBuilder.cpp, resulting in the current `_oop_count` not being updated if a flat field contains oops. This is OK, since the oop fields are handled at a later point in time. However, for contended fields, which are part of the contended group, embedded oop fields are only tracked if: `cg->oop_count() > 0`, which is not guaranteed to be true.

I've written a test to see if the oop field is included in the class' OopMap and also changed the `_oop_count` field `_has_embedded_oops`, and changed that to a boolean property, which is how we've used it so far.

Since embedded fields in flat paylods are not tracked in `cg->oop_fields()`, we can't do the assert that is currently there, so I just removed that.

Testing:
* New test fails without the fixes in fieldLayoutBuilder and passes with fixes
* Running Oracle's tier1-3

---------

---------
### Progress

### Issue
 * [JDK-8383415](https://bugs.openjdk.org/browse/JDK-8383415): [lworld] Flat contended oop field might not be included in OopMap (**Bug** - P4)

### Reviewers
 * [Frederic Parain](https://openjdk.org/census#fparain) (@fparain - Committer) ⚠️ Review applies to [83974110](https://git.openjdk.org/valhalla/pull/2371/files/83974110de00a3a3230603da937f17490e0e2135)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2371/head:pull/2371` \
`$ git checkout pull/2371`

`$ git checkout pull/2371` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2371/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2371`

`$ git pr show -t 2371`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2371.diff">https://git.openjdk.org/valhalla/pull/2371.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2371#issuecomment-4327838042)
</details>

### DIFF
--- a/src/hotspot/share/classfile/fieldLayoutBuilder.cpp
+++ b/src/hotspot/share/classfile/fieldLayoutBuilder.cpp
@@ -174,8 +174,7 @@ FieldGroup::FieldGroup(int contended_group) :
   _small_primitive_fields(nullptr),
   _big_primitive_fields(nullptr),
   _oop_fields(nullptr),
-  _contended_group(contended_group),  // -1 means no contended group, 0 means default contended group
-  _has_embedded_oops(false) {}
+  _contended_group(contended_group) {} // -1 means no contended group, 0 means default contended group
 
 void FieldGroup::add_primitive_field(int idx, BasicType type) {
   int size = type2aelembytes(type);
@@ -194,7 +193,6 @@ void FieldGroup::add_oop_field(int idx) {
     _oop_fields = new GrowableArray<LayoutRawBlock*>(INITIAL_LIST_SIZE);
   }
   _oop_fields->append(block);
-  _has_embedded_oops = true;
 }
 
 void FieldGroup::add_flat_field(int idx, InlineKlass* vk, LayoutKind lk) {
@@ -206,10 +204,6 @@ void FieldGroup::add_flat_field(int idx, InlineKlass* vk, LayoutKind lk) {
   block->set_layout_kind(lk);
   if (block->size() >= heapOopSize) {
     add_to_big_primitive_list(block);
-
-    if (vk->contains_oops()) {
-      _has_embedded_oops = true;
-    }
   } else {
     assert(!vk->contains_oops(), "Size of Inline klass with oops should be >= heapOopSize");
     add_to_small_primitive_list(block);
@@ -1344,8 +1338,7 @@ void FieldLayoutBuilder::compute_inline_class_layout() {
   epilogue();
 }
 
-void FieldLayoutBuilder::add_flat_field_oopmap(OopMapBlocksBuilder* nonstatic_oop_maps,
-                InlineKlass* vklass, int offset) {
+void FieldLayoutBuilder::add_flat_field_oopmap(OopMapBlocksBuilder* nonstatic_oop_maps, InlineKlass* vklass, int offset) {
   int diff = offset - vklass->payload_offset();
   const OopMapBlock* map = vklass->start_of_nonstatic_oop_maps();
   const OopMapBlock* last_map = map + vklass->nonstatic_oop_map_count();
@@ -1356,7 +1349,10 @@ void FieldLayoutBuilder::add_flat_field_oopmap(OopMapBlocksBuilder* nonstatic_oo
 }
 
 void FieldLayoutBuilder::register_embedded_oops_from_list(OopMapBlocksBuilder* nonstatic_oop_maps, GrowableArray<LayoutRawBlock*>* list) {
-  if (list == nullptr) return;
+  if (list == nullptr) {
+    return;
+  }
+
   for (int i = 0; i < list->length(); i++) {
     LayoutRawBlock* f = list->at(i);
     if (f->block_kind() == LayoutRawBlock::FLAT) {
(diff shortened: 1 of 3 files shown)


fparain (approved):
Good catch. LGTM.

src/hotspot/share/classfile/fieldLayoutBuilder.cpp
```diff
@@ -175,7 +175,7 @@ FieldGroup::FieldGroup(int contended_group) :
   _big_primitive_fields(nullptr),
   _oop_fields(nullptr),
   _contended_group(contended_group),  // -1 means no contended group, 0 means default contended group
-  _oop_count(0) {}
+  _has_embedded_oops(false) {}
```
xmas92: I wonder if we should just remove the `_has_embedded_oops` completely. And just unconditionally call `register_embedded_oops`. It already handles an empty (no embedded oops) list of sub-blocks. 

It makes the code less complex, but we miss an optimisation in the case where we have flattened a lot of values where no fields contains oops.

But seems like the better optimization if this was a problem would be to create a list of each raw block with oops, so we only walk those, rather than all if one happens to contain an oop.

xmas92 (commented):
I think the implementation looks good, but I wonder if we can simplify even more.

src/hotspot/share/classfile/fieldLayoutBuilder.cpp
jsikstro: I think this is a good idea and it does make the code a bit easier to reason about.

test/hotspot/jtreg/runtime/valhalla/inlinetypes/ContendedFlatFieldOopMap.java
```diff
@@ -0,0 +1,72 @@
...
+
+    public static void main(String[] args) {
+        Object[] obj = new Object[1024 * 1024];
+        WeakReference ref = new WeakReference(obj);
+        holder = new Holder(obj);
+        obj = null;
+
+        Asserts.assertNotEquals(ref.get(), null);
+
+        System.gc();
```
Arraying: Is there any reason we prefer `System.gc` over `WhiteBox`' GC?

test/hotspot/jtreg/runtime/valhalla/inlinetypes/ContendedFlatFieldOopMap.java
```diff
@@ -0,0 +1,72 @@
...
+package runtime.valhalla.inlinetypes;
+
+import jdk.test.lib.Asserts;
+import jdk.internal.vm.annotation.Contended;
+import java.lang.ref.WeakReference;
+
+/*
+ * @test ContendedFlatFieldOopMap
+ * @library /test/lib
+ * @requires vm.flagless
```
Arraying: This will ignore other GCs than the default (likely G1). Is that intentional?

Arraying (commented):
The code itself looks good, nice catch! Left some small comments about the test.

xmas92 (approved):
The implementation looks good.

_As @Arraying says the test can probably be improved, both in reproducibility and testing scope. (`System.gc()` may be concurrent and not actually trigger a GC synchronously. And the code should hold true regardless of flags, even if non flattened values never was a problem)._

test/hotspot/jtreg/runtime/valhalla/inlinetypes/ContendedFlatFieldOopMap.java
jsikstro: No, just a habit. I'll change to `WhiteBox.fullGC()`, since that is a bit stronger.

test/hotspot/jtreg/runtime/valhalla/inlinetypes/ContendedFlatFieldOopMap.java
jsikstro: I'll change this to `@requires vm.gc != "Z"`, since ZGC is the only GC that will provoke the earlier bug, but the test can really be run with any GC and should pass.

xmas92 (approved):

test/hotspot/jtreg/runtime/valhalla/inlinetypes/ContendedFlatFieldOopMap.java
```diff
@@ -0,0 +1,80 @@
...
+
+/*
+ * @test ContendedFlatFieldOopMap
+ * @library /test/lib
+ * @requires vm.gc != "Z"
+ * @modules java.base/jdk.internal.vm.annotation
+ * @build jdk.test.whitebox.WhiteBox
+ * @enablePreview
+ * @run driver jdk.test.lib.helpers.ClassFileInstaller jdk.test.whitebox.WhiteBox
+ * @run main/othervm -XX:-RestrictContended
```
Arraying, line 39: Nit: it might be good to guard against this being set:
`@requires vm.opt.RestrictContended == null`
Or maybe even:
`@requires vm.opt.RestrictContended == null | vm.opt.RestrictContended == "false"`

jsikstro: Good point. Fixed in latest commit.

Arraying (approved):
LGTM. Thank you.

xmas92 (approved):

jsikstro:
Thank you for the reviews everyone! 

We've had some offline discussion about not limiting the new test as much with `@requires`-properties. Right now running the test with ZGC won't provoke the original problem, and running without contended won't as well. The current `@requires`-properties make sure we run the test in a configuration that tests the originally affected code-path.

In the future we might not want to limit the test not to be run with ZGC. We should revisit the test when objects containing oops will be commonly flattened with ZGC.

/integrate

PR #2372, opened 2026-04-27
8383417: [lworld] Implement limitation to the size of instances due to flattening

fparain:
This patch adds a simple policy to put a limit on field flattening to prevent instances from becoming too large. It is easy to generate very huge instances with nested flat values, so huge that the JVM struggles to allocate its internal data structure to handle such classes. The new policy allocates a flattening budget to each class, and field flattening in this class stops when the budget is exhausted. The limit can be tuned with a VM parameter.

Tested on Mach5 tier1-4

---------

---------
### Progress

### Issue
 * [JDK-8383417](https://bugs.openjdk.org/browse/JDK-8383417): [lworld] Implement limitation to the size of instances due to flattening (**Enhancement** - P4)

### Reviewers
 * [Axel Boldt-Christmas](https://openjdk.org/census#aboldtch) (@xmas92 - Committer) ⚠️ Review applies to [30932333](https://git.openjdk.org/valhalla/pull/2372/files/3093233333a09930b7392312c49d4ee5bbdd07d0)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2372/head:pull/2372` \
`$ git checkout pull/2372`

`$ git checkout pull/2372` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2372/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2372`

`$ git pr show -t 2372`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2372.diff">https://git.openjdk.org/valhalla/pull/2372.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2372#issuecomment-4328111445)
</details>


test/hotspot/jtreg/runtime/valhalla/inlinetypes/TestFlatteningBudget.java
```diff
@@ -0,0 +1,139 @@
...
+ */
+
+/*
+ * @test TestFlatteningBudget
+ * @modules java.base/jdk.internal.vm.annotation
+ *          java.base/jdk.internal.value
+ *          java.base/jdk.internal.misc
+ * @library /test/lib
+ * @enablePreview
+ * @run main runtime.valhalla.inlinetypes.TestFlatteningBudget
```
xmas92, line 33: Should this be a driver tests as it creates new instances for the each test?

Not sure if this test should be flagless as it assumes it does not get in external flags that disable flattening. Alternatively have main method do a quick test for flattening, then throw a skipped exception if fields are not flattened as expected.

Maybe there is already some shared test utility for doing this sort of thing.

src/hotspot/share/runtime/globals.hpp
```diff
@@ -832,6 +832,9 @@ const int ObjectAlignmentInBytes = 8;
           "Max number of embedded object references in a value container "  \
           "before no flattening attempts are made, <0 indicates no limit")  \
                                                                             \
+  product(intx, FlatteningBudget, 1024,                                     \
+          "Maximum size dedicated to flat fields in an instance")           \
+          range(0, 1024 * 1024)                                             \
```
xmas92: Should this be part of the `DISABLE_FLAG_AND_WARN_IF_NOT_DEFAULT` type mechanism ?

Also I have not digged to deep into AOT / CDS. But it seems like this property should be stable from dump and load. `CDS_MUST_MATCH_FLAGS_DO`. However this might limit things to much. But I am guess if this flag affected layout decisions between dump and run, things would not work.

xmas92 (commented):
The concept looks good. I am wondering about a few interactions and corner cases.

src/hotspot/share/runtime/globals.hpp
fparain: Good point. The new VM option has an impact on flattening decisions, so it must match the value use to generate the CDS archive.

test/hotspot/jtreg/runtime/valhalla/inlinetypes/TestFlatteningBudget.java
fparain: The `--enable-preview` is not passed to the test when run in `driver` mode, causing the test to fail to execute.

Logic has been added to the test to silently skip the test if the flattening conditions are not satisfied ( in such case, a message is printed on the output and recorded in the .jtr file)

xmas92 (approved):
Alright. lgtm. 
Thanks for fixing.

dean-long:
Once you have a flattening budget, doesn't this bring up issues similar to method inlining on how to spend that budget?  For example, should it be spent eagerly on the first fields (depth first), or should top-level fields be flattened before nested fields (breadth first)?

xmas92:
> Once you have a flattening budget, doesn't this bring up issues similar to method inlining on how to spend that budget? For example, should it be spent eagerly on the first fields (depth first), or should top-level fields be flattened before nested fields (breadth first)?

I was thinking the same, but more along the lines of a bytes utilization knapsack problem. We probably don’t have any information about what is good to flatten unless we add something to the AOT training data. 

But feels like it would add a whole lot of complexity, and seems like a post 401 RFE.

src/hotspot/share/runtime/globals.hpp
```diff
@@ -832,6 +832,9 @@ const int ObjectAlignmentInBytes = 8;
           "Max number of embedded object references in a value container "  \
           "before no flattening attempts are made, <0 indicates no limit")  \
                                                                             \
+  product(intx, FlatteningBudget, 1024,                                     \
```
Arraying: The flag should be diagnostic to be consistent. Flags such as `UseArrayFlattening` are set to diagnostic. I don't think this is more "important" than `UseArrayFlattening`. Let me know if you disagree.

test/hotspot/jtreg/runtime/valhalla/inlinetypes/TestFlatteningBudget.java
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+
+
+    // Test 1 : impact of flattening budget on flattening decisions
+    static class Test1 {
+        static final Unsafe UNSAFE = Unsafe.getUnsafe();
+        static HotSpotDiagnosticMXBean hsDiag = ManagementFactory.getPlatformMXBean(HotSpotDiagnosticMXBean.class);
+
+        static int getIntVMOption(String vmOption) {
```
Arraying: FYI: We have a much nicer interface through `WhiteBox`, which would prevent having to write all the boilerplate around `ManagementFactory`.

src/hotspot/share/runtime/globals.hpp
```diff
@@ -832,6 +832,9 @@ const int ObjectAlignmentInBytes = 8;
...
+          "Maximum size dedicated to flat fields in an instance")           \
```
Arraying: Nit: it would be nice to specify the unit (bytes) here.

src/hotspot/share/classfile/fieldLayoutBuilder.cpp
```diff
@@ -787,6 +802,7 @@ FieldLayoutBuilder::FieldLayoutBuilder(const Symbol* classname, ClassLoaderData*
   _nullable_non_atomic_layout_size_in_bytes(-1),
   _fields_size_sum(0),
   _declared_nonstatic_fields_count(0),
+  _flattening_budget(FlatteningBudget),
```
Arraying: Nit: initialize this somewhere else, and then assert that the `FlatteningBudget` flag is `>0`. While this is constrained by the flag, I still think it's good to have an extra safeguard in place, just in case the flag changes later on.

test/hotspot/jtreg/runtime/valhalla/inlinetypes/TestFlatteningBudget.java
```diff
@@ -0,0 +1,172 @@
...
+            return Boolean.valueOf(option.getValue()).booleanValue();
+        }
+
+        public static void main(String[] args) throws Exception {
+
+            // Pre-conditions
+            boolean useFieldFlattening = getBooleanVMOption("UseFieldFlattening");
+            // Test cannot work properly if field flattening is disabled
+            if (!useFieldFlattening) {
+              System.out.println("Test skipped because field flattening is disabled");
```
Arraying: The "proper" way of doing this is by throwing a `SkippedException`.

Arraying: It would be nicer to do this through jtreg though. You can make a `@requires` where you can check if they're (a) set explicitly, and/or (b) the expected boolean value. See this example: https://github.com/openjdk/jdk/pull/29848/changes

Arraying (commented):
This seems useful to have, thanks! Left a comment about flag level, left some other remarks about test maintainability.

src/hotspot/share/runtime/globals.hpp
fparain: Disabling array flattening is a diagnostic feature to investigate if a particular issue is related to array flattening. The FlatteningBudget is different, it's a sizing parameter that some applications need to change, not because someone is investigating an issue, but because the design of the application's classes would not be allow them to be flatten otherwise. It's a flag that makes sense in production, not necessarily in a troubleshooting situtation.

xmas92 (approved):

fparain:
> > Once you have a flattening budget, doesn't this bring up issues similar to method inlining on how to spend that budget? For example, should it be spent eagerly on the first fields (depth first), or should top-level fields be flattened before nested fields (breadth first)?
> 
> I was thinking the same, but more along the lines of a bytes utilization knapsack problem. We probably don’t have any information about what is good to flatten unless we add something to the AOT training data.
> 
> But feels like it would add a whole lot of complexity, and seems like a post 401 RFE.

The flattening decision differs from the inlining decision in the fact that it doesn't make recursive decisions. When considering if a field should be flattened or not, the algorithm can only use layouts that have already be computed for this field's value class. Flattening decisions in this class have already been made when the class was loaded.

The current implementation of this flattening budget is inherently a greedy algorithm, the budget is being consumed in the order the fields are inserted in the layout, and those insertions are performed from the bigger value to the smaller value. It will prioritize the flattening of a few big values rather than many small values. This aspect of the algorithm was mentioned in the CR. However, it would be difficult to change this without rewriting the whole layout computation algorithm to add more passes over the fields information.

It is important to remember that the main purpose of this budget is not to tune the flattening decisions of the JVM, it is a far too coarse tool for that. Its goal is to simply put a limit on how far the JVM will flatten fields to prevent user code from easily reaching JVM internal resources limits and causing a VM crash. Most users will never have to deal with this VM option.

src/hotspot/share/classfile/fieldLayoutBuilder.cpp
```diff
@@ -787,6 +802,8 @@ FieldLayoutBuilder::FieldLayoutBuilder(const Symbol* classname, ClassLoaderData*
   _nullable_non_atomic_layout_size_in_bytes(-1),
   _fields_size_sum(0),
   _declared_nonstatic_fields_count(0),
+  _flattening_budget((int)FlatteningBudget),  // uint -> int convertion but FlatteningBudget value has
+                                              // been validated in arguments.cpp (range [0, 1024 * 1024]).
```
Arraying: It's defined in `globals.hpp` and we don't do any ergonomics/validation in `arguments.cpp`, the comment feels a bit misleading.

fparain:
```
$ ./build/lworld/images/jdk/bin/java --enable-preview -XX:+UnlockExperimentalVMOptions -XX:FlatteningBudget=-1 Test2
Improperly specified VM option 'FlatteningBudget=-1'
Error: Could not create the Java Virtual Machine.
Error: A fatal exception has occurred. Program will exit.

$ ./build/lworld/images/jdk/bin/java --enable-preview -XX:+UnlockExperimentalVMOptions -XX:FlatteningBudget=9999990 Test2
uint FlatteningBudget=9999990 is outside the allowed range [ 0 ... 1048576 ]
Improperly specified VM option 'FlatteningBudget=9999990'
Error: Could not create the Java Virtual Machine.
Error: A fatal exception has occurred. Program will exit.

$ ./build/lworld/images/jdk/bin/java --enable-preview -XX:+UnlockExperimentalVMOptions -XX:FlatteningBudget=1024 Test2
Run test1
Run test2
Run test3
Done
```

There's  a validation that the value is within the range specified in `globals.hpp`.

Arraying (approved):
Thank you, looks good.

fparain:
Thanks @xmas92 @Arraying for the reviews.
/integrate

PR #2374, opened 2026-04-28
8357249: Compiler task keeps --system files open

david-beaumont:
Draft request for discussion purposes only...

---------

---------
### Progress
- \[ \] Change must not contain extraneous whitespace
- \[ \] Commit message must refer to an issue

### Error
&nbsp;⚠️ OCA signatory status must be verified

### Issue
 * [JDK-8357249](https://bugs.openjdk.org/browse/JDK-8357249): Compiler task keeps --system files open (**Bug** - P4) ⚠️ Issue is already resolved. Consider making this a "backport pull request" by setting the PR title to `Backport <hash>` with the hash of the original commit. See [Backports](https://wiki.openjdk.org/display/SKARA/Backports).

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2374/head:pull/2374` \
`$ git checkout pull/2374`

`$ git checkout pull/2374` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2374/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2374`

`$ git pr show -t 2374`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2374.diff">https://git.openjdk.org/valhalla/pull/2374.diff</a>

</details>

### DIFF
--- a/src/java.base/share/classes/jdk/internal/jimage/BasicImageReader.java
+++ b/src/java.base/share/classes/jdk/internal/jimage/BasicImageReader.java
@@ -32,6 +32,7 @@ import java.lang.reflect.Method;
 import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
 import java.nio.IntBuffer;
+import java.nio.MappedByteBuffer;
 import java.nio.channels.FileChannel;
 import java.nio.file.Path;
 import java.nio.file.StandardOpenOption;
@@ -68,6 +69,7 @@ public class BasicImageReader implements AutoCloseable {
             isSystemProperty("sun.arch.data.model", "64", "32");
     private static final boolean USE_JVM_MAP =
             isSystemProperty("jdk.image.use.jvm.map", "true", "true");
+    // Whether to map the entire file contents.
     private static final boolean MAP_ALL =
             isSystemProperty("jdk.image.map.all", "true", IS_64_BIT ? "true" : "false");
 
@@ -77,7 +79,7 @@ public class BasicImageReader implements AutoCloseable {
     private final ByteBuffer memoryMap;
     private final FileChannel channel;
     private final ImageHeader header;
-    private final long indexSize;
+    private final int indexSize;
     private final IntBuffer redirect;
     private final IntBuffer offsets;
     private final ByteBuffer locations;
@@ -92,13 +94,15 @@ public class BasicImageReader implements AutoCloseable {
         this.byteOrder = Objects.requireNonNull(byteOrder);
         this.name = this.imagePath.toString();
 
+        // Only the runtime image is loaded with the root class-loader.
+        final boolean isRuntimeImage = BasicImageReader.class.getClassLoader() == null;
         ByteBuffer map;
 
-        if (USE_JVM_MAP && BasicImageReader.class.getClassLoader() == null) {
+        if (USE_JVM_MAP && isRuntimeImage) {
             // Check to see if the jvm has opened the file using libjimage
             // native entry when loading the image for this runtime
             map = NativeImageBuffer.getNativeMap(name);
-         } else {
+        } else {
             map = null;
         }
 
@@ -111,7 +115,7 @@ public class BasicImageReader implements AutoCloseable {
             AccessController.doPrivileged(new PrivilegedAction<Void>() {
                 @Override
                 public Void run() {
-                    if (BasicImageReader.class.getClassLoader() == null) {
+                    if (isRuntimeImage) {
                         try {
                             Class<?> fileChannelImpl =
                                 Class.forName("sun.nio.ch.FileChannelImpl");
@@ -134,33 +138,44 @@ public class BasicImageReader implements AutoCloseable {
 
         // If no memory map yet and 64 bit jvm then memory map entire file
         if (MAP_ALL && map == null) {
-            map = channel.map(FileChannel.MapMode.READ_ONLY, 0, channel.size());
+            if (isRuntimeImage) {
+                map = channel.map(FileChannel.MapMode.READ_ONLY, 0, channel.size());
+            } else {
+                // Non-runtime images are used for compiler tasks, and we avoid
+                // memory mapping them so the file handles can be closed when the task
+                // completes (rather than having to wait for garbage collection).
+                // This is important when long-lived server processes compile code
+                // across multiple runtime images. Once this code no longer requires
+                // JDK-8, we might be able to switch to using MemorySegment here.
+                int imageFileSize = (int) channel.size();
+                if ((long) imageFileSize != channel.size()) {
+                    throw new IOException("\"" + name + "\" too large");
+                }
+                map = readDirectBuffer(channel, imageFileSize, name);
+            }
         }
 
-        // Assume we have a memory map to read image file header
+        // headerBuffer is a temporary buffer for reading the heading (which
+        // may be the full buffer, or a scratch buffer we discard later).
         ByteBuffer headerBuffer = map;
         int headerSize = ImageHeader.getHeaderSize();
 
         // If no memory map then read header from image file
         if (headerBuffer == null) {
-            headerBuffer = ByteBuffer.allocateDirect(headerSize);
-            if (channel.read(headerBuffer, 0L) == headerSize) {
-                headerBuffer.rewind();
-            } else {
-                throw new IOException("\"" + name + "\" is not an image file");
-            }
+            headerBuffer = readDirectBuffer(channel, headerSize, name);
         } else if (headerBuffer.capacity() < headerSize) {
+            // A fully mapped file cannot be smaller than the header size.
             throw new IOException("\"" + name + "\" is not an image file");
         }
 
-        // Interpret the image file header
+        // Read the header and find the index size, which is the minimum buffer
+        // size we need to read if we haven't already mapped the entire file.
         header = readHeader(intBuffer(headerBuffer, 0, headerSize));
         indexSize = header.getIndexSize();
 
         // If no memory map yet then must be 32 bit jvm not previously mapped
         if (map == null) {
-            // Just map the image index
-            map = channel.map(FileChannel.MapMode.READ_ONLY, 0, indexSize);
+            map = readDirectBuffer(channel, indexSize, name);
         }
 
         memoryMap = map.asReadOnlyBuffer();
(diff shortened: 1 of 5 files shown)


src/java.base/share/classes/jdk/internal/jimage/BasicImageReader.java
```diff
@@ -68,6 +69,7 @@ public Boolean run() {
             isSystemProperty("sun.arch.data.model", "64", "32");
     private static final boolean USE_JVM_MAP =
             isSystemProperty("jdk.image.use.jvm.map", "true", "true");
+    // Whether to map the entire file contents.
```
david-beaumont: It took me ages to realise this isn't "map all image files that are opened", but rather "map an entire image file if it is being mapped at all"

src/java.base/share/classes/jdk/internal/jimage/BasicImageReader.java
```diff
@@ -77,7 +79,7 @@ public Boolean run() {
     private final ByteBuffer memoryMap;
     private final FileChannel channel;
     private final ImageHeader header;
-    private final long indexSize;
+    private final int indexSize;
```
david-beaumont, line 89: Just in passing since it's not a long from the getter.

src/java.base/share/classes/jdk/internal/jimage/BasicImageReader.java
```diff
@@ -92,13 +94,15 @@ protected BasicImageReader(Path path, ByteOrder byteOrder)
         this.byteOrder = Objects.requireNonNull(byteOrder);
         this.name = this.imagePath.toString();
 
+        // Only the runtime image is loaded with the root class-loader.
+        final boolean isRuntimeImage = BasicImageReader.class.getClassLoader() == null;
```
david-beaumont, line 107: Semantic readability/intent.

src/java.base/share/classes/jdk/internal/jimage/BasicImageReader.java
```diff
@@ -134,33 +138,44 @@ public Void run() {
 
         // If no memory map yet and 64 bit jvm then memory map entire file
         if (MAP_ALL && map == null) {
-            map = channel.map(FileChannel.MapMode.READ_ONLY, 0, channel.size());
+            if (isRuntimeImage) {
+                map = channel.map(FileChannel.MapMode.READ_ONLY, 0, channel.size());
+            } else {
+                // Non-runtime images are used for compiler tasks, and we avoid
```
david-beaumont: This is what fixes the non-runtime jimage closing issue.
It's nasty because it goes from memory mapped file to 100M of copied data.

An alternative (with performance cost) would be to do only the "partial" loading of the buffer, say up to the index, and then fetch file content with normal file reads. But this is all open to much discussion.

src/java.base/share/classes/jdk/internal/jimage/BasicImageReader.java
```diff
@@ -134,33 +138,44 @@ public Void run() {
...
+                // memory mapping them so the file handles can be closed when the task
+                // completes (rather than having to wait for garbage collection).
+                // This is important when long-lived server processes compile code
+                // across multiple runtime images. Once this code no longer requires
+                // JDK-8, we might be able to switch to using MemorySegment here.
+                int imageFileSize = (int) channel.size();
+                if ((long) imageFileSize != channel.size()) {
+                    throw new IOException("\"" + name + "\" too large");
+                }
+                map = readDirectBuffer(channel, imageFileSize, name);
```
david-beaumont: Pulled out a little helper since it's used in 3 places.

src/java.base/share/classes/jdk/internal/jimage/ImageReader.java
```diff
@@ -89,12 +91,11 @@
  * to the jimage file provided by the shipped JDK by tools running on JDK 8.
  */
 public final class ImageReader implements AutoCloseable {
-    private final SharedImageReader reader;
-
-    private volatile boolean closed;
+    // Set to null when closed to allow GC of underlying buffers to unmap files.
+    private final AtomicReference<SharedImageReader> reader;
```
david-beaumont, line 95: This class is the place where we've already got lifetime semantics.
By letting the reader be cleared, we can replace the `closed` flag and retain non-locking patterns of use.
The benefit is that now the underlying `reader` becomes GC-able when the `ImageReader` is closed.

While this may not be of benefit when using direct buffers, it does let any mapped buffers get GC-ed sooner, which is required for closing the underlying memory mapped file.

I think this change is reasonable in general and just improves the semantics of the way this class's lifetime is managed wrt GC.

src/java.base/share/classes/jdk/internal/jrtfs/JrtFileSystem.java
```diff
@@ -313,6 +314,12 @@ synchronized void cleanup() throws IOException {
             isOpen = false;
             image.close();
             image = null;
+
```
david-beaumont, line 317: This is the change that lets the JAR file be closed. I hate it!
However it shows that there's potential for passing a "closer" or some kind into the JrtFileSystem from the provider to close the underling `JarLoader` (which is the actual thing that needs closing).

More to look at here, but this is proof of concept at least.

src/jdk.compiler/share/classes/com/sun/tools/javac/code/ClassFinder.java
```diff
@@ -217,14 +219,20 @@ protected ClassFinder(Context context) {
         } else {
             useCtProps = false;
         }
-        if (useCtProps && JRTIndex.isAvailable()) {
+        JRTIndex index = null;
```
david-beaumont, line 222: Removes need for `JRTIndex.isAvailable()` which opened the uncloseable image in order to do its test.
We can just try and open the thing we want to open instead.
This should be identical in behaviour to what's there now and is necessary for preventing the non-runtime image being held open forever.

src/jdk.compiler/share/classes/com/sun/tools/javac/file/JRTIndex.java
```diff
@@ -98,16 +98,6 @@ public static JRTIndex instance(boolean previewMode) {
         }
     }
 
-    /** {@return whether the JRT file-system is available to create an index} */
```
david-beaumont, line 101: Removed just to prove it's not needed anywhere else.

david-beaumont (commented):
Discussion points and notes added.

src/jdk.compiler/share/classes/com/sun/tools/javac/file/JRTIndex.java
AlanBateman: I'm pretty sure this dates back to before the jrt file system worked with exploded builds.

src/java.base/share/classes/jdk/internal/jrtfs/JrtFileSystem.java
```diff
@@ -313,6 +314,12 @@ synchronized void cleanup() throws IOException {
...
+            // Closes the jrt-fs.jar !!!
+            ClassLoader cl = provider.getClass().getClassLoader();
+            if (cl instanceof URLClassLoader) {
+                ((URLClassLoader) cl).close();
+            }
```
AlanBateman, line 322: JrtFileSystemProvider creates the JrtFsLoader/URLClassLoader and that might be a better place to close it. A postClose or some other callback to JrtFileSystemProvider could do this. It's just a suggestion to encapsulate everything about the custom and closeable class loader in one place.

src/java.base/share/classes/jdk/internal/jimage/BasicImageReader.java
AlanBateman: The changes look right but "Non-runtime images" is a not great phase. You could speak of the current run-time image vs. a target run-time image. That way the comments are clear - use memory-mapping for the current run-time image, regular file I/O for other run-time images.

src/java.base/share/classes/jdk/internal/jimage/BasicImageReader.java
```diff
@@ -178,6 +193,16 @@ public Void run() {
...
 
+    private static ByteBuffer readDirectBuffer(FileChannel channel, int size, String name)
+            throws IOException {
+        ByteBuffer map = ByteBuffer.allocateDirect(size);
+        if (channel.read(map, 0L) != size) {
+            throw new IOException("\"" + name + "\" is not an image file");
+        }
+        map.rewind();
+        return map;
+    }
```
AlanBateman: I assume this doesn't really need to be a direct buffer, it will work equally well with a heap buffer.

In terms of naming, "readDirectBuffer" doesn't "read a direct buffer". It can be just "read". Same thing for the pre-existing "readBuffer".

As regards the `!= size` check. A short read is possible, meaning it would be valid for read to return a value < size.

src/java.base/share/classes/jdk/internal/jrtfs/JrtFileSystem.java
david-beaumont: I tried looking at this, but there's no nice way to pass the classloader, or some delegate callback across the boundary (since you can't easily add a new API because you might be talking to old versions of the code).
Unless you're will to pass the class-loader in the env map, but that's pretty nasty.
I do really want to neaten this up, but might not get time.

AlanBateman: Okay, I was thinking something simple like `provider.afterClose(this)` so it's local to the jrtfs implementation (so version mismatch issues). We can look at it another time.

src/java.base/share/classes/jdk/internal/jimage/BasicImageReader.java
```diff
@@ -402,6 +414,22 @@ private byte[] getBufferBytes(ByteBuffer buffer) {
         return bytes;
     }
 
+    /**
+     * Reads the image file to return a newly allocated buffer.
+     */
+    private ByteBuffer copyBuffer(int size)
```
AlanBateman, line 420: This reads "size" bytes from the start of the file into a buffer, and returns it. So something like readNBytes is a more appropriate name.

AlanBateman (commented):
Are you going to resolve the merge conflict and make the PR ready for review?

david-beaumont:
I'm not necessarily happy with this as it stands. It's more proof of concept.
I also don't like it being on lworld.
Once my preview mode changes are in, some/all of these changes can/should go directly into mainline.

PR #2375, opened 2026-04-29
8383567: [lworld] VM.class_print_layout should use external class names

caspernorrbin:
Hi everyone, `PrintClassLayout::print_class_layout` expects internal class names as its argument, but is called externally by users through `jcmd`. This leads to inconsistencies with other functions. For example:
```
jcmd <pid> VM.class_hierarchy java.lang.String
jcmd <pid> VM.class_print_layout java/lang/String
```

In addition, its output is also in the internal format. This should be changed to also use the external dotted class name instead. ~~For this fix I changed the `Closure` used for finding classes to instead use external names, and also swapped the output print to use the external name.~~ I added a conversion from external dotted names to internal slashed names before we start searching for classes. That way we avoid comparing strings and avoid allocations from calls to `external_name()`. As a side effect, this now works on both internal and external names, as slashes are left untouched.

Testing:
- Tier 1
- Manual inspection with `jcmd` 

---------

---------
### Progress

### Integration blocker
&nbsp;⚠️ Title mismatch between PR and JBS for issue [JDK-8383567](https://bugs.openjdk.org/browse/JDK-8383567)

### Issue
 * [JDK-8383567](https://bugs.openjdk.org/browse/JDK-8383567): VM.class_print_layout should use external class names (**Bug** - P4) ⚠️ Title mismatch between PR and JBS. ⚠️ Issue is already resolved. Consider making this a "backport pull request" by setting the PR title to `Backport <hash>` with the hash of the original commit. See [Backports](https://wiki.openjdk.org/display/SKARA/Backports).

### Reviewers
 * [David Holmes](https://openjdk.org/census#dholmes) (@dholmes-ora - Committer) ⚠️ Review applies to [eb127d34](https://git.openjdk.org/valhalla/pull/2375/files/eb127d3446954dce5e016f8bc90574d4f2b7e3a0)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2375/head:pull/2375` \
`$ git checkout pull/2375`

`$ git checkout pull/2375` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2375/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2375`

`$ git pr show -t 2375`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2375.diff">https://git.openjdk.org/valhalla/pull/2375.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2375#issuecomment-4343690088)
</details>


src/hotspot/share/memory/heapInspection.cpp
```diff
@@ -600,23 +600,19 @@ void PrintClassLayout::print_class_layout(outputStream* st, char* class_name) {
...
 
   for(int i = 0; i < klasses->length(); i++) {
     Klass* klass = klasses->at(i);
     if (!klass->is_instance_klass()) continue;  // Skip
     InstanceKlass* ik = InstanceKlass::cast(klass);
     int tab = 1;
-    st->print_cr("Class %s [@%s]:", klass->name()->as_C_string(),
+    st->print_cr("Class %s [@%s]:", klass->external_name(),
         klass->class_loader_data()->loader_name());
-    ResourceMark rm;
```
dholmes-ora, line 613: You may still need this to limit the ResourceArea usage, otherwise you will be keeping all allocations from the complete loop alive.

src/hotspot/share/memory/heapInspection.cpp
```diff
@@ -600,23 +600,19 @@ void PrintClassLayout::print_class_layout(outputStream* st, char* class_name) {
...
+  FindClassByExternalNameClosure fbnc(klasses, class_name);
   cit.iterate(&fbnc);
 
   for(int i = 0; i < klasses->length(); i++) {
     Klass* klass = klasses->at(i);
     if (!klass->is_instance_klass()) continue;  // Skip
     InstanceKlass* ik = InstanceKlass::cast(klass);
     int tab = 1;
-    st->print_cr("Class %s [@%s]:", klass->name()->as_C_string(),
+    st->print_cr("Class %s [@%s]:", klass->external_name(),
```
dholmes-ora, line 617: Doesn't it suffice to just make this change? If we convert to Symbol it seems more efficient than using `strcmp` (though I confess I'm not clear what this closure is actually doing and what set of classes it potentially generates).

caspernorrbin: I agree using `strcmp` is not very efficient, since we iterate the full `KlassInfoTable`. We still need to keep the external name as the argument however, but I reverted most changes for a different design. Now, we instead convert user input from dotted to slash form once, then use the same logic as before (probe the symbol table, then match on `klass->name()`). That way, we don't have to worry about comparing strings and we avoid lots of allocations from `external_name()`

As a side effect, this now also works on both internal and external names, as slashes would be left untouched.

src/hotspot/share/memory/heapInspection.cpp
```diff
@@ -600,12 +600,17 @@ void PrintClassLayout::print_class_layout(outputStream* st, char* class_name) {
...
 
-  Thread* THREAD = Thread::current();
-
-  Symbol* classname = SymbolTable::probe(class_name, (int)strlen(class_name));
+  ResourceMark rm;
+  char* normalized_name = NEW_RESOURCE_ARRAY(char, strlen(class_name) + 1);
+  strcpy(normalized_name, class_name);
+  for (char* p = normalized_name; *p != '\0'; p++) {
+    if (*p == '.') {
+      *p = '/';
```
dholmes-ora: This should use `JVM_SIGNATURE_DOT` and `JVM_SIGNATURE_SLASH` - though it makes me cringe that we have to do this ad-hoc conversion.

src/hotspot/share/memory/heapInspection.cpp
```diff
@@ -600,12 +600,17 @@ void PrintClassLayout::print_class_layout(outputStream* st, char* class_name) {
     return;
   }
 
-  Thread* THREAD = Thread::current();
-
-  Symbol* classname = SymbolTable::probe(class_name, (int)strlen(class_name));
+  ResourceMark rm;
+  char* normalized_name = NEW_RESOURCE_ARRAY(char, strlen(class_name) + 1);
+  strcpy(normalized_name, class_name);
```
dholmes-ora: Do we have to make a copy here rather than modifying in place? The user-supplied string is already in a mutable C-heap allocated buffer, as processed by the DCmd framework. If we think it inappropriate to modify at this level perhaps we can push the conversion up into the DCmd framework itself?

dholmes-ora (commented):
I'd missed the fact the user-supplied class name was also expected in internal form and needed to be changed to use Java form instead. I have some thoughts on that below but perhaps that is better handled in a separate RFE.

src/hotspot/share/memory/heapInspection.cpp
fparain: There's already a `Symbol::as_klass_external_name(char* buf, int size)` method to do the  '/'s into '.'s conversion. Should this code be moved to a new `Symbol::as_klass_internal_name(char* buf, int size)` method?

src/hotspot/share/memory/heapInspection.cpp
caspernorrbin: I didn't know these existed, but I swapped to use these now.

src/hotspot/share/memory/heapInspection.cpp
caspernorrbin: I kept it in the same location as of now, but removed the extra allocation. I don't think putting it in `Symbol` is the right place, because here we are just modifying a mutable `char*`. So the operation is not really on a symbol. But it could perhaps be moved to where we parse the DCmd instead, if you feel like it doesn't belong here.

src/hotspot/share/memory/heapInspection.cpp
```diff
@@ -614,9 +617,9 @@ void PrintClassLayout::print_class_layout(outputStream* st, char* class_name) {
     if (!klass->is_instance_klass()) continue;  // Skip
     InstanceKlass* ik = InstanceKlass::cast(klass);
     int tab = 1;
-    st->print_cr("Class %s [@%s]:", klass->name()->as_C_string(),
-        klass->class_loader_data()->loader_name());
     ResourceMark rm;
+    st->print_cr("Class %s [@%s]:", klass->external_name(),
+        klass->class_loader_data()->loader_name());
```
dholmes-ora:
```suggestion
    st->print_cr("Class %s [@%s]:", klass->external_name(),
                 klass->class_loader_data()->loader_name());
```
Pre-existing indent nit

dholmes-ora (approved):
Looks good.

Thanks

PR #2382, opened 2026-05-04
8380059: [lworld] zero: java/foreign/enablenativeaccess/TestEnableNativeAccessJarManifest.java failed with SIGSEGV in InterpreterRuntime::write_flat_field

johan-sjolen:
Hi,

The `zero` interpreter has fast paths for basic getters and setters, but these fail to take into account how flattened fields work. This causes flattened fields to be copied onto the stack directly, instead of wrapped in an oop (for getters), and flattened fields to be transformed into oops (for setters). Instead of adapting the fast paths, we bail to the slow path.

To convince you, this is `getter_entry`:

```c++
    switch (entry->tos_state()) {
      // SNIP!
      case atos: SET_STACK_OBJECT(object->obj_field(offset),    0); break;
```

and this is the basic bytecode dispatch (slow path):

```c++
switch (tos_type) {
  // SNIP!
case atos:
                oop val;
                if (entry->is_flat()) {
                  CALL_VM(InterpreterRuntime::read_flat_field(THREAD, obj, entry), handle_exception);
                  val = THREAD->vm_result_oop();
                  THREAD->set_vm_result_oop(nullptr);
                } else {
                  val = obj->obj_field(field_offset);
                }

```

I added a test in order to coax the getter/setter pattern optimization to run consistently and re-ran the previously disabled test along with the new one on linux-x64-zero, and they both pass. Running the new test with lworld -Xint linux-x64-zero crashes similarly to the reported bug.

## Everything below this header is bonus info for the curious

We fail within `init` of `TestSuite` which has a flattened field:

```
 - private value 'skipFailedInvocationCounts' 'Ljava/lang/Boolean;' @12  Flat inline type field 'java/lang/Boolean':
   - private final value 'value' 'Z' @12  false (0x00)
   - [null_marker] @13 Field marked as non-null
 ```

and in `SuiteRunner.java:183` we can see this:

```java
    skipFailedInvocationCounts = suite.skipFailedInvocationCounts(); // (suite has type XmlSuite)
```

and `XmlSuite` has the following getter

```java
  public Boolean skipFailedInvocationCounts() {
    return m_skipFailedInvocationCounts;
  }
```

This is probably triggering the buggy zero code.

Just some more discovery stuff:

Let's look at `si_addr` which is `0x0000000001000108`. Alright, we're running with compressed klass pointers but without `UseCompactObjectHeaders`, so when doing the `write_flat_field` we finally reach `return CompressedKlassPointers::decode_not_null(_compressed_klass);` and `_compressed_klass` has offset 8. So, the actual "oop" is at `0x1000100`. OK, I dunno, that could be a flattened Boolean I guess?

---------

---------
### Progress

### Issue
 * [JDK-8380059](https://bugs.openjdk.org/browse/JDK-8380059): [lworld] zero: java/foreign/enablenativeaccess/TestEnableNativeAccessJarManifest.java failed with SIGSEGV in InterpreterRuntime::write_flat_field (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2382/head:pull/2382` \
`$ git checkout pull/2382`

`$ git checkout pull/2382` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2382/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2382`

`$ git pr show -t 2382`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2382.diff">https://git.openjdk.org/valhalla/pull/2382.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2382#issuecomment-4386456607)
</details>

### DIFF
--- a/src/hotspot/cpu/zero/zeroInterpreter_zero.cpp
+++ b/src/hotspot/cpu/zero/zeroInterpreter_zero.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2003, 2025, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2003, 2026, Oracle and/or its affiliates. All rights reserved.
  * Copyright 2007, 2008, 2009, 2010, 2011 Red Hat, Inc.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
@@ -582,6 +582,12 @@ int ZeroInterpreter::getter_entry(Method* method, intptr_t UNUSED, TRAPS) {
     return normal_entry(method, 0, THREAD);
   }
 
+  // Flattened entries require handling beyond a direct field load.
+  // Bail to slow path.
+  if (entry->is_flat()) {
+    return normal_entry(method, 0, THREAD);
+  }
+
   ZeroStack* stack = thread->zero_stack();
   intptr_t* topOfStack = stack->sp();
 
@@ -670,6 +676,12 @@ int ZeroInterpreter::setter_entry(Method* method, intptr_t UNUSED, TRAPS) {
   ConstantPoolCache* cache = method->constants()->cache();
   ResolvedFieldEntry* entry = cache->resolved_field_entry_at(index);
   if (!entry->is_resolved(Bytecodes::_putfield)) {
+    return normal_entry(method, 0, THREAD);
+  }
+
+  // Flattened entries require handling beyond a direct field store.
+  // Bail to slow path.
+  if (entry->is_flat()) {
     return normal_entry(method, 0, THREAD);
   }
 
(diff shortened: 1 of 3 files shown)


Arraying:
>  - private value 'skipFailedInvocationCounts' 'Ljava/lang/Boolean;' @12  Flat inline type field 'java/lang/Boolean':

I'm not super familiar with the Zero internals or potential optimizations we have/could have baked in. However, we can't really yield that many field flattening benefits with just an interpreter. Perhaps we still enjoy a lower overall allocation rate when only buffering at points of field access? Would it make sense to revisit the flattening decisions when running Zero? As in, disable field flattening entirely and guard cases like these with assertions.

johan-sjolen:
tier1-tier5 with enable preview looks good.

johan-sjolen:
> > * private value 'skipFailedInvocationCounts' 'Ljava/lang/Boolean;' @12  Flat inline type field 'java/lang/Boolean':
> 
> I'm not super familiar with the Zero internals or potential optimizations we have/could have baked in. However, we can't really yield that many field flattening benefits with just an interpreter. Perhaps we still enjoy a lower overall allocation rate when only buffering at points of field access? Would it make sense to revisit the flattening decisions when running Zero? As in, disable field flattening entirely and guard cases like these with assertions.

We could always set `UseFieldFlattening` and `UseArrayFlattening` to false if we build with zero.

coleenp (approved):
This makes sense.

Arraying (approved):
The change looks good, thanks! Let's defer the discussion of if Zero should support flattening.

johan-sjolen:
/integrate

Thank you for the reviews, let's see what the future holds for Zero + Valhalla.

PR #2389, opened 2026-05-05
8383565: [lworld] NativeNMethodBarrier does not correctly handle multiple entry points

Arraying:
Hi all,

This patch fixes some performance issues with AArch64's `NMethodEntryBarrier`.

The purpose of a method entrypoint is to act as a barrier for potential GC work. Entrypoints can be "armed", such that when the method is called, a slow-path is triggered. This slow-path then does any necessary fixups. 

With the introduction of scalarized calling conventions in Valhalla, methods have multiple entrypoints, depending on the calling conventions that are used. This is due to the fact that `nmethod`s cannot be multiplexed when scalarized. There can be up to three entrypoints for a given method. The locations of these entrypoints is tracked via relocations. 

The problem lies within the usage of `RelocIterator`. When `static void set_value` is called (such as when arming entrypoints), it creates a `NativeNMethodBarrier` for each entrypoint. However, the `RelocIterator` only finds the first relocation, meaning each embodiment of `NativeNMethodBarrier` sees the same `_guard_addr`. Consequently, executions entering through the other entrypoints will not see the up-to-date armed value, and be unnecessarily slow-pathed. 

This patch introduces several improvements:
1. Makes `NativeNMethodBarrier` use the `ldr` instructions to determine the `_guard_addr`s, as opposed to relocations (contributed by @TobiHartmann). This gets the correct addresses for each of the guards, and avoids having to iterate through the `nmethod`'s relocations, which is expensive for large methods. 
2. Consolidates all the entrypoints as first-class citizens of `NativeNMethodBarrier`, instead of having an `NativeNMethodBarrier` per entrypoint. This reduces object allocation, and makes the code much easier to read for free! 
3. Stricter assertions. All entrypoints `ldr` instructions are verified (as opposed to just the scalarized ones previously). All guard addresses are verified to be within the `nmethod`'s instructions/stubs.

Testing: tiers 1-4 and some adhoc testing with `--enable-preview`. 

---------

---------
### Progress

### Issue
 * [JDK-8383565](https://bugs.openjdk.org/browse/JDK-8383565): [lworld] NativeNMethodBarrier does not correctly handle multiple entry points (**Bug** - P3)

### Reviewers
 * [Joel Sikström](https://openjdk.org/census#jsikstro) (@jsikstro - Committer) ⚠️ Review applies to [6ee68586](https://git.openjdk.org/valhalla/pull/2389/files/6ee6858667d03a249f5113f61c3a484ea5a3c6b0)
 * [Tobias Hartmann](https://openjdk.org/census#thartmann) (@TobiHartmann - Committer) ⚠️ Review applies to [6ee68586](https://git.openjdk.org/valhalla/pull/2389/files/6ee6858667d03a249f5113f61c3a484ea5a3c6b0)

### Contributors
 * Tobias Hartmann `<thartmann@openjdk.org>`

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2389/head:pull/2389` \
`$ git checkout pull/2389`

`$ git checkout pull/2389` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2389/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2389`

`$ git pr show -t 2389`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2389.diff">https://git.openjdk.org/valhalla/pull/2389.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2389#issuecomment-4379746428)
</details>

### DIFF
--- a/src/hotspot/cpu/aarch64/c2_CodeStubs_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/c2_CodeStubs_aarch64.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020, 2025, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2020, 2026, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -59,7 +59,6 @@ void C2EntryBarrierStub::emit(C2_MacroAssembler& masm) {
   __ b(continuation());
 
   __ bind(guard());
-  __ relocate(entry_guard_Relocation::spec());
   __ emit_int32(0);   // nmethod guard value
 }
 
--- a/src/hotspot/cpu/aarch64/gc/shared/barrierSetNMethod_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/gc/shared/barrierSetNMethod_aarch64.cpp
@@ -72,23 +72,44 @@ static int* decode_guard_from_instruction(nmethod* nm, address& instruction) {
   return result;
 }
 
+// The NativeNMethodBarrier class encapsulates up to three entrypoints and handles their
+// arming/verification.
+// An entrypoint is defined as a tuple of <instr. address, guard address>:
+// * The instr. address corresponds to the ldr of the guard value of that entrypoint.
+// * The guard address is the address where the guard value of that entrypoint resides.
+//
+// Each nmethod has at least one entrypoint. The default must always be well-defined
+// (neither instruction nor guard are nullptr).
+//
+// When using C1/C2 and scalarization, up to two additional (verified) entrypoints,
+// alt1 and alt2 can be present. The meaning of these depends on who compiled the nmethod.
+//
+// The mapping of C1-compiled methods (scalarization used) looks as follows:
+// * alt1: verified entry point
+// * alt2 (optional): verified inline ro entry point
+//
+// The mapping of C2-compiled methods (scalarization used) looks as follows:
+// * alt1: verified inline entry point
+// * alt2 (optional): verified inline ro entry point
+//
+// In other scenarios, neither alt1 nor alt2 are defined.
 class NativeNMethodBarrier {
  private:
   // The addresses of the instructions that act as the guards.
-  address _verified_entry_instruction;
+  address _default_entry_instruction;
   address _verified_alt1_instruction;
   address _verified_alt2_instruction;
   // Pointers representing the actual guard values themselves.
-  int* _verified_entry_guard;
+  int* _default_entry_guard;
   int* _verified_alt1_guard;
   int* _verified_alt2_guard;
 
  public:
   NativeNMethodBarrier(nmethod* nm) :
-    _verified_entry_instruction(nullptr),
+    _default_entry_instruction(nullptr),
     _verified_alt1_instruction(nullptr),
     _verified_alt2_instruction(nullptr),
-    _verified_entry_guard(nullptr),
+    _default_entry_guard(nullptr),
     _verified_alt1_guard(nullptr),
     _verified_alt2_guard(nullptr) {
 #if INCLUDE_JVMCI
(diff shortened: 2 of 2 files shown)


src/hotspot/cpu/aarch64/c2_CodeStubs_aarch64.cpp
```diff
@@ -59,7 +59,6 @@ void C2EntryBarrierStub::emit(C2_MacroAssembler& masm) {
   __ b(continuation());
 
   __ bind(guard());
-  __ relocate(entry_guard_Relocation::spec());
```
Arraying, line 62: Note that RISC-V still uses `entry_guard_Relocation` so the relocation itself cannot be cleaned up.

src/hotspot/cpu/aarch64/gc/shared/barrierSetNMethod_aarch64.cpp
```diff
@@ -63,99 +64,172 @@ static int entry_barrier_offset(nmethod* nm) {
...
 
-  bool check_barrier(err_msg& msg) const;
-  void verify() const {
-    err_msg msg("%s", "");
-    assert(check_barrier(msg), "%s", msg.buffer());
+  // Checks the validity of a single barrier.
+  bool check_barrier_impl(address& instruction, err_msg& msg) {
+    uint32_t* addr = (uint32_t*) instruction;
+    uint32_t inst = *addr;
+    if ((inst & 0xff000000) != 0x18000000) {
```
jsikstro, line 224: Pre-existing: We should probably follow up on this in mainline later on, this is hard to read.

src/hotspot/cpu/aarch64/gc/shared/barrierSetNMethod_aarch64.cpp
```diff
@@ -63,99 +64,172 @@ static int entry_barrier_offset(nmethod* nm) {
...
-              _guard_addr = reinterpret_cast<int*>(reloc->addr());
-              return;
-            }
-          }
-          ShouldNotReachHere();
+    {
+      // The default entry point has a known address. The guard address can be
+      // decoded from the literal in the instruction. Verification will confirm
+      // that this instruction corresponds to a load.
+      _default_entry_instruction = nm->code_begin() + nm->frame_complete_offset() + entry_barrier_offset(nm);
```
jsikstro, line 131: Now that this is the only place calling `entry_barrier_offset` it would be really nice to make it return a positive value and subtract here instead to make this easier to read and understand.

src/hotspot/cpu/aarch64/gc/shared/barrierSetNMethod_aarch64.cpp
```diff
@@ -240,7 +289,8 @@ void BarrierSetNMethod::set_guard_value(nmethod* nm, int value, int bit_mask) {
   // stub.
   MACOS_AARCH64_ONLY(ThreadWXEnable wx(WXWrite, Thread::current()));
 
-  set_value(nm, value, bit_mask);
+  NativeNMethodBarrier barrier(nm);
+  barrier.set_values(value, bit_mask);
```
jsikstro, line 293: I see, so this is where we initially only set the "default" guard value, and now we're setting all of them. Neat.

jsikstro (approved):
Looks good. This is an interesting area! It's unfortunate that we now have (up to) three separate guard values, it would be nice if we only had one.

I have some comments but we don't have to do them now if we want to get this in.

dean-long:
I agree it would be better to have only a single guard value word in each nmethod.  We could either emit one slot (probably in the stubs area) and have all the stubs share it, or have only one shared stub.  The latter would involve using a BLR to branch to the stub instead of using the continuation label.

src/hotspot/cpu/aarch64/gc/shared/barrierSetNMethod_aarch64.cpp
```diff
@@ -63,99 +64,172 @@ static int entry_barrier_offset(nmethod* nm) {
...
+// The NativeNMethodBarrier class encapsulates up to three entrypoints and handles their
+// arming/verification.
+// An entrypoint is defined as a tuple of <instr. address, guard address>:
+// * The instr. address corresponds to the ldr of the guard value of that entrypoint.
+// * The guard address is the address where the guard value of that entrypoint resides.
+//
+// Each nmethod has at least one entrypoint. The default must always be well-defined
+// (neither instruction nor guard are nullptr).
+//
+// When using C1/C2 and scalarization, up to two additional (verified) entrypoints,
```
TobiHartmann:
```suggestion
// When using the scalarized calling convention, up to two additional (verified) entrypoints,
```

TobiHartmann (approved):
This looks good to me as a point fix for the performance regression. I agree with @dean-long and @jsikstro though that it would be nice to optimize this to only use one guard value, but I think this can be done as a follow-up enhancement.

Arraying:
Thanks for the feedback everyone. I've filed [JDK-8384078](https://bugs.openjdk.org/browse/JDK-8384078) and [JDK-8384077](https://bugs.openjdk.org/browse/JDK-8384077) as follow-up work.

Arraying:
Thanks all for the reviews.
/integrate

PR #2390, opened 2026-05-05
8383559: [lworld] Autobox cache removal causes performance regression with Renaissance benchmarks

TobiHartmann:
To re-enable the autobox cache, this patch reverts (parts of) [JDK-8369921](https://bugs.openjdk.org/browse/JDK-8369921), [JDK-8378476](https://bugs.openjdk.org/browse/JDK-8378476), deopt parts of [JDK-8378531](https://bugs.openjdk.org/browse/JDK-8378531) and [JDK-8379148](https://bugs.openjdk.org/browse/JDK-8379148).

I had to modify the `newCacheArray` methods to return a non-flattened array to avoid buffering.

I wasn't sure about the comments above the `valueOf` methods:
https://github.com/openjdk/valhalla/commit/b3ccb6abfdb6e2bff59a7d79bf01b669ac319257#diff-6136f7a80110ec00bf9a0b7da9943eb2fc33025ab400bc1f70153c1b2ac36993R146

Should they be reverted too? For now, I left them as is.

Thanks,
Tobias

---------

---------
### Progress

### Issue
 * [JDK-8383559](https://bugs.openjdk.org/browse/JDK-8383559): [lworld] Autobox cache removal causes performance regression with Renaissance benchmarks (**Bug** - P2)

### Reviewers
 * [Ioi Lam](https://openjdk.org/census#iklam) (@iklam - Committer)
 * [Matias Saavedra Silva](https://openjdk.org/census#matsaave) (@matias9927 - Committer)

### Contributors
 * Ioi Lam `<iklam@openjdk.org>`

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2390/head:pull/2390` \
`$ git checkout pull/2390`

`$ git checkout pull/2390` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2390/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2390`

`$ git pr show -t 2390`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2390.diff">https://git.openjdk.org/valhalla/pull/2390.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2390#issuecomment-4380642501)
</details>


src/hotspot/share/opto/type.cpp
```diff
@@ -327,7 +327,7 @@ const Type* Type::make_from_constant(ciConstant constant, bool require_constant,
         } else {
           guarantee(require_constant || oop_constant->should_be_constant(), "con_type must get computed");
           con_type = TypeOopPtr::make_from_constant(oop_constant, require_constant);
-          if (Compile::current()->eliminate_boxing() && is_autobox_cache) {
+          if (is_autobox_cache) {
```
TobiHartmann:
```suggestion
          if (Compile::current()->eliminate_boxing() && is_autobox_cache) {
```

TobiHartmann: This should be reverted to mainline.

src/java.base/share/classes/java/lang/Short.java
viktorklang-ora, line 299: It's likely that this needs to get updated in this PR as well (divergence between spec and impl—also see the other methods that are affected in the same way).

viktorklang-ora: ( as discussed in the PR description )

src/hotspot/share/runtime/deoptimization.cpp
```diff
@@ -1256,10 +1255,6 @@ class BooleanBoxCache : public BoxCacheBase<java_lang_Boolean> {
 BooleanBoxCache* BooleanBoxCache::_singleton = nullptr;
 
 oop Deoptimization::get_cached_box(AutoBoxObjectValue* bv, frame* fr, RegisterMap* reg_map, bool& cache_init_error, TRAPS) {
```
Arraying, line 1261: None of our box caches should be flat arrays. This should also apply to `BooleanBoxCache`, whose fields `TRUE` and `FALSE` should not be flattened. This is currently the case because we don't flatten `static` fields, but I think it'd be good to explicitly express these invariants as assertions somewhere (does not have to be here). Or even a comment saying to look at the JDKland code, so HotSpot readers see the absence of flattening.

src/java.base/share/classes/java/lang/Character.java
```diff
@@ -9435,24 +9436,28 @@ private CharacterCache(){}
         static Character[] archivedCache;
 
         static {
-            if (!PreviewFeatures.isEnabled()) {
-                int size = 127 + 1;
+            int size = 127 + 1;
 
-                // Load and use the archived cache if it exists
+            // Load and use the archived cache if it exists
+            if (!PreviewFeatures.isEnabled()) {
```
Arraying: I don't think there is anything preventing us from using the CDS archive. We're currently archiving the `cache` even when running with `--enable-preview`, just not loading it.

src/java.base/share/classes/java/lang/Integer.java
```diff
@@ -939,18 +935,22 @@ private static void runtimeSetup() {
...
+                    // Load IntegerCache.archivedCache from archive, if possible
+                    CDS.initializeFromArchive(IntegerCache.class);
+                    precomputed = archivedCache;
+                }
             }
 
             cache = loadOrInitializeCache(precomputed);
-            archivedCache = cache; // Legacy CDS archive support (to be deprecated)
+            if (!PreviewFeatures.isEnabled()) {
+                archivedCache = cache; // Legacy CDS archive support (to be deprecated)
```
Arraying: I really wonder why this was guarded, if it's the same reason as the `CharacterCache` or because AOT specifically won't work with it.

Arraying (commented):
The code changes look good overall, thanks! Maybe @matias9927 knows more about the AOT-specifics here and can weigh in as well.

I don't really have the expertise on specification to weigh in, but it seems like Victor has already pointed out some mistakes.

src/java.base/share/classes/java/lang/Byte.java
```diff
@@ -123,25 +124,29 @@ private ByteCache() {}
...
-            } else {
-                cache = null;
-                assert archivedCache == null;
+                archivedCache = c;
             }
+            cache = archivedCache;
+            assert cache.length == size;
+        }
+
+        private static Byte[] newCacheArray(int size) {
```
liach, line 143: I don't think we need this - we should be able to revert to mainline code. The only code I think that need any update would be the old `new Byte[size]` should become `(Byte[]) ValueClass.newReferenceArray(Byte.class, size)` both with and without preview.

src/java.base/share/classes/java/lang/Byte.java
```diff
@@ -123,25 +124,29 @@ private ByteCache() {}
...
 
         static {
-            if (!PreviewFeatures.isEnabled()) {
-                final int size = -(-128) + 127 + 1;
+            final int size = -(-128) + 127 + 1;
 
-                // Load and use the archived cache if it exists
+            // Load and use the archived cache if it exists
+            if (!PreviewFeatures.isEnabled()) {
                 CDS.initializeFromArchive(ByteCache.class);
```
viktorklang-ora: What's the reason why CDS can't be used when Preview is enabled? Might make sense to document that in a comment here

liach:
For [JDK-8383894](https://bugs.openjdk.org/browse/JDK-8383894), I think we need to move the `@DeserializeConstructor` annotations from the `valueOf` methods to the constructors to eliminate that particular regression. Can you include that in your patch?

src/java.base/share/classes/java/lang/Character.java
iklam: I think it's not necessary. I have a patch that removes this check around the CDS code. I ran all the CDS tests that have `--enable-preview` in them and they all passed. I'll try to run more tests now.

https://github.com/iklam/valhalla/commit/a91c40735d760be7d160f2d72e53f48fecb08b49

matias9927: @iklam's patch seems to work but I want to point out that `ArchivedIntegerCacheTest.java` needs to be updated:
`@requires !java.enablePreview` should be removed.

TobiHartmann:
Thanks for the reviews! I've been working on https://git.openjdk.org/valhalla/pull/2406 which was blocking this PR and will get to this next week.

src/java.base/share/classes/java/lang/Character.java
TobiHartmann: Thanks! I applied Ioi's fix and updated `ArchivedIntegerCacheTest.java`.

src/hotspot/share/runtime/deoptimization.cpp
TobiHartmann: I added assertions that box caches are reference arrays and Boolean cache fields are initialized.

src/java.base/share/classes/java/lang/Byte.java
TobiHartmann: Fixed.

src/java.base/share/classes/java/lang/Short.java
TobiHartmann: I added missing preview-mode valueOf spec texts for Boolean, Float, and Double.

TobiHartmann: @viktorklang-ora Is there anything else that is missing?

TobiHartmann:
> I think we need to move the @DeserializeConstructor annotations from the valueOf methods to the constructors to eliminate that particular regression. Can you include that in your patch?

Done. Maybe @kuksenko could verify that this solves [JDK-8383894](https://bugs.openjdk.org/browse/JDK-8383894).

src/java.base/share/classes/java/lang/Byte.java
TobiHartmann: > new Byte[size] should become (Byte[]) ValueClass.newReferenceArray(Byte.class, size) both with and without preview.

@liach That's not possible because `ValueClass.newReferenceArray` throws when preview is disabled because boxes are identity classes then.

TobiHartmann:
Thanks for the reviews! I addressed all comments and will re-run testing.

test/hotspot/jtreg/runtime/cds/appcds/cacheObject/ArchivedIntegerCacheTest.java
```diff
@@ -25,7 +25,6 @@
 /*
  * @test
  * @summary Test primitive box caches integrity in various scenarios (IntegerCache etc)
- * @requires !java.enablePreview
```
viktorklang-ora, line 28: Are we sure that there are no other tests that need their requires-clauses updated as a part of this change?

src/java.base/share/classes/java/lang/Short.java
viktorklang-ora: Not that I can think of. I have been wrong before tho.

test/hotspot/jtreg/runtime/cds/appcds/cacheObject/ArchivedIntegerCacheTest.java
TobiHartmann: I had a quick look and nothing else stood out. But I defer to people more familiar in this area.

TobiHartmann:
If there are no strong objections, I propose to integrate this as is because some other work depends on it. I would leave it to the core-libs / CDS experts to follow-up with more cleanups.

iklam (approved):
CDS changes look OK.

matias9927 (approved):
CDS changes look good!

liach (approved):
Core libs updates look good.

TobiHartmann:
Thanks for the reviews. I'll wait for testing to all pass before integrating (probably only on Friday because Thursday is a public holiday here).

PR #2397, opened 2026-05-07
8383755: [lworld] Clean up Class.isValue and Class.isIdentity

dlsmith-dev:
Two changes:

- Remove the `Class.isIdentity` method, which is redundant given the `Class.isValue` method
- Clean up `Class.isValue` so that it only returns true for value classes, not other non-identity things (interfaces, primitives)

---------

---------
### Progress

### Issue
 * [JDK-8383755](https://bugs.openjdk.org/browse/JDK-8383755): [lworld] Clean up Class.isValue and Class.isIdentity (**Bug** - P2)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2397/head:pull/2397` \
`$ git checkout pull/2397`

`$ git checkout pull/2397` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2397/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2397`

`$ git pr show -t 2397`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2397.diff">https://git.openjdk.org/valhalla/pull/2397.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2397#issuecomment-4393367823)
</details>


src/java.base/share/classes/jdk/internal/value/ValueClass.java
```diff
@@ -43,13 +43,14 @@ public final class ValueClass {
...
+                (fieldType.isValue() ||
+                 fieldType.isInterface() ||
+                 fieldType == Object.class);
     }
 
     /// {@return whether an object of this exact class is a value object}
-    /// This excludes abstract value classes and primitives.
+    /// This excludes abstract value classes.
     public static boolean isConcreteValueClass(Class<?> clazz) {
         return clazz.isValue() && !Modifier.isAbstract(clazz.getModifiers());
```
liach, line 55: Now this can be `clazz.isValue() && Modifier.isFinal(clazz.getModifiers())`, before primitives would report they are value but they are `abstract final` and thus fails the old test.

src/java.base/share/classes/java/lang/Class.java
```diff
@@ -626,52 +626,21 @@ public static Class<?> forName(Module module, String name) {
...
      * @see AccessFlag#IDENTITY
      * @since Valhalla
      */
     @PreviewFeature(feature = PreviewFeature.Feature.VALUE_OBJECTS, reflective=true)
     public boolean isValue() {
-        if (!PreviewFeatures.isEnabled()) {
-            return false;
-        }
-        return !isIdentity();
+        return !identity && !primitive && !isInterface();
```
liach: We can update the VM representation to change `identity` injection to `value` injection as a separate RFE probably for post lworld.

liach (approved):

dlsmith-dev:
Interested in feedback: is this the right behavior for the method?

My thinking is:

- The reflection API is mainly focused on language-oriented terminology, even as it is rooted in class files. "Constructor", not "instance initializer". Awareness of anonymous and inner classes. Hence `isValue` rather than `isIdentity`.
- The reflection API tends to offer a simple _syntactic_ presentation rather than more complex _semantic_ properties. No subtyping or generic substitution. No inheritance or overriding relationships. Hence "is this a class with a value modifier" rather than "is this an entity that doesn't need identity?" or "is this a type that could have identityless values?".

src/java.base/share/classes/jdk/internal/value/ValueClass.java
dansmithcode: We don't have non-final concrete value classes today, but if we ever did, I guess it's best if the implementation matches the method name.

src/java.base/share/classes/java/lang/Class.java
dansmithcode: Yeah, it's possible, but I wasn't totally sure what other components might be inspecting this field directly. Didn't want to rock the boat. I agree it's worth considering as a future refactoring.

liach:
I think this is the right behavior for `isValue` to only report `true` for abstract and concrete value classes. I have no opinion on whether `isValue` or `isIdentity` is better.

src/java.base/share/classes/jdk/internal/misc/Unsafe.java
```diff
@@ -1730,7 +1731,7 @@ public final <V> boolean compareAndSetReference(Object o, long offset,
                                                     Class<?> type,
                                                     V expected,
                                                     V x) {
-        if (type.isValue() || isValueObject(expected)) {
+        if (isValueObject(expected)) {
```
dansmithcode, line 1734: Discussed with Fred: all we're trying to do here is quickly check for cases where a pointer comparison won't work as a stand-in for `==`. If `expected` is not a value object, a pointer comparison is just fine. No need to consider the type of the container.

src/java.base/share/classes/java/lang/Class.java
coleenp: I didn't really like injecting 'identity'.   java.lang.Class can figure this out optimally without the JVM's help.
The JVM just sets the value based on whether it's an array and the ACC_IDENTITY access flag is set.

  set_is_identity(mirror(), k->is_array_klass() || k->is_identity_class());

You could cache the answer if you'd like in a field:
   value = !isArray() && getClassAccessFlags() has ACC_IDENTITY (why not in modifiers??) && !primitive && !isInterface();

Anyways, you can check the performance and fix this later.

dansmithcode: Good to know! Somehow I was thinking it was filtered out of `modifiers`, but it's right there, so easy to check directly.

I tried to see what happens if I delete the field & constructor parameter, but stuff breaks without many breadcrumbs, so I'll file a bug and leave that to somebody who knows what they're doing. 😄

dansmithcode: Filed JDK-8384193

src/java.base/share/classes/java/lang/Class.java
```diff
@@ -640,7 +641,12 @@ public static Class<?> forName(Module module, String name) {
      */
     @PreviewFeature(feature = PreviewFeature.Feature.VALUE_OBJECTS, reflective=true)
     public boolean isValue() {
-        return !identity && !primitive && !isInterface();
+        if (!PreviewFeatures.isEnabled()) {
+            return false;
+        } else {
+            int mask = IDENTITY | Modifier.INTERFACE;
```
liach: Le's use ClassFile.ACC_ constants instead of Modifier ones or adding ones to Class.

dansmithcode: Ok, but if we do that, might as well get rid of all four of them. See latest push.

liach (approved):
Latest push look good. We can clean up mainline java.lang.Class in a pre-401 patch or as part of 401 integration.

forax:
Are you sure you want to add a dependency from java.lang.Class to java.lang.classfile.ClassFile ?

Or you are confident that because those a static final of primitive type, the compiler will not add that dependency,
in that case, maybe a comment would be useful

liach:
Even a compiler dependency would be fine - Class itself even uses Stream in its methods such as `getPermittedSubclasses`. As long as the dependencies are not triggered in class initialization I think we are safe.

PR #2405, opened 2026-05-07
8384068: [lworld] ValueClassPlugin misses some annotations

xmas92:
There are few places which uses the fully qualified name for the `AsValueClass` annotation. As a result the plugging does not modify these classes when compiling, so they are still identity classes.

I propose improving the robustness of this plugin. We need to do this during parsing so we cannot actually check 100% that it will resolve to the correct annotation. However we can do a best effort, which handles same package, fully qualified, imported and rejects other annotations with the same class name.

---------

---------
### Progress

### Integration blocker
&nbsp;⚠️ Title mismatch between PR and JBS for issue [JDK-8384068](https://bugs.openjdk.org/browse/JDK-8384068)

### Issue
 * [JDK-8384068](https://bugs.openjdk.org/browse/JDK-8384068): ValueClassPlugin misses some annotations (**Bug** - P4) ⚠️ Title mismatch between PR and JBS.

### Reviewers
 * [Johan Sjölen](https://openjdk.org/census#jsjolen) (@johan-sjolen - no project role)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2405/head:pull/2405` \
`$ git checkout pull/2405`

`$ git checkout pull/2405` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2405/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2405`

`$ git pr show -t 2405`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2405.diff">https://git.openjdk.org/valhalla/pull/2405.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2405#issuecomment-4394817561)
</details>


xmas92:
Just a friendly ping.

The only tests that are currently being affected is `gc/stress/gcbasher`. So we are missing some EnablePreview testing here. Would be nice to get this in both for test coverage and to ensure that we do not miss future uses of `@jdk.test.lib.valueclass.AsValueClass`.

jsikstro (commented):
I think this is going in the right direction. One question:

What if we have a case like below, that will match as true when it shouldn't. But I guess that's rare/bad code.
```
package jdk.test.lib.valueclass;

import other.AsValueClass;

@AsValueClass
public final class ShadowInSamePackage {}
```

jsikstro (approved):
This is not my area, but I think the revised logic around checking for the `AsValueClass` annotation is sound and seem to slot it nicely with the surrounding code.

johan-sjolen (approved):
This should obviously be done at semantic analysis-time (well, post-semantic analysis-time). I don't get why it has to be done at parse time, but OK, if you say so then I believe you. Besides, this is an improvement.

PR #2406, opened 2026-05-07
8384052: [lworld] Default refined klass invariant in ObjArrayKlass::klass_from_description is broken

TobiHartmann:
https://github.com/openjdk/valhalla/pull/2390 triggered some weird failures and it turned out that this invariant is broken when `ValueClass.newReferenceArray` is called before the default refined klass is created:
https://github.com/openjdk/valhalla/blob/7558cd73f557e34b0f11dd0a37952854ab2881f6/src/hotspot/share/oops/objArrayKlass.cpp#L410-L414

In that case, the first entry in the linked list will not be the default refined klass but the not-flat, nullable variant. This leads to the code emitted by `LibraryCallKit::load_default_refined_array_klass` returning the wrong refined klass and thus to wrong results with the `_copyOf` and `_newArray` intrinsics.

I added a targeted regression test for this that covers both intrinsics and randomizes the order in which refined klasses are created.

The fix was provided by @fparain.

Thanks,
Tobias

---------

---------
### Progress

### Issue
 * [JDK-8384052](https://bugs.openjdk.org/browse/JDK-8384052): [lworld] Default refined klass invariant in ObjArrayKlass::klass_from_description is broken (**Bug** - P3)

### Reviewers
 * [Frederic Parain](https://openjdk.org/census#fparain) (@fparain - Committer) ⚠️ Review applies to [51c2978d](https://git.openjdk.org/valhalla/pull/2406/files/51c2978da704b27406a91f0306dbf2346e8006bc)

### Contributors

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2406/head:pull/2406` \
`$ git checkout pull/2406`

`$ git checkout pull/2406` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2406/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2406`

`$ git pr show -t 2406`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2406.diff">https://git.openjdk.org/valhalla/pull/2406.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2406#issuecomment-4395238534)
</details>

### DIFF
--- a/src/hotspot/share/opto/subnode.cpp
+++ b/src/hotspot/share/opto/subnode.cpp
@@ -1636,8 +1636,8 @@ Node *BoolNode::Ideal(PhaseGVN *phase, bool can_reshape) {
   Node *cmp = in(1);
   if( !cmp->is_Sub() ) return nullptr;
   int cop = cmp->Opcode();
-  if( cop == Op_FastLock || cop == Op_FastUnlock ||
-      cmp->is_SubTypeCheck() || cop == Op_VectorTest ) {
+  if (cop == Op_FastLock || cop == Op_FastUnlock || cop == Op_FlatArrayCheck ||
+      cmp->is_SubTypeCheck() || cop == Op_VectorTest) {
     return nullptr;
   }
   Node *cmp1 = cmp->in(1);
--- a/src/hotspot/share/oops/objArrayKlass.cpp
+++ b/src/hotspot/share/oops/objArrayKlass.cpp
@@ -392,6 +392,7 @@ ObjArrayKlass* ObjArrayKlass::klass_from_description(ArrayDescription ad, TRAPS)
   element_klass()->validate_array_description(ad);
 
   const ArrayProperties props = ad._properties;
+  assert(props.is_valid(), "must be");
 
   if (properties() == props && kind() == ad._kind) {
     assert(is_refined_objArray_klass(), "Must be a refined array klass");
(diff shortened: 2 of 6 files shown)


src/hotspot/share/opto/subnode.cpp
```diff
@@ -1636,8 +1636,8 @@ Node *BoolNode::Ideal(PhaseGVN *phase, bool can_reshape) {
   Node *cmp = in(1);
   if( !cmp->is_Sub() ) return nullptr;
   int cop = cmp->Opcode();
-  if( cop == Op_FastLock || cop == Op_FastUnlock ||
-      cmp->is_SubTypeCheck() || cop == Op_VectorTest ) {
+  if (cop == Op_FastLock || cop == Op_FastUnlock || cop == Op_FlatArrayCheck ||
```
TobiHartmann, line 1639: Unrelated fix for an issue I hit during debugging.

fparain (approved):
LGTM
Thank you for finding this bug and taking care of the fix.

TobiHartmann:
Thanks Fred! The test still triggers an intermittent issue that I'm investigating and I'll therefore hold off with integration until I found the root cause.

TobiHartmann:
The logic in the copyOf intrinsic is broken and should be re-worked with [JDK-8251971](https://bugs.openjdk.org/browse/JDK-8251971). I removed the old logic and added (too strong) checks.

PR #2409, opened 2026-05-07
8382584: [lworld] Mandatory pre-loading still present in systemDictionary.cpp

caspernorrbin:
Hi everyone,

This change removes the remaining mandatory preloading path for null-restricted instance fields when loading classes from CDS. [JDK-8380053](https://bugs.openjdk.org/browse/JDK-8380053) already changed normal class loading so null-restricted field types are no longer loaded just because the field is null-restriced. Instead, the preloading is speculative and driven by the `LoadableDescriptor` attribute. The CDS path still had the older special case in `SystemDictionary::preload_from_null_free_field`, which could force loading of null-restricted field types.

This fix removes that CDS-only preload helper and routes non-static field signatures through the existing `LoadableDescriptor` path instead. This makes loading follow the same model as in `ClassFileParser::fetch_field_classes()` with speculative loading.

Testing:
- Tier 1-3

---------

---------
### Progress

### Issue
 * [JDK-8382584](https://bugs.openjdk.org/browse/JDK-8382584): [lworld] Mandatory pre-loading still present in systemDictionary.cpp (**Bug** - P2)

### Reviewers
 * [Coleen Phillimore](https://openjdk.org/census#coleenp) (@coleenp - Committer) ⚠️ Review applies to [b24ca875](https://git.openjdk.org/valhalla/pull/2409/files/b24ca875cd65033ca021620c12c61c5e7828c8c0)
 * [Frederic Parain](https://openjdk.org/census#fparain) (@fparain - Committer) ⚠️ Review applies to [20feb714](https://git.openjdk.org/valhalla/pull/2409/files/20feb714afc680758345ac34587409b65273fbce)
 * [Ioi Lam](https://openjdk.org/census#iklam) (@iklam - Committer)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2409/head:pull/2409` \
`$ git checkout pull/2409`

`$ git checkout pull/2409` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2409/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2409`

`$ git pr show -t 2409`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2409.diff">https://git.openjdk.org/valhalla/pull/2409.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2409#issuecomment-4396159444)
</details>


src/hotspot/share/classfile/systemDictionary.cpp
```diff
@@ -1203,14 +1167,7 @@ InstanceKlass* SystemDictionary::load_shared_class(InstanceKlass* ik,
       Symbol* sig = fs.signature();
       int field_index = fs.index();
 
-      if (fs.is_null_free_inline_type()) {
-        // A false return means that the class didn't load for other reasons than an exception.
-        bool check = preload_from_null_free_field(ik, class_loader, sig, field_index, CHECK_NULL);
-        if (!check) {
-          ik->set_shared_loading_failed();
```
liach, line 1210: Can we safely drop this failure marking?

coleenp: It's used for regular shared class loading so I don't think we can drop this shared_loading_failed marker.

coleenp (approved):
Looks good!

src/hotspot/share/classfile/systemDictionary.cpp
caspernorrbin: Didn't realise that one was needed. I've re-added it, along with an edit to `try_preload_from_loadable_descriptors` so we can see if we failed or not. It now also has a new parameter that decides if failing means we call `shared_loading_failed `.

caspernorrbin:
Following our offline discussions, I have now pushed a more robust version that:
- Validates archived flat/null-restricted field metadata by resolving the field class at runtime and ensuring it matches the archived dependency.
- Renames `check_shared_class_super_type` to `check_shared_class_dependency`, since the helper is now used for more than super classes/interfaces.
- Adds a regression test that strips LoadableDescriptors and verifies the archived class is still rejected when its field dependency is substituted.

src/hotspot/share/classfile/systemDictionary.cpp
```diff
@@ -1106,43 +1106,41 @@ bool SystemDictionary::check_shared_class_super_types(InstanceKlass* ik, Handle
   return true;
 }
 
-// Pre-load class referred to in non-static null-free instance field. These fields trigger MANDATORY loading.
-// Some pre-loading does not fail fatally
-bool SystemDictionary::preload_from_null_free_field(InstanceKlass* ik, Handle class_loader, Symbol* sig, int field_index, TRAPS) {
+// Pre-load class referred to in non-static fields with archived inline field metadata. These fields
+// must be checked against the resolved runtime class before the shared class can be used.
+bool SystemDictionary::preload_from_required_inline_field(InstanceKlass* ik, Handle class_loader, Symbol* sig, int field_index, TRAPS) {
   TempNewSymbol name = Signature::strip_envelope(sig);
```
fparain: This symbol is only used in logging, could it be possible to create it only when a message is logged?

src/hotspot/share/classfile/systemDictionary.cpp
```diff
@@ -1203,15 +1200,18 @@ InstanceKlass* SystemDictionary::load_shared_class(InstanceKlass* ik,
...
+      if (fs.is_flat() || fs.is_null_free_inline_type()) {
+        bool check = preload_from_required_inline_field(ik, class_loader, sig, field_index, CHECK_NULL);
         if (!check) {
           ik->set_shared_loading_failed();
           return nullptr;
         }
-      } else if (Signature::has_envelope(sig)) {
-        // Pending exceptions are cleared so we can fail silently
+      } else {
+        // Pending exceptions are cleared so we can fail silently.
```
DanHeidinga: Do we need this branch to attempt to preload?  As long as we have the classes for the flattened fields, we should be OK.  I would expect the other classes to be optional at this point

src/hotspot/share/classfile/systemDictionary.cpp
caspernorrbin: I have now wrapped the creation(s) inside `if (is_log_enabled(...))`.

src/hotspot/share/classfile/systemDictionary.cpp
caspernorrbin: This branch handles the speculative preloading from the `LoadableDescriptors` attribute for non-flat nullable fields, so it is not part of the required flat/null-restricted validation. I kept it to preserve the existing behavior where `LoadableDescriptors` can opportunistically load field classes. These loads are optional, failures are ignored and should not cause the shared loading to fail.

DanHeidinga: I'm fine with leaving this for now but we should think further about this after the merge to mainline.  With CDS, these field classes for non-flat fields will be loaded during `ik->link_class()` so this may be additional complexity we can simplify

src/hotspot/share/classfile/systemDictionary.cpp
DanHeidinga: @fparain As per our discussion about nullresticted fields, this will need to be updated as well to not skip static fields

liach: Hmm, does this mean NR static fields may have non-reference layout?

coleenp (commented):
I guess I should un-review this since it's changed since my review.

src/hotspot/share/classfile/systemDictionary.cpp
DanHeidinga: No, this is related to ensuring that NR fields at dump time and NR fields at restore time are consistent

fparain: To @DanHeidinga , yes, null-restricted static fields should not be skipped, but this PR would need JDK-8384756 to be fixed before being able to implement this change.

caspernorrbin:
Following the changes from [JDK-8384756](https://bugs.openjdk.org/browse/JDK-8384756), I have updated this PR to also validate archived metadata for static null-restricted fields.

The new version:
- Includes classes with `has_null_restricted_static_fields()` in the `load_shared_class` field metadata validation loop, and only skips static fields that are not null-restricted.
- Updates `get_inline_type_field_klass_or_null()` to allow the static null-restricted metadata case.
- Simplifies the flat/null-restricted preload failure logging to use one message with a computed failure reason.
- Replaces the old stripped LoadableDescriptors test with `ArchivedFieldMetadataMismatchTest`, which now covers both nullable flat instance fields and static null-restricted fields. The name did not really match what was actually tested, so this new name hopefully matches that better.
- Moves the LoadableDescriptors stripping helper into `TestCommon` and also works without the attribute, so that test classes can be copied through the same path whether or not they actually have the attribute and can also be accessed from other future tests.

src/hotspot/share/classfile/systemDictionary.cpp
iklam: The comments needs to be clarified. It says "we can fail silently", but we just ignore any failures and keep on executing.

src/hotspot/share/classfile/systemDictionary.cpp
```diff
@@ -1196,22 +1198,25 @@ InstanceKlass* SystemDictionary::load_shared_class(InstanceKlass* ik,
     return nullptr;
   }
 
-  if (ik->has_inlined_fields()) {
+  if (ik->has_inlined_fields() || ik->has_null_restricted_static_fields()) {
     for (AllFieldStream fs(ik); !fs.done(); fs.next()) {
-      if (fs.access_flags().is_static()) continue;
+      if (fs.access_flags().is_static() && !fs.is_null_free_inline_type()) continue;
```
iklam: `if` body should be in braces to be consistent with the style used below:

```
if (...) {
  continue;
}
```

src/hotspot/share/classfile/systemDictionary.cpp
```diff
@@ -1106,43 +1106,46 @@ bool SystemDictionary::check_shared_class_super_types(InstanceKlass* ik, Handle
...
+    TempNewSymbol name = Signature::strip_envelope(sig);
+    log_info(class, preload)("Preloading of class %s during loading of shared class %s "
+                             "(cause: archived flat/null-restricted field metadata) succeeded",
+                             name->as_C_string(), ik->name()->as_C_string());
+  }
   return true;
 }
 
-// Tries to pre-load classes referred to in non-static nullable instance fields if they are found in the
+// Tries to pre-load classes referred to in non-static instance fields if they are found in the
```
iklam: Comment should be updated to indicate that this is optional. If a failure is encountered, it's silently ignored.
Could you add a reference to the same code that happens when a class is loaded outside of archive?

src/hotspot/share/classfile/systemDictionary.cpp
```diff
@@ -1106,43 +1106,46 @@ bool SystemDictionary::check_shared_class_super_types(InstanceKlass* ik, Handle
...
-                           name->as_C_string(), ik->name()->as_C_string());
-  InstanceKlass* real_k = SystemDictionary::resolve_with_circularity_detection(ik->name(), name,
-                                                                               class_loader, false, CHECK_false);
-  if (HAS_PENDING_EXCEPTION) {
-    log_info(class, preload)("Preloading of class %s during loading of class %s "
-                             "(cause: null-free non-static field) failed: %s",
-                             name->as_C_string(), ik->name()->as_C_string(),
-                             PENDING_EXCEPTION->klass()->name()->as_C_string());
-    return false; // Exception is still pending
+// Pre-load class referred to in non-static fields with archived inline field metadata. These fields
```
DanHeidinga:
```suggestion
// Pre-load class referred to in fields with archived inline field metadata. These fields
```

This is called for both inline instance fields and also for null-free static fields

src/hotspot/share/classfile/systemDictionary.cpp
DanHeidinga:
```suggestion
// Tries to pre-load classes referred to in instance fields if they are found in the
```

non-static instance is redundent

fparain (approved):
Looks good to me.
Thank you for your patience when we were trying to define the right behavior and fixing all the missing pieces.

src/hotspot/share/classfile/systemDictionary.cpp
DanHeidinga: Shouldn't this function be using the `check_shared_class_dependency` helper as well?  It's doing basically the same thing - validating that the current view of `real_k` matches the view of `k` from CDS

iklam (changes requested):

iklam:
This patch is needed to avoid failures in tier 4 testing
```
$ git diff
diff --git a/test/hotspot/jtreg/TEST.groups b/test/hotspot/jtreg/TEST.groups
index 0ef426ef4ce..8ce8204d138 100644
--- a/test/hotspot/jtreg/TEST.groups
+++ b/test/hotspot/jtreg/TEST.groups
@@ -538,6 +538,7 @@ hotspot_aot_classlinking = \
  -runtime/cds/appcds/aotCode \
  -runtime/cds/appcds/aotFlags \
  -runtime/cds/appcds/aotProfile \
+ -runtime/cds/appcds/ArchivedFieldMetadataMismatchTest.java \
  -runtime/cds/appcds/BadBSM.java \
  -runtime/cds/appcds/cacheObject/ArchivedFlatArrayTest.java \
  -runtime/cds/appcds/cacheObject/ArchiveHeapTestClass.java \
```

test/hotspot/jtreg/runtime/cds/appcds/ArchivedFieldMetadataMismatchTest.java
```diff
@@ -0,0 +1,188 @@
...
+    }
+
+    private static String buildJarWithStrippedLoadableDescriptors(String jarName, String... classNames) throws Exception {
+        Path testClasses = Path.of(System.getProperty("test.classes", "."));
+        Path jarClasses = testClasses.resolve(jarName);
+        Files.createDirectories(jarClasses);
+
+        for (String className : classNames) {
+            TestCommon.stripLoadableDescriptors(testClasses, jarClasses, className);
+        }
```
iklam: Can you explain why this is necessary?

If I disable this, the test case fails.

iklam: This code doesn't seem necessary. I have simplified this test case and removed this part. Please see my version at:

https://github.com/iklam/valhalla/commit/a836109cb8b7c4edc77136937eff8e7c7358a83b

caspernorrbin:
Thanks for the quick feedback! I have pushed an update that updates various comments to be more clear, and changed `try_preload_from_loadable_descriptors` to also use `check_shared_class_dependency`. I also cherry-picked in @iklam's simplified test version, which is much cleaner while covering the same behavior.

DanHeidinga (approved):
Looks good.  Thanks Casper!

iklam (approved):

caspernorrbin:
Thank you everyone for your help on this! Let's get it in.

/integrate

PR #2410, opened 2026-05-07
8384081: [lworld] JVM_IHashCode caches hash after pending exception

jsikstro:
Hello,

This PR fixes an issue in `JVM_IHashCode` where we don't handle an exception occurring in the Java call to `ValueObjectMethods.valueObjectHashCode`. If an exception is thrown, we still proceed to install the (bad) hashcode that is based on a return value which is garbage data.

To fix this, we should make sure to return if there is a pending exception.

@xmas92 came up with a smart approach to encapsulate the behavior of wrapping an exception in an internal error into a separate helper, which is implemented in this PR as well. I've adapted other similar places where we wrap exceptions like this as well. These use-cases are all Valhalla-specific, so makes sense to change here IMO.

Testing:
* Running Oracle's tier1-4
* Test reproducer that fails before (installing a bad hashcode), and succeeds with the changes in this PR (never installs a bad hashcode).

---------

---------
### Progress

### Issue
 * [JDK-8384081](https://bugs.openjdk.org/browse/JDK-8384081): [lworld] JVM_IHashCode caches hash after pending exception (**Bug** - P4)

### Reviewers

### Contributors
 * Axel Boldt-Christmas `<aboldtch@openjdk.org>`
 * Paul Hübner `<phubner@openjdk.org>`

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2410/head:pull/2410` \
`$ git checkout pull/2410`

`$ git checkout pull/2410` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2410/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2410`

`$ git pr show -t 2410`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2410.diff">https://git.openjdk.org/valhalla/pull/2410.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2410#issuecomment-4398112046)
</details>

### DIFF
--- a/src/hotspot/share/utilities/exceptions.cpp
+++ b/src/hotspot/share/utilities/exceptions.cpp
@@ -476,8 +476,8 @@ Handle Exceptions::new_exception(JavaThread* thread, Symbol* name,
 }
 
 void Exceptions::wrap_exception_in_internal_error(const char* message, JavaThread* THREAD) {
-  // If there is a pending exception and that is not an internal error,
-  // clear it and wrap it in an internal error instead.
+  // If there is a pending exception that is not an Error, clear it and wrap it
+  // in an InternalError instead.
   if (THREAD->has_pending_exception()) {
     oop exception = THREAD->pending_exception();
     if (!exception->is_a(vmClasses::Error_klass())) {
--- a/src/hotspot/share/utilities/exceptions.cpp
+++ b/src/hotspot/share/utilities/exceptions.cpp
@@ -475,6 +475,19 @@ Handle Exceptions::new_exception(JavaThread* thread, Symbol* name,
                                    to_utf8_safe);
 }
 
+void Exceptions::wrap_exception_in_internal_error(const char* message, JavaThread* THREAD) {
+  // If there is a pending exception and that is not an internal error,
+  // clear it and wrap it in an internal error instead.
+  if (THREAD->has_pending_exception()) {
+    oop exception = THREAD->pending_exception();
+    if (!exception->is_a(vmClasses::Error_klass())) {
+      Handle e(THREAD, exception);
+      THREAD->clear_pending_exception();
+      THROW_MSG_CAUSE(vmSymbols::java_lang_InternalError(), message, e);
+    }
+  }
+}
+
 // invokedynamic uses wrap_dynamic_exception for:
 //    - bootstrap method resolution
 //    - post call to MethodHandleNatives::linkCallSite
--- a/test/hotspot/jtreg/runtime/valhalla/inlinetypes/HashOverflowTest.java
+++ b/test/hotspot/jtreg/runtime/valhalla/inlinetypes/HashOverflowTest.java
@@ -1,0 +1,67 @@
+/*
+ * Copyright (c) 2026, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+/*
+ * @test
+ * @summary Tests that JVM_IHashCode does not cache when a SOE occurs.
+ * @comment This test runs with the interpreter such that the value is always
+ *          buffered, meaning that the identity hash code computed will be
+ *          saved in the markWord.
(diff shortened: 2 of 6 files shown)


xmas92 (approved):
lgtm.

src/hotspot/share/utilities/exceptions.cpp
```diff
@@ -475,6 +475,19 @@ Handle Exceptions::new_exception(JavaThread* thread, Symbol* name,
                                    to_utf8_safe);
 }
 
+void Exceptions::wrap_exception_in_internal_error(const char* message, JavaThread* THREAD) {
+  // If there is a pending exception and that is not an internal error,
```
dholmes-ora:
```suggestion
  // If there is a pending exception that is not an Error,
```

src/hotspot/share/utilities/exceptions.cpp
```diff
@@ -475,6 +475,19 @@ Handle Exceptions::new_exception(JavaThread* thread, Symbol* name,
                                    to_utf8_safe);
 }
 
+void Exceptions::wrap_exception_in_internal_error(const char* message, JavaThread* THREAD) {
```
dholmes-ora, line 478: Nit: we use `THREAD` when we need it for the macros e.g. `HAS_PENDING_EXCEPTION`. If the code doesn't use the macros then it could just be `JavaThread* current`. Though arguably we should be using TRAPS as this can throw. I realize `wrap_dynamic_exception` set the precedent for this formulation though, so not insisting on any changes.

src/hotspot/share/utilities/exceptions.cpp
```diff
@@ -475,6 +475,19 @@ Handle Exceptions::new_exception(JavaThread* thread, Symbol* name,
...
+  // If there is a pending exception and that is not an internal error,
+  // clear it and wrap it in an internal error instead.
```
dholmes-ora:
```suggestion
  // clear it and wrap it in an InternalError instead.
```

dholmes-ora (approved):
Good catch on the missing return! The helper looks good.

src/hotspot/share/utilities/exceptions.cpp
jsikstro: Yeah I agree with you. I tried to be consistent with `wrap_dynamic_exception` (except for the comments which are now fixed). I'll take a note of this and maybe revisit this in mainline post-integration.

xmas92 (approved):

Arraying (commented):
The implementation looks good, thank you! I feel like it would be very valuable to introduce a regression test for this. Is that something that could be feasible?

xmas92:
> The implementation looks good, thank you! I feel like it would be very valuable to introduce a regression test for this. Is that something that could be feasible?

We can add one but I don't think it will be that useful. All we can really do is to check that these known runtime upcall places do things correctly. 

I think there is a mainline RFE here which refactors the `JavaCalls::call` style helpers and wrapper to return a type of `ReturnJavaValue`, which captures if an exception is pending (directly) after the call returned.

So we will do:
1. Check that there is no pending exception before the upcall
2. Store the pending exception state in the `ReturnJavaValue` type
3. Assert if one tries to get the result out of a `ReturnJavaValue` which was created from a call with a pending exception.

This would make us catch this and any future similar mistakes. But would require a bit of code churn / refactoring.

Arraying:
> > The implementation looks good, thank you! I feel like it would be very valuable to introduce a regression test for this. Is that something that could be feasible?
> 
> We can add one but I don't think it will be that useful. All we can really do is to check that these known runtime upcall places do things correctly.
> 
> I think there is a mainline RFE here which refactors the `JavaCalls::call` style helpers and wrapper to return a type of `ReturnJavaValue`, which captures if an exception is pending (directly) after the call returned.
> 
> So we will do:
> 
> 1. Check that there is no pending exception before the upcall
> 2. Store the pending exception state in the `ReturnJavaValue` type
> 3. Assert if one tries to get the result out of a `ReturnJavaValue` which was created from a call with a pending exception.
> 
> This would make us catch this and any future similar mistakes. But would require a bit of code churn / refactoring.

I'm prototyping a basic Java program that calls `System::identityHash` on a very large structure twice, and ensures that the second call fails if the first one does. AFAIU, this is _not_ the case before this fix.

xmas92:
> I'm prototyping a basic Java program that calls `System::identityHash` on a very large structure twice, and ensures that the second call fails if the first one does. AFAIU, this is _not_ the case before this fix.

Thats is correct, we would have read a garbage value out of the `JavaValue result` and write it into the markWord. 
The second call would just use that.

jsikstro:
I fully agree with Axel that there's room for improvement in how we're using return values from calls to Java, and that that is a broader enhancement/change.

For this specific scenario I've added a regression test authored py Paul, which fails before the changes in this PR and passes with them.

test/hotspot/jtreg/runtime/valhalla/inlinetypes/HashOverflowTest.java
```diff
@@ -0,0 +1,67 @@
...
+
+/*
+ * @test
+ * @summary Tests that JVM_IHashCode does not cache when a SOE occurs.
+ * @comment This test runs with the interpreter such that the value is always
+ *          buffered, meaning that the identity hash code computed will be
+ *          saved in the markWord.
+ * @enablePreview
+ * @compile HashOverflowTest.java
+ * @run main/othervm -Xint -Xss256K
```
dholmes-ora, line 33: If this has to run with -Xint should we exclude from e.g. Xcomp runs, or maybe just mark this as flagless?

jsikstro: Good catch, let's mark it flagless.

xmas92 (approved):

dholmes-ora (approved):

jsikstro:
Thank you for the reviews and contributions everyone! I merged with the tip of lworld locally and reran tier1-2 which is green.
/integrate

PR #2412, opened 2026-05-07
8380075: [lworld] Update JVMTI spec to clarify that ObjectFree events are not sent for value objects

sspitsyn:
This update address two enhancements:
 - [8380075](https://bugs.openjdk.org/browse/JDK-8380075) [lworld] Update JVMTI spec to clarify that ObjectFree events are not sent for value objects
 - [8384049](https://bugs.openjdk.org/browse/JDK-8384049) [lworld] Minor GetLocalObject/GetLocalInstance spec wording update

The first one adds an `ObjectFree` event spec clarification that only identity objects are supported by this event.
The second one just slightly reformulates previously added spec clarifications for the JVMTI  `GetLocalObject` and `GetLocalInstance` functions, underlying that they work only when preview features are enabled.

The `ObjectFree` implementation does not need any matching update. Strong references are used for value objects instead of weak references in the `JvmtiTagMap`, so the value objects are kept alive while they are listed in the `JvmtiTagMap`. This means there are no `ObjectFree` events can be generated for value objects.

Testing:
 - No testing is needed for pure spec wording changes

---------

---------
### Progress

### Issues
 * [JDK-8380075](https://bugs.openjdk.org/browse/JDK-8380075): [lworld] Update JVMTI spec to clarify that ObjectFree events are not sent for value objects (**Enhancement** - P4)
 * [JDK-8384049](https://bugs.openjdk.org/browse/JDK-8384049): [lworld] Minor GetLocalObject/GetLocalInstance spec wording update (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2412/head:pull/2412` \
`$ git checkout pull/2412`

`$ git checkout pull/2412` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2412/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2412`

`$ git pr show -t 2412`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2412.diff">https://git.openjdk.org/valhalla/pull/2412.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2412#issuecomment-4402005146)
</details>

### DIFF
--- a/src/hotspot/share/prims/jvmti.xml
+++ b/src/hotspot/share/prims/jvmti.xml
@@ -5888,8 +5888,8 @@ class C2 extends C1 implements I2 {
           <outptr><jobject/></outptr>
             <description>
               On return, points to the variable's value.
-              If the frame's method is a value object constructor and
-              the requested local is the "<code>this</code>" object, the
+              When preview features are enabled, the frame's method is a non-identity object
+              constructor and the requested local is the "<code>this</code>" object, the
               value_ptr is set to a snapshot of the "<code>this</code>" object.
             </description>
         </param>
@@ -5944,8 +5944,8 @@ class C2 extends C1 implements I2 {
           <outptr><jobject/></outptr>
             <description>
               On return, points to the variable's value.
-              If the frame's method is a value object constructor, the 
-              value_ptr is set to a snapshot of the "<code>this</code>" object.
+              When preview features are enabled and the frame's method is a non-identity object
+              constructor, the value_ptr is set to a snapshot of the "<code>this</code>" object.
             </description>
         </param>
       </parameters>
@@ -14137,6 +14137,7 @@ myInit() {
       must not use <jvmti/> functions except those which
       specifically allow such use (see the raw monitor, memory management,
       and environment local storage functions).
+      When preview features are enabled, this event supports identity object allocations only.
     </description>
     <origin>new</origin>
     <capabilities>
--- a/src/hotspot/share/prims/jvmti.xml
+++ b/src/hotspot/share/prims/jvmti.xml
@@ -5948,7 +5948,7 @@ class C2 extends C1 implements I2 {
               On return, points to the variable's value.
               When preview features are enabled, and the local instance is a value
               object under construction, the value_ptr is set to a snapshot
-              of the"<code>this</code>" object that represents the value
+              of the "<code>this</code>" object that represents the value
               object's state at the point the snapshot is taken.
             </description>
         </param>
--- a/src/hotspot/share/prims/jvmti.xml
+++ b/src/hotspot/share/prims/jvmti.xml
@@ -5888,9 +5888,11 @@ class C2 extends C1 implements I2 {
           <outptr><jobject/></outptr>
             <description>
               On return, points to the variable's value.
-              When preview features are enabled, the frame's method is a value class
-              constructor and the requested local is a "<code>this</code>" value object,
-              the value_ptr is set to a snapshot of the "<code>this</code>" object.
+              When preview features are enabled, if the requested local is the
+              "<code>this</code>" object, and "<code>this</code>" is a value
+              object under construction, the value_ptr is set to a snapshot
+              of the"<code>this</code>" object that represents the value
+              object's state at the point the snapshot is taken.
             </description>
         </param>
       </parameters>
@@ -5944,9 +5946,10 @@ class C2 extends C1 implements I2 {
           <outptr><jobject/></outptr>
             <description>
               On return, points to the variable's value.
-              When preview features are enabled, the frame's method is a value class
-              constructor and the local instance is a value object,
-              the value_ptr is set to a snapshot of the "<code>this</code>" object.
+              When preview features are enabled, and the local instance is a value
+              object under construction, the value_ptr is set to a snapshot
+              of the"<code>this</code>" object that represents the value
+              object's state at the point the snapshot is taken.
             </description>
         </param>
       </parameters>
--- a/src/hotspot/share/prims/jvmti.xml
+++ b/src/hotspot/share/prims/jvmti.xml
@@ -5888,9 +5888,9 @@ class C2 extends C1 implements I2 {
           <outptr><jobject/></outptr>
             <description>
               On return, points to the variable's value.
-              When preview features are enabled, the frame's method is an object
-              constructor and the requested local is a "<code>this</code>" value object, the
-              value_ptr is set to a snapshot of the "<code>this</code>" object.
+              When preview features are enabled, the frame's method is a value class
+              constructor and the requested local is a "<code>this</code>" value object,
+              the value_ptr is set to a snapshot of the object.
             </description>
         </param>
       </parameters>
@@ -5944,9 +5944,9 @@ class C2 extends C1 implements I2 {
           <outptr><jobject/></outptr>
             <description>
               On return, points to the variable's value.
-              When preview features are enabled, the frame's method is an object
-              constructor and the requested instance is a value object, the
-              value_ptr is set to a snapshot of the "<code>this</code>" object.
+              When preview features are enabled, the frame's method is a value class
+              constructor and the local instance is a value object,
+              the value_ptr is set to a snapshot of the "<code>this</code>" object.
             </description>
         </param>
       </parameters>
--- a/src/hotspot/share/prims/jvmti.xml
+++ b/src/hotspot/share/prims/jvmti.xml
@@ -5890,7 +5890,7 @@ class C2 extends C1 implements I2 {
               On return, points to the variable's value.
               When preview features are enabled, the frame's method is a value class
               constructor and the requested local is a "<code>this</code>" value object,
-              the value_ptr is set to a snapshot of the object.
+              the value_ptr is set to a snapshot of the "<code>this</code>" object.
             </description>
         </param>
       </parameters>
--- a/src/hotspot/share/prims/jvmti.xml
+++ b/src/hotspot/share/prims/jvmti.xml
@@ -5891,7 +5891,7 @@ class C2 extends C1 implements I2 {
               When preview features are enabled, if the requested local is the
               "<code>this</code>" object, and "<code>this</code>" is a value
               object under construction, the value_ptr is set to a snapshot
-              of the"<code>this</code>" object that represents the value
+              of the "<code>this</code>" object that represents the value
               object's state at the point the snapshot is taken.
             </description>
         </param>
--- a/src/hotspot/share/prims/jvmti.xml
+++ b/src/hotspot/share/prims/jvmti.xml
@@ -5888,8 +5888,8 @@ class C2 extends C1 implements I2 {
           <outptr><jobject/></outptr>
             <description>
               On return, points to the variable's value.
-              When preview features are enabled, the frame's method is a value object
-              constructor and the requested local is the "<code>this</code>" object, the
+              When preview features are enabled, the frame's method is an object
+              constructor and the requested local is a "<code>this</code>" value object, the
               value_ptr is set to a snapshot of the "<code>this</code>" object.
             </description>
         </param>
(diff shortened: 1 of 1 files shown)


src/hotspot/share/prims/jvmti.xml
```diff
@@ -5888,8 +5888,8 @@ class C2 extends C1 implements I2 {
           <outptr><jobject/></outptr>
             <description>
               On return, points to the variable's value.
-              If the frame's method is a value object constructor and
-              the requested local is the "<code>this</code>" object, the
+              When preview features are enabled, the frame's method is a non-identity object
+              constructor and the requested local is the "<code>this</code>" object, the
```
plummercj: I think this should stay as "value object", but I'll defer to @dansmithcode and @dholmes-ora

src/hotspot/share/prims/jvmti.xml
```diff
@@ -14144,6 +14137,7 @@ myInit() {
       must not use <jvmti/> functions except those which
       specifically allow such use (see the raw monitor, memory management,
       and environment local storage functions).
+      When preview features are enabled, this event supports identity object allocations only.
```
plummercj: I think maybe the CR and PR could have better titles since there is no code change here. Something along the lines of "Update JVMTI spec to clarify that ObjectFree events are not sent for value objects"

sspitsyn: Thanks, changed the JBS issue title.

src/hotspot/share/prims/jvmti.xml
sspitsyn: Thanks. I was thinking about the same. Changed to "value object" now for all 3 fragments.

sspitsyn: @dansmithcode and @dholmes-ora Please, let me know if you are okay with the latest update.

dansmithcode: Is the behavior special when the frame's method is declared by a _value class_, or when the requested local is a _value object_? (E.g., what happens when I'm looking at the (value class) `Number` constructor and `this` is a (value object) `BigInteger`?)

sspitsyn: @dansmithcode
I'm not sure I fully understand the question. The requested local is a value `this` object. The API behavior is special in a case when two conditions are `true`: (1) the frame belongs to a value object constructor (so the cltor's class is a value class) and (2) the requested local is the `THIS` object from the slot 0. It is per our previous discussion on the `project-jep401` slack channel: We do not want mutable JNI reference to a value object to escape while it is under construction, so we return a JNI reference to a snapshot of the 'THIS' object.

dansmithcode: An abstract value class can have identity and value subclasses.

```
abstract value class Number { ... }
value class Integer extends Number { ... }
class BigInteger extends Number { ... }
```

I'm asking what happens when you try to get `this` from a constructor of `Number`: do you check whether `this` is a value object (an Integer, not a BigInteger), or do you apply the special-case logic to _all_ values of `this`, whether they are Integers or BigIntegers?

dansmithcode: (I'm guessing a "snapshot" is an operation that only makes sense for value objects, so you would not do it for the BigInteger. But I'm not sure.)

sspitsyn: Got it, thank you for checking!

> do you check whether this is a value object (an Integer, not a BigInteger), or do you apply the special-case logic to all values of this, whether they are Integers or BigIntegers?

We check whether this is a value object and do not care about identity objects like BigIntegers.

> I'm guessing a "snapshot" is an operation that only makes sense for value objects, so you would not do it for the BigInteger.

Yes, exactly.

dansmithcode: Okay, then here's how I'd phrase it:

> When preview features are enabled, if the requested local is the "<code>this</code>" variable
> of a constructor, and it references a value object, the value_ptr is set to a snapshot of the
> value object.

dansmithcode: On a behavioral level, I wonder how you are deciding which locals are `this` variables; like what happens if somebody copies `this` into local slot 3?

It may make more sense to say and implement:

> When preview features are enabled, if the frame's method is a value class constructor
> and the requested local is a value object, the value_ptr is set to a snapshot of the
> value object.

sspitsyn: > On a behavioral level, I wonder how you are deciding which locals are this variables; like what happens if somebody copies this into local slot 3?

Good question. There is no check for this corner case now.

> It may make more sense to say and implement:

Good suggestion with one correction. We do not want to snapshot any local value object but only `this` object which is under construction.

What about this? :
```
 When preview features are enabled, the frame's method is a value class
 constructor and the requested local is a "<code>this</code>" value object, the
 value_ptr is set to a snapshot of the "<code>this</code>" object.
```

plummercj: I'll let Serguei clarify, but here's what I found. The logic to decide if a snapshot (clone) is needed can be found in this diff (from the PR that implemented this support):

https://github.com/openjdk/valhalla/pull/2363/changes#diff-41730062a596d5a0b6b2330022428347b22d71d1a654af76e9959be7a75f65fcR642

My read on this is that if you are in a constructor, accessing local 0 (this), and the current frame's method is for a value class, then the snapshot will be done. That would mean when executing in the Number constructor, this snapshot behavior would apply even if a BigInteger is being constructed. This seems to be consistent with what is done at the language level (not allowing "this" to escape from the Number constructor, even when constructing a BigInteger).

Now my question is, what about when executing in the BigInteger constructor? Even though it is not a value type, it appears java does not allow escape of "this", presumably because it extends a value type. If we are going to do the same in JVMTI, we would need to do some extra checks of the inheritance, but I don't think there is actually any harm in JVMTI exposing "this" of BigInteger before the Number constructor is executed since it is actually an identity object.

sspitsyn: > this snapshot behavior would apply even if a BigInteger is being constructed.

My understanding that it should not. There is a check for `obj->is_inline()`.

> but I don't think there is actually any harm in JVMTI exposing "this" of BigInteger before the Number constructor is executed since it is actually an identity object.

Agreed.

dansmithcode: I see three tests in that code:

- `_jvf->method()->is_object_constructor()`: we're looking at an `<init>` method (I _think_ that's all this means?)
- `_index == 0`: trying to limit the behavior to `this`, but local 0 might not be `this`, and a different local might be `this`
- `obj != nullptr && obj->is_inline()`: the value of the local is a value object (again, guessing at what this code does)

If I'm reading right, there's not test for a _value class_, but there is a test that the object is a _value object_. So that's the description I offered above:

> When preview features are enabled, if the requested local is the "this" variable
> of a constructor, and it references a value object, the value_ptr is set to a snapshot of the
> value object.

I'm concerned about "the this variable of a constructor", though, because that's not really what the code is doing. To more accurately specify it, it should be "if the 0th local variable of a constructor is requested".

At that point we've got a spec and implementation in alignment. Great! But the the goal of the behavior has not been achieved, which is to defend against observable mutation of a larval value object (i.e., an `uninitializedThis` local variable that is a value object). I don't think you have any reliable way to tell which local variables might be `this`, and so to be safe we've got to cast a wider net and clone _all_ value objects accessed through a value class constructor.

sspitsyn: > so to be safe we've got to cast a wider net and clone all value objects accessed through a value class constructor.

I like this suggestion.

Then we need to reformulate the spec clarification to something like this:
> When preview features are enabled, the frame's method is a value class
 constructor and the local is a value object, the value_ptr is set to a snapshot of the
value object.

The implementation can be corrected by replacing the check `_index == 0` with the check `method->method_holder()->is_inline_class()`

I'm thinking if we also want a match between method's class and the object's class.
The spec would be something like this:
> When preview features are enabled, the frame's method is a value class constructor
  and the local is an object of that class, the value_ptr is set to a snapshot of the value object.

plummercj: > My read on this is that if you are in a constructor, accessing local 0 (this), and the current frame's method is for a value class, then the snapshot will be done. 

Sorry, I misread the code. The "value" check is on the object, not the class.

plummercj: > I'm concerned about "the this variable of a constructor", though, because that's not really what the code is doing. To more accurately specify it, it should be "if the 0th local variable of a constructor is requested".

I think JVMTI (and by extension JDWP and JDI) assume that "this" is always in slot 0. Does valhalla allow it to be stored elsewhere before initialization is complete?

dansmithcode: javac will always leave `this` in slot 0, and I believe the language rules don't allow users to store it in a different local variable during early construction.

The verifier has no such requirement. Generated bytecode can easily do `aload_0; astore_1;`.

sspitsyn: > javac will always leave this in slot 0, and I believe the language rules don't allow users to store it in a different local variable during early construction.

Does it mean the check for slot 0 is right and we have to keep it? There are no problems with `this` object stored in different slots during late object construction because the value object has been fully initialized.

dansmithcode: You understand your requirements better than I do, but I would think the invariants you're trying to preserve need to work on all class files, javac-generated or not?

sspitsyn: The invariant about slot of `this` object was set a long time ago when JVMTI was introduced. Now is not a good time to change it. We normally keep the spec and implementation compatible if possible.

dansmithcode: I _think_ the new invariant here is that a value object must never be observed to mutate (because if that happened, it would be in an implementation-dependent way, and may not be reproducible).

If that needs to be strongly guaranteed, then you need to guard against unexpected but valid bytecode shapes.

If it's not really an invariant, and we're just doing this for user convenience, then it makes sense to design for conventional use and not worry about non-standard bytecode patterns.

(One way to think about this: how bad would it be if this cloning thing didn't happen at all? What would break?)

sspitsyn: > If it's not really an invariant, and we're just doing this for user convenience, then it makes sense to design for conventional use and not worry about non-standard bytecode patterns.

I prefer this approach. It is already pretty narrow corner case. If a larval value object escapes because non-standard bytecode patterns, the user may experience some inconsistency caused by debugging of a value object construction. It does not look that we have to add more complexity to support this kind of non-standard bytecode patterns.

sspitsyn: I've pushed the spec update I'm suggesting now. Please, let me know if you are okay with it, or not.
Dan, thank you a lot for this discussion, questions and considerations!

src/hotspot/share/prims/jvmti.xml
```diff
@@ -5888,9 +5888,9 @@ class C2 extends C1 implements I2 {
           <outptr><jobject/></outptr>
             <description>
               On return, points to the variable's value.
-              If the frame's method is a value object constructor and
-              the requested local is the "<code>this</code>" object, the
-              value_ptr is set to a snapshot of the "<code>this</code>" object.
+              When preview features are enabled, the frame's method is a value class
+              constructor and the requested local is a "<code>this</code>" value object,
+              the value_ptr is set to a snapshot of the "<code>this</code>" object.
```
plummercj: I'm not so sure about the change to use "a 'this' value object" instead of "the 'this' value object". Also, separating out the "value" part might help here:

```
When preview features are enabled, if the frame's method is a value class
constructor, and the requested local is the “this” object, and “this” is a value object,
then value_ptr is set to a snapshot of the "<code>this</code>" object.
```

and you might want to add the following clarifying statement at the end. I think Dean suggested something along these lines:

`… that represents the value object’s state at the point the snapshot is taken.`

plummercj: Also, I'm not sure why we need to separately state "the frame's method is a value class
constructor" and "the requested local is ... a value object" instead of just saying "the requested local is a value object under constructor":

```
When preview features are enabled, if the requested local is the “this” object,
and “this” is a value object under construction,
then value_ptr is set to a snapshot of the "<code>this</code>" object.
```

sspitsyn: I like the suggestions, thanks. Pushed the update.

plummercj (approved):
Looks good.

sspitsyn:
Chris, thank you for review.

PR #2420, opened 2026-05-11
8383809: [type-classes] Prototype UInt128

rgiulietti:
Prototype a fully functional `unsigned int128` type as a Valhalla value class. 

---------

---------
### Progress

### Issue
 * [JDK-8383809](https://bugs.openjdk.org/browse/JDK-8383809): [type-classes] Prototype UInt128 (**Enhancement** - P4)

### Reviewers
 * [Joe Darcy](https://openjdk.org/census#darcy) (@jddarcy - Committer)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2420/head:pull/2420` \
`$ git checkout pull/2420`

`$ git checkout pull/2420` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2420/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2420`

`$ git pr show -t 2420`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2420.diff">https://git.openjdk.org/valhalla/pull/2420.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2420#issuecomment-4422314139)
</details>

### DIFF
--- a/src/java.base/share/classes/java/math/UInt128.java
+++ b/src/java.base/share/classes/java/math/UInt128.java
@@ -1,0 +1,1117 @@
+/*
+ * Copyright (c) 2026, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+package java.math;
+
+import jdk.internal.vm.annotation.Stable;
+
+import static java.lang.Character.*;
+import static java.lang.Long.divideUnsigned;
+import static java.lang.Long.toUnsignedString;
+import static java.math.UInt64.*;
+
+/**
+ * Immutable unsigned 128-bit integers ({@link #SIZE} = 128).
+ * <p>
+ * All operations are performed modulo 2<sup>{@link #SIZE}</sup>,
+ * unless specified otherwise.
+ * <p>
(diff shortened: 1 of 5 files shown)


rgiulietti:
Subject to change and discussion:
* Use of factory methods rather than constructors.
* Adding/changing/removing functionality (e.g., adding `rotateLeft()`).
* Adding more general `*Exact()` methods that throw on overflow.
* Use of typeclass mechanisms.

src/java.base/share/classes/java/math/UInt128.java
```diff
@@ -0,0 +1,1117 @@
...
+     * {@return the high bits}
+     */
+    public long c1() {
+        return c1;
+    }
+
+    /**
+     * Returns the instance representing {@code v}.
+     *
+     * @param v must meet 0 ≤ {@code v} &lt; 2<sup>{@link #SIZE}</sup>.
```
jddarcy, line 178: As a small issue, I don't know if having "<=" as a single character, as it seems to be here, goes against general encoding rules for JDK sources.

src/java.base/share/classes/java/math/UInt64.java
```diff
@@ -0,0 +1,252 @@
...
+    }
+
+    /**
+     * Returns whether {@code x} &lt; {@code y}.
+     *
+     * @param x the 1st parameter
+     * @param y the 2nd parameter
+     * @return the result described above.
+     */
+    static boolean lessThan(long x, long y) {
```
jddarcy: Is it intentional that this method is _not_ public?

Having a `@see` link to `Long.compareUnsigned`, if the obvious relationship between the methods holds, would be helpful.

src/java.base/share/classes/java/math/UInt128.java
rgiulietti: JDK source code has [switched to UTF-8](https://bugs.openjdk.org/browse/JDK-8301971) some time ago.

src/java.base/share/classes/java/math/UInt64.java
rgiulietti: The idea is that the `public` API point is `compareTo()`.
But it's no deal to make the `static` comparison methods `public` as well.

rgiulietti: Ah, I know see what you mean.
Addressed in new commit.

jddarcy (approved):

rgiulietti:
There's a `jshell` related failure in `CompletionSuggestionTest::testGeneric`, but seems unrelated to the changes in this PR.

src/java.base/share/classes/java/math/UInt128.java
```diff
@@ -0,0 +1,1117 @@
...
+        }
+        c = s.charAt(i++);
+        int lc = c | 0b10_0000;  // to lower case
+        if (lc == 'x') {
+            return parse(s, i, end, i, 16, '_', -1);
+        }
+        if (lc == 'b') {
+            return parse(s, i, end, i, 2, '_', -1);
+        }
+        return parse(s, begin, end, i, 8, '_', c);
```
nlisker, line 313: This can be a pattern matching switch:
```java
		return switch (lc) {
			case 'x' -> parse(s, i, end, i, 16, '_', -1);
			case 'b' -> parse(s, i, end, i, 2, '_', -1);
			default -> parse(s, begin, end, i, 8, '_', c;
		};
```

rgiulietti: Indeed.
Maybe in a followup PR.
Thanks

PR #2421, opened 2026-05-11
8383658: [lworld] Add value class test coverage to java.util.Collections

bwhuang-us:
**Tests Updated**

**TEST.groups**
Add java/util/Collections to valhalla_adopted.

**test/lib**
Introduces a shared jdk.test.lib.valueclass.Tuple value class (two-int, Comparable, with equals/hashCode).

**AddAll.java**
Broaden the existing @AsValueClass Point path to cover the same collection types (LinkedList, HashSet, LinkedHashSet) and verify final contents using value equality.

**AsLifoQueue.java** 
Add a parallel LIFO-order block with value-class elements exercising the same add/peek/remove/poll operations.

**BigBinarySearch.java**
Add a minimal comparable value class backed by the existing sparse-list structure to verify huge binary search without new edge cases.

**BinarySearchNullComparator.java**
Add a comparable value-class list and confirm that a null comparator falls back to natural ordering correctly.

**CheckedListBash.java**
Add a checkedList(…, V.class) helper covering insert, contains/indexOf/toArray, and wrong-type raw-add rejection.

**CheckedListReplaceAll.java**
Verify that replaceAll on a checked value list accepts correct-type results and throws ClassCastException for wrong-type results.

**CheckedMapBash.java**
Add a value-key/value checked-map smoke path covering put/containsKey/containsValue/equals and raw wrong-type put rejection.

**CheckedMapReplaceAll.java**
Verify that Map.replaceAll on a checked value map accepts correct-type values and rejects wrong-type values.

**CheckedQueue.java**
Add a checkedQueue(…, V.class) block covering add/peek and wrong-type raw-add rejection.

**CheckedSetBash.java** 
Add a checked value-class set helper covering add/contains/remove, equality with a copy, and wrong-type raw-add rejection.

**Disjoint.java** 
Add overlapping and non-overlapping value-class collections to confirm that disjoint uses value-state equality.

**Enum.java** 
Add a value-class Vector<V> and assert that Collections.list produces an equal list in original order.

**EnumerationAsIterator.java** 
Add one value-class data-provider case using Collections.enumeration(Arrays.asList(…)) and verify element delivery.

**FindSubList.java** 
Add value-class source and target lists and assert correct first and last sublist indices.

**Frequency.java** 
Add a value-class list with repeated separately constructed equal values and confirm that frequency counts by value-state equality.

**NCopies.java** 
Add value-class versions of the indexOf/lastIndexOf/equals/hashCode/reversed checks using a repeated value element.

**NullComparator.java** 
Add value-class comparable list checks for Collections.sort, binarySearch, min, and max with a null comparator.

**ReplaceAll.java** 
Add one value-class list case mirroring the integer path to confirm equality-based replacement of value objects.

**ReverseOrder.java** 
Duplicate the custom comparable payload as a value class and run the same reverse-order sort assertion.

**ReverseOrder2.java** 
Add a comparable value-class list to the existing sort/reverse-sort checks only.

**Rotate.java** 
Add a value-class list and assert the expected rotated order for the existing rotation distances.

**Ser.java**
Add value-object payload checks for the singleton and nCopies serialization round-trips, leaving empty-collection cases unchanged.

**Shuffle.java** 
Add a value-class list and verify that size and per-element frequency are preserved after shuffle.

**SingletonIterator.java** 
Add a Collections.singleton(new V(…)).iterator() forEachRemaining check delivering the value object exactly once.

**Swap.java** 
Add a value-class marker element and perform the same sequential adjacent swaps, asserting final position by value equality.

**T6433170.java**
Add a value-class checked collection with a mixed-type addAll input and assert all-or-nothing behavior on type rejection.

**Tests With No Change**

**CheckedIdentityMap.java** 
The test is specifically about identity-based entry equality in IdentityHashMap, which is the opposite of value-class semantics and would require inverted expectations.

**CheckedNull.java** 
Null acceptance/rejection behavior is unchanged by JEP 401, so value-class elements add no distinct coverage.

**DelegatingIteratorForEachRemaining.java** 
The test is about iterator delegation and NPE propagation by wrapper classes, where element type is entirely incidental.

**EmptyCollectionSerialization.java** 
Empty singleton collections have no element payload, and the test is identity-sensitive with respect to the collection object itself, not its contents.

**EmptyIterator.java** 
There are no elements, and the existing == assertions target singleton iterator instances rather than user element identity.

**EmptyNavigableMap.java** 
The map is always empty, so value-class key/value semantics cannot be observed.

**EmptyNavigableSet.java** 
Same as EmptyNavigableMap: no elements are present to exercise value-class equality or ordering.

**EqualsTest.java** 
The test is about infinite recursion in equals for cyclic self-containing wrapper graphs, not element value semantics.

**MinMax.java** 
The bug under test is a malicious collection misreporting its size during iteration, making element kind entirely irrelevant.

**RacingCollections.java** 
The coverage is concurrent structural mutation and wrapper robustness, and extending it with value elements would require inventing new stress logic rather than conservatively extending existing behavior.

**RotateEmpty.java** 
The list is always empty, so no value element can participate, and adding one would convert it into non-empty rotate coverage.

**RotateHuge.java** 
The bug concerns distance/size arithmetic overflow using placeholder objects, not element equality or identity.

**SetFromMap.java** 
The key scenario depends on IdentityHashMap distinguishing equal-but-non-identical objects, which is incompatible with value-class identity-free semantics.

**SyncListBash.java** 
The lock is the synchronized wrapper object itself, not the list elements, so value-class payloads cannot affect synchronization correctness.

**SyncSubMutexes.java** 
The test is explicitly about monitor/mutex identity across synchronized views, and value objects cannot serve as synchronization monitors under JEP 401.

**T5078378.java** 
The test is a compile-time raw/generic type-inference regression, and adding value classes would require preview source changes unrelated to the original compilation check.

**UnmodifiableMapEntrySet.java**
Mutation is attempted on Map.Entry wrapper objects, not on the key/value payload, so element type is irrelevant to the coverage.

**ViewSynch.java**
Like SyncSubMutexes, the behavior under test is monitor identity and locking across sorted-map views, not element semantics.

**WrappedNull.java**
The scenario tests null wrapper arguments with no valid element instances, so there is nothing to substitute with a value class.

**WrappedUnmodifiableCollections.java**
The test checks whether Collections.unmodifiable* elides redundant wrapping based on the wrapper object's identity, which is unaffected by element type.

**Wrappers.java**
This is a reflection test inspecting whether wrapper classes override default methods, making element type completely irrelevant.

---------

---------
### Progress

### Issue
 * [JDK-8383658](https://bugs.openjdk.org/browse/JDK-8383658): [lworld] Add value class test coverage to java.util.Collections (**Task** - P4)

### Reviewers
 * [Chen Liang](https://openjdk.org/census#liach) (@liach - Committer) ⚠️ Review applies to [2bd89460](https://git.openjdk.org/valhalla/pull/2421/files/2bd8946080843d67287619e2f5f30f05b6a3edeb)
 * [Leonid Mesnik](https://openjdk.org/census#lmesnik) (@lmesnik - Committer)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2421/head:pull/2421` \
`$ git checkout pull/2421`

`$ git checkout pull/2421` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2421/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2421`

`$ git pr show -t 2421`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2421.diff">https://git.openjdk.org/valhalla/pull/2421.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2421#issuecomment-4425064840)
</details>

### DIFF
--- a/test/jdk/java/util/Collections/CheckedListBash.java
+++ b/test/jdk/java/util/Collections/CheckedListBash.java
@@ -28,10 +28,9 @@
  * @author  Josh Bloch
  * @key randomness
  * @library /test/lib
- * @run main CheckedListBash
  */
 
-import jdk.test.lib.valueclass.Tuple;
+import jdk.test.lib.valueclass.VClass;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -231,12 +230,12 @@ public class CheckedListBash {
     }
 
     static void testValueCheckedList() {
-        List<Tuple> list = Collections.checkedList(new ArrayList<>(), Tuple.class);
-        list.add(new Tuple(1, 1));
-        list.add(new Tuple(2, 2));
-        if (!list.contains(new Tuple(1, 1)) || list.indexOf(new Tuple(2, 2)) != 1)
+        List<VClass> list = Collections.checkedList(new ArrayList<>(), VClass.class);
+        list.add(new VClass(1, new int[] { 1 }));
+        list.add(new VClass(2, new int[] { 2 }));
+        if (!list.contains(new VClass(1, new int[] { 1 })) || list.indexOf(new VClass(2, new int[] { 2 })) != 1)
             fail("value checkedList lookup failed");
-        Tuple[] a = list.toArray(new Tuple[0]);
+        VClass[] a = list.toArray(new VClass[0]);
         if (!Arrays.asList(a).equals(list))
             fail("value checkedList toArray failed");
         try {
--- a/test/jdk/java/util/Collections/CheckedListBash.java
+++ b/test/jdk/java/util/Collections/CheckedListBash.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2003, 2024, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2003, 2026, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -27,8 +27,11 @@
  * @summary Unit test for Collections.checkedList
  * @author  Josh Bloch
  * @key randomness
+ * @library /test/lib
+ * @run main CheckedListBash
  */
 
+import jdk.test.lib.valueclass.Tuple;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -182,6 +185,7 @@ public class CheckedListBash {
             || !l.equals(Arrays.asList(ia).subList(0, listSize)))
             fail("toArray(Object[]) is hosed (3)");
 
+        testValueCheckedList();
     }
 
     // Done inefficiently so as to exercise toArray
@@ -224,5 +228,20 @@ public class CheckedListBash {
 
     static void fail(String s) {
         throw new RuntimeException(s);
+    }
+
+    static void testValueCheckedList() {
+        List<Tuple> list = Collections.checkedList(new ArrayList<>(), Tuple.class);
+        list.add(new Tuple(1, 1));
+        list.add(new Tuple(2, 2));
+        if (!list.contains(new Tuple(1, 1)) || list.indexOf(new Tuple(2, 2)) != 1)
+            fail("value checkedList lookup failed");
+        Tuple[] a = list.toArray(new Tuple[0]);
+        if (!Arrays.asList(a).equals(list))
+            fail("value checkedList toArray failed");
+        try {
+            ((List) list).add("not a Tuple");
+            fail("value checkedList accepted wrong type");
+        } catch (ClassCastException expected) { }
     }
 }
--- a/test/jdk/java/util/Collections/Ser.java
+++ b/test/jdk/java/util/Collections/Ser.java
@@ -27,7 +27,6 @@
  * @summary EMPTY_SET, EMPTY_LIST, and the collections returned by
  *          nCopies and singleton were spec'd to be serializable, but weren't.
  * @library /test/lib
- * @run main Ser
  */
 
 import jdk.test.lib.valueclass.AsValueClass;
--- a/test/jdk/java/util/Collections/Ser.java
+++ b/test/jdk/java/util/Collections/Ser.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 1999, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1999, 2026, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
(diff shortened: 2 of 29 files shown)


test/lib/jdk/test/lib/valueclass/Tuple.java
```diff
@@ -21,22 +21,24 @@
...
-
-    ValueArgumentsProvider(ArgumentValue[] argumentValues) {
-        this.argumentValues = argumentValues;
-    }
-
-    @Override
-    public Object[] getArguments(Object invocationTarget, int invocationCounter) {
-        return Arrays.stream(argumentValues).map(v -> v.getValue()).toArray();
+@AsValueClass
+public class Tuple implements Comparable<Tuple> {
```
liach: Let's declare this class final.

test/jdk/java/util/Collections/Ser.java
```diff
@@ -26,17 +26,36 @@
...
 import java.io.ObjectOutputStream;
+import java.io.Serializable;
 import java.util.Collections;
 import java.util.List;
 import java.util.Set;
 
 public class Ser {
+
+    @AsValueClass
+    static class SerV implements Serializable {
```
liach:
```suggestion
    static final class SerV implements Serializable {
```

liach (approved):

test/jdk/java/util/Collections/CheckedListBash.java
```diff
@@ -27,8 +27,11 @@
  * @summary Unit test for Collections.checkedList
  * @author  Josh Bloch
  * @key randomness
+ * @library /test/lib
+ * @run main CheckedListBash
```
lmesnik: Why
' * @run main CheckedListBash'
is needed now? 
Was it the same as default?

Also, there is a new main parameter, like
* @run main ${test.main.class}
to run current class if needed.
You might prefer to use.

test/lib/jdk/test/lib/valueclass/VClass.java
```diff
@@ -21,22 +21,24 @@
...
-    ValueArgumentsProvider(ArgumentValue[] argumentValues) {
-        this.argumentValues = argumentValues;
-    }
-
-    @Override
-    public Object[] getArguments(Object invocationTarget, int invocationCounter) {
-        return Arrays.stream(argumentValues).map(v -> v.getValue()).toArray();
+@AsValueClass
+public final class Tuple implements Comparable<Tuple> {
+    public int x;
```
lmesnik, line 37: Doesn't it makes sense to make not Tuple but some more interesting value class.
Like

@AsValueClass
public final class VClass {
 Integer i; <- value class, should be payload
 Integer[2] arrI; <- the value array that should be flattened
}

So it provoke more C2 optimizations.

lmesnik (changes requested):
in general changes looks very good and do what is expected.

There are few comments inline.

lmesnik (approved):
Thanks, the changes looks good.

PR #2423, opened 2026-05-12
8384359: [lworld] ConstraintCastNode::Ideal should not push incompatible casts through InlineTypeNodes

TobiHartmann:
The problem is that we are pushing casts through InlineTypeNodes without checking if their types are even compatible. In dead parts of the graph, when a type check wasn't folded yet, the type of the CheckCastPP can be completely unrelated to the type of the InlineTypeNode. Pushing the cast through leads to an incorrect type being used further below in the graph and subsequent failures.

I tried hard to come up with standalone reproducer for this but failed. It reproduces reliably with replay compilation though.

Thanks,
Tobias

---------

---------
### Progress

### Issue
 * [JDK-8384359](https://bugs.openjdk.org/browse/JDK-8384359): [lworld] ConstraintCastNode::Ideal should not push incompatible casts through InlineTypeNodes (**Bug** - P2)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2423/head:pull/2423` \
`$ git checkout pull/2423`

`$ git checkout pull/2423` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2423/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2423`

`$ git pr show -t 2423`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2423.diff">https://git.openjdk.org/valhalla/pull/2423.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2423#issuecomment-4430645745)
</details>

### DIFF
--- a/src/hotspot/share/opto/castnode.cpp
+++ b/src/hotspot/share/opto/castnode.cpp
@@ -118,7 +118,8 @@ Node *ConstraintCastNode::Ideal(PhaseGVN *phase, bool can_reshape) {
   // Push cast through InlineTypeNode but be careful because in dead parts of the graph,
   // an InlineTypeNode can be casted to some completely unrelated type.
   InlineTypeNode* vt = in(1)->isa_InlineType();
-  if (vt != nullptr && vt->is_allocated(phase) && vt->type()->inline_klass()->is_subtype_of(_type->is_instptr()->instance_klass())) {
+  if (vt != nullptr && vt->is_allocated(phase) && _type->isa_instptr() != nullptr &&
+      vt->type()->inline_klass()->is_subtype_of(_type->is_instptr()->instance_klass())) {
     Node* cast = clone();
     cast->set_req(1, vt->get_oop());
     vt = vt->clone()->as_InlineType();
--- a/src/hotspot/share/opto/castnode.cpp
+++ b/src/hotspot/share/opto/castnode.cpp
@@ -115,9 +115,10 @@ Node *ConstraintCastNode::Ideal(PhaseGVN *phase, bool can_reshape) {
     return this;
   }
 
-  // Push cast through InlineTypeNode
+  // Push cast through InlineTypeNode but be careful because in dead parts of the graph,
+  // an InlineTypeNode can be casted to some completely unrelated type.
   InlineTypeNode* vt = in(1)->isa_InlineType();
-  if (vt != nullptr && vt->is_allocated(phase)) {
+  if (vt != nullptr && vt->is_allocated(phase) && vt->type()->inline_klass()->is_subtype_of(_type->is_instptr()->instance_klass())) {
     Node* cast = clone();
     cast->set_req(1, vt->get_oop());
     vt = vt->clone()->as_InlineType();
--- a/src/hotspot/share/opto/inlinetypenode.cpp
+++ b/src/hotspot/share/opto/inlinetypenode.cpp
@@ -1220,6 +1220,7 @@ Node* InlineTypeNode::Ideal(PhaseGVN* phase, bool can_reshape) {
   Node* oop = get_oop();
   if (oop->isa_InlineType() && !phase->type(oop)->maybe_null()) {
     InlineTypeNode* vtptr = oop->as_InlineType();
+    assert(inline_klass() == vtptr->inline_klass(), "inconsistent types");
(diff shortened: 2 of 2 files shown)


marc-chevalier (approved):
That makes sense. If I understand well, we just find the inconsistency and leave it there. Can we be more active and remove the dead parts of the graph already? Or can we somehow make sure these nodes are disappearing "soon"? Maybe I'm biased toward early node killing...

Unfortunate we don't have a test, but I know the feeling...

TobiHartmann:
Thanks for the review Marc!

> If I understand well, we just find the inconsistency and leave it there

It's not really inconsistency. It's basically "cast A to B if subtype check C succeeds" but C would always fail and A can not be casted to B because they don't have a subtype relation. So the cast would be replaced by TOP and the subtype check would be folded to always false - but only later.

> Can we be more active and remove the dead parts of the graph already? Or can we somehow make sure these nodes are disappearing "soon"? Maybe I'm biased toward early node killing...

These dead data nodes depend on type checks like C in the example above that will be folded during IGVN but that depends on the order in which nodes are processed. In general, and especially with `-XX:+StressIGVN`, that order is basically random, so we have to make sure that all optimizations can handle such dead graphs. In this particular case, we should not assume that checkcast will always "pass" - they might even fail statically if they haven't been processed by IGVN yet.

> Unfortunate we don't have a test, but I know the feeling...

I wasted hours trying to come up with a targeted test and didn't succeed. It's really hard to get profiling + timing + node processing order right (even with -XX:+StressIGVN etc.). But it did reproduce reliably with replay and in various CI tests at least.

marc-chevalier:
I think we agree on what I called unconsistencies, but after that, I'm more puzzled. Indeed, we cannot assume the checkcast will pass, but if we can show it doesn't, we already know these nodes must be dead. I suppose right now, we are waiting for the condition to remove them later during the same IGVN by finding out that `C` never allows this branch to be taken. But I wonder whether we could be more aggressive and remove them now, so that we don't need to be able to simplify `C` (or alternatively, we can see it as another way to prove a path is dead). Isn't it similar to something Roland did?

merykitty (approved):
The fix looks good, thanks for your work.

merykitty:
@marc-chevalier Roland is working on it if I understand correctly. See https://github.com/openjdk/jdk/pull/30266.

TobiHartmann:
@marc-chevalier do you mean this?
https://github.com/openjdk/valhalla/blob/f39fa97e6905c6cad790b97aad663b6a4e2040a2/src/hotspot/share/opto/node.cpp#L3215-L3222

marc-chevalier:
Yes, this! Anyway, it's a bigger fish.

TobiHartmann:
Yes, an alternative fix would be to move the `TypeNode::Ideal` call to before this optimization to let it fold the CheckCastPP aggresively:

```
diff --git a/src/hotspot/share/opto/castnode.cpp b/src/hotspot/share/opto/castnode.cpp
index c46045c9810..aa7fae8369e 100644
--- a/src/hotspot/share/opto/castnode.cpp
+++ b/src/hotspot/share/opto/castnode.cpp
@@ -115,6 +115,13 @@ Node *ConstraintCastNode::Ideal(PhaseGVN *phase, bool can_reshape) {
     return this;
   }
 
+  if (in(1) != nullptr && phase->type(in(1)) != Type::TOP) {
+    Node* progress = TypeNode::Ideal(phase, can_reshape);
+    if (progress != nullptr) {
+      return progress;
+    }
+  }
+
   // Push cast through InlineTypeNode
   InlineTypeNode* vt = in(1)->isa_InlineType();
   if (vt != nullptr && vt->is_allocated(phase)) {
@@ -128,10 +135,6 @@ Node *ConstraintCastNode::Ideal(PhaseGVN *phase, bool can_reshape) {
     return vt;
   }
 
-  if (in(1) != nullptr && phase->type(in(1)) != Type::TOP) {
-    return TypeNode::Ideal(phase, can_reshape);
-  }
-
   return nullptr;
 }
```

However, we would fail again when running with `-XX:-KillPathsReachableByDeadTypeNode` which is "ok" but something we would like to avoid.

PR #2425, opened 2026-05-12
8382504: [lworld] TestIntrinsics.java triggers assert "must not cast to a different type"

TobiHartmann:
Mainline fix [JDK-8373248](https://bugs.openjdk.org/browse/JDK-8373248) added additional code to verify that a `CastPPNode` does not change the type and we now assert when casting the value of an unsafe flat access to the proper type. The `CastPP` should be replaced by a `CheckCastPP`.

Thanks,
Tobias

---------

---------
### Progress

### Issue
 * [JDK-8382504](https://bugs.openjdk.org/browse/JDK-8382504): [lworld] TestIntrinsics.java triggers assert "must not cast to a different type" (**Bug** - P3)

### Reviewers
 * [Marc Chevalier](https://openjdk.org/census#mchevalier) (@marc-chevalier - Committer) ⚠️ Review applies to [241244d1](https://git.openjdk.org/valhalla/pull/2425/files/241244d160779a95d634745584d6b668e15ec1fc)
 * [Quan Anh Mai](https://openjdk.org/census#qamai) (@merykitty - Committer) ⚠️ Review applies to [241244d1](https://git.openjdk.org/valhalla/pull/2425/files/241244d160779a95d634745584d6b668e15ec1fc)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2425/head:pull/2425` \
`$ git checkout pull/2425`

`$ git checkout pull/2425` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2425/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2425`

`$ git pr show -t 2425`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2425.diff">https://git.openjdk.org/valhalla/pull/2425.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2425#issuecomment-4431174037)
</details>

### DIFF
--- a/src/hotspot/share/opto/library_call.cpp
+++ b/src/hotspot/share/opto/library_call.cpp
@@ -2779,7 +2779,7 @@ bool LibraryCallKit::inline_unsafe_flat_access(bool is_store, AccessKind kind) {
     const Type* value_type = _gvn.type(value);
     if (!value_type->is_inlinetypeptr()) {
       value_type = Type::get_const_type(value_klass)->filter_speculative(value_type);
-      Node* new_value = _gvn.transform(new CastPPNode(control(), value, value_type, ConstraintCastNode::DependencyType::NonFloatingNarrowing));
+      Node* new_value = _gvn.transform(new CheckCastPPNode(control(), value, value_type, ConstraintCastNode::DependencyType::NonFloatingNarrowing));
       new_value = InlineTypeNode::make_from_oop(this, new_value, value_klass);
       replace_in_map(value, new_value);
       value = new_value;
--- a/src/hotspot/share/opto/library_call.cpp
+++ b/src/hotspot/share/opto/library_call.cpp
@@ -2745,6 +2745,7 @@ bool LibraryCallKit::inline_unsafe_flat_access(bool is_store, AccessKind kind) {
     if (layout == LayoutKind::REFERENCE) {
       if (!base_type->is_aryptr()->is_not_flat()) {
         const TypeAryPtr* array_type = base_type->is_aryptr()->cast_to_not_flat();
+        // TODO 8350865 This should be a CheckCastPP, can we add a test?
         Node* new_base = _gvn.transform(new CastPPNode(control(), base, array_type, ConstraintCastNode::DependencyType::NonFloatingNarrowing));
         replace_in_map(base, new_base);
         base = new_base;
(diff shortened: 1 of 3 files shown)


marc-chevalier (commented):
It seems reasonable, but I'm not sure I get it enough. What do we have in `value` vs `value_type` when that goes wrong?

TobiHartmann:
Thanks for the review Marc.

> What do we have in value vs value_type when that goes wrong?

The intrinsic has an Object type in its signature:
```
  do_signature(putFlatValue_signature,    "(Ljava/lang/Object;JILjava/lang/Class;Ljava/lang/Object;)V")                 \
```

https://github.com/openjdk/valhalla/blob/873f02d2d521108f8883341350315991b592e733/src/java.base/share/classes/jdk/internal/misc/Unsafe.java#L384

So `value` is an Object and `value_type` is a value class type. We cast `value` to that value class type.

TobiHartmann:
Of course, we could / should add type checks (at some point). But it's Unsafe and therefore good enough for now.

marc-chevalier:
Ah, of course! Yes, a `CastPPNode` won't do, as I recently learnt from Quan Anh.

marc-chevalier (approved):

TobiHartmann:
Thanks Marc!

merykitty:
Thanks for fixing this issue. There are other problematic CastPPNodes in LibraryCallKit::inline_unsafe_flat_access. I think they do not result in assert. So, maybe they can be solved later.

TobiHartmann:
Thanks Quan Anh!

> There are other problematic CastPPNodes in LibraryCallKit::inline_unsafe_flat_access. I think they do not result in assert

I looked at those two and they seem ok to me because they don't change the type but just its properties:

https://github.com/openjdk/valhalla/blob/f39fa97e6905c6cad790b97aad663b6a4e2040a2/src/hotspot/share/opto/library_call.cpp#L2747-L2748

https://github.com/openjdk/valhalla/blob/f39fa97e6905c6cad790b97aad663b6a4e2040a2/src/hotspot/share/opto/library_call.cpp#L2762-L2764

But I agree, let's fix those later if need be.

merykitty (approved):
Looks good to me.

TobiHartmann:
Thanks Quan Anh!

merykitty:
@TobiHartmann They are problematic, because when we remove the cast, the pointer, for example, which were `TwoBytes(+0)+bot`, becomes `TwoBytes(+bot)+bot`, which is a different memory because each field in a flat array has its own slice.

TobiHartmann:
I think you are right. I added a comment to follow up on those with JDK-8350865. Maybe we can construct a reproducer to hit the same assert - would be nice.

PR #2432, opened 2026-05-13
8384127: [lworld] C2 compilation failed with "graph should be schedulable" assert

marc-chevalier:
That was discovered when working on the intrinsic for isAtomic. It turns out it is actually not needed, but the shape of the graph to trigger the issue needs to be complex enough not to be simplified too hard, but simple enough that optimization triggers.

Let's see what happens. In this screenshot, let's look at
1  before
Nodes `569 LoadB` and `573 LoadB` are loading the fields from `nullFreeAtomicArray2[1]`, in the inlined `Asserts.assertEquals`. Since they are strict final fields, they cannot be observed to be mutated, and we can rewire the memory input a lot higher. We get
2  after

The difficulty comes from the node `563 LoadN` that corresponds to the access `nullFreeAtomicArray2[1]`: we can climb up to the initial value of the `355 AllocateArray`.
3  getting input

Here, the default value is the value `initVal1`, which is `340 LoadN`, and the type is opaque here. Then, the graph becomes
4  nope

then
5  nope

and here, it is not clear that these `LoadB` are for strict final fields. That starts to be wrong. Later, during GCM, precedence edges from `538 MemBarStoreStore` to the `LoadB` nodes, creating a cycle inside the same block, and then, the graph becomes not schedulable.

The fix is either not to subsume `563 LoadN` by `340 LoadN` (which is sad) because types don't match, or subsumes it by something correct. Here, we can just add a `CheckCastPPNode` under the initial value to subsumes by something correct, and all is fine again.

Should we check that the initial value indeed has the right type? Yes, but not here, and it is worked on (see [JDK-8383820: [lworld] Several native methods of ValueClass class lack proper argument validation](https://bugs.openjdk.org/browse/JDK-8383820)). I've just added a debug only check and halt, it shall be enough for now.

Thanks,
Marc

---------

---------
### Progress

### Issue
 * [JDK-8384127](https://bugs.openjdk.org/browse/JDK-8384127): [lworld] C2 compilation failed with "graph should be schedulable" assert (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2432/head:pull/2432` \
`$ git checkout pull/2432`

`$ git checkout pull/2432` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2432/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2432`

`$ git pr show -t 2432`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2432.diff">https://git.openjdk.org/valhalla/pull/2432.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2432#issuecomment-4441032283)
</details>

### DIFF
--- a/src/hotspot/share/opto/library_call.cpp
+++ b/src/hotspot/share/opto/library_call.cpp
@@ -4767,6 +4767,17 @@ bool LibraryCallKit::inline_newArray(bool null_free, bool atomic) {
                 init_val = init_val->as_InlineType()->buffer(this);
               }
             }
+#ifdef ASSERT
+            init_val = null_check(init_val);
+            Node* wrong_type_ctl = gen_subtype_check(init_val, makecon(TypeKlassPtr::make(array_klass->element_klass())));
+            {
+              PreserveJVMState pjvms(this);
+              set_control(wrong_type_ctl);
+              halt(control(), frameptr(), "incompatible type for initVal in newArray");
+              stop_and_kill_map();
+            }
+#endif
+            init_val = _gvn.transform(new CheckCastPPNode(control(), init_val, TypeOopPtr::make_from_klass(array_klass->element_klass())));
             // TODO 8350865 Should we add a check of the init_val type (maybe in debug only + halt)?
             // If we insert a checkcast here, we can be sure that init_val is an InlineTypeNode, so
             // when we folded a field load from an allocation (e.g. during escape analysis), we can
--- a/src/hotspot/share/opto/library_call.cpp
+++ b/src/hotspot/share/opto/library_call.cpp
@@ -4778,7 +4778,7 @@ bool LibraryCallKit::inline_newArray(bool null_free, bool atomic) {
                 stop_and_kill_map();
               }
 #endif
-              init_val = _gvn.transform(new CheckCastPPNode(control(), init_val, TypeOopPtr::make_from_klass(array_klass->element_klass())));
+              init_val = _gvn.transform(new CheckCastPPNode(control(), init_val, TypeOopPtr::make_from_klass(array_klass->element_klass()), ConstraintCastNode::DependencyType::NonFloatingNarrowing));
             }
           }
           Node* obj = new_array(makecon(array_klass_type), length, 0, nullptr, false, init_val);
--- a/src/hotspot/share/opto/library_call.cpp
+++ b/src/hotspot/share/opto/library_call.cpp
@@ -4767,21 +4767,19 @@ bool LibraryCallKit::inline_newArray(bool null_free, bool atomic) {
                 init_val = init_val->as_InlineType()->buffer(this);
               }
             }
+            if (init_val != nullptr) {
 #ifdef ASSERT
-            init_val = null_check(init_val);
-            Node* wrong_type_ctl = gen_subtype_check(init_val, makecon(TypeKlassPtr::make(array_klass->element_klass())));
-            {
-              PreserveJVMState pjvms(this);
-              set_control(wrong_type_ctl);
-              halt(control(), frameptr(), "incompatible type for initVal in newArray");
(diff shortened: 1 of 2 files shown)


src/hotspot/share/opto/library_call.cpp
```diff
@@ -4769,10 +4767,19 @@ bool LibraryCallKit::inline_newArray(bool null_free, bool atomic) {
...
+              init_val = null_check(init_val);
+              Node* wrong_type_ctl = gen_subtype_check(init_val, makecon(TypeKlassPtr::make(array_klass->element_klass())));
+              {
+                PreserveJVMState pjvms(this);
+                set_control(wrong_type_ctl);
+                halt(control(), frameptr(), "incompatible type for initVal in newArray");
+                stop_and_kill_map();
+              }
+#endif
+              init_val = _gvn.transform(new CheckCastPPNode(control(), init_val, TypeOopPtr::make_from_klass(array_klass->element_klass())));
```
merykitty: Please always remember to set the `dependency` argument of the constructors of `ConstraintCastNode`'s. It should only be `FloatingNarrowing` if we also just emit the corresponding check for that cast, otherwise, it must be `NonFloatingNarrowing`.

marc-chevalier: I tihnk in this case, it should rather be `NonFloatingNarrowing`. I'm running tests again.

Shouldn't this argument be made non-optional if we should always set it?

merykitty: Yes, I think it should, I am planning to do a round of audit regarding this argument next month. The same also applies to `LoadNode`.

dean-long:
I'm sure it's unrelated, but do you know if all the MemBarStoreStore nodes are necessary?  The way the graph is pruned it looks like there are 4 in a row.

src/hotspot/share/opto/library_call.cpp
```diff
@@ -4769,10 +4767,19 @@ bool LibraryCallKit::inline_newArray(bool null_free, bool atomic) {
                 init_val = init_val->as_InlineType()->buffer(this);
               }
             }
-            // TODO 8350865 Should we add a check of the init_val type (maybe in debug only + halt)?
-            // If we insert a checkcast here, we can be sure that init_val is an InlineTypeNode, so
-            // when we folded a field load from an allocation (e.g. during escape analysis), we can
-            // remove the check init_val->is_InlineType().
+            if (init_val != nullptr) {
```
merykitty, line 4770: Is it easier if we just emit a checkcast here?

merykitty:
Apparently, I just encountered a [similar issue in mainline](https://bugs.openjdk.org/browse/JDK-8382936) (screw you array polymorphism).

src/hotspot/share/opto/library_call.cpp
marc-chevalier: `gen_checkcast` would deopt or throw, and here, I was suggested to abort. I couldn't find a function that would do that magically for me, but I still delegate the hard part to `gen_subtype_check`. That's something.

marc-chevalier:
@dean-long No, I don't know I fear. Also, I display only a fraction of the graph. Maybe the look less redundant with the whole graph. I've checked some of them, and they seem to be legitimately added, but maybe some could be removed later as optimizations go. Also, the state in the screenshot is just in the middle some IGVN, before loop opts, things can still happen. I can't check very easily, I don't have the graph opened just there.

merykitty (approved):
Looks good to me, then. Thanks for making the fix.

marc-chevalier:
/integrate

Thanks @merykitty for the review, and for the extensive guidance during investigation!

PR #2433, opened 2026-05-13
8384202: [lworld] C2: assert(offset == field->offset_in_bytes()) failed: offset mismatch

TobiHartmann:
The original failure was in `test21` from `compiler/valhalla/inlinetypes/TestNullableInlineTypes.java#id5`. I extracted it into a standalone reproducer.

C2 asserts in `InlineTypeNode::field_value_by_offset` while looking up a field value by offset. The lookup reaches a flat field whose corresponding input is already `TOP`. This happens on the normal return path from `test2()` in `test()`. That path is dead because `test2()` always throws a `NullPointerException` when it tries to initialize a null restricted field with null. We only "know" this after incremental inlining though.

So at this point, with `-XX:+StressIGVN`, the impossible control path has not been removed yet, but the data path has already started dying: field values on the dead `InlineTypeNode` have been replaced by `TOP`. `InlineTypeNode::field_value_by_offset` already accounts for this with a `value->is_top()` check, but the offset assert in that case is too strict.

In this reproducer, the lookup is recursive: it asks for a field inside a flat field whose value is already `TOP`. Therefore, the requested offset is the offset of a nested field within the flattened payload, while `field->offset_in_bytes()` is the offset of the enclosing flat field. Those offsets are not expected to match.

I adjusted the assert accordingly.

Thanks,
Tobias

---------

---------
### Progress

### Issue
 * [JDK-8384202](https://bugs.openjdk.org/browse/JDK-8384202): [lworld] C2: assert(offset == field-&gt;offset_in_bytes()) failed: offset mismatch (**Bug** - P4)

### Reviewers
 * [Marc Chevalier](https://openjdk.org/census#mchevalier) (@marc-chevalier - Committer) ⚠️ Review applies to [426af5ba](https://git.openjdk.org/valhalla/pull/2433/files/426af5ba760b8cf7cbbbc3d65dc88ac02aac4a5f)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2433/head:pull/2433` \
`$ git checkout pull/2433`

`$ git checkout pull/2433` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2433/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2433`

`$ git pr show -t 2433`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2433.diff">https://git.openjdk.org/valhalla/pull/2433.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2433#issuecomment-4441274165)
</details>

### DIFF
--- a/src/hotspot/share/opto/inlinetypenode.cpp
+++ b/src/hotspot/share/opto/inlinetypenode.cpp
@@ -229,10 +229,14 @@ Node* InlineTypeNode::field_value_by_offset(int offset, bool recursive) const {
   ciField* field = this->field(index);
   assert(!field->is_flat() || field->type()->is_inlinetype(), "must be an inline type");
 
-  if (!recursive || !field->is_flat() || value->is_top()) {
-    // If the graph is dying (value is top), a load may still ask for a nested
-    // field inside a flattened field before the dead load itself is folded away.
-    assert((offset == field->offset_in_bytes()) || (value->is_top() && field->is_flat()), "offset mismatch");
+  if (value->is_top()) {
+    // The graph is dying but a load may still ask for a nested field
+    // inside a flattened field before the dead load itself is folded away.
+    assert(offset == field->offset_in_bytes() || field->is_flat(), "offset mismatch");
+    return value;
+  }
+  if (!recursive || !field->is_flat()) {
+    assert(offset == field->offset_in_bytes(), "offset mismatch");
     return value;
   }
 
--- a/src/hotspot/share/opto/inlinetypenode.cpp
+++ b/src/hotspot/share/opto/inlinetypenode.cpp
@@ -230,7 +230,9 @@ Node* InlineTypeNode::field_value_by_offset(int offset, bool recursive) const {
   assert(!field->is_flat() || field->type()->is_inlinetype(), "must be an inline type");
 
   if (!recursive || !field->is_flat() || value->is_top()) {
-    assert(offset == field->offset_in_bytes(), "offset mismatch");
+    // If the graph is dying (value is top), a load may still ask for a nested
+    // field inside a flattened field before the dead load itself is folded away.
+    assert((offset == field->offset_in_bytes()) || (value->is_top() && field->is_flat()), "offset mismatch");
     return value;
   }
 
--- a/test/hotspot/jtreg/compiler/valhalla/inlinetypes/TestDeadInlineTypeFieldOffset.java
+++ b/test/hotspot/jtreg/compiler/valhalla/inlinetypes/TestDeadInlineTypeFieldOffset.java
@@ -29,6 +29,7 @@ import jdk.internal.vm.annotation.NullRestricted;
 /*
  * @test
  * @summary Test loads from a dead InlineTypeNode with top field values.
+ * @bug 8384202
  * @requires vm.compiler2.enabled
  * @requires (vm.opt.PreloadClasses == null | vm.opt.PreloadClasses == "true")
  * @enablePreview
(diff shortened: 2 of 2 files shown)


test/hotspot/jtreg/compiler/valhalla/inlinetypes/TestDeadInlineTypeFieldOffset.java
```diff
@@ -0,0 +1,100 @@
...
+
+/*
+ * @test
+ * @summary Test loads from a dead InlineTypeNode with top field values.
+ * @requires vm.compiler2.enabled
+ * @enablePreview
+ * @modules java.base/jdk.internal.vm.annotation
+ * @run main ${test.main.class}
+ * @run main/othervm -XX:+UnlockDiagnosticVMOptions -XX:+IgnoreUnrecognizedVMOptions
+ *                   -XX:+StressIGVN -XX:StressSeed=8 -XX:+AlwaysIncrementalInline
```
marc-chevalier, line 39: The hardcoded is on purpose? It reproduces so rarely otherwise?

src/hotspot/share/opto/inlinetypenode.cpp
```diff
@@ -230,7 +230,9 @@ Node* InlineTypeNode::field_value_by_offset(int offset, bool recursive) const {
   assert(!field->is_flat() || field->type()->is_inlinetype(), "must be an inline type");
 
   if (!recursive || !field->is_flat() || value->is_top()) {
```
merykitty: This looks a bit cluttered, can we move `if (value->is_top())` into a separate case before this `if` instead?

marc-chevalier (approved):
That makes sense. Do we want the bug number in the test?

src/hotspot/share/opto/inlinetypenode.cpp
TobiHartmann: Sure. I had this version first and didn't like it but I don't have a strong opinion. Updated.

test/hotspot/jtreg/compiler/valhalla/inlinetypes/TestDeadInlineTypeFieldOffset.java
TobiHartmann: Yes, it's on purpose to make it reliably reproduce the issue and the usual way of how we design these tests. I.e. one run with fixed seed and one without. As we discussed offline, there are several alternatives to this approach, including `RepeatCompilation` etc. We decided on the current approach a while ago but we can surely revisit. It's out of scope of this PR though 🙂 

Testing also revealed that PreloadClasses is required to make sure that the `@NullRestricted` annotation is respected, so I added a corresponding `@requires`.

TobiHartmann:
Thanks for the reviews Quan Anh and Marc! I pushed an updated version.

> Do we want the bug number in the test?

I added a bug number to the test.

merykitty (approved):
Thanks, that looks good to me.

test/hotspot/jtreg/compiler/valhalla/inlinetypes/TestDeadInlineTypeFieldOffset.java
merykitty: I think that means this test lacks a test that does not use a fixed seed?

TobiHartmann:
Thanks Quan Anh!

test/hotspot/jtreg/compiler/valhalla/inlinetypes/TestDeadInlineTypeFieldOffset.java
TobiHartmann: > I think that means this test lacks a test that does not use a fixed seed?

There is the `@run main ${test.main.class}` that will reproduce this without a fixed seed, right? Testing in higher tiers will run the required stress options.

PR #2434, opened 2026-05-13
8384283: [lworld] Post-parse call devirtualization with unloaded return type fails

marc-chevalier:
As suggested by Tobias, it seems that the loop
https://github.com/openjdk/valhalla/blob/de6aff4ffe3344bfb886fa27385e432016bbbeef/src/hotspot/share/opto/graphKit.cpp#L2344-L2347
is wrong, even with the index (`[0]`) fixed.

Indeed, `final_state` is the `SafePointNode` that tracks the JVM state, that is the stack, slots... This kind of things. There is no good reason this should replace the output of the existing call. Instead, we should replace the projection of the old call with the output of the new call. The new call can be found as being the input of `result`. Of course, `result` can be something else than a projection from a call, typically a load, a constant... But then, I don't think we can be in the case of an unloaded return type. Testing seems to agree with that.

Another thing to take care of: `callprojs->resproj[i]` might be null, when the projection isn't used. Indeed, `store_inline_type_fields_to_buf` has a lot of `double` input, and so the `Half` part of these inputs are `top` (and thus, not wired from the old call). On the other hand, the projection of the new call must be there, but they should:
https://github.com/openjdk/valhalla/blob/de6aff4ffe3344bfb886fa27385e432016bbbeef/src/hotspot/share/opto/graphKit.cpp#L2167-L2181

Thanks,
Marc

---------

---------
### Progress

### Issue
 * [JDK-8384283](https://bugs.openjdk.org/browse/JDK-8384283): [lworld] Post-parse call devirtualization with unloaded return type fails (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2434/head:pull/2434` \
`$ git checkout pull/2434`

`$ git checkout pull/2434` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2434/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2434`

`$ git pr show -t 2434`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2434.diff">https://git.openjdk.org/valhalla/pull/2434.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2434#issuecomment-4441323833)
</details>

### DIFF
--- a/src/hotspot/share/opto/graphKit.cpp
+++ b/src/hotspot/share/opto/graphKit.cpp
@@ -2310,7 +2310,6 @@ void GraphKit::replace_call(CallNode* call, Node* result, bool do_replaced_nodes
   CallProjections* callprojs = call->extract_projections(true, do_asserts);
 
   Unique_Node_List wl;
-  Node* init_mem = call->in(TypeFunc::Memory);
   Node* final_mem = final_state->in(TypeFunc::Memory);
   Node* final_ctl = final_state->in(TypeFunc::Control);
   Node* final_io = final_state->in(TypeFunc::I_O);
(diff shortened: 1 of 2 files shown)


src/hotspot/share/opto/graphKit.cpp
```diff
@@ -2310,7 +2310,6 @@ void GraphKit::replace_call(CallNode* call, Node* result, bool do_replaced_nodes
   CallProjections* callprojs = call->extract_projections(true, do_asserts);
 
   Unique_Node_List wl;
-  Node* init_mem = call->in(TypeFunc::Memory);
```
marc-chevalier, line 2313: That was just unused.

test/hotspot/jtreg/compiler/valhalla/inlinetypes/TestCallDevirtualizationWithUnloadedReturn.java
```diff
@@ -0,0 +1,83 @@
...
+ * questions.
+ */
+
+/**
+ * @test
+ * @bug 8384283
+ * @summary Test post-parse call devirtualization with an unloaded return type.
+ * @enablePreview
+ * @run main ${test.main.class}
+ * @run main/othervm -XX:+IgnoreUnrecognizedVMOptions -XX:+UnlockDiagnosticVMOptions
```
TobiHartmann: I think both flags are not needed.

TobiHartmann (approved):
Looks good to me!

marc-chevalier:
/integrate

Thanks @TobiHartmann for review!

PR #2438, opened 2026-05-15
8383997: [lworld] Buffering via SharedRuntime::store_inline_type_fields_to_buf is slow

TobiHartmann:
We noticed that `VectorUnsignedMinMaxOperationsTest.java` timed out in the Valhalla CI with `--enable-preview`. The root cause is that the test has a large number of `@Run` methods that get excluded from compilation and call `Verify.checkEQ(int_out[i], expected)` with `int` arguments that are autoboxed by `Integer.valueOf`. I.e. we call `Integer.valueOf` which returns a scalarized Integer from the interpreter and thus have to buffer via the slow path (due to `-XX:-UseTLAB`) by calling `StubRoutines::store_inline_type_fields_to_buf`. This is slower than buffering in the callee which happens in the baseline:

Without `--enable-preview` the test runs for 6m27.851s vs. with `--enable-preview` the test runs for 9m9.105s.

The fix is to add a fast path to `SharedRuntime::store_inline_type_fields_to_buf` that delegates initialization of the buffer to the calling stub which calls the pack handler after the runtime call returned. I also moved the expensive `RegisterMap` creation to the slow path. This improves runtime significantly to 5m34.543s which is even faster than the baseline (don't ask me why :slightly_smiling_face:).

Thanks,
Tobias

---------

---------
### Progress

### Issue
 * [JDK-8383997](https://bugs.openjdk.org/browse/JDK-8383997): [lworld] Buffering via SharedRuntime::store_inline_type_fields_to_buf is slow (**Bug** - P4)

### Reviewers
 * [Marc Chevalier](https://openjdk.org/census#mchevalier) (@marc-chevalier - Committer) ⚠️ Review applies to [8ddaa819](https://git.openjdk.org/valhalla/pull/2438/files/8ddaa8197af83a0f6e43caa82bf9e4d18973ece8)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2438/head:pull/2438` \
`$ git checkout pull/2438`

`$ git checkout pull/2438` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2438/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2438`

`$ git pr show -t 2438`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2438.diff">https://git.openjdk.org/valhalla/pull/2438.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2438#issuecomment-4459154715)
</details>


marc-chevalier (approved):
It took me a bit, but I'm convinced now.

TobiHartmann:
Thanks for your review Marc! As we discussed offline, I removed the unused `callerFrame` local to make it more explicit that the `sender` call is to initialize the `reg_map`.

PR #2445, opened 2026-05-18
8384405: [lworld] C2: Load nodes inserted by InlineTypeNode::make_from_flat() need to be pinned

TobiHartmann:
@rwestrel (thanks!) noticed that flat array loads are control dependent on the flat-array layout checks, but they are not pinned strongly enough relative to the range check. New `TestFlatArrayLoadPinning.java` demonstrates this: With `-XX:+StressGCM`, C2 can schedule one of the field loads from the flat array to before the range check. When the array index is out of bounds, compiled code performs an out-of-bounds memory load from the flat array payload and crashes with a `SIGSEGV` instead of throwing the expected `ArrayIndexOutOfBoundsException`.

The fix is to use `C2_UNKNOWN_CONTROL_LOAD` for these loads.

Thanks,
Tobias

---------

---------
### Progress

### Issue
 * [JDK-8384405](https://bugs.openjdk.org/browse/JDK-8384405): [lworld] C2: Load nodes inserted by InlineTypeNode::make_from_flat() need to be pinned (**Bug** - P4)

### Reviewers
 * [Roland Westrelin](https://openjdk.org/census#roland) (@rwestrel - Committer)

### Contributors
 * Roland Westrelin `<roland@openjdk.org>`

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2445/head:pull/2445` \
`$ git checkout pull/2445`

`$ git checkout pull/2445` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2445/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2445`

`$ git pr show -t 2445`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2445.diff">https://git.openjdk.org/valhalla/pull/2445.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2445#issuecomment-4475931944)
</details>

### DIFF
--- a/src/hotspot/share/opto/inlinetypenode.cpp
+++ b/src/hotspot/share/opto/inlinetypenode.cpp
@@ -1456,7 +1456,8 @@ InlineTypeNode* InlineTypeNode::make_from_flat_impl(GraphKit* kit, ciInlineKlass
 InlineTypeNode* InlineTypeNode::make_from_flat_array(GraphKit* kit, ciInlineKlass* vk, Node* base, Node* idx) {
   assert(vk->maybe_flat_in_array(), "element type %s cannot be flat in array", vk->name()->as_utf8());
   PhaseGVN& gvn = kit->gvn();
-  DecoratorSet decorators = IN_HEAP | IS_ARRAY | MO_UNORDERED | C2_CONTROL_DEPENDENT_LOAD;
+  // The flat field loads are dependent on both the array layout checks as well as the range check.
+  DecoratorSet decorators = IN_HEAP | IS_ARRAY | MO_UNORDERED | C2_CONTROL_DEPENDENT_LOAD | C2_UNKNOWN_CONTROL_LOAD;
   kit->C->set_flat_accesses();
   InlineTypeNode* vt_nullable = nullptr;
   InlineTypeNode* vt_null_free = nullptr;
--- a/test/hotspot/jtreg/compiler/valhalla/inlinetypes/TestFlatArrayLoadPinning.java
+++ b/test/hotspot/jtreg/compiler/valhalla/inlinetypes/TestFlatArrayLoadPinning.java
@@ -1,0 +1,81 @@
+/*
+ * Copyright (c) 2026, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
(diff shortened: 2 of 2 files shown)


src/hotspot/share/opto/inlinetypenode.cpp
```diff
@@ -1456,7 +1456,8 @@ InlineTypeNode* InlineTypeNode::make_from_flat_impl(GraphKit* kit, ciInlineKlass
 InlineTypeNode* InlineTypeNode::make_from_flat_array(GraphKit* kit, ciInlineKlass* vk, Node* base, Node* idx) {
   assert(vk->maybe_flat_in_array(), "element type %s cannot be flat in array", vk->name()->as_utf8());
   PhaseGVN& gvn = kit->gvn();
-  DecoratorSet decorators = IN_HEAP | IS_ARRAY | MO_UNORDERED | C2_CONTROL_DEPENDENT_LOAD;
+  // The flat field loads are dependent on both the array layout checks as well as the range check.
+  DecoratorSet decorators = IN_HEAP | IS_ARRAY | MO_UNORDERED | C2_CONTROL_DEPENDENT_LOAD | C2_UNKNOWN_CONTROL_LOAD;
```
rwestrel, line 1460: Do we need both `C2_CONTROL_DEPENDENT_LOAD` and `C2_UNKNOWN_CONTROL_LOAD here?
Also, if the value is known null free and non atomic at compile time (the runtime checks control fold and the loads are control dependent on the range check), wouldn't we want to not pin the load?

TobiHartmann: > Do we need both C2_CONTROL_DEPENDENT_LOAD and `C2_UNKNOWN_CONTROL_LOAD here?

I took this from `ArrayCopyNode::load` where we have both as well (I think you did that with JDK-8323274 :slightly_smiling_face: ). We also have both in a few other places, for example in the `needs_range_check` branch in `Parse::array_load` (added with JDK-8333889).

> Also, if the value is known null free and non atomic at compile time (the runtime checks control fold and the loads are control dependent on the range check), wouldn't we want to not pin the load?

You mean if the array layout is known at compile time and the load is directly below the the range check? Yes, I think in that case we would want to not pin the load. Seems a bit risky / non-trivial though. Are you fine with deferring to a follow-up enhancement?

rwestrel: > > Do we need both C2_CONTROL_DEPENDENT_LOAD and `C2_UNKNOWN_CONTROL_LOAD here?
> 
> I took this from `ArrayCopyNode::load` where we have both as well (I think you did that with JDK-8323274 🙂 ). We also have both in a few other places, for example in the `needs_range_check` branch in `Parse::array_load` (added with JDK-8333889).

Right. We need both because the first one causes the control to be set and the second to set the dependency type.

> > Also, if the value is known null free and non atomic at compile time (the runtime checks control fold and the loads are control dependent on the range check), wouldn't we want to not pin the load?
> 
> You mean if the array layout is known at compile time and the load is directly below the the range check? Yes, I think in that case we would want to not pin the load. Seems a bit risky / non-trivial though. Are you fine with deferring to a follow-up enhancement?

Sure.

rwestrel (approved):

src/hotspot/share/opto/inlinetypenode.cpp
TobiHartmann: Thanks Roland, I filed [JDK-8384865](https://bugs.openjdk.org/browse/JDK-8384865) for this.

PR #2446, opened 2026-05-18
8328675: [lworld] Re-enable EliminateAutoBox

TobiHartmann:
This patch re-enables the `EliminateAutoBox` functionality when `--enable-preview` is set, i.e. when the primitive boxes are value classes. `EliminateAutoBox` has multiple effects, including an impact on inlining decisions but its main goal is to avoid boxing via `Integer.valueOf(int)` etc by replacing that call. I added support for scalarized returns from that method and also added support for removing unboxing via `Integer.intValue` etc.

With this fix, I'm seeing various improvements:

Screenshot from 2026-06-02 11-08-53

(The regressions are not significant and due to high variance of the benchmark)

There's more we could do here but I already spent way too much time on this, so I'll defer follow-up work to [JDK-8385809](https://bugs.openjdk.org/browse/JDK-8385809).

Thanks,
Tobias

---------

---------
### Progress

### Issue
 * [JDK-8328675](https://bugs.openjdk.org/browse/JDK-8328675): [lworld] Re-enable EliminateAutoBox (**Enhancement** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2446/head:pull/2446` \
`$ git checkout pull/2446`

`$ git checkout pull/2446` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2446/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2446`

`$ git pr show -t 2446`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2446.diff">https://git.openjdk.org/valhalla/pull/2446.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2446#issuecomment-4601190849)
</details>


src/hotspot/share/opto/macro.cpp
```diff
@@ -704,7 +704,7 @@ Node* PhaseMacroExpand::inline_type_from_mem(ciInlineKlass* vk, const TypeAryPtr
   };
 
   // Create a new InlineTypeNode and retrieve the field values from memory
-  InlineTypeNode* vt = InlineTypeNode::make_uninitialized(_igvn, vk, false);
+  InlineTypeNode* vt = InlineTypeNode::make_uninitialized(_igvn, vk, null_free);
```
TobiHartmann, line 707: Just spotted this when browsing the code.

src/hotspot/share/opto/macro.cpp
```diff
@@ -1557,47 +1557,55 @@ bool PhaseMacroExpand::eliminate_allocate_node(AllocateNode *alloc) {
...
+  _callprojs = call->extract_projections(false /*separate_io_proj*/, false /*do_asserts*/);
 
-  _callprojs = boxing->extract_projections(false /*separate_io_proj*/, false /*do_asserts*/);
+  process_users_of_allocation(call);
 
-  const TypeTuple* r = boxing->tf()->range_sig();
-  assert(r->cnt() > TypeFunc::Parms, "sanity");
-  const TypeInstPtr* t = r->field_at(TypeFunc::Parms)->isa_instptr();
-  assert(t != nullptr, "sanity");
+  const TypeInstPtr* t = nullptr;
```
marc-chevalier: That is used only for logging, right?

marc-chevalier (approved):
After second pass, I'm convinced.

I think it's unfortunate we have so much case split in `ProjNode::Identity` wrt the flatness of arguments and return. It's rather a general comment than something actionable here.

src/hotspot/share/opto/macro.cpp
TobiHartmann: Good point, I moved this code into the log specific branch below.

TobiHartmann:
Thanks for your review Marc! I adjusted the logging related code and also added a comment to [JDK-8385809](https://bugs.openjdk.org/browse/JDK-8385809) that we should consider updating the naming of methods and log entries.

marc-chevalier (approved):

TobiHartmann:
Thanks again!

PR #2452, opened 2026-05-19
8381373: [lworld] C2 failed assert(!flat || flat_array) failed: inconsistency

TobiHartmann:
C1/interpreter array-access profiling is racy and can publish the exact array klass before the flat/null-free flag is visible:
https://github.com/openjdk/valhalla/blob/f87d6d1cea3b578b458075a411f7aaa5dddb2cd9/src/hotspot/share/c1/c1_LIRGenerator.cpp#L1968-L1971

C2 reads profile information lock-free and can therefore, in rare cases, observe the inconsistencies reported in the JBS issue. For example, it can observe profiling for an array load with an `array_type` that is knowingly flat but does not have the `flat_array` property set (yet). The new regression test reliably reproduces such inconsistencies for array loads and stores.

Since a valid profiled flat array klass is stronger information than the flat flag, the fix is to prefer that information if available. With that fix, the asserts don't make much sense anymore, so I removed them.

Thanks,
Tobias

---------

---------
### Progress

### Issue
 * [JDK-8381373](https://bugs.openjdk.org/browse/JDK-8381373): [lworld] C2 failed assert(!flat || flat_array) failed: inconsistency (**Bug** - P4)

### Reviewers
 * [Quan Anh Mai](https://openjdk.org/census#qamai) (@merykitty - Committer) ⚠️ Review applies to [f693fc00](https://git.openjdk.org/valhalla/pull/2452/files/f693fc0013f800df702d85b712982de1e01871de)
 * [Dean Long](https://openjdk.org/census#dlong) (@dean-long - no project role) ⚠️ Review applies to [f693fc00](https://git.openjdk.org/valhalla/pull/2452/files/f693fc0013f800df702d85b712982de1e01871de)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2452/head:pull/2452` \
`$ git checkout pull/2452`

`$ git checkout pull/2452` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2452/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2452`

`$ git pr show -t 2452`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2452.diff">https://git.openjdk.org/valhalla/pull/2452.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2452#issuecomment-4488943374)
</details>

### DIFF
--- a/src/hotspot/share/ci/ciMethod.cpp
+++ b/src/hotspot/share/ci/ciMethod.cpp
@@ -666,7 +666,7 @@ bool ciMethod::parameter_profiled_type(int i, ciKlass*& type, ProfilePtrKind& pt
 
 // MDO updates are racy. C2 can observe the array type before the profiling code has updated the
 // corresponding flat/null-free flags. If the array type is known, prefer the properties it provides.
-static void sanitize_array_access_profile(ciKlass* array_type, bool& flat_array, bool& null_free_array) {
+static void update_flags_from_type(ciKlass* array_type, bool& flat_array, bool& null_free_array) {
   if (array_type != nullptr) {
     flat_array |= array_type->is_flat_array_klass();
     null_free_array |= array_type->as_array_klass()->is_elem_null_free();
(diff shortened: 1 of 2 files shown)


merykitty (approved):
Makes sense to me, thanks for investigating, this is a really tricky issue.

TobiHartmann:
Thanks for the quick review, Quan Anh!

src/hotspot/share/ci/ciMethod.cpp
```diff
@@ -664,6 +664,15 @@ bool ciMethod::parameter_profiled_type(int i, ciKlass*& type, ProfilePtrKind& pt
   return false;
 }
 
+// MDO updates are racy. C2 can observe the array type before the profiling code has updated the
+// corresponding flat/null-free flags. If the array type is known, prefer the properties it provides.
+static void sanitize_array_access_profile(ciKlass* array_type, bool& flat_array, bool& null_free_array) {
```
dean-long: Having "sanitize" in the name feels like it implies a level of confidence in the values that isn't really warranted.  We can miss flags updates as well as type updates, so the combined result is still best effort, therefore I would not call it sanitized.  Maybe something like "update_flags_from_type" is closer to the truth?

dean-long (approved):

src/hotspot/share/ci/ciMethod.cpp
TobiHartmann: I agree and renamed the method accordingly.

TobiHartmann:
Thanks for the review Dean!

PR #2464, opened 2026-05-21
8384756: [lworld] Exclude class from archive if class of inline field is excluded

matias9927:
Classes can be excluded from the CDS archive for a variety of different reasons, such as having old classfile versions or being stored in a signed jar. These exclusions can extend to other classes that depend on an excluded class such as supers or interfaces.

Flat and null-restricted fields have layouts that are calculated when a classfile is parsed and they rely on the klass of the field being consistent between dumptime and runtime, thus adding a new dependency. This patch adds inline flat and null-restricted fields as new dependencies that may lead to the holding class being excluded from the CDS archive. Verified with a new test and tier 1-5 tests.

---------

---------
### Progress

### Issue
 * [JDK-8384756](https://bugs.openjdk.org/browse/JDK-8384756): [lworld] Exclude class from archive if class of inline field is excluded (**Bug** - P2)

### Reviewers
 * [Ioi Lam](https://openjdk.org/census#iklam) (@iklam - Committer) ⚠️ Review applies to [e723f009](https://git.openjdk.org/valhalla/pull/2464/files/e723f009ff743f7f99c526a4fcc29a5b52c4fb53)

### Contributors

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2464/head:pull/2464` \
`$ git checkout pull/2464`

`$ git checkout pull/2464` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2464/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2464`

`$ git pr show -t 2464`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2464.diff">https://git.openjdk.org/valhalla/pull/2464.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2464#issuecomment-4520419293)
</details>


src/hotspot/share/classfile/systemDictionaryShared.cpp
```diff
@@ -297,6 +298,25 @@ class SystemDictionaryShared::ExclusionCheckCandidates
...
+    // the field holder.
+    if (k->has_inlined_fields()) {
+      for (AllFieldStream fs(k); !fs.done(); fs.next()) {
+        if (fs.access_flags().is_static()) continue;
+
+        if (fs.is_flat() || fs.is_null_free_inline_type()) {
+          Thread* current = Thread::current();
+          Handle loader(current, k->class_loader());
+          Symbol* field_klass_name = Signature::strip_envelope(fs.signature());
+          Klass* field_klass = SystemDictionary::find_instance_or_array_klass(current, field_klass_name, loader);
```
fparain: Why doing a system dictionary search for flat fields when a pointer is already directly available in the _inline_layout_info_array?

coleenp: if the field is a null_free_inline_type is it also available in the _inline_layout_info_array?
This search is pretty quick so would be good to have them done the same way.

matias9927: I was hoping the klasses were cached somewhere but I forgot about that array. Thanks for pointing it out!

fparain: null-restricted fields are not in the _inline_layout_info_array, but they should be rarer (@NullRestricted is not an officially supported annotation). We don't know yet what the future of null-restricted field will look like, so I would not add them to the _inline_layout_info-array for now.

fparain: Update: fields marked as null-restricted at CDS dump time are in fact in the `_inline_layout_info_array` because the class file parser fills the array with all known inline classes, and the @NullRestricted annotation is kept only for known value classes.

test/hotspot/jtreg/runtime/cds/appcds/test-classes/excluded/ExcludedValueClass.java
```diff
@@ -0,0 +1,40 @@
...
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ *
+ */
+
+package excluded;
+
+import jdk.internal.vm.annotation.LooselyConsistentValue;
+import jdk.internal.vm.annotation.NullRestricted;
```
iklam: Are these necessary

test/hotspot/jtreg/runtime/cds/appcds/InlineFieldExclusionTest.java
```diff
@@ -0,0 +1,76 @@
...
+ */
+
+/*
+ * @test
+ * @bug 8384756
+ * @summary Verify exclusion of classes if their types are also excluded
+ * @requires vm.cds
+ * @library /test/lib
+ * @enablePreview
+ * @modules java.base/jdk.internal.vm.annotation
```
iklam, line 31: Is this necessary?

test/hotspot/jtreg/runtime/cds/appcds/test-classes/InlineFieldExclusionApp.java
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ *
+ */
+
+import excluded.ExcludedValueClass;
+
+public class InlineFieldExclusionApp {
```
iklam, line 27: We usually just put this class inside InlineFieldExclusionTest.java to avoid having too many files.

matias9927: These classes are in difference packages so I wanted to make it clear that they will not be in the same JAR file.

fparain:
Here's a patch that addresses the issue with null-restricted static fields: it ensures that by the time the class is loaded, all static fields with a @NullRestricted annotation have been validated, and a pointer to the InlineKlass of their type has been set in the inline_layout_info_array.
[nr_patch.patch](https://github.com/user-attachments/files/28397621/nr_patch.patch)

src/hotspot/share/classfile/systemDictionaryShared.cpp
```diff
@@ -297,6 +298,20 @@ class SystemDictionaryShared::ExclusionCheckCandidates
         return true; // Keep iterating.
       });
     }
+
+    // Inline fields need to have their layouts preserved between dumptime and runtime.
+    // To ensure this, the types of the fields must be stored in the archive along with
+    // the field holder.
+    if (k->has_inlined_fields()) {
```
fparain: This condition is missing the case of the static null-restricted fields (same for test at line 540).
So far, we don't have a flag to indicate the presence of such fields.

src/hotspot/share/classfile/systemDictionaryShared.cpp
```diff
@@ -297,6 +298,20 @@ class SystemDictionaryShared::ExclusionCheckCandidates
...
+      for (AllFieldStream fs(k); !fs.done(); fs.next()) {
+        if (fs.access_flags().is_static()) continue;
```
DanHeidinga: Similar to Fred's comment this should be:
```
if (fieldinfo.access_flags().is_static() && !fieldinfo.field_flags().is_null_free_inline_type()) continue;
```

src/hotspot/share/classfile/systemDictionaryShared.cpp
```diff
@@ -539,6 +539,8 @@ bool SystemDictionaryShared::check_dependencies_exclusion(InstanceKlass* k, Dump
   // field layouts will be consistent at runtime.
   if (k->has_inlined_fields()) {
     for (AllFieldStream fs(k); !fs.done(); fs.next()) {
+      if (fs.access_flags().is_static() && !fs.field_flags().is_null_free_inline_type()) continue;
+
```
DanHeidinga: I think we can remove this check entirely.  It used to be a fast path to skip static fields but now we end up repeating a check we do below.  The fast path is now half the work of the normal path.

src/hotspot/share/classfile/systemDictionaryShared.cpp
```diff
@@ -304,7 +304,7 @@ class SystemDictionaryShared::ExclusionCheckCandidates
     // the field holder.
     if (k->has_inlined_fields()) {
       for (AllFieldStream fs(k); !fs.done(); fs.next()) {
-        if (fs.access_flags().is_static()) continue;
+        if (fs.access_flags().is_static() && !fs.field_flags().is_null_free_inline_type()) continue;
```
DanHeidinga: Same here. I think we're probably better off removing this clause and letting the general case handle it

test/hotspot/jtreg/runtime/cds/appcds/InlineFieldExclusionTest.java
```diff
@@ -0,0 +1,77 @@
...
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ *
+ */
+
+/*
+ * @test
+ * @bug 8384756
+ * @summary Verify exclusion of classes if their types are also excluded
```
iklam: How about:

```
@summary A class should be excluded if the type of one of its inlined fields is excluded.
```

src/hotspot/share/classfile/systemDictionaryShared.cpp
```diff
@@ -519,6 +532,20 @@ bool SystemDictionaryShared::check_dependencies_exclusion(InstanceKlass* k, Dump
...
   }
 
+  // If any of the null restricted or flat field types are excluded, the current
+  // klass must be excluded as well, otherwise there is no guarantee that the
+  // field layouts will be consistent at runtime.
+  if (k->has_inlined_fields()) {
+    for (AllFieldStream fs(k); !fs.done(); fs.next()) {
+      if (fs.is_flat() || fs.is_null_free_inline_type()) {
+        InlineKlass* field_klass = k->get_inline_type_field_klass(fs.index());
+        if (is_dependency_excluded(k, InstanceKlass::cast(field_klass), "inline field")) {
```
iklam: "inline field" should be "inline field type", so the error message looks like:

```
Skipping InlineFieldExclusionApp: inline field type ExcludedValueClass is excluded
```

iklam (commented):
CDS changes look good to me. Some nits.

iklam (approved):

src/hotspot/share/classfile/systemDictionaryShared.cpp
```diff
@@ -297,6 +298,18 @@ class SystemDictionaryShared::ExclusionCheckCandidates
         return true; // Keep iterating.
       });
     }
+
+    // Inline fields need to have their layouts preserved between dumptime and runtime.
+    // To ensure this, the types of the fields must be stored in the archive along with
+    // the field holder.
+    if (k->has_inlined_fields()) {
```
DanHeidinga: @fparain were you expecting to see an updated `if` statement with two conditions here?

```
if (k->has_inlined_fields() || k->has_null_restricted_static_fields()) {...}
```

src/hotspot/share/classfile/classFileParser.cpp
```diff
@@ -6366,11 +6386,20 @@ void ClassFileParser::post_process_parsed_stream(const ClassFileStream* const st
...
     Symbol* sig = fieldinfo.signature(cp);
     if (Signature::has_envelope(sig)) {
       TempNewSymbol name = Signature::strip_envelope(sig);
-      if (name == _class_name) continue;
+      if (name == _class_name) {
+        if (fieldinfo.field_flags().is_null_free_inline_type() && !is_inline_type()) {
+          fieldinfo.field_flags_addr()->update_null_free_inline_type(false);
+        } else {
+          // dummy setting to triggger the allocation of the inline_layout_info array
+          // the real pointer will be set later in ::fill_instance_klass, once the InlineKlass has been allocated
```
coleenp:
```suggestion
          // Dummy setting to trigger the allocation of the inline_layout_info array -
          // the real pointer will be set later in fill_instance_klass, once the InlineKlass has been allocated.
```

src/hotspot/share/classfile/classFileParser.cpp
```diff
@@ -6313,7 +6333,7 @@ void ClassFileParser::post_process_parsed_stream(const ClassFileStream* const st
   // If it turned out that we didn't inline any of the fields, we deallocate
   // the array of InlineLayoutInfo since it isn't needed, and so it isn't
   // transferred to the allocated InstanceKlass.
-  if (_inline_layout_info_array != nullptr && !_layout_info->_has_inlined_fields) {
+  if (_inline_layout_info_array != nullptr && !(_layout_info->_has_inlined_fields || _has_null_restricted_static_fields)) {
```
coleenp, line 6336:
```suggestion
  if (_inline_layout_info_array != nullptr && !_layout_info->_has_inlined_fields &&  !_has_null_restricted_static_fields)) {
```

Was there a reason to group the last two restrictions?

src/hotspot/share/classfile/classFileParser.cpp
```diff
@@ -5475,6 +5478,22 @@ void ClassFileParser::fill_instance_klass(InstanceKlass* ik,
...
 
+  if (_has_null_restricted_static_fields) {
+    for (int i = 0; i < _temp_field_info->length(); i++) {
+      FieldInfo& fieldinfo = _temp_field_info->at(i);
+      if (fieldinfo.access_flags().is_static() && fieldinfo.field_flags().is_null_free_inline_type()) {
+        Symbol* sig = fieldinfo.signature(_cp);
+        assert(Signature::has_envelope(sig), "Must already have been checked");
+        TempNewSymbol name = Signature::strip_envelope(sig);
+        if (name == _class_name) {
+          // replace the nullptr previously stored now that we have the InstanceKlass for this klass
```
coleenp:
```suggestion
          // Replace the nullptr previously stored now that we have the InstanceKlass for this klass.
```

coleenp (commented):
I haven't been following all of this bug fix but it looks good to me although I have minor comment suggestions.

src/hotspot/share/classfile/systemDictionaryShared.cpp
fparain: Yes, null restricted static fields must be in the candidate list along with the inlined fields.

src/hotspot/share/classfile/systemDictionaryShared.cpp
```diff
@@ -519,6 +532,20 @@ bool SystemDictionaryShared::check_dependencies_exclusion(InstanceKlass* k, Dump
     }
   }
 
+  // If any of the null restricted or flat field types are excluded, the current
+  // klass must be excluded as well, otherwise there is no guarantee that the
+  // field layouts will be consistent at runtime.
+  if (k->has_inlined_fields()) {
```
fparain: null-restricted static fields must be included too.

src/hotspot/share/classfile/classFileParser.cpp
matias9927: They are equivalent expressions that were suggested offline.

src/hotspot/share/classfile/classFileParser.hpp
```diff
@@ -576,6 +577,7 @@ class ClassFileParser {
   bool is_inline_type() const { return !_access_flags.is_identity_class() && !_access_flags.is_interface() && !_access_flags.is_abstract(); }
   bool is_identity_class() const { return _access_flags.is_identity_class(); }
   bool has_inlined_fields() const { return _layout_info->_has_inlined_fields; }
+  bool has_static_null_restricted_fields() const { return _has_null_restricted_static_fields; }
```
DanHeidinga: Not called and flipped placement of "static".  Can we remove this function?

fparain (approved):
Looks good to me.

DanHeidinga (approved):

matias9927:
Thanks for the reviews, comments, and discussions @coleenp @fparain @DanHeidinga and @iklam!
/integrate

PR #2468, opened 2026-05-22
8385159: [lworld] C2 parts of JDK-8350865 - Part. 1

marc-chevalier:
### `LibraryCallKit::inline_getArrayProperties`
https://github.com/openjdk/valhalla/blob/f09dea673b68e51ff3604d7c6f0f0561dd0f8c98/src/hotspot/share/opto/library_call.cpp#L4805-L4807

Done.

This one uncovered a bug, then I ran into something unrelated, but polluting testing. It looks good now. The logic is not quite simple, but it is exactly as it is described in the runtime version of it.

### `LibraryCallKit::inline_getArrayProperties`
https://github.com/openjdk/valhalla/blob/f09dea673b68e51ff3604d7c6f0f0561dd0f8c98/src/hotspot/share/opto/library_call.cpp#L4797-L4798

This one, we need to keep. `inline_getArrayProperties` and `flat_array_test` builds a graph like
```
CMoveI(Bool(FlatArrayCheck(mem, array_klass), cmp), 0, 1, BOOL)
```
which is fine for
https://github.com/openjdk/valhalla/blob/f09dea673b68e51ff3604d7c6f0f0561dd0f8c98/src/hotspot/share/opto/macro.cpp#L3047-L3049
and we take the else-branch. But if instead of `array_klass`, we had just an `array`, we would take the if-branch. And this one expects a different shape:
- it looks for the `BoolNode` (let it be `bol`) under the `FlatArrayCheckNode`
- it looks at outputs of `bol`, and for each, expects it to be an `IfNode`
- use the control of the `IfNode` to build more control flow (with the lock/unlock paths...)

Clearly, our graph has a `CMoveINode` under the `BoolNode`, so this crashes. It's not just a matter of `expand_flatarraycheck_node` being too strict, but rather the problem is that we need a control input for this expansion, and we don't have with the `CMoveINode`.

I can see two easy solutions:
- change `inline_getArrayProperties` to generate something more like
```
  Phi(
    Region(
        r,
        IfTrue(If(ctrl, Bool(FlatArrayCheck(mem, array_klass), cmp)) as iff),
        IfFalse(iff)
    ) as r,
    0,
    1
  )
```
  which would allow the expansion to work.
- not change anything.

I went with not changing anything because loading the class seems less bad (and might already happen, and thus be share, as it is used by some other operations) than making a more complicate structure, that is control-pinned.

### `LibraryCallKit::inline_getFieldMap`
https://github.com/openjdk/valhalla/blob/f09dea673b68e51ff3604d7c6f0f0561dd0f8c98/src/hotspot/share/opto/library_call.cpp#L3186-L3187

Yes, sir! Testing seems to be happy.

### `MulNode::Value`
https://github.com/openjdk/valhalla/blob/f09dea673b68e51ff3604d7c6f0f0561dd0f8c98/src/hotspot/share/opto/mulnode.cpp#L202-L213

There is no test that seems to be on fire without that. It's not clear to me why it'd be useful, so I removed...

### `TypeAryPtr::empty`
https://github.com/openjdk/valhalla/blob/f09dea673b68e51ff3604d7c6f0f0561dd0f8c98/src/hotspot/share/opto/type.cpp#L5734-L5737

Here, I went for fixing the underlying `TypeAry::empty`, as fundamentally, that's where the magic happens, and here too, if we have flat and non-flat, we should end-up with an empty type.

https://github.com/openjdk/valhalla/blob/f09dea673b68e51ff3604d7c6f0f0561dd0f8c98/src/hotspot/share/opto/type.cpp#L2589-L2591

Thanks to [Add unsigned bounds and known bits to TypeInt/Long](https://bugs.openjdk.org/browse/JDK-8315066), `TypeInt::empty` is constantly false, so `_size->empty()` is redundant. There is the question of how we end-up with something that is both flat and non-flat. If I remember well, it is because of an intersection of a speculative type (that can be flat), and a signature type (assumed non-flat), which must give an empty value. It would be nice if we had bottom-coalescence, that is if one of the component of a product of abstract value is bottom (Hotspot's top), we put everything to bottom (HS's top) since the concretization is empty anyway. That is a simple case of reduction. In our case, it would be nice if, when we see such clash at construction type, to simply go with setting `_elem` to something empty, and not test more than one thing/have a consistent view, whichever part we are looking at. But if we do that in the dual world, we would reduce to bottom (HS's top), something that is the dual of non-empty, and thus, losing information. The easiest fix would simply be to come back later, once we have a proper join, and we get rid of the dual. That will be trivial then.

Thanks,
Marc

---------

---------
### Progress

### Issue
 * [JDK-8385159](https://bugs.openjdk.org/browse/JDK-8385159): [lworld] C2 parts of JDK-8350865 - Part. 1 (**Sub-task** - P4)

### Reviewers
 * [Tobias Hartmann](https://openjdk.org/census#thartmann) (@TobiHartmann - Committer) ⚠️ Review applies to [d63f36f3](https://git.openjdk.org/valhalla/pull/2468/files/d63f36f3d42ff6fdfa53222878a5631a9c248976)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2468/head:pull/2468` \
`$ git checkout pull/2468`

`$ git checkout pull/2468` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2468/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2468`

`$ git pr show -t 2468`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2468.diff">https://git.openjdk.org/valhalla/pull/2468.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2468#issuecomment-4517316630)
</details>

### DIFF
--- a/src/hotspot/share/opto/type.cpp
+++ b/src/hotspot/share/opto/type.cpp
@@ -2588,7 +2588,7 @@ bool TypeAry::singleton(void) const {
 
 bool TypeAry::empty(void) const {
   assert(!_size->empty(), "TypeInt is never empty");
-  // TODO 8370914 This should be simplified at construction time once we get rid of dual
+  // TODO 8385426 This should be simplified at construction time once we get rid of dual
   // Doing it with the dual-based join is annoying. TypeAry::empty tests whether the
   // element type is empty. When computing the dual of an array that can be flat or not,
   // we will get an element type that is empty, and doesn't need more. We even shouldn't
--- a/src/hotspot/share/opto/type.cpp
+++ b/src/hotspot/share/opto/type.cpp
@@ -5709,6 +5709,14 @@ bool TypeAryPtr::empty(void) const {
   if (_ary->empty())       return true;
   // TODO 8350865 This should go to the meet implementation
   if (is_flat() && is_not_flat()) {
+#ifndef PRODUCT
+    stringStream ss;
+    ss.print("this: ");
+    dump_on(&ss);
+    assert(false, "something is wrong: %s", ss.base());
+#else
+    assert(false, "something is wrong");
+#endif
     return true;
   }
   return TypeOopPtr::empty();
--- a/src/hotspot/share/opto/type.cpp
+++ b/src/hotspot/share/opto/type.cpp
@@ -2587,7 +2587,16 @@ bool TypeAry::singleton(void) const {
 }
 
 bool TypeAry::empty(void) const {
-  return _elem->empty() || _size->empty();
+  assert(!_size->empty(), "TypeInt is never empty");
+  // TODO 8370914 This should be simplified at construction time once we get rid of dual
+  // Doing it with the dual-based join is annoying. TypeAry::empty tests whether the
+  // element type is empty. When computing the dual of an array that can be flat or not,
+  // we will get an element type that is empty, and doesn't need more. We even shouldn't
+  // do more otherwise, we can't make the dual involutive. But if we compute the
+  // intersection of a flat and a non-flat array, we could change the element type to an
+  // empty type to reduce the abstract value. And we must be careful not to do that in
+  // the dual world.
+  return _elem->empty() || (_flat && _not_flat);
 }
 
 //--------------------------ary_must_be_exact----------------------------------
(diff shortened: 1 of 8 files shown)


src/hotspot/share/opto/type.cpp
```diff
@@ -2587,7 +2587,16 @@ bool TypeAry::singleton(void) const {
 }
 
 bool TypeAry::empty(void) const {
-  return _elem->empty() || _size->empty();
+  assert(!_size->empty(), "TypeInt is never empty");
+  // TODO 8370914 This should be simplified at construction time once we get rid of dual
```
TobiHartmann: We might miss this TODO when fixing JDK-8370914, so maybe it's better to create a separate (Valhalla) bug for this. Also, I think this task is out of scope of JDK-8370914.

TobiHartmann (approved):
Thanks for working on this Marc! Looks good to me.

marc-chevalier:
/integrate

Thanks for your review!

PR #2474, opened 2026-05-24
8370769: [lworld] C2 is too aggressive in scalarizing large value objects

merykitty:
Hi,

Currently, C2 eagerly expands an oop leaf field of a value class if that leaf field has its type being a value class, this happens recursively, as long as there is no cycle. This eagerness may help the compiler have a better chance optimizing merging value types, as the `InlineTypeNode` is pushed down through the `Phi`s, allows C2 to fold the loads better. However, this expansion is unbounded, which means at any parsing or IGVN step, the compiler may emit an unbounded amount of nodes.

This PR removes this eagerness, that is, it makes the compiler not try expanding a value object subfield, unless the field is loaded. This removes the unbounded nature of value object expansion, it also makes the compiler emit less node, which alleviate the pressure on other optimizations in extreme cases.

In most case, this does not change the behaviour of C2, as value objects are aggressively flattened inside other value objects, unless their `LooselyConsistentValue` mismatch, which is not an issue for JEP 401. It also means that, we have no choice but stopping flattening at this level, because going even 1 level deeper may lead to issues in extreme cases, as a value object can have hundreds of leaf fields. This is demonstrated in the added regression test. `V1` is fully flattened, `V0` flattens only its first field, so if we also expand the leaf fields of `V0`, we will expand the object fully, leading to an explosion of node number.

There are mutliple test failures due to the compiler not able to fold through a load from a `Phi` of `InlineTypeNode`s. This will need fixing after JEP 401. I have tried keeping these IR verification, at least for some combinations of flags. Unfortunately, changing how `MyValue1` is flattened will affect many other tests. As a result, I plan to leave it to [JDK-8384979](https://bugs.openjdk.org/browse/JDK-8384979).

I encounted a tricky issue with load folding from a value object, which is exposed by this patch. That is, when we push a `CastPP` through an `InlineTypeNode`, we need to cast all of its oop field that is null-free to not null. This is because we inject the information that those fields are null-free into the loads, but the inputs of the `InlineTypeNode` can be `null`, because the `InlineTypeNode` itself is nullable. As a result, when the load looks through the `Cast` to the `InlineTypeNode`, it misses the dependency of the `Cast` that it just looked through, dropping the constraint that the loaded value must be not null. As a result, it can be scheduled before the null check, resulting in a crash. The fix is to unconditionally push a cast to not-null through its `InlineTypeNode` input, to add `CastPP` of null-free inputs to not-null, and to remove the ability of `see_through_inline_type` to look through `ConstraintCast`s.

Testing:

Please take a look and leave your review, thanks a lot.

---------

---------
### Progress

### Issue
 * [JDK-8370769](https://bugs.openjdk.org/browse/JDK-8370769): [lworld] C2 is too aggressive in scalarizing large value objects (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2474/head:pull/2474` \
`$ git checkout pull/2474`

`$ git checkout pull/2474` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2474/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2474`

`$ git pr show -t 2474`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2474.diff">https://git.openjdk.org/valhalla/pull/2474.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2474#issuecomment-4529403716)
</details>

### DIFF
--- a/src/hotspot/share/opto/castnode.cpp
+++ b/src/hotspot/share/opto/castnode.cpp
@@ -115,19 +115,9 @@ Node *ConstraintCastNode::Ideal(PhaseGVN *phase, bool can_reshape) {
     return this;
   }
 
-  // Push cast through InlineTypeNode but be careful because in dead parts of the graph,
-  // an InlineTypeNode can be casted to some completely unrelated type.
-  InlineTypeNode* vt = in(1)->isa_InlineType();
-  if (vt != nullptr && vt->is_allocated(phase) && _type->isa_instptr() != nullptr &&
-      vt->type()->inline_klass()->is_subtype_of(_type->is_instptr()->instance_klass())) {
-    Node* cast = clone();
-    cast->set_req(1, vt->get_oop());
-    vt = vt->clone()->as_InlineType();
-    if (!_type->maybe_null()) {
-      vt->as_InlineType()->set_null_marker(*phase);
-    }
-    vt->set_oop(*phase, phase->transform(cast));
-    return vt;
+  // Push cast through InlineTypeNode
+  if (in(1)->is_InlineType()) {
+    return ideal_inline_type_node(phase);
   }
 
   if (in(1) != nullptr && phase->type(in(1)) != Type::TOP) {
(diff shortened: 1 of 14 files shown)


test/hotspot/jtreg/compiler/valhalla/inlinetypes/TestLWorld.java
```diff
@@ -4914,9 +4915,9 @@ public void test174_verifier() {
         Asserts.assertFalse(test174(x, y));
         Asserts.assertFalse(test174(x, null));
         Asserts.assertFalse(test174(null, x));
-        Asserts.assertFalse(test174(x, z));
-        Asserts.assertFalse(test174(z, x));
-        Asserts.assertFalse(test174(z, new AllPrimitives(rI, 0)));
+        // Asserts.assertFalse(test174(x, z));
```
TobiHartmann, line 4918: Why is this needed? (Same below)

src/hotspot/share/opto/castnode.cpp
```diff
@@ -232,6 +222,49 @@ Node* ConstraintCastNode::pin_node_under_control_impl() const {
   return make_cast_for_type(in(0), in(1), bottom_type(), _dependency.with_pinned_dependency(), _extra_types);
 }
 
+Node* ConstraintCastNode::ideal_inline_type_node(PhaseGVN* phase) {
```
TobiHartmann: The name of this method is a bit misleading.

TobiHartmann (commented):
> it makes the compiler not try expanding a value object subfield, unless the field is loaded

Do you mean "unless the field is **flattened**"?

> It also means that, we have no choice but stopping flattening at this level, because going even 1 level deeper may lead to issues in extreme cases, as a value object can have hundreds of leaf fields

But since we know the layout(s) statically, couldn't we just set a limit/threshold on the number of fields that get scalarized or on the depth?

test/hotspot/jtreg/compiler/valhalla/inlinetypes/TestLWorld.java
merykitty: Because I changed the field from an `Integer` to an `int`, so there is no guarantee that `x` and `z` are different values anymore.

src/hotspot/share/opto/castnode.cpp
merykitty: Renamed it.

merykitty:
@TobiHartmann Thanks for your review.

> Do you mean "unless the field is flattened"?

Yes, that's right, my mistake.

> But since we know the layout(s) statically, couldn't we just set a limit/threshold on the number of fields that get scalarized or on the depth?

I think that's possible, but since it makes the code more complicated and it is not relevant for JEP401, I think we can leave it to lw401-post, what do you think?

test/hotspot/jtreg/compiler/valhalla/inlinetypes/TestLWorld.java
TobiHartmann: Ah, I missed that. Maybe add a `TODO` comment so we don't miss to re-enable these or add it to the `This should be Integer` comment.

TobiHartmann (commented):
> I think that's possible, but since it makes the code more complicated and it is not relevant for JEP401, I think we can leave it to lw401-post, what do you think?

Right, I think it's fine to leave this to follow-up enhancements. But why is it not relevant for JEP 401?

merykitty:
> Right, I think it's fine to leave this to follow-up enhancements. But why is it not relevant for JEP 401?

Because since there is no `LooselyConsistentValue` mismatch, all value objects are flattened if the holder is also a value object (apart from `Double` and `Long` I think), unless the layout builder sees that the holder size will become too large.

merykitty:
> Ah, I missed that. Maybe add a TODO comment so we don't miss to re-enable these or add it to the This should be Integer comment.

Done.

TobiHartmann (approved):
Thanks for the offline discussion! This looks good to me. Please make sure to also run testing in the higher tiers.

merykitty:
Thanks a lot for your review, I have run tests upto tier7.
/integrate

PR #2481, opened 2026-05-28
8384897: [lworld] this.staticField should be restricted in early construction context

mcimadamore:
This PR moves the early access checks back to `Resolve` from `Attr`.

This simplifies the code, as we no longer need a separate visitor to check for early references.

To do this, the code uses a simple record called `EarlyConstructorContext` which is used to keep track "facts" about where we are in attribution with respect to early initialization. This means tracking:
* whether we need to issue warnings or errors
* whether we are inside a lambda, or a nested class
* who is the owner of the early construction context

An instance of this class is stored in `AttrContext`, generalizing the current `ctorPrologue` field. `Attr` updates this field -- so, for instance, inside `visitLambda` a "nested" early context is derived and stored in `AttrContext`.

The most intricate bit has to do with how to handle stuff like `Foo.this.x`.  There are many complected issues here:
* when we see `Foo.this` we don't know if `x` is a method or a field (so we might report issues too early)
* when, later, we see `x` in Resolve, we don't know if the qualifier was `Foo.this` (so we might miss some issues).

To overcome these problems, some coupling between `Attr` and `Resolve` is unavoidable. In this PR, `Attr::visitSelect` checks whether the qualifier happens to be `this/super` or `Foo.this/super`. If so, it _unsets_ the early construction context. Effectively this means that `Resolve` will NOT report an error.

Then, in such cases, `Attr` will also pass the `selected` part to `selectSym` which will pass it to `Resolve::findIdentInType`. This means that, at this point, we can finally check `x`, but knowing that the qualifier was `Foo.this`.

Another improvement is that there's no longer duplication between `Resolve` and `Attr::checkAssignable`. This duplication is present in mainline, but the current lworld repo (thanks to the work @vicente-romero-oracle  did) removed that. I wanted to maintain that cleanup. To do that, now Attr passes the expected kind to some methods in `Resolve`. This kind is then used to tell whether we're inside an assignment or not.

This strategy doesn't require looking inside `env.tree`, so it naturally side-steps issues such as https://bugs.openjdk.org/browse/JDK-8385431.

I've also simplified the test harness to remove `super` from some of the tests (this harness is used to check the warning code path). It is now a simple annotation processor, so it can focus on just rewriting the tree and nothing else.

---------

---------
### Progress

### Issue
 * [JDK-8384897](https://bugs.openjdk.org/browse/JDK-8384897): [lworld] this.staticField should be restricted in early construction context (**Bug** - P3)

### Reviewers
 * [Chen Liang](https://openjdk.org/census#liach) (@liach - Committer) ⚠️ Review applies to [194998cb](https://git.openjdk.org/valhalla/pull/2481/files/194998cb355a2e8bfe8e1360013772fb39b1409b)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2481/head:pull/2481` \
`$ git checkout pull/2481`

`$ git checkout pull/2481` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2481/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2481`

`$ git pr show -t 2481`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2481.diff">https://git.openjdk.org/valhalla/pull/2481.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2481#issuecomment-4565183713)
</details>


src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Lower.java
```diff
@@ -142,8 +142,7 @@ protected Lower(Context context) {
         useMatchException = Feature.PATTERN_SWITCH.allowedInSource(source) &&
                             (preview.isEnabled() || !preview.isPreview(Feature.PATTERN_SWITCH));
         typePairToName = TypePairs.initialize(syms);
-        this.allowValueClasses = (!preview.isPreview(Feature.VALUE_CLASSES) || preview.isEnabled()) &&
-                Feature.VALUE_CLASSES.allowedInSource(source);
+        this.allowValueClasses = preview.isEnabled() && Feature.VALUE_CLASSES.allowedInSource(source);
```
mcimadamore, line 145: I've simplified all these checks as this is the more idiomatic style I think

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Resolve.java
```diff
@@ -423,50 +424,45 @@ public boolean isAccessible(Env<AttrContext> env, Type site, Symbol sym, boolean
...
-                                    &&
-                                    isAccessible(env, site, checkInner)
-                                    &&
-                                    notOverriddenIn(site, sym);
-                default: // this case includes erroneous combinations as well
-                    return isAccessible(env, site, checkInner) && notOverriddenIn(site, sym);
-            }
-        } finally {
-            env.enclClass.sym = enclosingCsym;
+        switch ((short)(sym.flags() & AccessFlags)) {
```
mcimadamore, line 427: I believe this change was unrelated, so I've reverted it

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/AttrContext.java
```diff
@@ -155,7 +153,6 @@ AttrContext dup(WriteableScope scope) {
         info.preferredTreeForDiagnostics = preferredTreeForDiagnostics;
         info.visitingServiceImplementation = visitingServiceImplementation;
         info.allowProtectedAccess = allowProtectedAccess;
-        info.instanceInitializerBlock = instanceInitializerBlock;
```
mcimadamore, line 158: I believe this change was unrelated, so I've reverted it

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java
```diff
@@ -5928,7 +5583,7 @@ void attribClass(ClassSymbol c) throws CompletionFailure {
                             log.error(TreeInfo.diagnosticPositionFor(c, env.tree), Errors.NonSealedWithNoSealedSupertype(c));
                         }
                     }
-                } else if ((c.flags_field & Flags.COMPOUND) == 0) {
+                } else {
```
mcimadamore, line 5586: I believe this change was unrelated, so I've reverted it

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java
```diff
@@ -1829,11 +1455,7 @@ public void visitBlock(JCBlock tree) {
             final Env<AttrContext> localEnv =
                 env.dup(tree, env.info.dup(env.info.scope.dupUnshared(fakeOwner)));
 
-            if ((tree.flags & STATIC) != 0) {
-                localEnv.info.staticLevel++;
-            } else {
-                localEnv.info.instanceInitializerBlock = true;
-            }
+            if ((tree.flags & STATIC) != 0) localEnv.info.staticLevel++;
```
mcimadamore, line 1458: I believe this change was unrelated, so I've reverted it

test/micro/org/openjdk/bench/jdk/classfile/Transforms.java
```diff
@@ -134,7 +134,8 @@ public enum NoOpTransform {
                     shared
                     ? options
                     : Stream.concat(Stream.of(options), Stream.of(ClassFile.ConstantPoolSharingOption.NEW_POOL)).toArray(ClassFile.Option[]::new));
-            this.transform = bytes -> cc.transformClass(cc.parse(bytes), classTransform);
+            ClassFile tempcc = cc;
```
mcimadamore: This emitted a warning now, so I had to update it (to keep the build/test from failing)

test/langtools/tools/javac/SuperInit/EarlyLocalClass.out
```diff
@@ -1,2 +1,2 @@
-EarlyLocalClass.java:13:37: compiler.err.cant.ref.before.ctor.called: hashCode()
+EarlyLocalClass.java:17:32: compiler.err.cant.ref.before.ctor.called: this
```
mcimadamore, line 1: Note -- in most cases, the differences in golden outputs are caused by two factors:
* when reverting the code to what master had, if we have `this.m()` we now report an issue on `this`, not `m()`
* there are line differences due to the fact that test headers had to be updated to work with the new harness
* there's an extra trailing `-Werror` issue reported, because we're no longer using the harness to do the comparison

test/langtools/tools/javac/SuperInit/SuperInitFails.out
```diff
@@ -1,43 +1,46 @@
...
+SuperInitFails.java:138:9: compiler.err.non.canonical.constructor.invoke.another.constructor: SuperInitFails.Record1
+SuperInitFails.java:143:9: compiler.err.non.canonical.constructor.invoke.another.constructor: SuperInitFails.Record2
+SuperInitFails.java:160:50: compiler.err.cant.ref.before.ctor.called: x
+SuperInitFails.java:164:15: compiler.err.cant.ref.before.ctor.called: this
+SuperInitFails.java:173:13: compiler.err.cant.ref.before.ctor.called: x
+SuperInitFails.java:177:17: compiler.err.cant.ref.before.ctor.called: x
+SuperInitFails.java:181:24: compiler.err.cant.ref.before.ctor.called: x
+SuperInitFails.java:185:18: compiler.err.cant.ref.before.ctor.called: x
+SuperInitFails.java:191:32: compiler.err.cant.ref.before.ctor.called: x
+SuperInitFails.java:200:36: compiler.err.cant.ref.before.ctor.called: xx
```
mcimadamore, line 21: This error is new -- but legitimate, as access occurred inside a class nested inside the early constructor context

test/langtools/tools/javac/DefiniteAssignment/DA_DUConstructors.java
```diff
@@ -16,12 +16,13 @@ class C1 {
...
+    class C2_Base {
+        C2_Base(int i) {}
+    }
+    class C2 extends C2_Base {
         final int x;
         C2() {
-            this(x = 3); // error
-        }
-        C2(int i) {
+            super(x = 3); // error
```
mcimadamore, line 25: We needed to change the test here, as early reference from this(...) is no longer allowed. This causes a failure in `Attr` which then prevents `Flow` to run -- which is what this test wanted to check

test/langtools/tools/javac/SuperInit/SuperInitFailsWarnings.out
```diff
@@ -1,38 +1,41 @@
...
+SuperInitFails.java:124:22: compiler.err.call.must.only.appear.in.ctor
+SuperInitFails.java:138:9: compiler.err.non.canonical.constructor.invoke.another.constructor: SuperInitFails.Record1
+SuperInitFails.java:143:9: compiler.err.non.canonical.constructor.invoke.another.constructor: SuperInitFails.Record2
+SuperInitFails.java:160:50: compiler.err.cant.ref.before.ctor.called: x
+SuperInitFails.java:164:15: compiler.err.cant.ref.before.ctor.called: this
+SuperInitFails.java:200:36: compiler.err.cant.ref.before.ctor.called: xx
+SuperInitFails.java:216:33: compiler.err.ctor.calls.not.allowed.here
+SuperInitFails.java:256:13: compiler.err.cant.ref.before.ctor.called: x
+SuperInitFails.java:269:18: compiler.err.cant.ref.before.ctor.called: x
+SuperInitFails.java:296:20: compiler.err.cant.ref.before.ctor.called: this
```
mcimadamore, line 12: This used to be:

```
SuperInitFails.java:293:25: compiler.warn.would.not.be.allowed.in.prologue: testMethod()
```

But now it turned to error because the constructor with this call had a `this()` constructor call, not `super`. The updated test annotation processor correctly does _not_ remove the `this(...)` which means this stays an error.

liach (approved):
This definitely is more maintainable than that visitor in the long run. Peeked through and saw no obvious problem, but I am no professional.

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Resolve.java
```diff
@@ -3933,6 +3962,117 @@ private List<Type> pruneInterfaces(Type t) {
...
+        }
+        if (field.isStatic()) {
+            // early this can only qualify instance field accesses
+            return false;
+        }
+        return isSimpleEarlyFieldRefAllowed(pos, env, context, field, writeOnlyTarget);
+    }
+
+    /** Implements early access checks for unqualified field references (6.5.6.1) */
+    private boolean isSimpleEarlyFieldRefAllowed(DiagnosticPosition pos,
```
mcimadamore, line 4016: There's more comments than code here -- but what I found when working with this code (and the one it replaces in master) is that it's very easy to forget as to _why_ a check was there, so I thought it would be better to capture. If we think it's too much I can compress

vicente-romero-oracle (approved):
looks great to me

PR #2487, opened 2026-05-29
8385629: [lworld] UnsetFieldInfo should be replaced by bit sets in Code.State

mcimadamore:
This PR cleans the handling of unset strict fields which is needed for stackmap generation.
In the current impl, this is done with UnsetFieldsInfo which is prepared by Flow, and then propagated to Gen/Code.
The problem with this approach is that we introduce unnecessary coupling, as we need to keep track where, at the AST level, an assignment to a strict uninit variable takes place.
This leads to issues as the AST-oriented view is too high-level for the kind of manipulation that Code needs.
By modelling unset fields as a bitset in Code.State (same as what we do to keep track of local variable ranges), state merges are dealt with more naturally and succinctly.

---------

---------
### Progress

### Issue
 * [JDK-8385629](https://bugs.openjdk.org/browse/JDK-8385629): [lworld] UnsetFieldInfo should be replaced by bit sets in Code.State (**Enhancement** - P3)

### Reviewers
 * [Vicente Romero](https://openjdk.org/census#vromero) (@vicente-romero-oracle - Committer) ⚠️ Review applies to [4796d864](https://git.openjdk.org/valhalla/pull/2487/files/4796d864ddd35713d55a9706999d85b7db1f9a10)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2487/head:pull/2487` \
`$ git checkout pull/2487`

`$ git checkout pull/2487` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2487/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2487`

`$ git pr show -t 2487`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2487.diff">https://git.openjdk.org/valhalla/pull/2487.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2487#issuecomment-4573156326)
</details>


src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/Code.java
```diff
@@ -236,6 +231,16 @@ public Code(MethodSymbol meth,
         this.generateEarlyLarvalFrame = generateEarlyLarvalFrame;
     }
 
+    void initUnsetStrictFields(ClassSymbol csym) {
+        int strictFieldOffset = 0;
```
mcimadamore, line 235: Here we collect all strict instance fields in the class scope. These symbols are normally returned in reversed order, so using `prepend` restores natural source order.

Also, we override `adr` -- which is ok, because we only need that for local variables (and strict fields are... fields).

src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/Code.java
```diff
@@ -1057,7 +1062,13 @@ public void emitop2(int op, int od, PoolConstant data) {
             markDead();
             break;
         case putfield:
-            state.pop(((Symbol)data).erasure(types));
+            VarSymbol field = (VarSymbol)data;
+            state.pop(field.erasure(types));
+            if (field.isStrictInstance() &&
+                    field.owner == meth.owner &&
+                    state.peek().hasTag(UNINITIALIZED_THIS)) {
+                state.unsetStrict.excl(field.adr);
```
mcimadamore, line 1070: This is where all the magic happens -- if we see a putfield targeting a strict instance field of this class and the "receiver" is uninitialized, we transition that field to "init".

src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/Code.java
```diff
@@ -1837,6 +1846,7 @@ void markInitialized(UninitializedType old) {
 
         State join(State other) {
             defined.andSet(other.defined);
+            unsetStrict.orSet(other.unsetStrict);
```
mcimadamore, line 1849: This is important: unset field in a merged frame is obtaining by OR-ing the unset state of the parent frames.

src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/Code.java
vicente-romero-oracle: clever

src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/Code.java
vicente-romero-oracle: nice

vicente-romero-oracle (approved):
lgtm

src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/Code.java
liach: This actually much resemble the stack map generation of classfile api :+1:

PR #2488, opened 2026-05-29
8385634: [lworld] Remove unnecessary coupling between Lower, Gen and LocalProxyVarsGen

mcimadamore:
This PR removes some unnecessary coupling between Lower, Gen and LocalProxyVarsGen.

It does so by making LocalProxyVarsGen no longer a standalone pass but, rather, a helper for Gen::normalizeMethod.

The main idea is that we can make things more regular, by having Gen always inserting variable initializer in the correct place.
Then, LocalProxyVarGen will create a blank proxy, and use its visitor to inspect the rest of the constructor body.
Since the visitor _already_ rewires assignments to real fields as assignments to proxies, this new arrangement has the desired effect of generating the same code as before, but w/o too much coupling.

Some massaging to `Gen::normalizeMethod` was needed because now we need to make sure it calls the proxy step for all constructors, not just in case there's some pending var initializers.

Finally, when cleaning up `Lower` I noticed a likely bug: `freevardefs` was no longer preserving the `LOCAL_CAPTURE_FIELD` -- sometimes it was replacing it with `STRICT`. But `LOCAL_CAPTURE_FIELD` is used by LambdaToMethod, so changing this probably results in bad downstream lowering.
I've fixed this by adding both `STRICT` _and_ `LOCAL_CAPTURE_FIELD` to the captured sym.

---------

---------
### Progress

### Issue
 * [JDK-8385634](https://bugs.openjdk.org/browse/JDK-8385634): [lworld] Remove unnecessary coupling between Lower, Gen and LocalProxyVarsGen (**Enhancement** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2488/head:pull/2488` \
`$ git checkout pull/2488`

`$ git checkout pull/2488` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2488/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2488`

`$ git pr show -t 2488`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2488.diff">https://git.openjdk.org/valhalla/pull/2488.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2488#issuecomment-4575397914)
</details>


mcimadamore:
I've pushed a new revision that deals with cases where we have early field initializers that refer to `Lower`-generated synthetic fields. Example:

```
class Test {
              class Inner { }
              value class V {
                  Object o = new Inner(); // this$0 ref
                  V() { super(); }
              }
}
```

Here, the field initializer needs `this$0`, but since that initializer is being "copied" into the constructor prologue, that `this$0` becomes an "early read". I've tried several approaches, and almost everything failed. What worked was to resurrect the initializer visitor @vicente-romero-oracle had, and generalize it a bit, to also deal with local captures.

mcimadamore:
Alternatives considered:
* move `normalizeMethod` earlier. On paper this sounds clean, but it's not so straightforward. If the initializer contains nested classes or lambdas there's a real risk to generate redundant artifacts, as e.g. Lower might attempt to generate the same inner class from different constructors
* Mark access to `this$0` and friends from early initializers as "early access". In principle this works but (a) we need to find some way so that an early access in a field initializer is an early access _in all constructors_ at once (but we already have something like that in `Attr`/`Resolve`). And, we need to be mindful of the order in which things appear in the normalized constructor -- if early initializers are added at the beginning of the constructors, no amount of local proxy patching can save us.

mcimadamore:
> * In principle this works but (a) we need to find some way so that an early access in a field initializer is an early access _in all constructors_ at once (but we already have something like that in `Attr`/`Resolve`)

Spoke too fast -- early access from field initializer is NOT proxied correctly. This compiles, but then fails with verifier error (this is not a regression, apparently this never worked):

```
class EarlyFieldInitRead {
      static int seed() { return 42; }

      static value class V {
          int x = seed();
          int y = x;
          V() { super(); }
      }

          new V();
      }
  }
```

The issue is that when we record early reads in `Attr`/`Resolve` we use `env.enclMethod` -- which if we're in a field is not something meaningful. As a result, `LocalProxyVarsGen` doesn't pick that up, and doesn't proxy the access.

A simple solution could be to associate early reads from field initializers to the enclosing class.

mcimadamore:
I've tightened the code to make sure `Gen` doesn't rewrite more than it should:
* I've added a new explicit flag for the`this$0` field (`OUTER_THIS_FIELD`)
* This means that the rewriter can check whether the sym being accessed is either `OUTER_THIS_FIELD` or `CAPTURED_LOCAL_FIELD`
* If there's a match we look into `md.sym.extraParams` or `md.sym.capturredLocals` for the symbol to replace
* additionally, `Gen::rewriteEarlyInitializersIfNeeded` doesn't do anything if the above two lists of synthetic params are empty

mcimadamore:
Found another issue in cases like this:

```
class Outer {
      class Inner {}

      value class V {
          Supplier<Object> s = () -> new Inner();
          V() { super(); }
      }
  }
 ```
Here we will try to emit an instance lambda method that captures `this` (and accesses `this.this$0` in its body). This won't fly if we're in early construction context.

To workaround this, when we determine lambda captures, if we see `this$0` reference and we're inside an early field initializer, we capture `this$0` directly (instead of setting `captureThis = true`). This will make the generated lambda method static, and will means that the client will pass `this$0` explicitly as a capture param. Then `Gen` will see that and patch access with access to the synthetic ctor param.

Note: rewriting the lambda as a static lambda works here because the lambda should have no direct dependencies on `this` or `super` (as those are forbidden by language rules). What's left can be dealt with in the context of a static lambda with some extra capture params.

vicente-romero-oracle (approved):
lgtm

PR #2489, opened 2026-05-29
8385328: [lworld] C2: assert(phi->_idx >= nodes_size()) failed: only new Phi per instance memory slice still happens after JDK-8374742

rwestrel:
The bug is caused by a mismatch in the way `LoadFlat` and `StoreFlat`
are handled in `ConnectionGraph::find_inst_mem()` and
`MemNode::optimize_simple_memory_chain()`. Once EA has run a first
time, found some value array allocations as non escaping, and removed
their `LoadFlat`/`StoreFlat`, igvn runs and is able to further
transform the memory subgraph which introduces inconsistencies that
are then caught the next time EA runs.

The fix tries to make sure `ConnectionGraph::find_inst_mem()` and
`MemNode::optimize_simple_memory_chain()` handle `LoadFlat` and
`StoreFlat` similarly.

---------

---------
### Progress

### Issue
 * [JDK-8385328](https://bugs.openjdk.org/browse/JDK-8385328): [lworld] C2: assert(phi-&gt;_idx &gt;= nodes_size()) failed: only new Phi per instance memory slice still happens after JDK-8374742 (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2489/head:pull/2489` \
`$ git checkout pull/2489`

`$ git checkout pull/2489` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2489/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2489`

`$ git pr show -t 2489`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2489.diff">https://git.openjdk.org/valhalla/pull/2489.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2489#issuecomment-4576191888)
</details>


src/hotspot/share/opto/escape.cpp
```diff
@@ -4610,6 +4610,12 @@ Node* ConnectionGraph::find_inst_mem(Node *orig_mem, int alias_idx, GrowableArra
         result = proj_in->in(TypeFunc::Memory);
       } else if (proj_in->is_LoadFlat()) {
         result = proj_in->in(TypeFunc::Memory);
+      } else if (proj_in->is_StoreFlat()) {
+        Node* base = proj_in->as_StoreFlat()->base();
+        const TypeOopPtr* t_store_base =igvn->type(base)->is_oopptr();
+        if (t_store_base->instance_id() != toop->instance_id()) {
```
dean-long: A comment explaining what is happening here would be helpful.  Why does StoreFlat need this check but LoadFlat does not?

TobiHartmann (approved):
Looks good to me.

src/hotspot/share/opto/escape.cpp
rwestrel: I added some comments. Do they look ok to you?

src/hotspot/share/opto/memnode.cpp
```diff
@@ -321,10 +321,10 @@ Node* MemNode::optimize_simple_memory_chain(Node* mchain, const TypeOopPtr* t_oo
         }
         result = proj_in->in(TypeFunc::Memory);
       } else if (proj_in->is_LoadFlat() || proj_in->is_StoreFlat()) {
-        if (is_strict_final_load) {
-          // LoadFlat and StoreFlat cannot happen to strict final fields
-          result = proj_in->in(TypeFunc::Memory);
-        }
+        assert(is_strict_final_load || is_known_instance, "tested above");
+        // LoadFlat and StoreFlat cannot happen to strict final fields
+        // LoadFlat and StoreFlat to known instances are removed at the end of EA: this one is unrelated
```
dean-long: I asked a particular LLM about this, and it thinks there could be a problem here because a C2_MISMATCHED StoreFlat is not removed, and EA only marks the fields as global escape and not the based object.  Do you agree?

src/hotspot/share/opto/escape.cpp
```diff
@@ -4609,7 +4609,21 @@ Node* ConnectionGraph::find_inst_mem(Node *orig_mem, int alias_idx, GrowableArra
...
+        // - or it is a LoadFlat for some object unrelated to alias_idx
+        // In both cases, it's safe to assume this LoadFlat doesn't modify the memory for alias_idx
         result = proj_in->in(TypeFunc::Memory);
+      } else if (proj_in->is_StoreFlat()) {
+        // Either:
+        // - this is a StoreFlat for alias_idx's non escaping allocation that does modify the memory state for alias_idx
+        // - or it is a StoreFlat for some object unrelated to alias_idx that can't modify the memory state for alias_idx
+        Node* base = proj_in->as_StoreFlat()->base();
+        const TypeOopPtr* t_store_base =igvn->type(base)->is_oopptr();
+        if (t_store_base->instance_id() != toop->instance_id()) {
```
dean-long: Is it impossible to encounter InstanceBot here?  This compare is only valid if both sides are known instances, right?

src/hotspot/share/opto/escape.cpp
dean-long: Yes, the comments help, thanks!

src/hotspot/share/opto/memnode.cpp
rwestrel: A `StoreFlat` is not a regular `Store` so it can't be `C2_MISMATCHED`.

src/hotspot/share/opto/escape.cpp
rwestrel: It can nbe `InstanceBot` on the left hand side. AFAIU, `InstanceBot` doesn't mean any instance, it means any instance other than the ones that are known. So comparing `InstanceBot` with a known instance returns false as expected.
This said, one problem with this logic is that it gets types from igvn and they may not have propagated yet. I reworked it to not use the types from igvn.

src/hotspot/share/opto/escape.cpp
```diff
@@ -4532,6 +4532,41 @@ void ConnectionGraph::move_inst_mem(Node* n, Unique_Node_List& orig_phis) {
...
+    return true;
+  }
+  assert(ptn->is_LocalVar(), "sanity");
+  for (EdgeIterator i(ptn); i.has_next(); i.next()) {
+    Node* jobj_base = get_map(i.get()->idx());
+    if (jobj_base != nullptr && _igvn->type(jobj_base)->is_oopptr()->instance_id() == toop->instance_id()) {
+      return true;
+    }
+  }
+  assert(t_store_base->instance_id() != toop->instance_id(), "should not alias");
```
dean-long: I'm not an EA expert, but I'm still not convinced that we can compare instance_id() her directly here without checking for is_known_instance() or unknown edge pointing to `phantom_obj`.  Also, I think such a test would be better as TypeOopPtr helper method.  (It is a pet peeve of mine that we expose the implementation detail of how instance IDs and alias IDs are integers.)
Maybe an EA expert like @vnkozlov could check this too.

src/hotspot/share/opto/memnode.cpp
dean-long: If that's the case then we can add assert to check that in the StoreFlatNode ctor, and remove the check for C2_MISMATCHED at the beginning of StoreFlatNode::expand_non_atomic.

rwestrel: Right. I missed that. It seems `C2_MISMATCHED` accesses come from unsafe.
Not sure why `C2_MISMATCHED` would be a problem with `StoreFlat` but for `LoadFlat`, I suppose it's safer to not step over a `C2_MISMATCHED` `LoadFlat` for the same allocation as `alias_idx` as that one is not going away.

src/hotspot/share/opto/escape.cpp
```diff
@@ -4532,6 +4532,41 @@ void ConnectionGraph::move_inst_mem(Node* n, Unique_Node_List& orig_phis) {
 // is the specified alias index.
 //
 #define FIND_INST_MEM_RECURSION_DEPTH_LIMIT 1000
+
+bool ConnectionGraph::flat_access_aliases_with(Node* flat_access, const TypeOopPtr *toop) {
```
vnkozlov: Add comment about what this method does.

src/hotspot/share/opto/escape.cpp
```diff
@@ -4532,6 +4532,41 @@ void ConnectionGraph::move_inst_mem(Node* n, Unique_Node_List& orig_phis) {
...
+  Node* base = flat_access->is_StoreFlat() ? flat_access->as_StoreFlat()->base() : flat_access->as_LoadFlat()->base();
```
vnkozlov, line 4555: What type of `base` node could be here?

src/hotspot/share/opto/escape.cpp
```diff
@@ -4609,7 +4644,23 @@ Node* ConnectionGraph::find_inst_mem(Node* orig_mem, int alias_idx, Unique_Node_
         }
         result = proj_in->in(TypeFunc::Memory);
       } else if (proj_in->is_LoadFlat()) {
-        result = proj_in->in(TypeFunc::Memory);
+        // Either:
+        // - this is a LoadFlat for alias_idx's non escaping allocation: it will get removed by
+        // ConnectionGraph::optimize_flat_accesses() and has no effect on the memory state
+        // - or it is a LoadFlat for some object unrelated to alias_idx
+        // In both cases, it's safe to assume this LoadFlat doesn't modify the memory for alias_idx
```
vnkozlov: Which flat load can modify memory?  I surprise that this node has Memory projection.

src/hotspot/share/opto/escape.cpp
```diff
@@ -4609,7 +4644,23 @@ Node* ConnectionGraph::find_inst_mem(Node* orig_mem, int alias_idx, Unique_Node_
...
+        // If the LoadFlat is mismatched, it's not removed so don't step over it if it's a flat access to the alias_idx
+        // known instance
+        if (!proj_in->as_LoadFlat()->is_mismatched() || !flat_access_aliases_with(proj_in, toop)) {
```
vnkozlov, line 4675: When it can be `mismatched`? Your whole comment above just confuse me.

What I see is in normal case (not mismatched) we always step over loads regardless memory slice it is on. 
In mismatching case we skip if it is on different memory slice. That is it.

src/hotspot/share/opto/escape.cpp
```diff
@@ -4532,6 +4532,41 @@ void ConnectionGraph::move_inst_mem(Node* n, Unique_Node_List& orig_phis) {
...
+    return false;
+  }
+  PointsToNode* ptn = ptnode_adr(idx);
+  if (ptn == nullptr) {
+    return false;
+  }
+  PointsToNode::EscapeState es = ptn->escape_state();
+  if (es >= PointsToNode::GlobalEscape) {
+    return false;
+  }
```
vnkozlov, line 4567: Why not check here?  Asserts will not be needed then.
```
  const TypeOopPtr* t_base = _igvn->type(base)->is_oopptr();
  if (t_base->instance_id() == toop->instance_id()) {
    return true;
  }
```

src/hotspot/share/opto/escape.cpp
```diff
@@ -4532,6 +4532,41 @@ void ConnectionGraph::move_inst_mem(Node* n, Unique_Node_List& orig_phis) {
 // is the specified alias index.
 //
 #define FIND_INST_MEM_RECURSION_DEPTH_LIMIT 1000
+
+bool ConnectionGraph::flat_access_aliases_with(Node* flat_access, const TypeOopPtr *toop) {
+  Node* base = flat_access->is_StoreFlat() ? flat_access->as_StoreFlat()->base() : flat_access->as_LoadFlat()->base();
+  const TypeOopPtr* t_base = _igvn->type(base)->is_oopptr();
```
vnkozlov: This statement is too early - far before uses.

vnkozlov (commented):
Few comments/questions.

src/hotspot/share/opto/escape.cpp
rwestrel: `base` can be anything that's a pointer (at offset 0) but not an `AddP` AFAICT

src/hotspot/share/opto/escape.cpp
rwestrel: Whey they are expanded they need memory barriers which is why I think they have a memory projection. If the `LoadFlat` is from a non escaping object, then it doesn't need memory barriers when it's expanded AFAIU.

src/hotspot/share/opto/escape.cpp
rwestrel: It's `mismatched` when it comes from an unsafe call, for instance, if a flat field is read with unsafe at an offset that's not a constant so we don't really know what field is being read.

src/hotspot/share/opto/escape.cpp
rwestrel: That method should return the same as:
```
t_base->instance_id() == toop->instance_id()
```

The problem is that igvn hasn't run so types haven't propagated and `t_base` has a good chance to be a value that's not up to date and can't be relied on.

So yes, what you suggest would work but if feels safer to me to checks that the types and the result of the method are consistent.

rwestrel:
> Few comments/questions.

Thanks for looking at this. I pushed a small update.

rwestrel:
I updated the change again so that during igvn, if a `LoadFlat` or `StoreFlat` that is mismatched is encountered then the memory chain is not optimized further. Given those are not eliminated during EA, they could be for the `t_oop` instance.

src/hotspot/share/opto/escape.cpp
```diff
@@ -4532,6 +4532,42 @@ void ConnectionGraph::move_inst_mem(Node* n, Unique_Node_List& orig_phis) {
...
+  if (ptn == nullptr) {
+    return false;
+  }
+  PointsToNode::EscapeState es = ptn->escape_state();
+  if (es >= PointsToNode::GlobalEscape) {
+    return false;
+  }
+  if (ptn->is_JavaObject()) {
+    Node* jobj_base = get_map(ptn->idx());
+    if (jobj_base == nullptr || _igvn->type(jobj_base)->is_oopptr()->instance_id() != toop->instance_id()) {
```
dean-long: I still don't like the raw instance_id() comparisons.   Could we have something like _igvn->type(jobj_base)->is_oopptr()->same_instance(toop) based on something like meet_instance_id() or that asserts both are is_known_instance()?

rwestrel: I added a helper method. Does that look ok to you?

src/hotspot/share/opto/type.hpp
```diff
@@ -1547,6 +1547,7 @@ class TypeOopPtr : public TypePtr {
   bool is_known_instance()       const { return _instance_id > 0; }
   int  instance_id()             const { return _instance_id; }
   bool is_known_instance_field() const { return is_known_instance() && _offset.get() >= 0; }
+  bool same_instance_as(const TypeOopPtr* t) const { return instance_id() == t->instance_id(); }
```
dean-long: Can't we assert is_known_instance() here?  If that only makes sense for EA code, then this could be moved to a static helper in escape.cpp.  My main concern is comparing two instance IDs if either side can be <= 0.  Other code doing a similar comparison checks is_known_instance() first.  To make this general purpose, I think it would need to use something like
```
meet_instance_id(t->instance_id()) == t->instance_id()
```

dcubed-ojdk:
@rwestrel - Please note that the ProblemListing for tools/javac/platform/CanHandleClassFilesTest.java
has been widened to include all platforms.

I have not ProblemListed the BigApps tests, but we are seeing noise there also,
especially with applications/renaissance/RenaissanceStressTest.java.

Is there an ETA for this fix to land in repo-valhalla?

src/hotspot/share/opto/escape.cpp
```diff
@@ -4546,6 +4546,42 @@ void ConnectionGraph::move_inst_mem(Node* n, Unique_Node_List& orig_phis) {
 // is the specified alias index.
 //
 #define FIND_INST_MEM_RECURSION_DEPTH_LIMIT 1000
+
```
dean-long, line 4551: I suggest a helper function that adds asserts, something like:
```suggestion
static bool same_instance_as(const TypeOopPtr* t1, const TypeOopPtr* t2) {
  assert(t1->is_known_instance(), "known instance expected during EA phase");
  assert(t2->is_known_instance(), "known instance expected during EA phase");
  return t1->instance_id() == t2->instance_id();
}
```
and using it below.

rwestrel: It would have to be:
```
assert(t1->is_known_instance() || t2->is_known_instance(), "known instance expected during EA phase");
```
There's no guarantee both of them are known instances and that code is correct if only one is as, I think, we discussed before. Would you be ok with this?

dean-long: I'm OK with this, but I think there should be a comment that explains why it is safe to do the comparison.  It works because EA would not assign an instance id if it was possible for it to alias with any other type, right?

rwestrel: I added a comment. Does it look ok to you?

dean-long: Yes, thanks.

dean-long (approved):

dcubed-ojdk:
@rwestrel  - Is there an ETA for this fix to land in repo-valhalla?

dcubed-ojdk:
I'm doing some Mach5 testing with the v08 version of this fix.

I'm seeing some new failures in Tier6:

compiler/valhalla/inlinetypes/TestLWorld.java#id0
compiler/valhalla/inlinetypes/TestLWorld.java#id5
compiler/valhalla/inlinetypes/TestLWorld.java#id6

fail with this C2 crash:

```
#
#
#  Internal Error (/opt/mach5/mesos/work_dir/slaves/a9102fc6-1add-4658-abb6-501aa00dd79e-S28240/frameworks/1735e8a2-a1db-478c-8104-60c8b0af87dd-0196/executors/6145271d-45c3-4415-9f21-e7c3a927c3e0/runs/753b04f7-d4eb-4ea7-b728-7cf1952598b7/workspace/open/src/hotspot/share/opto/type.hpp:2574), pid=111669, tid=111685
#  assert(_base >= OopPtr && _base <= AryPtr) failed: Not a Java pointer
#
# JRE version: Java(TM) SE Runtime Environment (28.0) (fastdebug build 28-internal-2026-06-23-1927384.daniel.daugherty.valhallaexp.git)
# Java VM: Java HotSpot(TM) 64-Bit Server VM (fastdebug 28-internal-2026-06-23-1927384.daniel.daugherty.valhallaexp.git, compiled mode, sharing, tiered, compressed oops, compact obj headers, g1 gc, linux-amd64)
# Problematic frame:
# V  [libjvm.so+0xe19ff2]  ConnectionGraph::flat_access_aliases_with(Node*, TypeOopPtr const*)+0x3c2
#
# Core dump will be written. Default location: Determined by the following: "/opt/core.sh %p" (alternatively, falling back to /opt/mach5/mesos/work_dir/slaves/a9102fc6-1add-4658-abb6-501aa00dd79e-S28234/frameworks/1735e8a2-a1db-478c-8104-60c8b0af87dd-0196/executors/68f8f122-f8fb-48e2-9922-f75434682798/runs/0bc758ce-a68b-4838-9540-1552bd82e62c/testoutput/test-support/jtreg_open_test_hotspot_jtreg_hotspot_valhalla/scratch/5/core.111669)
#
Unsupported internal testing APIs have been used.
```

with this call stack:

```
---------------  T H R E A D  ---------------

Current thread (0x00007f3090287d70):  JavaThread "C2 CompilerThread0" daemon [_thread_in_native, id=111685, stack(0x00007f3075355000,0x00007f3075455000) (1024K)]


Current CompileTask:
C2:57558 10443    b  4       compiler.valhalla.inlinetypes.TestLWorld::test119 (104 bytes)

Stack: [0x00007f3075355000,0x00007f3075455000],  sp=0x00007f307544e940,  free space=998k
V  [libjvm.so+0xe19ff2]  ConnectionGraph::flat_access_aliases_with(Node*, TypeOopPtr const*)+0x3c2  (type.hpp:2574)
V  [libjvm.so+0xe1bbfc]  ConnectionGraph::find_inst_mem(Node*, int, Unique_Node_List&, unsigned int)+0x57c  (escape.cpp:4680)
V  [libjvm.so+0xe1adec]  ConnectionGraph::split_memory_phi(PhiNode*, int, Unique_Node_List&, unsigned int)+0x22c  (escape.cpp:4412)
V  [libjvm.so+0xe1c142]  ConnectionGraph::find_inst_mem(Node*, int, Unique_Node_List&, unsigned int)+0xac2  (escape.cpp:4758)
V  [libjvm.so+0xe1b99c]  ConnectionGraph::find_inst_mem(Node*, int, Unique_Node_List&, unsigned int)+0x31c  (escape.cpp:4690)
V  [libjvm.so+0xe1adec]  ConnectionGraph::split_memory_phi(PhiNode*, int, Unique_Node_List&, unsigned int)+0x22c  (escape.cpp:4412)
V  [libjvm.so+0xe1c142]  ConnectionGraph::find_inst_mem(Node*, int, Unique_Node_List&, unsigned int)+0xac2  (escape.cpp:4758)
V  [libjvm.so+0xe1b99c]  ConnectionGraph::find_inst_mem(Node*, int, Unique_Node_List&, unsigned int)+0x31c  (escape.cpp:4690)
V  [libjvm.so+0xe1adec]  ConnectionGraph::split_memory_phi(PhiNode*, int, Unique_Node_List&, unsigned int)+0x22c  (escape.cpp:4412)
V  [libjvm.so+0xe1c142]  ConnectionGraph::find_inst_mem(Node*, int, Unique_Node_List&, unsigned int)+0xac2  (escape.cpp:4758)
V  [libjvm.so+0xe1b99c]  ConnectionGraph::find_inst_mem(Node*, int, Unique_Node_List&, unsigned int)+0x31c  (escape.cpp:4690)
V  [libjvm.so+0xe1b99c]  ConnectionGraph::find_inst_mem(Node*, int, Unique_Node_List&, unsigned int)+0x31c  (escape.cpp:4690)
V  [libjvm.so+0xe1adec]  ConnectionGraph::split_memory_phi(PhiNode*, int, Unique_Node_List&, unsigned int)+0x22c  (escape.cpp:4412)
V  [libjvm.so+0xe1c142]  ConnectionGraph::find_inst_mem(Node*, int, Unique_Node_List&, unsigned int)+0xac2  (escape.cpp:4758)
V  [libjvm.so+0xe1b99c]  ConnectionGraph::find_inst_mem(Node*, int, Unique_Node_List&, unsigned int)+0x31c  (escape.cpp:4690)
V  [libjvm.so+0xe1adec]  ConnectionGraph::split_memory_phi(PhiNode*, int, Unique_Node_List&, unsigned int)+0x22c  (escape.cpp:4412)
V  [libjvm.so+0xe1c142]  ConnectionGraph::find_inst_mem(Node*, int, Unique_Node_List&, unsigned int)+0xac2  (escape.cpp:4758)
V  [libjvm.so+0xe25694]  ConnectionGraph::split_unique_types(GrowableArray<Node*>&, GrowableArray<ArrayCopyNode*>&, GrowableArray<MergeMemNode*>&, Unique_Node_List&)+0x2144  (escape.cpp:5311)
V  [libjvm.so+0xe334c3]  ConnectionGraph::compute_escape()+0x2c33  (escape.cpp:420)
V  [libjvm.so+0xe33c97]  ConnectionGraph::do_analysis(Compile*, PhaseIterGVN*)+0x177  (escape.cpp:121)
V  [libjvm.so+0xc0956c]  Compile::Optimize()+0x66c  (compile.cpp:3154)
V  [libjvm.so+0xc0d04d]  Compile::Compile(ciEnv*, ciMethod*, int, Options, DirectiveSet*)+0x1f2d  (compile.cpp:893)
V  [libjvm.so+0x9f3868]  C2Compiler::compile_method(ciEnv*, ciMethod*, int, bool, DirectiveSet*)+0x498  (c2compiler.cpp:149)
V  [libjvm.so+0xc1c704]  CompileBroker::invoke_compiler_on_method(CompileTask*)+0x844  (compileBroker.cpp:2041)
V  [libjvm.so+0xc1db48]  CompileBroker::compiler_thread_loop()+0x788  (compileBroker.cpp:1748)
V  [libjvm.so+0x12061bb]  JavaThread::thread_main_inner()+0x13b  (javaThread.cpp:651)
V  [libjvm.so+0x1c29df6]  Thread::call_run()+0xb6  (thread.cpp:243)
V  [libjvm.so+0x1823138]  thread_native_entry(Thread*)+0x118  (os_linux.cpp:931)
```

rwestrel:
@dcubed-ojdk thanks for the testing. Should be fixed now.

dean-long (approved):

src/hotspot/share/opto/escape.cpp
```diff
@@ -4548,6 +4548,44 @@ void ConnectionGraph::move_inst_mem(Node* n, Unique_Node_List& orig_phis) {
...
+// Does LoadFlat/StoreFlat flat_access alias with memory acess with type toop?
```
TobiHartmann:
```suggestion
// Does LoadFlat/StoreFlat flat_access alias with memory access with type toop?
```

TobiHartmann (approved):
This looks reasonable to me. I submitted some testing and will report back once it finished.

dcubed-ojdk:
I did Mach5 Tier[1-6] testing on e96fb69c73a92a6ebc4b8c84be25ebc492bfd851
which I think is one commit back. See https://bugs.openjdk.org/browse/JDK-8385328?focusedId=14889276&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-14889276

Short version: no unexpected failures.

@TobiHartmann is also running testing so please wait for his results.

dean-long (approved):

TobiHartmann:
All green - ship it!

TobiHartmann (approved):

rwestrel:
@dean-long @TobiHartmann thanks for the reviews
@dcubed-ojdk thanks for the testing

PR #2496, opened 2026-06-01
8382503: [lworld] compiler/c2/ReachabilityFenceFlagsTest.java fails post jdk-27+18 with preview enabled

merykitty:
Hi,

When `StressReachabilityFences` is `true`, the compiler tries to insert a reachability fence for all incoming states of the compilation. This misses the fact that an incoming `InlineTypeNode` may be unbuffered, so this PR fixes that by not inserting reachability fences for incoming `InlineTypeNode`s. In addition, when `reachabilityFence` is called on an `InlineTypeNode`, we need to buffer it.

---------

---------
### Progress

### Issue
 * [JDK-8382503](https://bugs.openjdk.org/browse/JDK-8382503): [lworld] compiler/c2/ReachabilityFenceFlagsTest.java fails post jdk-27+18 with preview enabled (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2496/head:pull/2496` \
`$ git checkout pull/2496`

`$ git checkout pull/2496` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2496/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2496`

`$ git pr show -t 2496`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2496.diff">https://git.openjdk.org/valhalla/pull/2496.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2496#issuecomment-4591744801)
</details>


TobiHartmann (approved):
Looks good to me.

src/hotspot/share/opto/library_call.cpp
```diff
@@ -7882,6 +7882,9 @@ bool LibraryCallKit::inline_reference_clear0(bool is_phantom) {
 // bool java.lang.ref.Reference.reachabilityFence();
 bool LibraryCallKit::inline_reference_reachabilityFence() {
   Node* referent = argument(0);
+  if (referent->is_InlineType()) {
+    referent = referent->as_InlineType()->buffer(this);
```
dean-long: Please explain.  The test doesn't trigger this code, so how do we test it?  I thought the whole point of reachabilityFence was to prevent a finalizer from running too early, but what does that even mean for an InlineType that might not be heap allocated?  Wouldn't it be a no-op or even better, an error, like trying to synchronize on a value object?  Or are we now saying that reachabilityFence on an inline type really means "do not scalarize, make sure finalizer gets run, but not until we get here"?

dean-long: In other words, I think reachabilityFence() requires an identity object for correct semantics.  I think there could be scenarios where buffering a copy is not enough.  Does that mean that instead of the finalizer running exactly 1 time, now it will run once for every buffered copy?

merykitty: A `reachabilityFence` keeps the object alive, which, in turn, keeps its fields alive. We can implement this by looking at all of the oop fields of the `InlineTypeNode` and keeping them alive instead. I decided to go for the simplest solution, which is to ensure that the whole object is kept alive.

merykitty:
Since there are still uncertainties in how `reachabiltyFence` should work with value objects, I have removed the change in `library_call.cpp`.

merykitty:
I also added a small change to fix the compilation failure on Shenandoah due to a merge conflict with master.

TobiHartmann (approved):
That makes sense. Could you please file a follow-up tracking issue for the "reachability fence on value objects" discussion?

merykitty:
@TobiHartmann Thanks for your review, I have filed [JDK-8385864](https://bugs.openjdk.org/browse/JDK-8385864).
/integrate

PR #2501, opened 2026-06-02
8384611: [lworld] Add value class test coverage to java.util HashMap and HashSet

bwhuang-us:
Changed Tests:

HashMap/KeySetRemove.java: equal ValueClass keys mapped to null can be removed from HashMap and TreeMap.

HashMap/NullKeyAtResize.java: null-key resize behavior remains correct when surrounding keys are ValueClass instances.

HashMap/PutNullKey.java: colliding comparable tree-bin keys are value-capable via @AsValueClass.

HashMap/ReplaceExisting.java: replacing an existing ValueClass key during active iteration does not corrupt the iterator.

HashMap/SetValue.java: Map.Entry.setValue() returns the old ValueClass value.

HashMap/ToArray.java: toArray() coverage is added for ValueClass keys, values, and set elements across hash and linked variants.

HashMap/TreeBinAssert.java: tree-bin iterator-removal coverage now uses a value-capable key.

Hashtable/EqualsCast.java: Provider/Hashtable equality now includes matching ValueClass key/value entries.

Hashtable/SimpleSerialization.java: serialization round-trip now also covers Hashtable<Integer,Integer>.

LinkedHashMap/ComputeIfAbsentAccessOrder.java: access-order behavior is repeated with Integer keys.

LinkedHashMap/EmptyMapIterator.java: fail-fast iterator behavior is repeated with Integer key/value entries.

LinkedList/AddAll.java: append-order behavior is repeated with ValueClass elements.

LinkedList/Clone.java: clone/equality checks for LinkedList, TreeSet, and TreeMap subclasses now include ValueClass contents.

Unchanged Tests:

HashMap/HashMapCloneLeak.java: unchanged because the test depends on WeakReference reachability. Value objects are not valid weak-reference targets, so adding VClass would not match the regression being tested.

HashMap/OverrideIsEmpty.java: unchanged because the test is about HashMap subclass method dispatch. Value classes cannot extend HashMap, and changing only the key/value payload would not add meaningful value-class coverage.

HashMap/WhiteBoxResizeTest.java: unchanged because it is a white-box/internal capacity and table-sizing test using reflection/VarHandles over HashMap, LinkedHashMap, HashSet, and WeakHashMap internals. Its purpose is sizing/lazy allocation/resize arithmetic, not key/value equality or value-object behavior.

HashMap/ToString.java: unchanged because it specifically verifies that HashMap.Entry.toString() does not throw when the map contains null keys or values. Adding value-class keys or values would not extend the original null-handling regression in a meaningful way.

HashSet/Serialization.java: unchanged because the existing HashSet<Integer> serialization round trip already provides platform value-class coverage when run under the Valhalla preview configuration.

Hashtable/DeserializedLength.java: unchanged because it already uses Integer keys and values. Under preview, that gives platform value-class coverage while preserving the test’s original focus on deserialized table capacity.

Hashtable/HashCode.java: unchanged because it intentionally tests the hash code of an empty Hashtable. Adding value-class entries would create a different non-empty-map test.

Hashtable/IllegalLoadFactor.java: unchanged because it validates constructor load-factor arguments and does not insert keys or values. Value-class semantics are irrelevant to this regression.

Hashtable/ReadObject.java: unchanged because it tests that deserialization does not call an overridable put method in a Hashtable subclass. Value classes cannot be Hashtable subclasses.

Hashtable/SelfRef.java: unchanged because it tests self-referential mutable identity maps in toString and hashCode. Value objects are not the self-referential identity object under test.

Hashtable/SerializationDeadlock.java: unchanged because it tests synchronization/deadlock behavior during serialization of mutually referential Hashtable identity objects. Adding value-class entries would be incidental.

LinkedHashMap/Basic.java: unchanged because it already uses Integer keys and values throughout, giving broad value-class coverage under preview.

LinkedHashMap/Cache.java: unchanged because it already uses Integer keys, so the cache/eldest-entry behavior is covered with platform value-class keys under preview.

LinkedHashSet/Basic.java: unchanged because the existing Integer element coverage already exercises value-class set behavior under preview, including equality, hashing, cloning, serialization, iteration, and clear behavior.

LinkedList/ComodifiedRemove.java: unchanged because it already uses Integer elements. The fail-fast iterator behavior is therefore covered with platform value-class elements under preview.

LinkedList/Remove.java: unchanged because it already uses an Integer element. The list-iterator remove/add behavior is covered with platform value-class elements under preview.

---------

---------
### Progress

### Issue
 * [JDK-8384611](https://bugs.openjdk.org/browse/JDK-8384611): [lworld] Add value class test coverage to java.util HashMap and HashSet (**Task** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2501/head:pull/2501` \
`$ git checkout pull/2501`

`$ git checkout pull/2501` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2501/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2501`

`$ git pr show -t 2501`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2501.diff">https://git.openjdk.org/valhalla/pull/2501.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2501#issuecomment-4607633890)
</details>


test/jdk/java/util/HashMap/KeySetRemove.java
```diff
@@ -40,5 +42,12 @@ public static void main(String[] args) throws Exception {
             if (!m[i].keySet().remove("bananas"))
                 throw new Exception("Yes, we have no bananas: "+i);
         }
+
+        Map[] valueMaps = {new HashMap<>(), new TreeMap<>()};
+        for (int i=0; i<valueMaps.length; i++) {
```
lmesnik: please update indentation (spaces)

test/jdk/java/util/HashMap/PutNullKey.java
```diff
@@ -45,6 +47,7 @@ public class PutNullKey {
     // A value 1.0 will ensure that a new threshold == capacity
     static final float LOAD_FACTOR = 1.0f;
 
+    @AsValueClass
```
lmesnik: Can you please add the second testcase and new class
CollidingHashValue 
that is '   @AsValueClass' 

You fix is going to work perfectly now, but later the plan is to rewrite    @AsValueClass with 'value class' and don't run the test with and without preview. So 2 testcases would be reuiqred.

test/jdk/java/util/HashMap/SetValue.java
```diff
@@ -37,11 +39,27 @@ public class SetValue {
...
+    private static void testStringValue() {
         Map m = new HashMap();
         m.put(key, oldValue);
         Map.Entry e = (Map.Entry) m.entrySet().iterator().next();
         Object returnVal = e.setValue(newValue);
         if (!returnVal.equals(oldValue))
             throw new RuntimeException("Return value: " + returnVal);
     }
+
+    private static void testTupleValue() {
```
lmesnik: Not sure if tuple is good name here, might be testVClassValue?

test/jdk/java/util/HashMap/ToArray.java
```diff
@@ -31,27 +31,26 @@
...
-            return Integer.compare(intValue, o.intValue);
-        }
-    }
 
     public static void main(String[] args) {
         checkMap(false);
         checkMap(true);
         checkSet(false);
         checkSet(true);
+        checkTupleMap(false);
```
lmesnik: Same about tuple in name.
"Not sure if tuple is good name here, might be testVClassValue?"

test/jdk/java/util/HashMap/TreeBinAssert.java
```diff
@@ -154,6 +156,7 @@ private void doTest(Object collection, int[] hashes,
     /**
      * Class that will have specified hash code in a HashMap.
      */
+    @AsValueClass
```
lmesnik: The same comment about having 2 testcases.

lmesnik (changes requested):

test/jdk/java/util/HashMap/KeySetRemove.java
bwhuang-us: Just to confirm, are you referring to the spacing inside the for loop parentheses, e.g. changing for (int i=0; i<valueMaps.length; i++) to for (int i = 0; i < valueMaps.length; i++)?

test/jdk/java/util/HashMap/PutNullKey.java
lmesnik: yes

test/jdk/java/util/HashMap/KeySetRemove.java
lmesnik: Yes, exactly.

PR #2507, opened 2026-06-03
8385930: [lworld] Update proxy variables generation in order to avert verifier errors

vicente-romero-oracle:
When dealing with strict fields with initializers at the time we track them there is no constructor that we can use to later retrieve them at lowering time. Those fields can drop on the floor and no proxies are created for them. This PR fixes this issue,

TIA

---------

---------
### Progress

### Issue
 * [JDK-8385930](https://bugs.openjdk.org/browse/JDK-8385930): [lworld] Update proxy variables generation in order to avert verifier errors (**Bug** - P4)

### Reviewers
 * [Chen Liang](https://openjdk.org/census#liach) (@liach - Committer) ⚠️ Review applies to [1f1d90d9](https://git.openjdk.org/valhalla/pull/2507/files/1f1d90d918b986af9bcdb8b8af76688fb7c32ca8)
 * [Maurizio Cimadamore](https://openjdk.org/census#mcimadamore) (@mcimadamore - Committer) ⚠️ Review applies to [56f51440](https://git.openjdk.org/valhalla/pull/2507/files/56f514400c4c8aa09e0fa503be750a00a2dac839)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2507/head:pull/2507` \
`$ git checkout pull/2507`

`$ git checkout pull/2507` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2507/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2507`

`$ git pr show -t 2507`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2507.diff">https://git.openjdk.org/valhalla/pull/2507.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2507#issuecomment-4615985432)
</details>


src/jdk.compiler/share/classes/com/sun/tools/javac/comp/LocalProxyVarsGen.java
```diff
@@ -106,14 +106,22 @@ protected LocalProxyVarsGen(Context context) {
...
-        fieldsReadInPrologue.put(constructor, fieldSet);
+        fieldsReadInPrologue.put(tree, fieldSet);
     }
 
-    public void patchConstructor(JCMethodDecl tree, TreeMaker make) {
+    public void patchConstructor(JCTree.JCClassDecl classDecl, JCMethodDecl tree, TreeMaker make) {
+        /* if some fields have initializers those probably were added using the enclosing class as their map key
+         * we need to recover those now and add them to this constructor
+         */
+        if (fieldsReadInPrologue.get(classDecl) != null) {
```
mcimadamore: We never clear this entry up, right?

mcimadamore: I wonder if it would be cleaner to first collect all the stuff that needs patching (and clear entries in the map) and then pass them to the addLocalProxiesFor method -- that method already accepts a Set anyway

src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/Gen.java
```diff
@@ -419,10 +419,9 @@ boolean hasFinally(JCTree target, Env<GenContext> env) {
 
     /** Distribute member initializer code into constructors and {@code <clinit>}
      *  method.
-     *  @param defs         The list of class member declarations.
-     *  @param c            The enclosing class.
+     *  @param classDecl         The class declaration to normalize.
      */
-    List<JCTree> normalizeDefs(List<JCTree> defs, ClassSymbol c) {
+    List<JCTree> normalizeDefs(JCClassDecl classDecl) {
```
mcimadamore, line 424: I don't oppose this change -- but note that I think changes can be minimized by taking the GenEnv (which should have an enclosing class?)

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Resolve.java
```diff
@@ -4055,9 +4055,9 @@ private boolean isSimpleEarlyFieldRefAllowed(DiagnosticPosition pos,
             } else {
                 // Read early ref, this is only allowed under JEP 401, and requires special codegen support
                 preview.checkSourceLevel(pos, Feature.VALUE_CLASSES);
-                if (context.ctorPrologue() && env.enclMethod != null) {
+                if (context.ctorPrologue()) {
                     // Track the early read for codegen
-                    localProxyVarsGen.addFieldReadInPrologue(env.enclMethod, field);
+                    localProxyVarsGen.addFieldReadInPrologue(env.enclMethod != null ? env.enclMethod : env.enclClass, field);
```
mcimadamore: maybe worth pulling `env.enclMethod != null ? env.enclMethod : env.enclClass` in a separate local, for making the expression more readable?

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/LocalProxyVarsGen.java
mcimadamore: I'm also noticing (but that's just an observation) that maybe symbols could be used as map keys too -- e.g. maybe we don't really need trees?

vicente-romero-oracle: yes we can probably use symbols but it won't make a difference I think

src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/Gen.java
vicente-romero-oracle: right but it has a reference to top classes, not inner classes

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/LocalProxyVarsGen.java
```diff
@@ -106,20 +106,32 @@ protected LocalProxyVarsGen(Context context) {
...
-    public void patchConstructor(JCMethodDecl tree, TreeMaker make) {
+    public void patchConstructor(JCTree.JCClassDecl classDecl, JCMethodDecl tree, TreeMaker make) {
+        /* if some fields have initializers those probably were added using the enclosing class as their map key
+         * we need to recover those now and add them to this constructor
+         */
+        Set<Symbol> earlyReads = null;
+        if (fieldsReadInPrologue.get(classDecl) != null) {
+            earlyReads = fieldsReadInPrologue.get(classDecl);
+            fieldsReadInPrologue.remove(classDecl);
+        }
```
liach, line 123:
```suggestion
        Set<Symbol> earlyReads = fieldsReadInPrologue.remove(classDecl);
```

However I wonder if this removing for one constructor is right - we might need to duplicate initializers to multiple constructors.

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/LocalProxyVarsGen.java
```diff
@@ -106,20 +106,32 @@ protected LocalProxyVarsGen(Context context) {
...
-            Set<Symbol> earlyReads = fieldsReadInPrologue.get(tree);
-            if (earlyReads != null && !noLocalProxyVars) {
-                addLocalProxiesFor(tree, earlyReads, make);
-                fieldsReadInPrologue.remove(tree);
+            if (earlyReads == null) {
+                earlyReads = fieldsReadInPrologue.get(tree);
+            } else {
+                earlyReads.addAll(fieldsReadInPrologue.get(tree));
             }
+            fieldsReadInPrologue.remove(tree);
```
liach:
```suggestion
            Set<Symbol> constructorEarlyReads = fieldsReadInPrologue.remove(tree);
            if (earlyReads == null) {
                earlyReads = constructorEarlyReads;
            } else {
                earlyReads.addAll(constructorEarlyReads);
            }
```

src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/Gen.java
mcimadamore: Inner classes should be gone by Gen time?

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/LocalProxyVarsGen.java
vicente-romero-oracle: yes, it is not right to remove it here, we need it for other constructors. We can remove once the class has been completely generated

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/LocalProxyVarsGen.java
mcimadamore: I agree this can be consolidated a bit -- using `remove` to avoid `get` + `remove` is nice. My general feeling is that the null logic is a bit confusing as well. E.g. maybe starting with an empty set and calling addAll a couple of times is clearer -- too bad that we have to deal with nulls from remove.

This might be nicer:

```
Stream.of(classDecl, tree)
           .flatMap(this::earlyReadsForTree)
           .collect(Collectors.toSet());
```

With a helper method:

```
Set<Symbol> earlyReadsForTree(JCTree tree) {
     Set<Symbol> earlyReads = fieldsReadInPrologue.remove(tree);
     return earlyReads == null ? earlyReads : Set.of();
}
```

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/LocalProxyVarsGen.java
mcimadamore: Oof -- good point

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/LocalProxyVarsGen.java
```diff
@@ -106,21 +106,33 @@ protected LocalProxyVarsGen(Context context) {
...
 
-    public void patchConstructor(JCMethodDecl tree, TreeMaker make) {
+    public void patchConstructor(JCTree.JCClassDecl classDecl, JCMethodDecl tree, TreeMaker make) {
+        /* if some fields have initializers those probably were added using the enclosing class as their map key
+         * we need to recover those now and add them to this constructor
+         */
+        Set<Symbol> earlyReads = null;
+        if (fieldsReadInPrologue.get(classDecl) != null) {
+            earlyReads = fieldsReadInPrologue.get(classDecl);
+            fieldsReadInPrologue.remove(classDecl);
```
mcimadamore: Doesn't this suffer from the problem you were trying to avoid -- e.g. don't remove field early reads before you have seen _all_ constructors?

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/LocalProxyVarsGen.java
mcimadamore: I suppose we can safely remove after we finished normalizing constructors in Gen

mcimadamore: Another thing we discussed few days ago would be to store early reads in a list on the AST instead of using a side map -- this avoids the problem of "when to remove"

src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/Gen.java
vicente-romero-oracle: right but for some reason we don't update it to point to the current class being generated. At least at the moment this method is called. See: Gen::genClass we create a localEnv that points to the current class, but it is a local variable of this method. Probably a bug? but something we shouldn't fix as part of this patch

liach (approved):
Current version looks ok to me

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/LocalProxyVarsGen.java
```diff
@@ -106,35 +106,39 @@ protected LocalProxyVarsGen(Context context) {
...
                 earlyReads.addAll(constructorEarlyReads);
             }
+            fieldsReadInPrologue.remove(tree.sym);
         }
         if (earlyReads != null && !noLocalProxyVars) {
             addLocalProxiesFor(tree, earlyReads, make);
         }
     }
 
+    public void allConstNormalized(Symbol.ClassSymbol csym) {
```
mcimadamore: should be allFieldNormalized?

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/LocalProxyVarsGen.java
```diff
@@ -106,35 +106,39 @@ protected LocalProxyVarsGen(Context context) {
...
-        if (fieldsReadInPrologue.get(tree) != null) {
-            Set<Symbol> constructorEarlyReads = fieldsReadInPrologue.remove(tree);
+        if (fieldsReadInPrologue.get(tree.sym) != null) {
+            Set<Symbol> constructorEarlyReads = fieldsReadInPrologue.remove(tree.sym);
             if (earlyReads == null) {
                 earlyReads = constructorEarlyReads;
             } else {
                 earlyReads.addAll(constructorEarlyReads);
             }
+            fieldsReadInPrologue.remove(tree.sym);
```
mcimadamore: you are calling remove twice here -- this seems not needed. And also it seems like you can consolidate

```
 Set<Symbol> constructorEarlyReads = fieldsReadInPrologue.remove(tree.sym);
 if (constructorEarlyReads != null) {
            if (earlyReads == null) {
                earlyReads = constructorEarlyReads;
            } else {
                earlyReads.addAll(constructorEarlyReads);
            }
}
```

mcimadamore (approved):
Looks good!

PR #2510, opened 2026-06-04
8385943: [lworld] Redundant ValueBased annotation on Record

liach:
`@ValueBased` is an annotation only meaningful for concrete value classes.  We should fix javac and remove the redundant annotation.

Apparently javac is not reading the `MIGRATED_VALUE_CLASS` flag from class files correctly. This is fine for most Java compilation jobs but affects the microbenchmark building, which I think reads from the non-preview set of classes. We can fix the reading of that flag to bypass microbenchmark building issues.

---------

---------
### Progress

### Issue
 * [JDK-8385943](https://bugs.openjdk.org/browse/JDK-8385943): [lworld] Redundant ValueBased annotation on Record (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2510/head:pull/2510` \
`$ git checkout pull/2510`

`$ git checkout pull/2510` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2510/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2510`

`$ git pr show -t 2510`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2510.diff">https://git.openjdk.org/valhalla/pull/2510.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2510#issuecomment-4618572739)
</details>

### DIFF
--- a/src/java.base/share/classes/java/lang/Record.java
+++ b/src/java.base/share/classes/java/lang/Record.java
@@ -98,7 +98,6 @@ package java.lang;
  * @since 16
  */
 @jdk.internal.MigratedValueClass
-@jdk.internal.ValueBased
 public abstract class Record {
     /**
      * Constructor for record classes to call.
--- a/src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Check.java
+++ b/src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Check.java
@@ -2581,7 +2581,7 @@ public class Check {
         Type identitySuper = null;
         for (Type t : types.closure(c)) {
             if (t != c) {
-                if (t.isIdentityClass() && (t.tsym.flags() & VALUE_BASED) == 0)
+                if (t.isIdentityClass() && (t.tsym.flags() & MIGRATED_VALUE_CLASS) == 0)
                     identitySuper = t;
                 if (c.isValueClass() && identitySuper != null && identitySuper.tsym != syms.objectType.tsym) { // Object is special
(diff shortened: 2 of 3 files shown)


vicente-romero-oracle (approved):
lgtm

liach:
Thanks for the review! I will go ahead and clean the annotation and integrate the javac fixes.

/integrate

src/java.base/share/classes/java/lang/Record.java
```diff
@@ -98,7 +98,6 @@
  * @since 16
  */
 @jdk.internal.MigratedValueClass
-@jdk.internal.ValueBased
```
ExE-Boss, line 101: This annotation has been here since the introduction of **Records** in **JDK 16**

liach: Huh?

https://github.com/openjdk/valhalla/blob/58c9fbc93db6c27b00d1e53d7ee8bf6e382eea82/src/java.base/share/classes/java/lang/Record.java#L89-L90

ExE-Boss: Ughh, that’s what I get for using **GitHub**’s `git blame` on mobile, where the UI is confusing.

It was actually added in commit <https://github.com/openjdk/valhalla/commit/b883f80f4253c0bc3fd2734595f8f25311bc22a1>, which was before `MigratedValueClass` existed.

PR #2517, opened 2026-06-05
8385966: [lworld] Remove ThreadInVMfromUnknown from get_supers

coleenp:
This change refactors the super class scalar argument checking code and deoptimization so that it is only called at link time and not for frame::describe() and other places that are only retrieving the calling convention.  The only time that we detect a mismatch is during class linking, before the overriding subclass method can run.

This avoids running vm code while thread_in_Java, which is unsafe, and removes the ThreadInVMfromUnknown, which was even more unsafe.

Tested with tier1-4.

---------

---------
### Progress

### Issue
 * [JDK-8385966](https://bugs.openjdk.org/browse/JDK-8385966): [lworld] Remove ThreadInVMfromUnknown from get_supers (**Enhancement** - P4)

### Reviewers
 * [Dan Heidinga](https://openjdk.org/census#heidinga) (@DanHeidinga - no project role) ⚠️ Review applies to [df36a6c9](https://git.openjdk.org/valhalla/pull/2517/files/df36a6c9843106fc7b7a2a2503de96df23f897b0)
 * [Patricio Chilano Mateo](https://openjdk.org/census#pchilanomate) (@pchilano - no project role) ⚠️ Review applies to [959ca7d0](https://git.openjdk.org/valhalla/pull/2517/files/959ca7d03353ecfb5702e0a158b1aec03fc4e3f4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2517/head:pull/2517` \
`$ git checkout pull/2517`

`$ git checkout pull/2517` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2517/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2517`

`$ git pr show -t 2517`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2517.diff">https://git.openjdk.org/valhalla/pull/2517.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2517#issuecomment-4632938101)
</details>

### DIFF
--- a/src/hotspot/share/runtime/sharedRuntime.cpp
+++ b/src/hotspot/share/runtime/sharedRuntime.cpp
@@ -2900,6 +2900,7 @@ bool CompiledEntrySignature::check_supers_and_deoptimize(int arg_num) {
 // Iterate over arguments and compute scalarized and non-scalarized signatures
 void CompiledEntrySignature::compute_calling_conventions(bool link_time) {
   assert(JavaThread::current()->thread_state() != _thread_in_native, "must not be in native");
+  assert(link_time || (_method != nullptr && _method->adapter() != nullptr), "invariant");
   bool has_scalarized = false;
   if (_method != nullptr) {
     InstanceKlass* holder = _method->method_holder();
@@ -2921,7 +2922,7 @@ void CompiledEntrySignature::compute_calling_conventions(bool link_time) {
       arg_num++;
     }
     for (SignatureStream ss(_method->signature()); !ss.at_return_type(); ss.next()) {
-      BasicType bt = ss.type();
+      const BasicType bt = ss.type();
       if (InlineTypePassFieldsAsArgs && bt == T_OBJECT) {
         InlineKlass* vk = ss.as_inline_klass(holder);
         if (vk != nullptr && vk->can_be_passed_as_fields() && (link_time || _method->is_scalarized_arg(arg_num))) {
@@ -2948,7 +2949,6 @@ void CompiledEntrySignature::compute_calling_conventions(bool link_time) {
           SigEntry::add_entry(_sig_cc, T_OBJECT, ss.as_symbol());
           SigEntry::add_entry(_sig_cc_ro, T_OBJECT, ss.as_symbol());
         }
-        bt = T_OBJECT;
       } else {
         SigEntry::add_entry(_sig_cc, ss.type(), ss.as_symbol());
         SigEntry::add_entry(_sig_cc_ro, ss.type(), ss.as_symbol());
--- a/src/hotspot/share/runtime/sharedRuntime.cpp
+++ b/src/hotspot/share/runtime/sharedRuntime.cpp
@@ -2816,7 +2816,6 @@ GrowableArray<Method*>* CompiledEntrySignature::get_supers() {
   Symbol* signature = _method->signature();
   const Klass* holder = _method->method_holder()->super();
   Symbol* holder_name = holder->name();
-  ThreadInVMfromUnknown tiv;
   JavaThread* current = JavaThread::current();
   HandleMark hm(current);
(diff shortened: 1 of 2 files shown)


fparain:
As this is a change related to compiler code, did you run valhalla-comp-stress on Mach5?

src/hotspot/share/runtime/sharedRuntime.cpp
```diff
@@ -2845,16 +2844,68 @@ GrowableArray<Method*>* CompiledEntrySignature::get_supers() {
   return _supers;
 }
 
+bool CompiledEntrySignature::check_supers_and_deoptimize(int arg_num) {
```
fparain, line 2847: compute_calling_conventions() checks the thread state vs the link_time argument, but shouldn't this method, check_supers_and_deoptimize(), be protected by an assert ensuring current thread state is _thread_in_vm?

coleenp: Yeah, I could move the assert down.
And run the valhalla-comp-stress tests too.  I ran local tests with -Xlog:continuations=trace:rt.log since freezing and thawing calls frame::describe a lot.

DanHeidinga (approved):
This looks reasonable

src/hotspot/share/runtime/sharedRuntime.cpp
coleenp: I moved the assert to check_supers_and_deoptimize() for thread_in_vm, but added one that asserted not thread_in_native since that was the problematic case.
The valhalla-comp-stress test group passed.

src/hotspot/share/runtime/sharedRuntime.cpp
```diff
@@ -2845,16 +2844,70 @@ GrowableArray<Method*>* CompiledEntrySignature::get_supers() {
...
+
+  return non_scalar_super;
+}
+
 // Iterate over arguments and compute scalarized and non-scalarized signatures
-void CompiledEntrySignature::compute_calling_conventions(bool init) {
+void CompiledEntrySignature::compute_calling_conventions(bool link_time) {
+  assert(JavaThread::current()->thread_state() != _thread_in_native, "must not be in native");
   bool has_scalarized = false;
   if (_method != nullptr) {
```
pchilano, line 2905: Given that the non-link case relies on the method being linked already to compute the signatures, should we assert that the adapter is available?

`assert(link_time || _method->adapter() != nullptr, "invariant");`

src/hotspot/share/runtime/sharedRuntime.cpp
pchilano: Preexistent: this can be removed since it’s one of the branch conditions.

pchilano (approved):
Thanks for fixing this. The refactoring also makes the code much more readable.

src/hotspot/share/runtime/sharedRuntime.cpp
coleenp: Yes, this is a good improvement and helps with understanding that the code is called for both link time to create the adapters and runtime to print registers.  Something like
assert(link_time || (_method != nullptr && _method->adapter() != nullptr), "invariant");

src/hotspot/share/runtime/sharedRuntime.cpp
coleenp: Added const to make it clear that nothing is changing bt.
Retesting with these changes, thanks.

fparain (approved):
Looks good to me.

coleenp:
Thank you for reviewing Patricio and Fred. Thanks for the discussion yesterday Patricio.
/integrate

coleenp:
Also, thanks for reviewing Dan.

PR #2526, opened 2026-06-08
8386250: [lworld] JVMTI GetClassModifiers

AlanBateman:
The specification of JVMTI GetClassModifiers needs some TLC. The spec was carried over from the equivalent function in JVMDI but doesn't really specify that it works like Class::getModifiers and returns Java language modifiers and not necessarily the access flags from the class file. The change in behavior to only set ACC_IDENTITY when the class is an identity class or array makes it a good time to re-write the spec to align with existing and future behavior.

A new test is added to test this function with value classes. The tests are added to the "newer" serviceability/jvmti tree. There are some older tests for GetClassModifiers in vmTestbase/nsk suite so there is a bit of an overlap.

---------

---------
### Progress

### Issue
 * [JDK-8386250](https://bugs.openjdk.org/browse/JDK-8386250): [lworld] JVMTI GetClassModifiers (**Enhancement** - P4)

### Reviewers
 * [Chen Liang](https://openjdk.org/census#liach) (@liach - Committer) ⚠️ Review applies to [412b624e](https://git.openjdk.org/valhalla/pull/2526/files/412b624e9a504b6fa1f4cdc54abbe00c8f66b3ac)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2526/head:pull/2526` \
`$ git checkout pull/2526`

`$ git checkout pull/2526` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2526/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2526`

`$ git pr show -t 2526`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2526.diff">https://git.openjdk.org/valhalla/pull/2526.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2526#issuecomment-4658546039)
</details>


src/hotspot/share/prims/jvmti.xml
```diff
@@ -7361,23 +7361,39 @@ class C2 extends C1 implements I2 {
...
-        If the class is an array class or a primitive class then its final
-        modifier is always true and its interface modifier is always false.
-        The values of its other modifiers are not determined by this specification.
-
+        When preview features are enabled, the <code>ACC_IDENTITY</code> access flag
+        is set if <code>klass</code> is an identity class or an array class.
+        <p/>
+        When preview features are disabled, the <code>ACC_SUPER</code> access flag may
+        be set if <code>klass</code> is not a primitive type or void. The flag has no
+        effective meaning.
```
liach, line 7396: I think we can be more assertive, like:
"the ACC_IDENTITY access flag is set if and only if klass is an identity class or an array class"

"the ACC_SUPER access flag is set if and only if klass is not a primitive type or void"

AlanBateman: For ACC_IDENTITY then using "if and only if" would be better, thanks!

For ACC_SUPER, the context is optional implementation-specific/undocumented behavior so I don't think "if and only if" would work there. I've tweaked the wording from "may be set" to "may be set only" to try to make this clearer.

liach (approved):
This new spec is an accurate reflection of how the new modifier flags work in JEP 401.

test/hotspot/jtreg/serviceability/jvmti/GetClassModifiers/libGetClassModifiers.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+JNIEXPORT jint JNICALL Java_GetClassModifiers_getClassModifiers(JNIEnv *env, jclass ignore, jclass target) {
+    jvmtiError err;
+    jint modifiers;
+    err = (*jvmti)->GetClassModifiers(jvmti, target, &modifiers);
+    if (err != JVMTI_ERROR_NONE) {
+        (*env)->FatalError(env, "GetClassModifiers failed");
+    }
+    return modifiers;
+}
```
sspitsyn, line 55: A couple of nits:
 - indent is better to be 2
 - converting to .cpp would be good

sspitsyn (approved):
This looks good. I've posted a couple of nits.

sspitsyn:
Chris has a concern related to the integration schedule and would prefer to separate the Valhalla specific changes from the long-standing improvement of the JVMTI `GetClassModifiers()` as it may impact the JDI and JDWP specs.
It feels like there has to be a way to avoid detailed JDI and JDWP spec changes.

AlanBateman:
> Chris has a concern related to the integration schedule and would prefer to separate the Valhalla specific changes from the long-standing improvement of the JVMTI `GetClassModifiers()` as it may impact the JDI and JDWP specs. It feels like there has to be a way to avoid detailed the JDI and JDWP spec changes.

There are no implementation changes here and there is nothing that specifies that the JDWP ReferenceType.Modifiers command be implemented with JVMTI GetClassModifiers. That allows us to separate the  concern with the JDWP and JDI specs. (BTW: At one point there were certified implementation that supported JDWP natively, no JVMTI).

So I think we should create a follow-up issue to re-examine the spec of the JDWP ReferenceType.Modifiers command, and the JDI methods. Would you be okay with that?

sspitsyn:
> So I think we should create a follow-up issue to re-examine the spec of the JDWP ReferenceType.Modifiers command, and the JDI methods. Would you be okay with that?

Yes, we are okay with this. Thanks!

test/hotspot/jtreg/serviceability/jvmti/GetClassModifiers/libGetClassModifiers.cpp
AlanBateman: Okay, moved it to .cpp. It was a bit of a coin toss initially as the tests in serviceability/jvmti have both .c and .cpp.

sspitsyn: Thanks. The plan is to make it consistent, at list in the `serviceability/jvmti` folder.

sspitsyn (approved):

PR #2537, opened 2026-06-11
8386342: [lworld] Large overlapping flat arraycopy overflows payload offset

caspernorrbin:
Hi everyone,

This fixes an overflow in `FlatArrayPayload::advance_index`. The method advanced an array payload offset by multiplying two `int` values before adding the result to a `ptrdiff_t`. For large flat arrays, the backward-overlap path in `FlatArrayKlass::copy_array` can call `advance_index(length - 1)`, and that multiplication can overflow before the value is widened.

The fix widens the operands before multiplying, so the offset adjustment is computed in a `ptrdiff_t`.

I also added a regression test that performs large overlapping copies on a flat value array. Before the fix, the backward overlapping copy fails in debug builds with an out-of-bounds assert. With the fix, both backward and forward overlapping copies complete successfully.

Testing:
- New test testing same-array flat copy behavior
- Tiers 1-3

---------

---------
### Progress

### Issue
 * [JDK-8386342](https://bugs.openjdk.org/browse/JDK-8386342): [lworld] Large overlapping flat arraycopy overflows payload offset (**Bug** - P2)

### Reviewers
 * [Johan Sjölen](https://openjdk.org/census#jsjolen) (@johan-sjolen - no project role) ⚠️ Review applies to [9c158119](https://git.openjdk.org/valhalla/pull/2537/files/9c1581192382f1b31a5df5f03cae3b8b36fbcfef)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2537/head:pull/2537` \
`$ git checkout pull/2537`

`$ git checkout pull/2537` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2537/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2537`

`$ git pr show -t 2537`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2537.diff">https://git.openjdk.org/valhalla/pull/2537.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2537#issuecomment-4680405607)
</details>

### DIFF
--- a/test/hotspot/jtreg/runtime/valhalla/inlinetypes/FlatArrayLargeOverlapCopyTest.java
+++ b/test/hotspot/jtreg/runtime/valhalla/inlinetypes/FlatArrayLargeOverlapCopyTest.java
@@ -37,6 +37,8 @@ package runtime.valhalla.inlinetypes;
 import jdk.internal.value.ValueClass;
 
 public class FlatArrayLargeOverlapCopyTest {
+    // Value[] is flattened with 8-byte elements.
+    // Copying 290M elements makes length * elementSize exceed Integer.MAX_VALUE.
     private static final int LENGTH = 290_000_000;
 
     public static value record Value(int value) {}
--- a/test/hotspot/jtreg/runtime/valhalla/inlinetypes/FlatArrayLargeOverlapCopyTest.java
+++ b/test/hotspot/jtreg/runtime/valhalla/inlinetypes/FlatArrayLargeOverlapCopyTest.java
@@ -1,0 +1,75 @@
+/*
+ * Copyright (c) 2026, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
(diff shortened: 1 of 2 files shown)


test/hotspot/jtreg/runtime/valhalla/inlinetypes/FlatArrayLargeOverlapCopyTest.java
```diff
@@ -0,0 +1,75 @@
...
+ * @run main/othervm/timeout=240 -Xint -Xmx6G
+ *                               runtime.valhalla.inlinetypes.FlatArrayLargeOverlapCopyTest
+ */
+
+package runtime.valhalla.inlinetypes;
+
+import jdk.internal.value.ValueClass;
+
+public class FlatArrayLargeOverlapCopyTest {
+    private static final int LENGTH = 290_000_000;
```
johan-sjolen, line 42: Add a comment for why this particular value was chosen.

johan-sjolen (approved):

fparain (approved):
LGTM

jsikstro (approved):
I agree that we need larger types for the calculation in `FlatArrayPayload::advance_index`. However, we could change the type of `FlatArrayPayload::Storage::_element_size` to `ptrdiff_t` since we only seem to use it that way now. I'm fine with us addressing that in a follow-up.

Also the test does use a non-trivial amount of memory, but I think the array being ~2200MB is reasonable and shouldn't clash with other tests running concurrently.

caspernorrbin:
Thank you for the quick reviews!

I don't know if I quite agree that a `ptrdiff_t` would be the right type to use to represent a size, but that can be discussed in a follow-up. For now, let's get this in.

/integrate

PR #2545, opened 2026-06-15
8386627: [lworld] Cleanup ExplodedImage jrtfs support

liach:
Introduce a `PathResolution` to `ExplodedImage` to remove a lot of ad-hoc treatments of preview path discovery. Also some general comments and test additions.

One interesting thing is the mainline jimage is not mapping `META-INF/preview/file` to `file` when preview is enabled; it does map `META-INF/preview/dir/file` to `dir/file`. Don't know if this is intended; exploded image performs the first mapping, which differs from jimage.

---------

---------
### Progress

### Issue
 * [JDK-8386627](https://bugs.openjdk.org/browse/JDK-8386627): [lworld] Cleanup ExplodedImage jrtfs support (**Enhancement** - P4)

### Reviewers
 * [Xueming Shen](https://openjdk.org/census#sherman) (@xuemingshen-oracle - no project role)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2545/head:pull/2545` \
`$ git checkout pull/2545`

`$ git checkout pull/2545` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2545/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2545`

`$ git pr show -t 2545`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2545.diff">https://git.openjdk.org/valhalla/pull/2545.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2545#issuecomment-4704486502)
</details>


liach:
@xuemingshen-oracle @dbalek Can you help review this cleanup and comment on that jimage behavior I identified in the PR description?

src/java.base/share/classes/jdk/internal/jrtfs/ExplodedImage.java
```diff
@@ -136,8 +146,8 @@ public PathNode resolveLink(boolean recursive) {
 
         private byte[] getContent() throws IOException {
             if (!getFileAttributes().isRegularFile())
-                throw new FileSystemException(getName() + " is not file");
-            return Files.readAllBytes(modulesDir.resolve(relPath));
+                throw new FileSystemException(getName() + " is not a resource");
```
AlanBateman: The original exception message was correct, it should say "file" or "regular file".

src/java.base/share/classes/jdk/internal/jrtfs/ExplodedImage.java
```diff
@@ -59,56 +60,65 @@ class ExplodedImage extends SystemImage {
...
+        /**
+         * Creates a completed directory node based a list of child nodes.
+         * Used for the root, /modules, /packages, and /packages/... non-leaf
+         * directories, all created in initNodes().
+         */
+        private PathNode(String name, List<PathNode> children) {
             super(name, modulesDirAttrs);
-            this.childNames = children.stream().map(Node::getName).collect(toList());
+            this.resolution = null;
+            this.childNames = children.stream().map(Node::getName).sorted().collect(Collectors.toList());
```
AlanBateman: None of APIs require the list of entries to be sorted.

liach: The other assignments to `childNames` all ensure the list is sorted, such as in the lazy computation. I find it weird that we sort inconsistently; I am open to removing the sort in the lazy computation method if that is better.

src/java.base/share/classes/jdk/internal/jrtfs/ExplodedImage.java
liach: This guy is only called from `SystemImage::getResource`, and the other implementation reports "not a resource" when it fails: https://github.com/openjdk/valhalla/blob/2523a4f43c8d865574693fff3a8944b2c1c64fe8/src/java.base/share/classes/jdk/internal/jimage/ImageReader.java#L788

AlanBateman: The exploded image doesn't have a jimage file. Also file system APIs (that jrtfs implements) speaks of files rather than resources.

A jlink generated run-time image has a jimage file. Some APIs built on that, e.g. Class::getResource, maps resource names to entries. The jrtfs implementation for the run-time image speaks files.

src/java.base/share/classes/jdk/internal/jrtfs/ExplodedImage.java
```diff
@@ -368,29 +394,32 @@ private void processModuleDirectory(Path moduleDir, Map<String, List<String>> pa
...
         String moduleName = moduleDir.getFileName().toString();
         // Make sure "/modules/<moduleName>" is created
-        Objects.requireNonNull(createModulesNode(MODULES + moduleName, moduleDir));
-        // Skip the first path (it's always the given root directory).
-        try (Stream<Path> contentsStream = Files.walk(moduleDir).skip(1)) {
+        Objects.requireNonNull(findNode(MODULES + moduleName));
+        UnaryOperator<Path> previewExtractor = isPreviewMode
+                ? (p -> p.startsWith(PREVIEW_DIR) ? PREVIEW_DIR.relativize(p) : p)
+                : UnaryOperator.identity();
+        try (Stream<Path> contentsStream = Files.walk(moduleDir)) {
```
AlanBateman: The re-worked processModuleDirectory looks okay. It could use Files.find to find directories rather than Files.walk and filtering the entries.

src/java.base/share/classes/jdk/internal/jrtfs/ExplodedImage.java
```diff
@@ -210,63 +215,94 @@ public synchronized Node findNode(String name) {
...
     }
 
     /**
-     * Returns the expected file path for name in the "/modules/..." namespace,
+     * Returns the expected file paths for name in the "/modules/..." namespace,
      * or {@code null} if the name is not in the "/modules/..." namespace or the
      * path does not reference a file.
      */
-    private Path underlyingModulesPath(String name) {
+    private PathResolution resolveModulesPath(String name) {
```
AlanBateman: Some of the naming is a bit confusing. This method locates a file in an exploded module. The parameter is jrt file path of the form "/modules/$MODULE/$PATH". I think find a better name, like findFile, as "modules path" is too close to "module path".

AlanBateman:
> One interesting thing is the mainline jimage is not mapping META-INF/preview/file to file when preview is enabled; it does map META-INF/preview/dir/file to dir/file. Don't know if this is intended; exploded image performs the first mapping, which differs from jimage.

There shouldn't be any cases right now that lead to resources going into the "top-level" META-INF/preview directory but I do see this would need attention, maybe with jlink too. We can create an issue in JBS to track this and maybe Sherman or Dusan can look at it.

So to your question, the behavior with the exploded image looks right.

test/jdk/jdk/internal/jrtfs/SystemImageTest.java
```diff
@@ -109,7 +104,15 @@ enum ImageType {SYSTEM, EXPLODED}
...
                     // Two new packages in preview mode (new symbolic links).
                     "@com.bar.preview.stuff.Foo",
-                    "@com.bar.preview.stuff.Bar"),
+                    "@com.bar.preview.stuff.Bar",
+                    // File cannot be preview-overridden
+                    "!META-INF/one.txt",
+                    // File cannot be discovered out-of-preview
+                    // Note: somehow this doesn't work for jimage if file is in
+                    // root directory
+                    "!META-INF/preview/com/two.txt",
```
AlanBateman, line 111: The comment is very confusing. It's just a file that exists only when running with preview features enabled.

My preference would be to not include tests that don't work with the run-time image, let's put them in the JBS issue to re-examine the jimage implementation.

test/jdk/jdk/internal/jrtfs/SystemImageTest.java
```diff
@@ -109,7 +104,15 @@ enum ImageType {SYSTEM, EXPLODED}
...
+                    // File cannot be discovered in any scenario
```
AlanBateman: Just say that this file is never found.

test/jdk/jdk/internal/jrtfs/SystemImageTest.java
```diff
@@ -233,6 +243,11 @@ public void testPreviewResources_disabled() throws IOException {
             assertAbsent(image, "/modules/modfoo/com/foo/bar/IsPreviewOnly.class");
             assertDirContents(image, "/modules/modfoo/com/foo", "HasPreviewVersion.class", "NormalFoo.class", "bar");
             assertDirContents(image, "/modules/modfoo/com/foo/bar", "NormalBar.class");
+
+            // No preview activation
```
AlanBateman: The use of "activation" in the comments is very confusing.  It's just checking that a file is not found.

src/java.base/share/classes/jdk/internal/jrtfs/ExplodedImage.java
```diff
@@ -364,33 +391,34 @@ private void initNodes() throws IOException {
...
 
-    private void processModuleDirectory(Path moduleDir, Map<String, List<String>> packageToModules)
+    private void findPackagesInModule(Path moduleDir, Map<String, List<String>> packageToModules)
             throws IOException {
         String moduleName = moduleDir.getFileName().toString();
         // Make sure "/modules/<moduleName>" is created
-        Objects.requireNonNull(createModulesNode(MODULES + moduleName, moduleDir));
-        // Skip the first path (it's always the given root directory).
-        try (Stream<Path> contentsStream = Files.walk(moduleDir).skip(1)) {
+        Objects.requireNonNull(findNode(MODULES + moduleName));
```
AlanBateman: This will throw NPE if the module is not found in the /modules tree. Pre-existing issue, that I don't think is possible right now, but would be better to throw IAE here.

src/java.base/share/classes/jdk/internal/jrtfs/ExplodedImage.java
```diff
@@ -210,90 +215,122 @@ public synchronized Node findNode(String name) {
...
         } catch (IOException x) {
-            // Since the path references a file errors should not be ignored.
+            // Since the path references a file, errors should not be ignored.
             throw new UncheckedIOException(x);
         }
     }
 
-    private static boolean isNonEmptyModulesName(String name) {
+    // Ensures this is a name taking form /modules/... with no trailing slash.
+    private static boolean isRealPathName(String name) {
```
AlanBateman: "real path" has connotations, suggesting it might be testing if the path is a canonicalized path. Something simple isModulesPath is okay or leave it as it was.

src/java.base/share/classes/jdk/internal/jrtfs/ExplodedImage.java
```diff
@@ -210,90 +215,122 @@ public synchronized Node findNode(String name) {
...
+        PathResolution resolution = resolveRealPath(name);
+        if (resolution == null) {
             return null;
         }
         // This can still return null for hidden files.
-        return createModulesNode(name, path);
+        return createRealPathNode(name, resolution);
+    }
+
+    static final class PathResolution {
```
AlanBateman: The naming is a bit awkward, using something like ResolvedPaths would make it easier to read, and would make the field in PathNode easier to understand too.

src/java.base/share/classes/jdk/internal/jrtfs/ExplodedImage.java
liach: Hmm, I don't know why this should throw IAE - this is used at initialization of Jrtfs, throwing an IAE through and wrapping in an InternalError is not helpful either.

I think we can safely eliminate this call as the findNode call in packages loop should accomplish the same effect.

src/java.base/share/classes/jdk/internal/jrtfs/ExplodedImage.java
liach: I think `PathInModules` is the best way to describe these paths.

test/jdk/jdk/internal/jrtfs/SystemImageTest.java
liach: Removed comment, created tracking issue https://bugs.openjdk.org/browse/JDK-8386842

xuemingshen-oracle:
I see the value of ResolvedPaths as a bundle for resolving a logical /modules/... name into regular/preview/selected physical paths. It makes the resolution result explicit and avoids reconstructing preview paths in multiple places.

However, I wonder if we need ResolvedPaths as a persistent layer on PathNode. The regular-vs-preview policy already seems centralized in resolvePathInModules()/candidatePath(), with createPathInModulesNodeIfValid()
handling node creation.

The ResolvedPaths is mostly a tuple. For file/resource nodes, only selected is used after construction: content and size come from selected, and the node never re-resolves between regular and preview. For directory nodes, regular/preview are only needed to lazily initialize childNames; after childNames is initialized, later listings use the cached childNames and no longer need those paths for enumeration.

So I think the important cleanup is the centralized resolution logic itself, not necessarily carrying regular/preview/selected on every real /modules PathNode. Could this be simplified so that the resolved paths stay local to node creation and/or directory completion, instead of becoming a long-lived PathNode field?

That said, I have not touched this part of the code for 10+ years :-)  so please take my comment with the appropriate grain of salt. This is more of a design question than an objection.

liach:
I have teared down `ResolvedPaths` per your suggestion. Indeed, in the end, the detection of resource, directory, and link node types becomes more straightforward.

AlanBateman (approved):
The latest version, after Sherman's suggestions, looks much better.

vicente-romero-oracle (approved):
looks sensible to me

src/java.base/share/classes/jdk/internal/jrtfs/ExplodedImage.java
```diff
@@ -210,90 +231,109 @@ public synchronized Node findNode(String name) {
...
+        PathNode node;
+        if (attrs.isRegularFile()) {
+            Path f = selected.getFileName();
+            // Only reject "marker files", doesn't apply to directories
+            if (f.toString().startsWith("_the.")) {
+                return null;
+            }
+            node = new PathNode(name, selected, attrs);
+        } else if (attrs.isDirectory()) {
+            List<Path> directories = Stream.of(regularPath, previewPath)
```
xuemingshen-oracle, line 324: Should we handle file-vs-directory conflicts like the following here?

```
modules/m/foo/                         // regular directory
modules/m/META-INF/preview/foo         // preview regular file
```

or the reverse case:

```
modules/m/foo                          // regular file
modules/m/META-INF/preview/foo/        // preview directory
```

For example, should we reject these as an invalid layout during node creation, instead of letting `Files.newDirectoryStream(absPath)` throw later from `collectChildNodeNames()`?

I would expect us to detect and handle this earlier.

Not a blocker, though.

liach: Do you recall how ImageReader is handling this situation right now? Does it ignore such inconsistencies or throw eagerly or fail latently like ExplodedImage?

I'm hesitant to change ExplodedImage when we can't ensure parity with ImageReader.

AlanBateman: > I would expect us to detect and handle this earlier.

The only files in META-INF/preview will be the class files for the wrapper classes, optional and java.time. There aren't any non-class resources. A bug in the JDK build or something making changes outside of build could create these weird scenario but I don't think we should worry about that now. We can do improvements on this once the code is in main line.

xuemingshen-oracle (approved):

src/java.base/share/classes/jdk/internal/jrtfs/ExplodedImage.java
liach: Created https://bugs.openjdk.org/browse/JDK-8386905 to further investigate this potential parity in both contexts.

liach:
Thanks for the reviews!

/integrate

PR #2547, opened 2026-06-15
8385886: [lworld] C2 VectorAPI/value type: bad deopt state leads to wrong result

marc-chevalier:
The whole problem is there:

https://github.com/openjdk/valhalla/blob/1ca1c8d9301215d43eb0d012cdc42ddbe7549d14/src/hotspot/share/opto/vector.cpp#L204-L210

We are rebuilding the top of the bytecode stack before the call by pushing the arguments. This will end up as the debug inputs for the allocation in `kit.box_vector` (throught the chain `GraphKit::box_vector` -> `GraphKit::set_edges_for_java_call` -> `GraphKit::add_safepoint_edges`). Having the wrong inputs on the allocation means that if it deopts, the bytecode state will be restored wrong.

In a case such as in the test, one of the local is scalarized and takes only one slot on the expression stack (in the bytecode). Meanwhile, for C2, the value is passed scalarized to the call, so we have inputs for the buffer, the null marker, and each field. The current implementation has two issues, the first one is that it iterates with the size of `domain_sig()`, that is the signature as it appears in the Java code. In the example, that will be `MyValue` and `IntVector`, thus a length of 2. But the inputs of `call` will reflect the signature `domain_cc()`, that is with the scalarized calling convention, which will be in the example: oop buffer for `MyValue`, null marker, the integer field, and `IntVector` (length = 4). So, the current code fails to gather all locals entirely. But replacing `domain_sig` with `domain_cc` isn't enough either. If we do that, we do collect everything, but they end up as independent values on the bytecode expression stack. This is visible in the opto assembly:
```
19b     call,static  wrapper for: new_array_blob (C2 runtime)
        # ...::test @ bci:54 (line 45) L[0]=RBP L[1]=rsp + #4 L[2]=#ScObj0 L[3]=#ScObj1 L[4]=#ScObj2 L[5]=#ScObj3 STK[0]=#null STK[1]=#1 STK[2]=rsp + #4 STK[3]=#ScObj3
        # ScObj0 ...$MyValue={ [null marker :-1]=#1 [a :0]=rsp + #4 }
        # ScObj1 jdk/incubator/vector/IntVector128={ [payload :0]=rsp + #32 }
        # ScObj2 jdk/incubator/vector/IntVector128={ [payload :0]=rsp + #16 }
        # ScObj3 jdk/incubator/vector/IntVector128={ [payload :0]=rsp + #48 }
        # OopMap {off=416/0x1a0}
```
Here, it looks like the stack should have 4 slots (instead of 2), and we can recognize the slots 0, 1 and 2 that forms actually the value `myVal`. But as a result, after deoptimization, the interpreter is given a state that is not correct.

So, on top of actually iterating over all the inputs of `call` we also need to group them inside an `InlineType` (that will be scalarized in safepoint later, thanks to `InlineTypeNode` logic). The logic is quite simple: we iterate over the Java arguments, and when it is scalarized, we gather the pieces using the type to know how many fields to get. Otherwise, the next input is the argument by itself. Doing so, we can see that the opto assembly looks more reasonable:
```
19b     call,static  wrapper for: new_array_blob (C2 runtime)
        # ...::test @ bci:54 (line 37) L[0]=RBP L[1]=rsp + #4 L[2]=#ScObj0 L[3]=#ScObj1 L[4]=#ScObj2 L[5]=#ScObj3 STK[0]=#ScObj0 STK[1]=#ScObj3
        # ScObj0 ...$MyValue={ [null marker :-1]=#1 [a :0]=rsp + #4 }
        # ScObj1 jdk/incubator/vector/IntVector128={ [payload :0]=rsp + #32 }
        # ScObj2 jdk/incubator/vector/IntVector128={ [payload :0]=rsp + #16 }
        # ScObj3 jdk/incubator/vector/IntVector128={ [payload :0]=rsp + #48 }
        # OopMap {off=416/0x1a0}
```

The bytecode is given a stack containing the buffered `MyValue` and the `IntVector`: all is fine!

Thanks,
Marc

---------

---------
### Progress

### Issue
 * [JDK-8385886](https://bugs.openjdk.org/browse/JDK-8385886): [lworld] C2 VectorAPI/value type: bad deopt state leads to wrong result (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2547/head:pull/2547` \
`$ git checkout pull/2547`

`$ git checkout pull/2547` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2547/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2547`

`$ git pr show -t 2547`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2547.diff">https://git.openjdk.org/valhalla/pull/2547.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2547#issuecomment-4707682791)
</details>

### DIFF
--- a/src/hotspot/share/opto/vector.cpp
+++ b/src/hotspot/share/opto/vector.cpp
@@ -207,19 +207,9 @@ void PhaseVector::scalarize_vbox_node(VectorBoxNode* vec_box) {
       uint in_idx = TypeFunc::Parms;
       for (uint parm_idx = 0; parm_idx < nargs; parm_idx++) {
         if (call->method()->is_scalarized_arg(static_cast<int>(parm_idx))) {
-          bool nullable = call->tf()->domain_sig()->field_at(in_idx)->maybe_null();
-          ciInlineKlass* vk = call->tf()->domain_sig()->field_at(in_idx)->inline_klass();
-          InlineTypeNode* it = InlineTypeNode::make_uninitialized(gvn, vk, !nullable);
-          it->set_oop(gvn, call->in(in_idx));
-          in_idx++;
-          if (nullable) {
-            it->set_null_marker(gvn, call->in(in_idx));
-            in_idx++;
-          }
-          for (int field_idx = 0; field_idx < vk->nof_nonstatic_fields(); field_idx++) {
-            it->set_field_value(field_idx, call->in(in_idx));
-            in_idx++;
-          }
+          bool nullable = call->tf()->domain_sig()->field_at(TypeFunc::Parms + parm_idx)->maybe_null();
+          ciInlineKlass* vk = call->tf()->domain_sig()->field_at(TypeFunc::Parms + parm_idx)->inline_klass();
+          InlineTypeNode* it = InlineTypeNode::make_from_multi(&kit, call, vk, in_idx, true, !nullable);
           kit.push(gvn.transform(it));
         } else {
           kit.push(call->in(in_idx));
--- a/src/hotspot/share/opto/vector.cpp
+++ b/src/hotspot/share/opto/vector.cpp
@@ -204,9 +204,29 @@ void PhaseVector::scalarize_vbox_node(VectorBoxNode* vec_box) {
       // Adjust JVMS from post-call to pre-call state: put args on stack
       uint nargs = call->method()->arg_size();
       kit.ensure_stack(kit.sp() + nargs);
-      for (uint i = TypeFunc::Parms; i < call->tf()->domain_sig()->cnt(); i++) {
-        kit.push(call->in(i));
+      uint in_idx = TypeFunc::Parms;
+      for (uint parm_idx = 0; parm_idx < nargs; parm_idx++) {
+        if (call->method()->is_scalarized_arg(static_cast<int>(parm_idx))) {
+          bool nullable = call->tf()->domain_sig()->field_at(in_idx)->maybe_null();
+          ciInlineKlass* vk = call->tf()->domain_sig()->field_at(in_idx)->inline_klass();
+          InlineTypeNode* it = InlineTypeNode::make_uninitialized(gvn, vk, !nullable);
+          it->set_oop(gvn, call->in(in_idx));
+          in_idx++;
+          if (nullable) {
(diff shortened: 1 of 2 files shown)


src/hotspot/share/opto/vector.cpp
```diff
@@ -204,9 +204,29 @@ void PhaseVector::scalarize_vbox_node(VectorBoxNode* vec_box) {
       // Adjust JVMS from post-call to pre-call state: put args on stack
       uint nargs = call->method()->arg_size();
       kit.ensure_stack(kit.sp() + nargs);
-      for (uint i = TypeFunc::Parms; i < call->tf()->domain_sig()->cnt(); i++) {
-        kit.push(call->in(i));
+      uint in_idx = TypeFunc::Parms;
+      for (uint parm_idx = 0; parm_idx < nargs; parm_idx++) {
+        if (call->method()->is_scalarized_arg(static_cast<int>(parm_idx))) {
+          bool nullable = call->tf()->domain_sig()->field_at(in_idx)->maybe_null();
```
merykitty: We have a method in `InlineTypeNode` that collects call inputs just like this, it is used when initializing the state of a `Parse`. Can we reuse it?

marc-chevalier: Can you point to what you have in mind? I looked for such a function (because I could not believe we weren't doing that many times already), but all the similar code required some significant transformation wrt their size. But it is very likely I've missed some!

marc-chevalier: Ah yes, `InlineTypeNode::make_from_multi` might do the trick. I've missed that, my bad. I'll try.

marc-chevalier: There, simpler, nicer, correcter. Tests look happy enough.

merykitty (approved):
Really nice, it looks reasonable to me.

dean-long (approved):

marc-chevalier:
/integrate

Thanks @merykitty and @dean-long!

PR #2548, opened 2026-06-15
8386702: [lworld] Reflection APIs should not allow strictly-initialized final fields to be mutated

AlanBateman:
Using the reflection APIs to mutate a strictly-initialized final instance field should fail with IllegalAccessException.

The list of conditions in Field.set and setAccessible are also updated.

---------

---------
### Progress

### Issue
 * [JDK-8386702](https://bugs.openjdk.org/browse/JDK-8386702): [lworld] Reflection APIs should not allow strictly-initialized final fields to be mutated (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2548/head:pull/2548` \
`$ git checkout pull/2548`

`$ git checkout pull/2548` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2548/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2548`

`$ git pr show -t 2548`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2548.diff">https://git.openjdk.org/valhalla/pull/2548.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2548#issuecomment-4718207085)
</details>

### DIFF
--- a/src/java.base/share/classes/java/lang/reflect/Field.java
+++ b/src/java.base/share/classes/java/lang/reflect/Field.java
@@ -1523,12 +1523,6 @@ class Field extends AccessibleObject implements Member {
     private void preSetFinal(Class<?> caller, boolean unreflect) throws IllegalAccessException {
         assert isFinalInstanceInNormalClass();
 
-        // strictly-initialized final fields cannot be mutated
-        if (isStrictInit()) {
-            throw new IllegalAccessException(cannotSetFieldMessage("Can not", unreflect)
-                    + " (strictly-initialized field)");
-        }
-
         if (caller != null) {
             // check if declaring class in package that is open to caller, or public field
             // and declaring class is public in package exported to caller
--- a/src/java.base/share/classes/java/lang/reflect/Field.java
+++ b/src/java.base/share/classes/java/lang/reflect/Field.java
@@ -190,6 +190,7 @@ class Field extends AccessibleObject implements Member {
      * <li>static final fields declared in any class or interface</li>
      * <li>final fields declared in a {@linkplain Class#isRecord() record}</li>
      * <li>final fields declared in a {@linkplain Class#isHidden() hidden class}</li>
+     * <li>{@linkplain #isStrictInit() strictly-initialized} final fields</li>
      * </ul>
      * <p>If this reflected object represents a non-static final field in a class that
      * is not a record class or hidden class, then enabling access will enable read
@@ -841,6 +842,7 @@ class Field extends AccessibleObject implements Member {
      * <li>{@code D} is not a {@linkplain Class#isHidden() hidden class}.</li>
      * <li>{@code D} is not a {@linkplain Class#isValue() value class}.</li>
      * <li>The field is non-static.</li>
+     * <li>The field is not a {@linkplain #isStrictInit() strictly-initialized} field. </li>
      * </ul>
      *
      * <p>If any of the above conditions is not met, this method throws an
@@ -870,6 +872,7 @@ class Field extends AccessibleObject implements Member {
      * <li>{@code D} is not a {@linkplain Class#isRecord() record class}.</li>
      * <li>{@code D} is not a {@linkplain Class#isHidden() hidden class}.</li>
      * <li>The field is non-static.</li>
+     * <li>The field is not a {@linkplain #isStrictInit() strictly-initialized} field. </li>
      * </ul>
      *
      * <p>If any of the above conditions is not met, this method throws an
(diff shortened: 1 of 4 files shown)


src/java.base/share/classes/java/lang/reflect/Field.java
```diff
@@ -190,6 +190,7 @@ Field copy() {
      * <li>static final fields declared in any class or interface</li>
      * <li>final fields declared in a {@linkplain Class#isRecord() record}</li>
      * <li>final fields declared in a {@linkplain Class#isHidden() hidden class}</li>
+     * <li>{@linkplain #isStrictInit() strictly-initialized} final fields</li>
```
liach, line 194: The other examples from identity requirements use a preview block, or as in `SwitchBootstraps`, prefixed with "if preview features are enabled, ..."

src/java.base/share/classes/java/lang/reflect/Field.java
```diff
@@ -1520,6 +1523,12 @@ void checkAllowedToUnreflectFinalSetter(Class<?> caller) throws IllegalAccessExc
     private void preSetFinal(Class<?> caller, boolean unreflect) throws IllegalAccessException {
         assert isFinalInstanceInNormalClass();
 
+        // strictly-initialized final fields cannot be mutated
+        if (isStrictInit()) {
```
liach: I think the check is best done here, which should fix all MethodHandle too: https://github.com/AlanBateman/valhalla/blob/1890ec9eb8038d92f5fba3107085cafabdc9f21e/src/java.base/share/classes/java/lang/invoke/MethodHandles.java#L3443

AlanBateman: This is a good question. We went around this a few times during JEP 500 and decided to send the conditional cases through checkAllowedToUnreflectFinalSetter, the benefit being that everything related to the CLI options to allow mutation of finals is in one place. You are right that the strict final fields case can't be changed by the CLI options so isStrictInit can be checked eariler. It means that newFieldAccessor must be created without a setter, which is a good thing.

src/java.base/share/classes/java/lang/reflect/Field.java
AlanBateman: This is a list of conditions and the "strictly-initialized" will have "PREVIEW" superscript.  So it's like Field.set which links to value class with the  "PREVIEW" superscript. I think it looks okay but I agree some prefixing with "if preview features" might be a bit clearer.

src/java.base/share/classes/java/lang/reflect/Field.java
ExE-Boss: I take it that “best done here” refers to having `Field​::isTrustedFinal()` return `true` for strictly initialised fields, the correct fix for which would be to update `fieldDescriptor​::is_trusted_final()` to include `is_strict()` in the `||` portion of:
https://github.com/openjdk/valhalla/blob/9cd13e6903a025ee82be54217ab76256b392a9ce/src/hotspot/share/runtime/fieldDescriptor.cpp#L46-L50

e.g.:
```c++
bool fieldDescriptor::is_trusted_final() const {
  InstanceKlass* ik = field_holder();
  return is_final() && (is_static() || is_strict()
                        || ik->is_hidden() || ik->is_record() || ik->is_inline_klass()
                        || (ik->is_abstract() && !ik->is_identity_class() && !ik->is_interface()));
}
```

liach: I don't think we should expand this "trusted final" concept.

test/jdk/java/lang/reflect/Field/StrictInitFinalFields.java
```diff
@@ -0,0 +1,148 @@
...
+        // Field.get/set after setAccessible
+        field.setAccessible(true);
+        assertEquals(initialValue, field.get(obj));
+        assertThrows(IllegalAccessException.class, () -> field.set(obj, newValue));
+
+        // VarHandle produced from reflected field
+        var lookup = MethodHandles.lookup();
+        VarHandle vh = lookup.unreflectVarHandle(field);
+        assertEquals(initialValue, vh.get(obj));
+        assertThrows(UnsupportedOperationException.class, () -> vh.set(obj, newValue));
```
liach, line 135: Luckily VarHandle already rejects all final fields, as I last checked in JEP 500 - it has some problem with non-final strict-init, addressed in #2550.

src/java.base/share/classes/java/lang/reflect/Field.java
liach: Fair, we can assume strictly-initialized fields do not exist without preview features. We can leave it as is and change if CSR review requests so.

test/jdk/java/lang/reflect/Field/StrictInitFinalFields.java
AlanBateman: Yes, VarHandle doesn't allow it. This test is checking Field.set behavior and then unreflects to test method handle and VarHandle behavior too. I think that is okay to test here.

liach (approved):

PR #2550, opened 2026-06-15
8386677: [lworld] VarHandle does not implement strict static checks

liach:
The checks for strict statics for VarHandle was never implemented. Luckily we can reuse LazyInitializingVarHandle so the patch is not too complex.

---------

---------
### Progress

### Issue
 * [JDK-8386677](https://bugs.openjdk.org/browse/JDK-8386677): [lworld] VarHandle does not implement strict static checks (**Bug** - P2)

### Reviewers
 * [Jorn Vernee](https://openjdk.org/census#jvernee) (@JornVernee - no project role)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2550/head:pull/2550` \
`$ git checkout pull/2550`

`$ git checkout pull/2550` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2550/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2550`

`$ git pr show -t 2550`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2550.diff">https://git.openjdk.org/valhalla/pull/2550.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2550#issuecomment-4710230823)
</details>

### DIFF
--- a/src/java.base/share/classes/java/lang/invoke/LazyInitializingVarHandle.java
+++ b/src/java.base/share/classes/java/lang/invoke/LazyInitializingVarHandle.java
@@ -1,38 +1,37 @@
 /*
- *  Copyright (c) 2023, Oracle and/or its affiliates. All rights reserved.
- *  DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ * Copyright (c) 2023, 2026, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
- *  This code is free software; you can redistribute it and/or modify it
- *  under the terms of the GNU General Public License version 2 only, as
- *  published by the Free Software Foundation.  Oracle designates this
- *  particular file as subject to the "Classpath" exception as provided
- *  by Oracle in the LICENSE file that accompanied this code.
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
  *
- *  This code is distributed in the hope that it will be useful, but WITHOUT
- *  ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
- *  FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
- *  version 2 for more details (a copy is included in the LICENSE file that
- *  accompanied this code).
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
  *
- *  You should have received a copy of the GNU General Public License version
- *  2 along with this work; if not, write to the Free Software Foundation,
- *  Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
  *
- *   Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
- *  or visit www.oracle.com if you need additional information or have any
- *  questions.
- *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
  */
 
 package java.lang.invoke;
 
+import jdk.internal.vm.annotation.DontInline;
 import jdk.internal.vm.annotation.ForceInline;
 import jdk.internal.vm.annotation.Stable;
 
 import java.util.Optional;
 
 import static java.lang.invoke.MethodHandleStatics.UNSAFE;
-import static java.lang.invoke.MethodHandleStatics.uncaughtException;
 
 /**
  * A lazy initializing var handle. It lazily initializes the referenced class before
@@ -41,17 +40,14 @@ import static java.lang.invoke.MethodHandleStatics.uncaughtException;
  */
 final class LazyInitializingVarHandle extends VarHandle {
 
-    // Implementation notes:
-    // We put a barrier on both target() (for VH form impl direct invocation)
-    // and on getMethodHandle() (for indirect VH invocation, toMethodHandle)
-    private final VarHandle target;
-    private final Class<?> refc;
-    private @Stable boolean initialized;
+    private final VarHandles.StaticFieldVarHandle target;
+    private final boolean strictInit;
+    private @Stable boolean fullyInitialized;
 
-    LazyInitializingVarHandle(VarHandle target, Class<?> refc) {
+    LazyInitializingVarHandle(VarHandles.StaticFieldVarHandle target, boolean strictInit) {
         super(target.vform, target.exact);
         this.target = target;
-        this.refc = refc;
+        this.strictInit = strictInit;
     }
 
     @Override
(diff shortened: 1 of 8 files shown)


src/java.base/share/classes/java/lang/invoke/MethodHandles.java
```diff
@@ -3961,7 +3961,7 @@ private VarHandle getFieldVarHandleCommon(byte getRefKind, byte putRefKind,
                 refc = lookupClass();
             }
             return VarHandles.makeFieldHandle(getField, refc,
-                                              this.allowedModes == TRUSTED && !getField.isTrustedFinalField());
+                                              this.allowedModes == TRUSTED);
```
JornVernee, line 3969: Is this no longer needed? Are all trusted final fields strict init now?

src/java.base/share/classes/java/lang/invoke/LazyInitializingVarHandle.java
```diff
@@ -61,31 +57,50 @@ MethodType accessModeTypeUncached(AccessType at) {
...
+            initialize(read, write);
+        }
         return target;
     }
 
-    @Override
-    @ForceInline
-    VarHandle target() {
-        ensureInitialized();
-        return target;
```
JornVernee, line 72: What about static method access? `onStaticFieldAccess` seems to be only used for field access, and the regular access to the target seems to skip the `ensureInitialized` check now.

src/java.base/share/classes/java/lang/invoke/MethodHandles.java
liach: No, I moved this check to the beginning of `VarHandles.makeFieldHandle` so IMPL_LOOKUP cannot create strict-final writing VarHandle instances:

```
boolean noWriting = f.isFinal() && (!trusted || f.isStrictInit());
```

src/java.base/share/classes/java/lang/invoke/LazyInitializingVarHandle.java
liach: `target` is completely eliminated now. `IndirectVarHandle` uses `asDirect` before passing a VH to the implementation MH; `LazyInitializingVarHandle` uses `onStaticFieldAccess` or old `target` once you are within the implementation MH/MemberName, so the two mechanisms were quite different and better separated. All other VHs don't need this class initialization barrier.

JornVernee: Ah you're right. I'm not reading the changes correctly. (that's what I get for trying to review this on very little sleep yesterday :/)

src/java.base/share/classes/java/lang/invoke/MethodHandles.java
JornVernee: Right, but are `isStrictInit` and `isTrustedFinalField` equivalent now? The latter also seems to include non-strict `static final` fields for instance. 

The latter is set by the VM when resolving the member name based on the criteria at [1](https://github.com/openjdk/valhalla/blob/ce9ae1864dd19e0dd88ba2b65e8828936df01508/src/hotspot/share/runtime/fieldDescriptor.cpp#L46).

Shouldn't it be:

```
boolean noWriting = f.isFinal() && (!trusted || f.isStrictInit() || f.isTrustedFinalField());
```

This would exclude `IMPL_LOOKUP` from creating var handles that can write to trusted finals as it does now.

`MethodHandles::unreflectField` also still uses `isTrustedFinalField`.

liach: No, `isTrustedFinalField` never includes strict fields. It is a blanket flag set by the VM for record/hidden class fields, but that doesn't respect the ciField package trusting, and thus is only a partial filter.

JornVernee: I gave it a try, and with this patch `IMPL_LOOKUP` can create var handles that can write to `static final` fields, whereas with the current `lworld` code it fails with an `UnsupportedOperationException`. This seems like an unintended change.

AlanBateman: I have to admit that when I initially saw the parameter to makeFieldHandle named "trusted" that I assumed it meant a trusted-final when it actually means produced from IMPL_LOOKUP.

liach: > This seems like an unintended change.

I intentionally made that change to secure against modifying strict or trusted finals even with IMPL_LOOKUP. I will roll it back if you think it's too strict.

JornVernee: It's not too strict, but it seems like you've not implemented it correctly. It should be:

```
boolean noWriting = f.isFinal() && (!trusted || f.isStrictInit() || f.isTrustedFinalField());
```

Otherwise trusted finals are writable.

Note that `IMPL_LOOKUP` currently _does not_ allow writing to trusted finals already (thanks to the `!getField.isTrustedFinalField()` here).

liach: I have fixed the missing isTrustedFinalField call in the latest update and added cases for those.

src/java.base/share/classes/java/lang/invoke/X-VarHandle.java.template
```diff
@@ -622,7 +616,7 @@ final class VarHandle$InputType$s {
 
         @ForceInline
         static boolean compareAndSet(VarHandle ob, $type$ expected, $type$ value) {
-            FieldStaticReadWrite handle = (FieldStaticReadWrite) ob.target();
+            FieldStaticReadWrite handle = (FieldStaticReadWrite) ob.onStaticFieldAccess(true, true);
```
JornVernee: CompareAndSet do both a read and write, so here you pass `true, true`. This will end up calling `notifyStrictStaticAccess` which then clears the 'unset' bit of the field for writes. However, a CAS is not guaranteed to do a write. Isn't that problematic? The VM would consider the field initialized after a failed CAS.

Would be nice if you could add a test for that.

src/java.base/share/classes/java/lang/invoke/VarHandles.java
```diff
@@ -130,74 +132,105 @@ else if (type == double.class) {
...
                 }
             } else {
-                return f.isFinal() && !isWriteAllowedOnFinalFields
+                return noWriting
                         ? new VarHandleReferences.FieldStaticReadOnly(decl, base, foffset, type, f.isNullRestricted())
                         : new VarHandleReferences.FieldStaticReadWrite(decl, base, foffset, type, f.isNullRestricted());
             }
         }
         else if (type == boolean.class) {
-            return maybeAdapt(f.isFinal() && !isWriteAllowedOnFinalFields
```
JornVernee, line 163: Shouldn't we keep the `maybeAdapt` here? It's used for testing indirect var handles.

src/java.base/share/classes/java/lang/invoke/LazyInitializingVarHandle.java
```diff
@@ -95,42 +110,9 @@ public Optional<VarHandleDesc> describeConstable() {
...
-
-        try {
-            return MH_ensureInitialized = MethodHandles.lookup().findVirtual(
-                    LazyInitializingVarHandle.class,
-                    "ensureInitialized",
-                    MethodType.methodType(void.class));
-        } catch (Throwable ex) {
-            throw uncaughtException(ex);
-        }
+        return super.getMethodHandleUncached(accessMode);
```
JornVernee: `target.getMethodHandle` goes to the target, but `super.getMethodHandleUncached` goes to our `vform`, which is the target's `vform`. What's the difference? (besides the lack of caching  in the targets `methodHandleTable` in the latter case) Both will end up returning the same method handle right?

JornVernee (commented):
Had a few more questions (inline).

src/java.base/share/classes/java/lang/invoke/LazyInitializingVarHandle.java
JornVernee: After looking with a debugger, it seems that we're safe from skipping the init check when calling `toMethodHandle` because the `get` of the target will end up calling `onStaticFieldAccess` for the original VarHandle, which is this one, so we end up doing the init check.

Looks like the old code was doing a redundant `ensureInitialized` call in that case.

I'm still a little puzzled about why the update to `methodHandleTable` in `initialize` is needed. It looks like this prevents two copies of identical tables floating around (one in the `target`, and one in this VH). It would be great if you could put a comment on `this.methodHandleTable = target.methodHandleTable;` about that.

src/java.base/share/classes/java/lang/invoke/X-VarHandle.java.template
liach: I can degrade the calls to be reporting only read for cas: after the first write, we don't really care about future writes.

JornVernee: Right, but shouldn't we be able to read after a _successful_ CAS? Reporting the write seems needed as well.

I just had a look with a debugger, and it seems the ordering of the read check before the write check in `initialize` saves us. The read check will throw an exception, so the write check (which clears the bit) never actually happens.

I think it would still be good to add a test case for CAS, and maybe a comment that the read check should be ordered before the write check in `initialize`.

JornVernee: Ah, I see what you're saying now: CAS only works on already initialized fields any way, since it does a read first. Okay, that makes sense.

src/java.base/share/classes/java/lang/invoke/VarHandles.java
liach: It destroys the StaticFieldVarHandle type. I think keeping that type through so LazyInitializingVarHandle can use attributes from StaticFieldVarHandle is more valuable.

src/java.base/share/classes/java/lang/invoke/LazyInitializingVarHandle.java
liach: I will remove the methodHandleTable sharing path. It indeed isn't that useful.

src/java.base/share/classes/java/lang/invoke/VarHandles.java
ExE-Boss: The `maybeAdapt` call happens in `makeFieldHandle`, passing the wrapping `LazyInitializingVarHandle` or the target handle if the class is fully initialised.

JornVernee: True. Ok, let's keep it like this then

liach: The maybeAdapt ideally happens wherever a VarHandle is returned. Before the adaption wraps both the underlying FieldStaticReadOnly/Write and the LazyInitializingVarHandle.

Unfortunately, now we need this `StaticFieldVarHandle` type, so we can't wrap the FieldStaticReadOnly/Write; we are still wrapping the LazyInitializingVarHandle.

JornVernee (approved):
Thanks for the cleanups, the latest version looks much simpler!

vicente-romero-oracle (approved):
looks sensible to me

liach:
Thanks for the reviews!

/integrate

PR #2554, opened 2026-06-16
8386237: [lworld] compiler/valhalla/inlinetypes/TestIntrinsics.java sub-tests fail with merge_jdk-28+1

marc-chevalier:
When https://github.com/openjdk/jdk/pull/31162 landed in Valhalla, it added this bit of code in `GrraphKit::gen_checkcast`:

https://github.com/openjdk/valhalla/blob/a339024e1f415bfd7c6875538f2adff77bf087f0/src/hotspot/share/opto/graphKit.cpp#L3914-L3916

While it manifests as a memory problem (and indeed, the memory graph is all fucked up), the causes lies in the type system, and more precisely, in `as_instance_type` invoked from `gen_checkcast` as well:

https://github.com/openjdk/valhalla/blob/a339024e1f415bfd7c6875538f2adff77bf087f0/src/hotspot/share/opto/graphKit.cpp#L3692

The problem is that `TypeAryKlassPtr::as_instance_type` will trust all the `is_flat()`, `is_not_flat()`, `is_not_null_free()`, `is_atomic()`, `null_free()` from an unrefined type:
https://github.com/openjdk/valhalla/blob/a339024e1f415bfd7c6875538f2adff77bf087f0/src/hotspot/share/opto/type.cpp#L6975-L6990

But unrefined types represent Java types: they don't express notions such as flatness, atomicity... These flags are arbitrary in unrefined types.

We should not! Before https://github.com/openjdk/jdk/pull/31162, this was unoticed because when parsing some array load or store, we were not reusing the casted value from `gen_checkcast`: it was indeed found to be stupidly small (like null or empty), because it was the intersection of some actual type, and the type wrongly created by `TypeAryKlassPtr::as_instance_type`, but it had no impact that we were not replacing anything in the map, this code in `Parse::array_store`
https://github.com/openjdk/valhalla/blob/a339024e1f415bfd7c6875538f2adff77bf087f0/src/hotspot/share/opto/parse2.cpp#L212-L214
would get the uncasted (more correct) version, freshly from the map.

So, we need to fix `TypeAryKlassPtr::as_instance_type`. While putting everything to "unknown" is sound, it is not a solution here: a lot of asserts expect us to be way more specific. I've managed to reuse `TypeInstKlassPtr::flat_in_array()`, but I had to make it more precise not to hit asserts such as
https://github.com/openjdk/valhalla/blob/a339024e1f415bfd7c6875538f2adff77bf087f0/src/hotspot/share/opto/parse2.cpp#L91-L92

I think we should consider changing the default values for unrefined types to some clearly incorrect combinations, such as `_flat == _not_flat == true`, so that if we use unrefined types as refined accidentally, it would probably be a train wreck immediately, and not accidentally correct, or plausible.

I also had to fix `TypePtr::meet_aryptr`: in the case `BotPTR`, we used to set `ptr` to `NotNull` in the case the result is not exact. This was breaking the verification of meet properties. Once again, because of damn dual!!! But clearly, we had a `TopPTR` in the non-dual, and I don't see a reason why `meet_aryptr` between bottom and anything should be less than bottom. Bottom is the universe! We shouldn't get less by adding something in it. And tests seem ok with it...

Thanks,
Marc

---------

---------
### Progress

### Issue
 * [JDK-8386237](https://bugs.openjdk.org/browse/JDK-8386237): [lworld] compiler/valhalla/inlinetypes/TestIntrinsics.java sub-tests fail with merge_jdk-28+1 (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2554/head:pull/2554` \
`$ git checkout pull/2554`

`$ git checkout pull/2554` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2554/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2554`

`$ git pr show -t 2554`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2554.diff">https://git.openjdk.org/valhalla/pull/2554.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2554#issuecomment-4727347829)
</details>

### DIFF
--- a/src/hotspot/share/opto/node.cpp
+++ b/src/hotspot/share/opto/node.cpp
@@ -3201,7 +3201,7 @@ void TypeNode::create_halt_path(PhaseIterGVN* igvn, Node* c, PhaseIdealLoop* loo
   }
 
   stringStream ss;
-  ss.print("dead path discovered by TypeNode during %s: this:%d; c:%d  method:%s", phase_str, this->_idx, c->_idx, igvn->C->method()->name()->as_quoted_ascii());
+  ss.print("dead path discovered by TypeNode during %s", phase_str);
 
   Node* halt = new HaltNode(c, frame, ss.as_string(igvn->C->comp_arena()));
   if (loop == nullptr) {
--- a/src/hotspot/share/opto/node.cpp
+++ b/src/hotspot/share/opto/node.cpp
@@ -3201,7 +3201,7 @@ void TypeNode::create_halt_path(PhaseIterGVN* igvn, Node* c, PhaseIdealLoop* loo
   }
 
   stringStream ss;
-  ss.print("dead path discovered by TypeNode during %s", phase_str);
+  ss.print("dead path discovered by TypeNode during %s: this:%d; c:%d  method:%s", phase_str, this->_idx, c->_idx, igvn->C->method()->name()->as_quoted_ascii());
 
   Node* halt = new HaltNode(c, frame, ss.as_string(igvn->C->comp_arena()));
   if (loop == nullptr) {
--- a/src/hotspot/share/opto/type.cpp
+++ b/src/hotspot/share/opto/type.cpp
@@ -6989,7 +6989,10 @@ const TypeOopPtr* TypeAryKlassPtr::as_instance_type(bool klass_change) const {
     // Unrefined types aren't trustworthy! Let's not mistake their ignorance for information.
     flat = false;
     // There are asserts that expect us to not be entirely naive about flatness.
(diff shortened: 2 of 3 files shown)


merykitty (approved):
Looks good to me, thanks a lot for working on this hard issue.

marc-chevalier:
/integrate

Thanks @merykitty for help and review!

PR #2562, opened 2026-06-17
8386815: [lworld] Reduce the number of constructors in IdentityException

jaikiran:
Can I please get a review of this change which specifies that the `IdentityException` throws a `NullPointerException` if the `Class` passed to its constructor is null?

This change is being proposed as part of the JEP-401 standard library changes that are currently under review at https://github.com/openjdk/jdk/pull/31123.

A new jtreg test has been introduced to verify the basic semantics of this exception class.

---------

---------
### Progress

### Issue
 * [JDK-8386815](https://bugs.openjdk.org/browse/JDK-8386815): [lworld] Reduce the number of constructors in IdentityException (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2562/head:pull/2562` \
`$ git checkout pull/2562`

`$ git checkout pull/2562` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2562/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2562`

`$ git pr show -t 2562`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2562.diff">https://git.openjdk.org/valhalla/pull/2562.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2562#issuecomment-4727751575)
</details>

### DIFF
--- a/src/java.base/share/classes/java/lang/IdentityException.java
+++ b/src/java.base/share/classes/java/lang/IdentityException.java
@@ -48,7 +48,11 @@ public final class IdentityException extends RuntimeException {
     /**
      * Create an {@code IdentityException} with the class name and default message.
      *
-     * @param clazz the class of the object
+     * @apiNote
+     * This constructor does not verify that the given {@code clazz} is a value class.
+     *
+     * @param clazz the class whose name will be included in the exception message
+     * @throws NullPointerException if {@code clazz} is null
      */
     public IdentityException(Class<?> clazz) {
         super(clazz.getName() + " is not an identity class");
--- a/src/java.base/share/classes/java/lang/IdentityException.java
+++ b/src/java.base/share/classes/java/lang/IdentityException.java
@@ -48,45 +48,11 @@ public final class IdentityException extends RuntimeException {
     }
 
     /**
-     * Create an {@code IdentityException} with the class name and default message.
-     *
-     * @apiNote
-     * This constructor does not verify that the given {@code clazz} is a value class.
-     *
-     * @param clazz the class whose name will be included in the exception message
-     * @throws NullPointerException if {@code clazz} is null
-     */
-    public IdentityException(Class<?> clazz) {
-        super(clazz.getName() + " is not an identity class");
-    }
-
-    /**
      * Create an {@code IdentityException} with a message.
      *
      * @param  message the detail message; can be {@code null}
      */
     public IdentityException(String message) {
         super(message);
-    }
-
-    /**
-     * Create an {@code IdentityException} with a cause.
-     *
-     * @param  cause the cause; {@code null} is permitted, and indicates
-     *               that the cause is nonexistent or unknown.
-     */
-    public IdentityException(Throwable cause) {
-        super(cause);
-    }
-
-    /**
-     * Create an {@code IdentityException} with a message and cause.
-     *
-     * @param  message the detail message; can be {@code null}
-     * @param  cause the cause; {@code null} is permitted, and indicates
-     *               that the cause is nonexistent or unknown.
-     */
-    public IdentityException(String message, Throwable cause) {
-        super(message, cause);
     }
 }
--- a/test/jdk/java/lang/IdentityExceptionBasicTest.java
+++ b/test/jdk/java/lang/IdentityExceptionBasicTest.java
@@ -1,0 +1,82 @@
+/*
+ * Copyright (c) 2026, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.
+ *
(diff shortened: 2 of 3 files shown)


jaikiran:
Hello Joe @jddarcy, in context of the discussion about the constructors in this exception class https://github.com/openjdk/jdk/pull/31123#discussion_r3416119896, I can also remove the unused constructors in this PR, if you and others suggest so.

test/jdk/java/lang/IdentityExceptionBasicTest.java
liach, line 1: I thought the jtreg directive block is supposed to appear before imports instead of after. Is there a convention on this?

liach (approved):

test/jdk/java/lang/IdentityExceptionBasicTest.java
jaikiran: Hello Chen, there are several tests in the JDK which place the test definition right above the test class declaration. There are many other tests too which place this before the imports.

To me, keeping it closer to the test class makes it easier to spot the test details.

jaikiran:
> Hello Joe @jddarcy, in context of the discussion about the constructors in this exception class [openjdk/jdk#31123 (comment)](https://github.com/openjdk/jdk/pull/31123#discussion_r3416119896), I can also remove the unused constructors in this PR, if you and others suggest so.

On second thoughts, when working on something else in this area, I realized that there are places in the native (VM) code which can throw the `IdentityException` and I haven't verified that the constructors that I was considering to remove aren't being called from that layer. So I won't do anything related to constructor removals in this PR and will reconsider it at a later date after.

src/java.base/share/classes/java/lang/IdentityException.java
```diff
@@ -48,7 +48,11 @@ public IdentityException() {
     /**
      * Create an {@code IdentityException} with the class name and default message.
      *
-     * @param clazz the class of the object
+     * @apiNote
+     * This constructor does not verify that the given {@code clazz} is a value class.
```
AlanBateman: URISyntaxException and InvalidPathException are examples of exceptions with specialized constructors (not one of the usual 4) that throw IAE when invoked with a bad parameter. I don't know if this was looked at, and rejected, at some point.

jaikiran: I think it's reasonable to throw `IllegalArgumentException`, when `clazz` is not null and is an identity class. I had hesitated doing that for the same reason as you note - I don't have prior knowledge of these discussions.

I'll update this PR to throw the IAE, run existing tests and then wait for a few days to see if anyone says we should not throw the IAE.

liach: I think Roger maybe added this constructor out of convenience. I checked some VM code and they seem to be never using this constructor; they have various ways of constructing messages, often incorporating the action that requires identity. Maybe we can consider removing this constructor too.

AlanBateman: That might be the simplest, it would always be added later if there is a needed.

jaikiran: That makes sense. I've now updated the PR to remove that constructor and since we are now removing this constructor, I looked for references of the other 2 which take a `Throwable`. Those aren't used either. So I've removed them as well.

I've run tier1, tier2 and tier3 with this change and it completed without any related issues.

liach (approved):
Thanks for the updates!

AlanBateman (approved):
Good, and maybe change the description of the JBS/PR now as it doesn't match where this ended up.

jaikiran:
Thank you for the reviews Chen and Alan. I've updated the JBS issue title and description to match the current changes in this PR.

I'll go ahead and integrate this now.

src/java.base/share/classes/java/lang/IdentityException.java
```diff
@@ -64,25 +55,4 @@ public IdentityException(Class<?> clazz) {
...
         super(message);
     }
-
-    /**
-     * Create an {@code IdentityException} with a cause.
-     *
-     * @param  cause the cause; {@code null} is permitted, and indicates
-     *               that the cause is nonexistent or unknown.
-     */
-    public IdentityException(Throwable cause) {
```
ExE-Boss, line 74: I’d rather we kept these, as it’s annoying to have to use [`Throwable​::initCause(…)`].

[`Throwable​::initCause(…)`]: https://docs.oracle.com/en/java/javase/26/docs/api/java.base/java/lang/Throwable.html#initCause(java.lang.Throwable)

ExE-Boss (changes requested):

src/java.base/share/classes/java/lang/IdentityException.java
jaikiran: Hello @ExE-Boss, these can be reintroduced at some point, if there's a genuine need for them. For new APIs, it's better to start off small, including for exception types.

AlanBateman: I think initially it will be more likely that IdentityException will another exception's cause rather than the other way around. As you say, additional constructors could be re-added if needed.

PR #2568, opened 2026-06-19
8386963: [lworld] Improve the exception message from Object synchronization methods on value objects

jaikiran:
Can I please get a review of this change which proposes to improve the exception message in `IllegalMonitorStateException` when it gets thrown for value objects?

With this change, we now report the following exception message when value objects are used to invoke `Object.wait()` and related methods:

```
jshell> o.wait()
|  Exception java.lang.IllegalMonitorStateException: current thread is not owner
|        at Object.wait0 (Native Method)
|        at Object.wait (Object.java:433)
|        at Object.wait (Object.java:387)
|        at (#2:1)
```
There were some thoughts about using an exception message of the form `Exception java.lang.IllegalMonitorStateException: object's type java.lang.Integer is not an identity class`. But since value objects cannot be synchronized upon, we decided it would be fine (and accurate) to reuse the same "current thread is not owner" exception message that we use for identity objects.

Given the nature of this change, I haven't added any new jtreg tests for this. Existing tests in tier1, tier2 and tier3 continue to pass after this change.

---------

---------
### Progress

### Issue
 * [JDK-8386963](https://bugs.openjdk.org/browse/JDK-8386963): [lworld] Improve the exception message from Object synchronization methods on value objects (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2568/head:pull/2568` \
`$ git checkout pull/2568`

`$ git checkout pull/2568` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2568/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2568`

`$ git pr show -t 2568`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2568.diff">https://git.openjdk.org/valhalla/pull/2568.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2568#issuecomment-4749580982)
</details>


src/hotspot/share/runtime/synchronizer.cpp
```diff
@@ -321,14 +321,24 @@ static bool _no_progress_skip_increment = false;
   if ((obj)->mark().is_inline_type()) {  \
     JavaThread* THREAD = current;           \
     ResourceMark rm(THREAD);                \
```
AlanBateman: Is the RM still needed?

jaikiran: I wasn't aware what a `ResourceMark` is. Reading the code and the code comment on `ResourceMarkImpl`, I believe it was here for tracking the resource allocated in the implementation of `obj->klass()->external_name()` (please correct me if I'm wrong). With that now gone, I think the `ResourceMark` is no longer needed. I've updated the PR accordingly.

I have triggered tier testing to be sure that this doesn't cause any issues.

jaikiran: tier1, tier2, tier3 testing of the latest changes in this PR completed without any related failures.

jaikiran:
There's been some discussion on whether to throw an `IdentityException` instead of `IllegalMonitorStateException` from these `Object.wait()/notify/notifyAll` methods when invoked on value objects. There hasn't been any consensus yet and for now it seems that we will continue with the `IllegalMonitorStateException` being thrown.

I don't have any more changes planned for this PR. The change here will merely improve the exception message contained in the `IllegalMonitorStateException`. I think that part is trivial and several of us agree that the new message is more appropriate.

The other change in this PR is the removal of `ResourceMark` from these if blocks. I don't have prior knowledge of that construct, so I would like a review from Alan and/or someone who knows more about this so that we can have this change integrated into `lworld` in preparation for the JEP 401 integration.

dholmes-ora (approved):
This seems fine to me. The ResourceMarks are no longer needed.

AlanBateman (approved):

jaikiran:
Thank you David and Alan for the reviews.

PR #2569, opened 2026-06-19
8382226: [lworld] C2: Fix _copyOf/_copyOfRange intrinsic for flat abstract value class arrays

chhagedorn:
The provided test cases fail when inlining the `Array.copyOf/copyOfRange()` intrinsics where the source array is flat and from an abstract value class.

The current code checks whether the source array or the destination array klass contain oops by assuming that a flat value class array is always concrete and thus an `InlineKlass` (i.e. can call `inline_klass()`). However, we could also have abstract value class arrays that are known to be flat (see test cases). This leads to a cast assertion failure because abstract value classes are represented by an `InstanceKlass` and not an `InlineKlass`.

To fix this, I added a simple bailout when detecting an abstract flat value class array. This is a conservative correctness fix and should be revisited again post-Valhalla-integration. We have [JDK-8251971](https://bugs.openjdk.org/browse/JDK-8251971) in place for that which should also tackle other issues around the arraycopy intrinsics and also address performance problems.

Thanks,
Christian

---------

---------
### Progress

### Issue
 * [JDK-8382226](https://bugs.openjdk.org/browse/JDK-8382226): [lworld] C2: Fix _copyOf/_copyOfRange intrinsic for flat abstract value class arrays (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2569/head:pull/2569` \
`$ git checkout pull/2569`

`$ git checkout pull/2569` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2569/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2569`

`$ git pr show -t 2569`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2569.diff">https://git.openjdk.org/valhalla/pull/2569.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2569#issuecomment-4751674149)
</details>


src/hotspot/share/opto/library_call.cpp
```diff
@@ -5197,15 +5197,22 @@ bool LibraryCallKit::inline_array_copyOf(bool is_copyOfRange) {
     // write barrier. Conservatively, go to slow path.
     // TODO 8251971: Optimize for the case when flat src/dst are later found
     // to not contain oops (i.e., move this check to the macro expansion phase).
-    // TODO 8382226: Revisit for flat abstract value class arrays
     BarrierSetC2* bs = BarrierSet::barrier_set()->barrier_set_c2();
     const TypeAryPtr* orig_t = _gvn.type(original)->isa_aryptr();
     const TypeKlassPtr* tklass = _gvn.type(klass_node)->is_klassptr();
-    bool exclude_flat = UseArrayFlattening && bs->array_copy_requires_gc_barriers(true, T_OBJECT, false, false, BarrierSetC2::Parsing) &&
+    const bool is_src_abstract_flat_value_array = orig_t != nullptr && !orig_t->elem()->is_inlinetypeptr() && orig_t->is_flat();
```
merykitty: Should we check for `!orig_t->is_not_flat()` instead?

chhagedorn: That's a good idea to cover all the cases where it could be flat. Updated this and the check for `tklass` as well accordingly.

src/hotspot/share/opto/library_call.cpp
```diff
@@ -5197,15 +5197,22 @@ bool LibraryCallKit::inline_array_copyOf(bool is_copyOfRange) {
...
+    const bool is_abstract_flat_value_class_array_involved = can_src_be_abstract_flat_value_class_array || can_dest_be_abstract_flat_value_class_array;
+    const bool exclude_flat = UseArrayFlattening &&
+                        // We do not know the exact layout of an abstract flat value class array. Bail out.
+                        (is_abstract_flat_value_class_array_involved ||
+                         (bs->array_copy_requires_gc_barriers(true, T_OBJECT, false, false, BarrierSetC2::Parsing) &&
                         // Can src array be flat and contain oops?
                         (orig_t == nullptr || (!orig_t->is_not_flat() && (!orig_t->is_flat() || orig_t->elem()->inline_klass()->contains_oops()))) &&
                         // Can dest array be flat and contain oops?
-                        tklass->can_be_inline_array() && (!tklass->is_flat() || tklass->is_aryklassptr()->elem()->is_instklassptr()->instance_klass()->as_inline_klass()->contains_oops());
+                        can_dest_be_value_class_array && (!tklass->is_flat() || tklass->is_aryklassptr()->elem()->is_instklassptr()->instance_klass()->as_inline_klass()->contains_oops())));
```
merykitty: This should also check for `tklass->is_not_flat()` instead.

chhagedorn: I agree. I also took another look at the code again and refactored it a bit to make `exclude_flat` easier to understand. Let me know, what you think.

src/hotspot/share/opto/library_call.cpp
```diff
@@ -5193,19 +5193,30 @@ bool LibraryCallKit::inline_array_copyOf(bool is_copyOfRange) {
...
     const TypeAryPtr* orig_t = _gvn.type(original)->isa_aryptr();
-    const TypeKlassPtr* tklass = _gvn.type(klass_node)->is_klassptr();
-    bool exclude_flat = UseArrayFlattening && bs->array_copy_requires_gc_barriers(true, T_OBJECT, false, false, BarrierSetC2::Parsing) &&
-                        // Can src array be flat and contain oops?
-                        (orig_t == nullptr || (!orig_t->is_not_flat() && (!orig_t->is_flat() || orig_t->elem()->inline_klass()->contains_oops()))) &&
-                        // Can dest array be flat and contain oops?
-                        tklass->can_be_inline_array() && (!tklass->is_flat() || tklass->is_aryklassptr()->elem()->is_instklassptr()->instance_klass()->as_inline_klass()->contains_oops());
+    const TypeAryKlassPtr* dest_klass_t = _gvn.type(klass_node)->is_klassptr()->isa_aryklassptr();
+    const bool can_src_be_abstract_flat_value_class_array = orig_t != nullptr && !orig_t->elem()->is_inlinetypeptr() && !orig_t->is_not_flat();
+    const bool can_dest_be_value_class_array = dest_klass_t != nullptr && dest_klass_t->can_be_inline_array();
```
merykitty: What if `dest` is not an `aryklassptr`? I think in that case this should also be true, right?

src/hotspot/share/opto/library_call.cpp
```diff
@@ -5193,19 +5193,30 @@ bool LibraryCallKit::inline_array_copyOf(bool is_copyOfRange) {
...
     BarrierSetC2* bs = BarrierSet::barrier_set()->barrier_set_c2();
     const TypeAryPtr* orig_t = _gvn.type(original)->isa_aryptr();
-    const TypeKlassPtr* tklass = _gvn.type(klass_node)->is_klassptr();
-    bool exclude_flat = UseArrayFlattening && bs->array_copy_requires_gc_barriers(true, T_OBJECT, false, false, BarrierSetC2::Parsing) &&
-                        // Can src array be flat and contain oops?
-                        (orig_t == nullptr || (!orig_t->is_not_flat() && (!orig_t->is_flat() || orig_t->elem()->inline_klass()->contains_oops()))) &&
-                        // Can dest array be flat and contain oops?
-                        tklass->can_be_inline_array() && (!tklass->is_flat() || tklass->is_aryklassptr()->elem()->is_instklassptr()->instance_klass()->as_inline_klass()->contains_oops());
+    const TypeAryKlassPtr* dest_klass_t = _gvn.type(klass_node)->is_klassptr()->isa_aryklassptr();
+    const bool can_src_be_abstract_flat_value_class_array = orig_t != nullptr && !orig_t->elem()->is_inlinetypeptr() && !orig_t->is_not_flat();
```
merykitty: This name is a little misleading, I think you want to catch the case `orig_t == nullptr` below, but if the static type is `j.l.O`, the runtime type can still be an abstract flat array.

chhagedorn: That's true. Though, I'm wondering if `orig_t` could even be a non-array pointer at all. The `Arrays.copyOf()` takes a `U[]`, so after type erasure, we should know it's an `Object[]` and thus have an array 
pointer. It looks like the original code was just defensive by checking `orig_t` as well. I removed it and testing passed.

src/hotspot/share/opto/library_call.cpp
chhagedorn: See general comment sent above.

chhagedorn (commented):
Sorry for taking some more time to respond here. Thanks for your additional questions @merykitty. I've had another closer look at the code and wrote some more tests and discovered more issues:
1. When passing in primitive type mirrors like `int.class` or for `void.class` to `Arrays.copyOf()`, there is no klass field and we return null:
https://github.com/openjdk/valhalla/blob/dc8b7c5c63155cd324933499895cef2507dfe55f/src/hotspot/share/opto/memnode.cpp#L2834-L2837
As a result, we crash here because `klass_node` returned from `load_klass_from_mirror()` is `top`:
https://github.com/openjdk/valhalla/blob/dc8b7c5c63155cd324933499895cef2507dfe55f/src/hotspot/share/opto/library_call.cpp#L5195
  - Solution: Bail out if `stopped()` after `load_klass_from_mirror()`.
2. When passing in an instance class mirror like `A.class` we should throw because it's not an array class. However, with the checks in the current code, we only emit a type array guard which does not bail out when having an instance klass:
https://github.com/openjdk/valhalla/blob/dc8b7c5c63155cd324933499895cef2507dfe55f/src/hotspot/share/opto/library_call.cpp#L5196-L5201

We then try to load the refined array klass from an instance klass pointer in `load_default_refined_array_klass()` where we read from `ObjArrayKlass::next_refined_array_klass_offset()` which is garbage.
  - Solution: Emit `generate_non_refArray_guard()` when the klass pointer is not an array klass.
3. We need to check `UseArrayFlattening` and the GC barrier availability after the inserted check for for 2., otherwise, we hit the same problem again when disabling `UseArrayFlattening` or not requiring GC barriers at all.


I pushed an update with the following changes:
- Refactored bailout logic again to a separate method where I check all the cases systematically:
  - Check dest klass to be an array klass.
  - Check all cases for maybe flat and bail out which also covers `j.l.Object`.
  - Check write barriers.
- Added more comments. 
- Added a lot more tests by using the Template Framework.

Testing this through t1-4 + stress looked good. Since we now bail out in more cases, I might also run some performance testing.

Let me know, what you think.

merykitty (approved):
Thanks for taking further look, it looks much better now.

chhagedorn:
Thanks @merykitty for your careful review!

src/hotspot/share/opto/library_call.cpp
```diff
@@ -5177,32 +5177,33 @@ bool LibraryCallKit::inline_array_copyOf(bool is_copyOfRange) {
...
-    Node* klass_node = load_klass_from_mirror(array_type_mirror, false, nullptr, 0);
+    RegionNode* bailout = new RegionNode(2);
+    record_for_igvn(bailout);
+
+    Node* klass_node = load_klass_from_mirror(array_type_mirror, false, bailout, 1);
+    if (stopped()) {
+      // Arrays.copyOf() uses a generic Class parameter which is erased to the raw type Class. This also allows
+      // passing in primitive class mirrors like int.class which do not have corresponding Klass* pointers.
+      // In these cases, klass_node will be top. Emit a trap to throw in the interpreter in this case.
+      bail_out_from_array_copyOf(bailout);
```
merykitty, line 5188: What if the `j.l.Class` is not a constant, but the runtime value is `int.class`, it will return `null`, right? Which explains the null check below. In that case, is it an inconsistency? Can it be an issue if a `j.l.Class` is narrowed to the constant `int.class` during IGVN, then the node will become `top` and the whole graph is incorrectly killed.

chhagedorn: Yes, if it's not a constant, then we will have type `j.l.Object`. When later folding to `int.class`, the `LoadKlass` will be replaced by `null` and the null check makes sure the graph is folded correctly.

TobiHartmann (approved):
Looks good to me too. Nice tests!

src/hotspot/share/opto/library_call.cpp
merykitty: Ah, I see the reason here, inside `LibraryCallKit::load_klass_from_mirror`, we null-check the result we get from the load. This means that if we have a constant `int.class`, the load will be a constant `null`, and the return value is `top`. But this also means that `klass_node` has already been null-checked, so the null-check we do right below seems redundant.

chhagedorn: That's a good observation and also seems to be redundant in mainline. We could follow-up with an RFE there, what do you think?

merykitty: Yes, that's good to me. I was confused because I thought the function results in different runtime value when the input is a constant vs when it is not (`top` vs `null`).

chhagedorn: Sounds good, filed [JDK-8387680](https://bugs.openjdk.org/browse/JDK-8387680). Thanks again for your careful review!

chhagedorn:

chhagedorn:
Performance testing with `--enable-preview` for some common benchmarks also looked good.

PR #2572, opened 2026-06-22
8387121: [lworld] ISE thrown when attempting to reflectively read strictly-initialized static field before init not specified in API docs

AlanBateman:
getstatic is specified/implement to throw ISE if a strictly-initialized static field is unset.  The reflection APIs are implemented to do the same but it doesn't surface in the API docs.

Field.getXXX is updated to specify the exception. Each getXXX method already specifies EIIE so it doesn't look out of place, and seems preferable to putting in the class description of isStrictInit. Doing the equivalent in MH/VH would add a lot of clutter so instead factory methods that produce a MH/VH to read a static field are updated. These methods already specify that they may provoke the field's class to be initialized so an additional sentence on invoking/operating-on while initializing the class doesn't look out of place.

StrictInitializingTest tests MH/VH access to a field that is unset but we don't seem to have tests for Field. Adding a simple test for Field.get and the MH/VH produced from a Field, and another test for FIeld.isStrictInit so its behavior when preview features are disabled and enabled is tested.

---------

---------
### Progress

### Issue
 * [JDK-8387121](https://bugs.openjdk.org/browse/JDK-8387121): [lworld] ISE thrown when attempting to reflectively read strictly-initialized static field before init not specified in API docs (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2572/head:pull/2572` \
`$ git checkout pull/2572`

`$ git checkout pull/2572` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2572/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2572`

`$ git pr show -t 2572`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2572.diff">https://git.openjdk.org/valhalla/pull/2572.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2572#issuecomment-4781769278)
</details>


src/java.base/share/classes/java/lang/reflect/Field.java
```diff
@@ -282,11 +282,11 @@ public boolean isSynthetic() {
     }
 
     /**
-     * Returns {@code true} if this field is a strictly
-     * initialized field; returns {@code false} otherwise.
+     * Returns {@code true} if this field is a strictly-initialized field;
+     * returns {@code false} otherwise.
```
liach, line 286: Related to previous preview label discussion with Jaikiran, I think we might indicate for all reflective preview APIs what their non-preview fallback is.

For instance, a non-preview user sees this new ISE block, click on "strictly-initialized", and reads this definition. That user has no idea how a preexisting field appears in non-preview.

If we add something like "no field is strictly-initialized if preview features are not enabled", users can see the ISE won't be thrown more easily.

liach (commented):
The test and spec changes looks good.

src/java.base/share/classes/java/lang/reflect/Field.java
AlanBateman: Fair point, this one is a reflective preview API so frameworks/libraries can get a head start. I've expanded the wording.

(With preview APIs to date it seems like that framework maintainers are using reflection to access rather than compiling and have static references to the preview APIs. Probably compiling to an older release. At least "reflective=true" communicates the intent.)

test/jdk/java/lang/reflect/Field/IsStrictInitTest.java
```diff
@@ -0,0 +1,105 @@
...
+
+    @ParameterizedTest
+    @ValueSource(booleans = { false, true })
+    void testIsStrictInit(boolean strict) throws Exception {
+        boolean preview = PreviewFeatures.isEnabled();
+        byte[] classBytes = buildClass(preview, strict);
+        Class<?> clazz = MethodHandles.lookup()
+                .defineHiddenClass(classBytes, false)
+                .lookupClass();
+        boolean expectStictInit = preview && strict;
```
liach: `expectStictInit` -> `expectStrictInit` typo

test/jdk/java/lang/reflect/Field/IsStrictInitTest.java
```diff
@@ -0,0 +1,105 @@
...
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+/*
+ * @test
+ * @summary Test Field.isStrictInit with fields that have ACC_STRICT_INIT set and not set
```
liach: I think the summary should indicate the ACC_STRICT_INIT is classfile version sensitive and ignored for older majors even when the bit is set in class files. Like "ACC_STRICT_INIT should not be recognized in older class files"?

AlanBateman: Okay, I'll see if I can improve the `@summary` line to make this clearer.

liach (approved):
Thanks, my rationale for the classfile version request is that when Strict Field Initialization graduates from preview, maintainers may know how to update `IsStrictInitTest` instead of accidentally deleting it, thinking it tests something only applicable to preview.

src/java.base/share/classes/java/lang/reflect/Field.java
```diff
@@ -484,6 +490,10 @@ public String toGenericString() {
      *              and the field is an instance field.
      * @throws    ExceptionInInitializerError if the initialization provoked
      *              by this method fails.
+     * @throws    IllegalStateException if the current thread is initializing the
```
jaikiran, line 493: Hello Alan, is the mention of "current thread" intentional here? The case I've in mind is something like:

```java
// get an uninitialized class
final Class k = Class.forName("Foo", false, someClassLoader); 
// get hold of the field of interest
final Field f = k.getDeclaredField("someStaticStrictInitField");

// trigger class initialization in Thread T1
Thread t1 = () -> {doSomethingToTriggerClassInit(k);}
t1.start();

// access the (static) strict init field in thread T2, expect it to fail with IllegalStateException, because some other thread is initializing the class currently
Thread t2 = () -> {f.get(null);}
t2.start();

```

I think this scenario where some other thread calls `Field.get()` when the field's `Class` is being initialized is possible, isn't it?

AlanBateman: > II think this scenario where some other thread calls Field.get() when the field's Class is being initialized is possible, isn't it?

The thread executing the class initializer can freely access the class fields (or methods). Attempts by other threads to access the class fields (or methods) will block until the class is initialized. With strictly-initialized fields, the field must be initialized before it can be read. For strictly-initialized static fields, if the thread initializing the class attempts to read the field before it has been initialized then the getstatic will throw ISE. The reflective APIs to the same. So yes, "current thread" is intentional here.

jaikiran: > Attempts by other threads to access the class fields (or methods) will block until the class is initialized. 
> ...
> The reflective APIs to the same. 

I wasn't aware of this behaviour of reflective APIs (and hadn't experimented either). Thank you.

jaikiran (approved):
The changes in `Field` and the new tests look good to me. My familiarity with VarHandle/MethodHandles is limited and I haven't reviewed that part thoroughly.

PR #2575, opened 2026-06-23
8386999: [lworld] C2: assert(is_dead_loop_safe()) failed: shouldn't be cleared yet

rwestrel:
The problem, here, is that, when a `Proj` is created for a late
inlining call and transformed, it is replaced by a `Proj` for a
dominating call. As a consequence, the code tries to clear an already
cleared flag.

Beyond the fix for the crash here, 8375694 (that added that code)
expects a single return projection but there can be several with the
scalarized calling convention. I filed 8387131 to investigate if the
change from 8375694 needs to be reworked.

---------

---------
### Progress

### Issue
 * [JDK-8386999](https://bugs.openjdk.org/browse/JDK-8386999): [lworld] C2: assert(is_dead_loop_safe()) failed: shouldn't be cleared yet (**Bug** - P2)

### Reviewers
 * [Vladimir Ivanov](https://openjdk.org/census#vlivanov) (@iwanowww - Committer)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2575/head:pull/2575` \
`$ git checkout pull/2575`

`$ git checkout pull/2575` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2575/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2575`

`$ git pr show -t 2575`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2575.diff">https://git.openjdk.org/valhalla/pull/2575.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2575#issuecomment-4779881298)
</details>

### DIFF
--- a/src/hotspot/share/opto/callGenerator.cpp
+++ b/src/hotspot/share/opto/callGenerator.cpp
@@ -184,7 +184,7 @@ JVMState* DirectCallGenerator::generate(JVMState* jvms) {
   kit.set_arguments_for_java_call(call);
   kit.set_edges_for_java_call(call, false, _separate_io_proj);
   Node* ret = kit.set_results_for_java_call(call, _separate_io_proj);
-  if (is_late_inline() && !call->is_boxing_method() && ret->is_Proj()) {
+  if (is_late_inline() && !call->is_boxing_method() && ret->is_Proj() && ret->in(0) == call) {
     // If late inlining for this call happens in a dead part of the graph it can leave a dead loop behind
     ret->mark_not_dead_loop_safe();
   }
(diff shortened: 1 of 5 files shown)


merykitty (approved):
Looks reasonable to me.

src/hotspot/share/opto/callGenerator.cpp
```diff
@@ -184,6 +184,10 @@ JVMState* DirectCallGenerator::generate(JVMState* jvms) {
   kit.set_arguments_for_java_call(call);
   kit.set_edges_for_java_call(call, false, _separate_io_proj);
   Node* ret = kit.set_results_for_java_call(call, _separate_io_proj);
+  if (is_late_inline() && !call->is_boxing_method() && ret->is_Proj() && ret->in(0) == call) {
```
iwanowww, line 187: Why do you need `!call->is_boxing_method()` here?

rwestrel: That was in the mainline fix:
https://github.com/openjdk/jdk/pull/30088

The reason if I remember correctly is that `Node::is_dead_loop_safe()` has some custom logic for `is_boxing_method()`

iwanowww: Oh, I overlooked it's a backport. Sorry for the noise.

iwanowww (approved):

rwestrel:
@merykitty @iwanowww thanks for the reviews

PR #2576, opened 2026-06-24
8387192: [lworld] Review comment drop for core libs

liach:
1. `weakCompareAndSetReferenceXxx` in `Unsafe` were delegating to wrong methods
2. Replace use of `0` for reference layout with `Unsafe.NON_FLAT_LAYOUT`
3. Replace use of `31 - Integer.numberOfLeadingZeros(ascale)` with `numberOfTrailingZeros`
4. Remove null check for VarHandle getting from NR fields (parity with ArrayVarHandle and DirectMethodHandle), add missing null check for flat value var handles
5. Remove the weird directives and old flags of RecursiveValueClass

Thanks to @JornVernee and @AlanBateman for spotting these issues.

---------

---------
### Progress

### Issue
 * [JDK-8387192](https://bugs.openjdk.org/browse/JDK-8387192): [lworld] Review comment drop for core libs (**Bug** - P4)

### Reviewers
 * [Jorn Vernee](https://openjdk.org/census#jvernee) (@JornVernee - no project role)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2576/head:pull/2576` \
`$ git checkout pull/2576`

`$ git checkout pull/2576` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2576/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2576`

`$ git pr show -t 2576`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2576.diff">https://git.openjdk.org/valhalla/pull/2576.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2576#issuecomment-4785399001)
</details>

### DIFF
--- a/src/java.base/share/classes/java/io/ObjectStreamClass.java
+++ b/src/java.base/share/classes/java/io/ObjectStreamClass.java
@@ -1933,7 +1933,7 @@ public final class ObjectStreamClass implements Serializable {
                 writeKeys[i] = usedKeys.add(key) ?
                     key : Unsafe.INVALID_FIELD_OFFSET;
                 offsets[i] = f.getOffset();
-                layouts[i] = rf != null && !f.isPrimitive() ? UNSAFE.fieldLayout(rf) : 0;
+                layouts[i] = rf != null && !f.isPrimitive() ? UNSAFE.fieldLayout(rf) : Unsafe.NON_FLAT_LAYOUT;
                 typeCodes[i] = f.getTypeCode();
                 if (!f.isPrimitive()) {
                     typeList.add((rf != null) ? rf.getType() : null);
--- a/src/java.base/share/classes/java/io/ObjectStreamClass.java
+++ b/src/java.base/share/classes/java/io/ObjectStreamClass.java
@@ -2029,9 +2029,9 @@ public final class ObjectStreamClass implements Serializable {
             for (int i = numPrimFields; i < fields.length; i++) {
                 vals[offsets[i]] = switch (typeCodes[i]) {
                     case 'L', '[' ->
-                            layouts[i] != 0
-                                    ? UNSAFE.getFlatValue(obj, readKeys[i], layouts[i], types[i - numPrimFields])
-                                    : UNSAFE.getReference(obj, readKeys[i]);
+                            layouts[i] == Unsafe.NON_FLAT_LAYOUT
+                                    ? UNSAFE.getReference(obj, readKeys[i])
+                                    : UNSAFE.getFlatValue(obj, readKeys[i], layouts[i], types[i - numPrimFields]);
                     default       -> throw new InternalError();
                 };
             }
@@ -2082,10 +2082,10 @@ public final class ObjectStreamClass implements Serializable {
                                 obj.getClass().getName());
                         }
                         if (!dryRun) {
-                            if (layouts[i] != 0) {
-                                UNSAFE.putFlatValue(obj, key, layouts[i], types[i - numPrimFields], val);
-                            } else {
+                            if (layouts[i] == Unsafe.NON_FLAT_LAYOUT) {
                                 UNSAFE.putReference(obj, key, val);
+                            } else {
+                                UNSAFE.putFlatValue(obj, key, layouts[i], types[i - numPrimFields], val);
                             }
                         }
                     }
--- a/src/java.base/share/classes/java/lang/invoke/ArrayVarHandle.java
+++ b/src/java.base/share/classes/java/lang/invoke/ArrayVarHandle.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2025, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2025, 2026, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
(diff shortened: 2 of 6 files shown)


src/java.base/share/classes/java/io/ObjectStreamClass.java
```diff
@@ -1933,7 +1933,7 @@ private static final class FieldReflector {
                 writeKeys[i] = usedKeys.add(key) ?
                     key : Unsafe.INVALID_FIELD_OFFSET;
                 offsets[i] = f.getOffset();
-                layouts[i] = rf != null && !f.isPrimitive() ? UNSAFE.fieldLayout(rf) : 0;
+                layouts[i] = rf != null && !f.isPrimitive() ? UNSAFE.fieldLayout(rf) : Unsafe.NON_FLAT_LAYOUT;
```
viktorklang-ora, line 1936: nice

src/java.base/share/classes/java/lang/invoke/ArrayVarHandle.java
```diff
@@ -33,7 +33,14 @@
...
-/// The var handle for polymorphic arrays.
+/// The var handle for all object arrays.
+///
+/// Field layouts can be determined ahead-of-time from their declaring class.
+/// Array layouts cannot be determined: all arrays of the same component type
+/// and different layouts share the same `Class`.
+///
+/// This var handle must check each incoming array instance to determine the
+/// layout and the correct operation for that array.
 final class ArrayVarHandle extends VarHandle {
```
viktorklang-ora, line 44: Given the repetition of logic, and the risk of missing some change down the line, at what point do we consider using a template to generate this file?

src/java.base/share/classes/jdk/internal/misc/Unsafe.java
```diff
@@ -1717,8 +1717,10 @@ public final native boolean compareAndSetReference(Object o, long offset,
                                                        Object expected,
                                                        Object x);
 
-    private final boolean isValueObject(Object o) {
-        return o != null && o.getClass().isValue();
+    @ForceInline
+    private final boolean isValueObject(Class<?> valueType, Object o) {
+        return ValueClass.isValueObjectCompatible(valueType)
+                && o != null && o.getClass().isValue();
```
viktorklang-ora, line 1731: I suspect having the null-check first is going to be cheaper?

src/java.base/share/classes/jdk/internal/misc/Unsafe.java
```diff
@@ -1848,7 +1850,7 @@ public final <V> boolean weakCompareAndSetReferencePlain(Object o, long offset,
                                                              Class<?> valueType,
                                                              V expected,
                                                              V x) {
-        if (isValueObject(expected)) {
+        if (isValueObject(valueType, expected)) {
             return compareAndSetReference(o, offset, valueType, expected, x);
```
viktorklang-ora, line 1863: My preference would be that all of the accesses that are stronger than necessary have a small comment stating that they are stronger than necessary until an appropriately-weaker implementation is added. This is to make it clear that the code is intentional, and under which circumstances it could/should change.

viktorklang-ora (commented):
LGTM (after comments being handled)

src/java.base/share/classes/jdk/internal/misc/Unsafe.java
liach: I think the `isValueObjectCompatible` is likely to be cheaper because it is going to be eliminated if `valueType` is a constant (it mostly is in the context of var handles and method handles)

src/java.base/share/classes/java/lang/invoke/ArrayVarHandle.java
liach: I think if that's the case, we should update all X-VarHandleXxx.java.template templates to be similarly generaetd. Even though they are templates, they do feature a lot of duplications. And given they are preexisting, I think this is better handled in mainline.

liach: Created https://bugs.openjdk.org/browse/JDK-8387216 to track this effort.

src/java.base/share/classes/jdk/internal/misc/Unsafe.java
liach: Sure. I think we did not go for that before because we have things like "storeStoreFence" deferring to stronger "storeFence" or "fullFence" without comments, but more comments would always be good.

src/java.base/share/classes/java/lang/invoke/X-VarHandle.java.template
```diff
@@ -107,11 +107,6 @@ final class VarHandle$InputType$s {
             FieldInstanceReadOnly handle = (FieldInstanceReadOnly)ob;
             $type$ value = UNSAFE.get$Type$(Objects.requireNonNull(handle.receiverType.cast(holder)),
                                  handle.fieldOffset{#if[FlatValue]?, handle.layout, handle.fieldType});
-#if[Reference]
-            if (value == null && handle.nullRestricted) {
-                throw new NullPointerException("Uninitialized null-restricted field");
-            }
-#end[Reference]
```
JornVernee, line 114: Just for my understanding: why is it okay to remove this?

liach: The VM and low level libraries ensure null restricted fields never carry null upon write (with strict field initialization). The checks upon read are thus redundant. DirectMethodHandle getters and ArrayVarHandle both have no null check when they read values out of a null-restricted structure. So I think we don't need to be extra cautious here, too.

JornVernee (approved):

vicente-romero-oracle (approved):
looks sensible to me

liach:
Thanks for the reviews!

/integrate

PR #2578, opened 2026-06-24
8387304: [lworld] Review the specs of reflective preview APIs

liach:
Reflective preview APIs can be used out of preview. Thus it is meaningful for us to note how they behave when preview features are disabled.

I hope this can clarify some common misconceptions and help onboarding to the JEP 401 reflection model.

---------

---------
### Progress

### Issue
 * [JDK-8387304](https://bugs.openjdk.org/browse/JDK-8387304): [lworld] Review the specs of reflective preview APIs (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2578/head:pull/2578` \
`$ git checkout pull/2578`

`$ git checkout pull/2578` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2578/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2578`

`$ git pr show -t 2578`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2578.diff">https://git.openjdk.org/valhalla/pull/2578.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2578#issuecomment-4801919591)
</details>


src/java.base/share/classes/java/lang/Class.java
```diff
@@ -1379,9 +1386,39 @@ private Class<?> elementType() {
      * arrays, the values of other modifiers are {@code false} other
      * than as specified above.
      *
+     * <div class="preview-block">
+     *      <div class="preview-comment">
+     *          When preview features are enabled, if this {@code Class} object
+     *          represents a class that does not use preview features or an
```
ExE-Boss: The old wording can be read as a class that doesn’t use arrays:
```suggestion
     *          represents a class that does not use preview features or is an
```

src/java.base/share/classes/java/lang/Class.java
```diff
@@ -1415,6 +1452,36 @@ private Class<?> elementType() {
      * For {@code Class} objects representing void, primitive types, and
      * arrays, access flags are absent other than as specified above.
      *
+     * <div class="preview-block">
+     *      <div class="preview-comment">
+     *          When preview features are enabled, if this {@code Class} object
+     *          represents a class that does not use preview features or an
```
ExE-Boss:
```suggestion
     *          represents a class that does not use preview features or is an
```

src/java.base/share/classes/java/lang/Class.java
liach: "represents a class ... or an array type". Note here "a class" is intentionally distinct from "a Class object", that a "class" means the "class" in "class or interface", in its narrowest sense.

liach: I have reworded this so:

1. It's the class file that does not use preview features
2. "or represents" makes wording more clear.

src/java.base/share/classes/java/lang/Class.java
```diff
@@ -1369,7 +1376,7 @@ private Class<?> elementType() {
      *      {@code true}
      * <li> its interface modifier is always {@code false}, even when
      *      the component type is an interface
-     * <li> when preview features are enabled, its {@linkplain
+     * <li> when preview features are enabled, its {@link
      *      AccessFlag#IDENTITY identity} modifier is always true
```
AlanBateman, line 1380: As it's now a link then maybe "identity" should be "IDENTITY".  I realize the other non-link usages are "identity".

src/java.base/share/classes/java/util/Objects.java
```diff
@@ -223,7 +235,7 @@ public static boolean hasIdentity(Object obj) {
      * @throws IdentityException if {@code obj} is not an identity object
      * @since 28
      */
-    @PreviewFeature(feature = PreviewFeature.Feature.VALUE_OBJECTS)
+    @PreviewFeature(feature = PreviewFeature.Feature.VALUE_OBJECTS, reflective = true)
```
AlanBateman, line 238: Changing the requireIdentity methods to be reflective preview APIs looks okay, as is the update to their docs.  My guess is that the libraries that want to get a head start will access the preview APIs with core reflection so they won't care if the methods are just normal preview APIs.

src/java.base/share/classes/java/lang/classfile/attribute/StackMapFrameInfo.java
```diff
@@ -82,6 +82,11 @@ public sealed interface StackMapFrameInfo
 
     /**
      * {@return the expanded unset fields}
+     * <p>
+     * If a stack map frame in a {@code class} file has a non-empty list of
+     * unset fields, the {@code class} file must declare it uses preview
+     * features.  The list of unset fields is always empty for a stack map frame
+     * declared in a {@code class} file that does not use preview features.
```
AlanBateman: It might be a bit more readable to swap the sentences so that it's initially clear that the list is always empty when they class does not use preview features. It may be non-empty when the class uses preview features. Just a suggestion as I had to read the paragraph a few times to digest it.

src/java.base/share/classes/java/lang/reflect/Field.java
```diff
@@ -285,11 +285,12 @@ public boolean isSynthetic() {
...
-     * enabled and this field is a strictly-initialized field. The
-     * {@link AccessFlag#STRICT_INIT ACC_STRICT_INIT} flag is considered not set
-     * when preview features are disabled; consequently, this method always
-     * returns {@code false} when preview features are disabled.
+     * <p>This method returns {@code true} if and only if the class or interface
+     * that declares this field uses preview features and this field is a
+     * strictly-initialized field. The {@link AccessFlag#STRICT_INIT
+     * ACC_STRICT_INIT} flag is considered not set for a field declared in a
+     * class or interface that does not use preview features; consequently,
+     * this method always returns {@code false} when preview features are disabled.
```
AlanBateman, line 293: Good, this is an improvement on what we had previously.

src/java.base/share/classes/java/util/Objects.java
```diff
@@ -196,6 +196,12 @@ public static String toIdentityString(Object o) {
      * {@code false}. All other objects, including arrays, are identity objects
      * and the result will be {@code true}.
      *
+     * <p>This method returns {@code false} if and only if the parameter is
+     * {@code null} or if the parameter represents a value object when preview
+     * features are enabled.  Value objects do not exist when preview features
+     * are disabled; consequently, this method behaves the same as {@link
+     * #nonNull Objects.nonNull} when preview features are disabled.
```
AlanBateman: "Value objects do not exist when preview features are disabled".  What would you think about just saying that "All objects are identity objects when preview features are disabled"?

src/java.base/share/classes/java/lang/Class.java
```diff
@@ -631,6 +631,13 @@ public static Class<?> forName(Module module, String name) {
      * {@code Class} object represents an interface, array type, primitive type,
      * or {@code void}, the result is {@code false}.
      *
+     * <p>This method returns {@code true} if and only if this {@code Class}
+     * object represents a class that uses preview features, and the class does
+     * not have the {@link AccessFlag#IDENTITY ACC_IDENTITY} flag set.
+     * The {@code ACC_IDENTITY} flag is considered always set for a class that
+     * does not use preview features; consequently, this method always returns
+     * {@code false} when preview features are disabled.
```
AlanBateman, line 639: Good.

src/java.base/share/classes/java/lang/Class.java
```diff
@@ -1379,9 +1386,40 @@ private Class<?> elementType() {
      * arrays, the values of other modifiers are {@code false} other
      * than as specified above.
      *
+     * <div class="preview-block">
+     *      <div class="preview-comment">
+     *          When preview features are enabled, if this {@code Class} object
+     *          represents a class whose {@code class} file does not use preview
```
AlanBateman: It might be smoother to start with "When preview features are enabled and this Class object ..."

src/java.base/share/classes/java/lang/Class.java
```diff
@@ -1379,9 +1386,40 @@ private Class<?> elementType() {
...
      * of <cite>The Java Virtual Machine Specification</cite>.
      *
+     * @apiNote
+     * <div class="preview-block">
+     *      <div class="preview-comment">
+     *          Developers should be aware that the presence of the {@code
+     *          identity} modifier is dependent on whether preview features are
+     *          enabled, a status not accessible to programs.  The preferred way
+     *          to check for the identity status is to use {@link #isValue()
+     *          Class.isValue()}.
```
AlanBateman: I think it would be better to drop "a status not accessible to programs", it will just anger developers.

Can the second sentence just use "Use the isValue method to test if a value class or identity class"?

src/java.base/share/classes/java/lang/Class.java
```diff
@@ -1379,9 +1386,40 @@ private Class<?> elementType() {
...
+     *          <p>
+     *          This snippet below checks whether a given {@code Class<?> clazz}
+     *          would have its {@code identity} modifier set when preview
+     *          features are enabled, yet behaves consistently regardless of
+     *          whether preview features are enabled.
+     *          {@snippet lang=java :
+     *          !clazz.isPrimitive() && !clazz.isValue() && !clazz.isInterface()
```
AlanBateman, line 1417: Good idea to include this snippet as it is useful to have.

src/java.base/share/classes/java/util/Objects.java
ExE-Boss: For a similar reason, `IdentityException` should probably also be made reflective.

src/java.base/share/classes/java/lang/Class.java
ExE-Boss: “a status not accessible to programs” is not even true, as user code can just try to load a hidden preview class for the current `ClassFileFormatVersion` and check whether it succeeds, like what the `defineHiddenClass​/PreviewHiddenClass` **JDK** and `ClassFile​/PreviewVersion` **JTReg** tests do:

```java
// `@Namespace` is an internal annotation that signals that the class is intended
// to be used to only hold `static` fields and methods
public final @Namespace class PreviewFeatures {
	private PreviewFeatures() throws InstantiationException { throw new InstantiationException(PreviewFeatures.class.getName()); }

	/// {@return whether preview features are enabled}
	public static boolean isEnabled() {
		return Holder.IS_PREVIEW_ENABLED;
	}

	// Can be replaced with `LazyConstant` once that exits preview
	private static final @Namespace class Holder {
		private Holder() throws InstantiationException { throw new InstantiationException(Holder.class.getName()); }
		private static final boolean IS_PREVIEW_ENABLED = computeIsPreviewEnabled();
	}

	private static boolean computeIsPreviewEnabled() {
		final var lookup = MethodHandles.lookup();
		final byte[] bytecodes = getCurrentPreviewBytecodes();
		try {
			lookup.defineHiddenClass(bytecodes, false, new MethodHandles.Lookup.ClassOption[] {
				// no `ClassOption.STRONG`, so that the temporary hidden class
				// can be collected after this method exits
			});
			return true;
		} catch (final UnsupportedClassVersionError _) {
			return false;
		} catch (final IllegalAccessException ex) {
			// `ExceptionUtil.fromReflectiveOperationException` behaves like
			// `java.lang.invoke.MethodHandleNatives::mapLookupExceptionToError`
			throw ExceptionUtil.fromReflectiveOperationException(ex);
		}
	}

	/// {@return the minimum bytecode needed to check whether preview features are enabled}
	private static final byte[] getCurrentPreviewBytecodes() {
		return ClassFile.of().build(
			ClassDesc.of(
				PreviewFeatures.class.getName()
					.replace('/', '$$')
					.concat("$PreviewCheck")
			),
			cb -> cb
				.withVersion(
					ClassFileFormatVersion.latest().major(),
					ClassFile.PREVIEW_MINOR_VERSION
				)
				.withFlags(ACC_PUBLIC | ACC_FINAL | ACC_SUPER)
				.withSuperclass(ConstantDescs.CD_Object)
		);
	}
}
```

src/java.base/share/classes/java/lang/Class.java
liach: I saw the other modifier references in lowercase being in code block - like in `{@code public}, {@code protected},` above, so making this "identity" appearance non-code font feels weird.

src/java.base/share/classes/java/util/Objects.java
liach: I don't think programs have a similarly strong reason to handle `IdentityException`, which is the only case I deem making that class valuable.

src/java.base/share/classes/java/lang/Class.java
liach: Done. Also used "either a class whose.... or an array type" to make clear what "or" is switching on.

src/java.base/share/classes/java/util/Objects.java
liach: Done.

src/java.base/share/classes/java/lang/classfile/attribute/StackMapFrameInfo.java
liach: Reordered so it now flows like "if non-preview empty; otherwise preview if non-empty"

src/java.base/share/classes/java/lang/Class.java
liach: Done. Indeed preview status is accessible to programs that can declare classes, but last time I checked this status is so uninteresting that people did not even discuss about this on stackoverflow.

src/java.base/share/classes/java/lang/Class.java
```diff
@@ -1388,9 +1388,9 @@ private Class<?> elementType() {
      *
      * <div class="preview-block">
      *      <div class="preview-comment">
-     *          When preview features are enabled, if this {@code Class} object
-     *          represents a class whose {@code class} file does not use preview
-     *          features or represents an array type, its {@code identity}
+     *          When preview features are enabled and this {@code Class} object
+     *          represents either a class whose {@code class} file does not
+     *          depend on preview features or an array type, its {@code identity}
      *          modifier is always true.
```
ExE-Boss:
```suggestion
     *          represents either a class whose {@code class} file does not
     *          depend on preview features or represents an array type,
     *          its {@code identity} modifier is always true.
```

src/java.base/share/classes/java/lang/Class.java
```diff
@@ -1455,10 +1454,10 @@ private Class<?> elementType() {
...
      * <div class="preview-block">
      *      <div class="preview-comment">
-     *          When preview features are enabled, if this {@code Class} object
-     *          represents a class whose {@code class} file does not use preview
-     *          features or represents an array type, its flags always include
-     *          {@code IDENTITY}.
+     *          When preview features are enabled and this {@code Class} object
+     *          represents either a class whose {@code class} file does not
+     *          depend on preview features or an array type, its flags always
+     *          include {@code IDENTITY}.
```
ExE-Boss:
```suggestion
     *          depend on preview features or represents an array type, its flags
     *          always include {@code IDENTITY}.
```

ExE-Boss (commented):
Personally, I’d still keep “represents” here as well, to make the meanings wholly unambiguous:

src/java.base/share/classes/java/lang/classfile/attribute/StackMapFrameInfo.java
```diff
@@ -82,6 +82,11 @@ public sealed interface StackMapFrameInfo
 
     /**
      * {@return the expanded unset fields}
+     * <p>
+     * If this stack map frame is declared in a {@code class} file that does not
+     * depend on preview features, the list of unset fields is always empty;
+     * otherwise, the {@code class} file must declare that it depends on preview
+     * features if the unset fields list contains any item.
```
AlanBateman: The first part is fine. I think the second part would be easier to read if you drop "otherwise" and just say "If the class file depends on preview features then it returns the list of unset fields".

src/java.base/share/classes/java/lang/Class.java
```diff
@@ -1379,9 +1386,39 @@ private Class<?> elementType() {
...
      *
+     * <div class="preview-block">
+     *      <div class="preview-comment">
+     *          When preview features are enabled and this {@code Class} object
+     *          represents either a class whose {@code class} file does not
+     *          depend on preview features or an array type, its {@code identity}
+     *          modifier is always true.
+     *          <p>
+     *          When preview features are disabled, such a {@code Class} object
+     *          does not have its {@code identity} modifier set.
```
AlanBateman, line 1397: You should be able to use the definite article in this sentence so it reads "the Class object" rather "such a Class object". I think that would be easier to read here.

(Same thing in Class::accessFlags preview block)

src/java.base/share/classes/java/lang/Class.java
liach: "represents either" need to become "either represents".

AlanBateman (approved):
You've addressed all my comments and I think this is good now.

vicente-romero-oracle (approved):
lgtm

liach:
Thank you for the reviews, Alan and Vicente!

/integrate

PR #2580, opened 2026-06-25
8387131: [lworld] C2: revisit 8375694 for scalarized return values

rwestrel:
In mainline, a single value can be returned from a call. With
valhalla, multiple values can be returned so the fix for 8375694 needs
to be adjusted.

---------

---------
### Progress

### Issue
 * [JDK-8387131](https://bugs.openjdk.org/browse/JDK-8387131): [lworld] C2: revisit 8375694 for scalarized return values (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2580/head:pull/2580` \
`$ git checkout pull/2580`

`$ git checkout pull/2580` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2580/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2580`

`$ git pr show -t 2580`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2580.diff">https://git.openjdk.org/valhalla/pull/2580.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2580#issuecomment-4797210150)
</details>

### DIFF
--- a/src/hotspot/share/opto/callGenerator.cpp
+++ b/src/hotspot/share/opto/callGenerator.cpp
@@ -184,9 +184,31 @@ JVMState* DirectCallGenerator::generate(JVMState* jvms) {
   kit.set_arguments_for_java_call(call);
   kit.set_edges_for_java_call(call, false, _separate_io_proj);
   Node* ret = kit.set_results_for_java_call(call, _separate_io_proj);
-  if (is_late_inline() && !call->is_boxing_method() && ret->is_Proj() && ret->in(0) == call) {
+  if (is_late_inline() && !call->is_boxing_method()) {
     // If late inlining for this call happens in a dead part of the graph it can leave a dead loop behind
-    ret->mark_not_dead_loop_safe();
+    if (ret->is_Proj() && ret->in(0) == call) {
+      ret->mark_not_dead_loop_safe();
+    } else if (ret->isa_InlineType()) {
+      InlineTypeNode* vt = ret->as_InlineType();
+      Node* oop = vt->get_oop();
+      if (oop->is_Proj()) {
+        assert(oop->in(0) == call, "");
+        oop->mark_not_dead_loop_safe();
+      }
+      Node* null_marker = vt->get_null_marker();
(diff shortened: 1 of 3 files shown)


TobiHartmann (approved):
Looks good to me. I'll report back once testing passed.

TobiHartmann:
Testing is clean.

src/hotspot/share/opto/callGenerator.cpp
```diff
@@ -62,6 +62,37 @@ bool CallGenerator::is_inlined_method_handle_intrinsic(ciMethod* symbolic_info,
...
+    return;
+  }
+  CallNode* call = call_node();
+  if (ret->is_Proj() && ret->in(0) == call) {
+    ret->mark_not_dead_loop_safe();
+  } else if (ret->isa_InlineType()) {
+    InlineTypeNode* vt = ret->as_InlineType();
+    Node* oop = vt->get_oop();
+    if (oop->is_Proj()) {
+      assert(oop->in(0) == call, "projection was replaced by dominating one");
```
merykitty: I don't quite understand this part. Why do we check `ret` being a `Proj` of `call` in the previous case, but in this case, we assume if it is a `Proj`, it must be a `Proj` of call?

rwestrel: The mainstream change assumes `Proj` is from `call`. That caused an issue with valhalla and I wanted to make sure we would catch such an issue. I took another look at `ProjNode::Identity()` again and, it seems possible that some projections for returned value can be replaced by a dominating projection. So  I replaced the assert with an actual check.

merykitty (approved):
Thanks, it looks good to me, then.

TobiHartmann (approved):
Good catch. Looks good to me too.

rwestrel:
@merykitty @TobiHartmann thanks for the reviews

PR #2588, opened 2026-06-26
8387417: [lworld] API doc improvements for java.lang.Object

AlanBateman:
Updates to java.lang.Object API docs:

- Improve preview-comment in class description that defines value class and value object
- Update spec for the finalize method to make it clear when finalize method of a value class is never invoked by the garbage collector
- Update the implSpec for the clone to specify behavior of Object.clone for value objects. Also add paragraph to method description for value objects, and add API note to advise using copy constructors or static factory methods instead
- Improve preview-comment in wait/notify methods and drop confusing addition to IMSE exception

A sanity test is added, this depends on the fix to finalizer registration in pull/2589.

---------

---------
### Progress

### Issue
 * [JDK-8387417](https://bugs.openjdk.org/browse/JDK-8387417): [lworld] API doc improvements for java.lang.Object (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2588/head:pull/2588` \
`$ git checkout pull/2588`

`$ git checkout pull/2588` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2588/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2588`

`$ git pr show -t 2588`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2588.diff">https://git.openjdk.org/valhalla/pull/2588.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2588#issuecomment-4835969672)
</details>


src/java.base/share/classes/java/lang/Object.java
liach: Note this convention is unrealistic for value objects - I recall we discussed about this with Dan Smith participating, forgot if we reached any conclusion about actions to take.

test/jdk/java/lang/Object/ValueObjects.java
```diff
@@ -97,6 +97,37 @@ protected void finalize() {
...
+            }
+        }
+
+        var latch = new CountDownLatch(1);
+        var obj = new C(latch);
+        obj = null;
+        for (int i = 0; i < 3; i++) {
+            System.gc();
+            // latch should not count down
+            assertFalse(latch.await(1, TimeUnit.SECONDS));
```
liach: 1 second is too long for a tier 1 test I think; we should also use Utils.adjustTimeout to respect jtreg timeouts.

AlanBateman: 1s isn't too bad for a tier1 test, it could be smaller of course. Using Utils.adjustTimeout here is a bit overkill but okay.

src/java.base/share/classes/java/lang/Object.java
AlanBateman: Object.clone is a flawed mechanism and its API docs has many issues. This convention is just re-framed in the context of identity objects. Kinda like constructor chaining, this convention is important for non-final classes to allow some subclass to override clone. It's not the main focus on the update though. The important bit is updating the implSpec so the method of the Object class is specified for value objects.

src/java.base/share/classes/java/lang/Object.java
```diff
@@ -35,12 +35,24 @@
...
  *      <div class="preview-comment">
- *          When preview features are enabled, subclasses of {@code java.lang.Object} can be either
- *          identity classes or {@linkplain Class#isValue value classes}.
- *          See The Java Language Specification {@jls value-objects-8.1.1.5 Value Classes}.
- *          Use of value class instances for synchronization or with
- *          {@linkplain java.lang.ref.Reference object references} results in
- *          {@link IdentityException}.
+ *          When preview features are enabled, subclasses of {@code java.lang.Object}
+ *          are either {@linkplain Class#isValue value classes} or identity classes.
+ *          A <em>value object</em> is an instance of a non-abstract value class. All
```
jaikiran, line 40: Hello Alan, in several other places in the other JDK, we seem to refer to such classes as concrete value class. Should we do the same here instead of saying non-abstract value class?

src/java.base/share/classes/java/lang/Object.java
```diff
@@ -225,12 +249,16 @@ public boolean equals(Object obj) {
...
-     * object and initializes all its fields with exactly the contents of
+     * <p>
+     * For an identity object, this method creates a new instance of the class of
+     * this object and initializes all its fields with exactly the contents of
      * the corresponding fields of this object, as if by assignment; the
      * contents of the fields are not themselves cloned. Thus, this method
      * performs a "shallow copy" of this object, not a "deep copy" operation.
      * <p>
+     * For a value object, this method returns an object that is indistinguishable
+     * from the original.
```
jaikiran: Given that statewise equivalence term is defined in the JLS, should we instead say "... returns an object that is statewise equivalent to the original"?

src/java.base/share/classes/java/lang/Object.java
AlanBateman: Defining a value object as an instance of non-abstract value classes works here, it's similar to the wording in the changes for the JLS.

src/java.base/share/classes/java/lang/Object.java
AlanBateman: Yes, that might be more precise (the JEP uses "indistinguishable" the JLS uses "statewise-equivalent").

src/java.base/share/classes/java/lang/Object.java
liach: I would prefer to phrase it like "the class of a value object is a value class", more in line with code like `o.getClass().isValue()`.

However, Valhalla models "identity" as the extra thing, so I guess it is better to go like "An identity object is an object whose class is an identity class or an array type. All other objects are value objects."

AlanBateman: I would prefer to keep this sentence as proposed as it keeps it consistent with the JLS and keeps the ordering that the terms are introduces consistent with the first sentence. I would be a bit wary of "All other objects are value objects" as it hints that there are other things that are also value classes.

test/jdk/java/lang/Object/ValueObjects.java
```diff
@@ -104,6 +105,7 @@ protected void finalize() {
      * Test that the finalize method on an abstract value value is not invoked by the GC.
```
jaikiran, line 105: Typo - should have been "abstract value class ..."

jaikiran (approved):
Thank you for the update. This looks good to me.

test/jdk/java/lang/Object/ValueObjects.java
AlanBateman: Well spotted.

liach (approved):
The introductory paragraph in class spec might be subject to further discussion to introduce people to "identity is the extra thing" view, but that shouldn't be a blocker.

PR #2589, opened 2026-06-26
8386972: [lworld] Finalizer should not be registered for value objects

fparain:
Change in finalizers registration to prevent it for value classes.

Tested tier1-4.

---------

---------
### Progress

### Issue
 * [JDK-8386972](https://bugs.openjdk.org/browse/JDK-8386972): [lworld] Finalizer should not be registered for value objects (**Bug** - P2)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2589/head:pull/2589` \
`$ git checkout pull/2589`

`$ git checkout pull/2589` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2589/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2589`

`$ git pr show -t 2589`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2589.diff">https://git.openjdk.org/valhalla/pull/2589.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2589#issuecomment-4857939848)
</details>


src/hotspot/share/classfile/classFileParser.cpp
```diff
@@ -4157,7 +4160,8 @@ void ClassFileParser::set_precomputed_flags(InstanceKlass* ik) {
   const Method* const m = ik->lookup_method(vmSymbols::finalize_method_name(),
                                            vmSymbols::void_method_signature());
   if (InstanceKlass::is_finalization_enabled() &&
-      (m != nullptr) && !m->is_empty_method()) {
+      (m != nullptr) && !m->is_empty_method() &&
+      m->method_holder()->access_flags().is_identity_class()) {
```
DanHeidinga, line 4164: Could we call ClassFileParser's ` is_identity_class()` here instead of going through the method_holder?

DanHeidinga: Fred pointed out my proposed change actually changes the meaning as my proposal checks the current class, not the class that defines the finalize method.  This is different for a finalize method from an abstract value class

ExE-Boss: A finalizer in an abstract value class should probably still be registered for identity subclasses; as otherwise, changing an abstract identity class with a finalizer to an abstract value class with a finalizer would be a breaking change as long as finalizers continue to exist.

AlanBateman: An abstract class with final fields may be good candidate to migrate to an abstract value class. If the abstract class relies on finalization then it is probably not a good candidate. If someone were to migrate without removing the dependency on finalization then it's not going to work as they might think as presumably the motivation for migrating the abstract class is to allow for subclasses to be value classes.

Right now, compiling an abstract value class with a finalize method emits a warning at compile-time to say that class should not have a finalize method as it will not be invoked (this is in addition to the removal warning that finalization is deprecated for removal). This will hopefully make developers aware that it doesn't work as they might think.

AlanBateman:
There are two tests in test/jdk/java/lang/Object/ValueObjects.java that are `@Disabled` for now. The `@Disabled` annotations can be removed when the changes here are integrated.

coleenp (approved):
Looks good.

fparain:
Thanks @coleenp for the review.
/integrate

src/hotspot/share/classfile/classFileParser.cpp
```diff
@@ -2250,6 +2250,7 @@ void ClassFileParser::copy_method_annotations(ConstMethod* cm,
 
 Method* ClassFileParser::parse_method(const ClassFileStream* const cfs,
                                       bool is_interface,
+                                      bool is_identity_class,
```
dholmes-ora, line 2253: Why change the signature to pass this in when we can just call `is_identity_class()` at line 2831??

src/hotspot/share/classfile/classFileParser.cpp
dholmes-ora: I tend to agree with @ExE-Boss here. Vagaries of finalizers aside, if the subclass can be an identity class then the superclass's finalizer should be registered for instances of the subclass. Otherwise the JEP needs to state clearly that this is the case.

dholmes-ora: FTR the behaviour was ratified by a spec update through JDK-8387417

jaikiran: For future reference, https://github.com/openjdk/valhalla/pull/2588/ updated the specification of Object.finalize() to specify:

```
* <div class="preview-block">
*          The {@code finalize} method of a value class is never directly
*          invoked by the garbage collector. This includes the case where an
*          abstract value class declares a {@code finalize} method and the
*          class is extended by an identity class; the garbage collector never
*          directly invokes the {@code finalize} method declared by the
*          abstract value class.
*      </div>
* </div>
```

PR #2593, opened 2026-06-29
8386067: [lworld] C2 asserts with "Missed Ideal optimization opportunity in PhaseIterGVN for CmpL"

TobiHartmann:
The `PushInlineTypeDown` implementation from [JDK-8302217](https://bugs.openjdk.org/browse/JDK-8302217) rewires `PhiNode` uses to a new `InlineTypeNode` but only adds direct users to the IGVN worklist. Nodes whose transformations depend on deeper input type changes might not be re-enqueued.

In the failing case, the offending shape looks like this:
```CmpL(OrL(CastP2X(phi_or_inline_type), CastP2X(other)), 0L)```

After the Phi input is replaced by a scalarized inline type or a not-null value, [CmpLNode::Ideal](https://github.com/openjdk/valhalla/blob/a0dc49c20ee6d20e925a107b0d3dabc7e0ba8863/src/hotspot/share/opto/subnode.cpp#L899-L922) can fold the corresponding `CastP2X` to the null marker or `1L`. Without notifying users through the normal IGVN replacement path, `VerifyIterativeGVN` will report a missed `CmpL` idealization.

The fix makes `replace_in_uses` notify users before rewiring, and uses it from `PushInlineTypeDown`.

Thanks,
Tobias

---------

---------
### Progress

### Issue
 * [JDK-8386067](https://bugs.openjdk.org/browse/JDK-8386067): [lworld] C2 asserts with "Missed Ideal optimization opportunity in PhaseIterGVN for CmpL" (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2593/head:pull/2593` \
`$ git checkout pull/2593`

`$ git checkout pull/2593` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2593/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2593`

`$ git pr show -t 2593`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2593.diff">https://git.openjdk.org/valhalla/pull/2593.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2593#issuecomment-4833716416)
</details>

### DIFF
--- a/src/hotspot/share/opto/phaseX.cpp
+++ b/src/hotspot/share/opto/phaseX.cpp
@@ -2486,6 +2486,7 @@ void PhaseIterGVN::subsume_node( Node *old, Node *nn ) {
   temp->destruct(this);     // reuse the _idx of this little guy
 }
 
+// Replaces n with m in all uses, including self-loops.
 void PhaseIterGVN::replace_in_uses(Node* n, Node* m) {
   assert(n != nullptr, "sanity");
   add_users_to_worklist(n);
--- a/src/hotspot/share/opto/phaseX.cpp
+++ b/src/hotspot/share/opto/phaseX.cpp
@@ -2491,11 +2491,9 @@ void PhaseIterGVN::replace_in_uses(Node* n, Node* m) {
   add_users_to_worklist(n);
   for (DUIterator_Fast imax, i = n->fast_outs(imax); i < imax; i++) {
     Node* u = n->fast_out(i);
-    if (u != n) {
-      rehash_node_delayed(u);
-      int nb = u->replace_edge(n, m);
-      --i, imax -= nb;
-    }
+    rehash_node_delayed(u);
+    int nb = u->replace_edge(n, m);
+    --i, imax -= nb;
   }
   assert(n->outcnt() == 0, "all uses must be deleted");
 }
--- a/src/hotspot/share/opto/phaseX.cpp
+++ b/src/hotspot/share/opto/phaseX.cpp
@@ -2488,6 +2488,7 @@ void PhaseIterGVN::subsume_node( Node *old, Node *nn ) {
 
 void PhaseIterGVN::replace_in_uses(Node* n, Node* m) {
   assert(n != nullptr, "sanity");
+  add_users_to_worklist(n);
   for (DUIterator_Fast imax, i = n->fast_outs(imax); i < imax; i++) {
     Node* u = n->fast_out(i);
     if (u != n) {
--- a/test/hotspot/jtreg/compiler/valhalla/inlinetypes/TestPushThroughPhiNotification.java
+++ b/test/hotspot/jtreg/compiler/valhalla/inlinetypes/TestPushThroughPhiNotification.java
@@ -1,0 +1,64 @@
+/*
+ * Copyright (c) 2026, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: test/hotspot/jtreg/compiler/valhalla/inlinetypes/TestPushThroughPhiNotification.java [hunk 1]
```diff
@@ -0,0 +1,64 @@
...
+
+/*
+ * @test
+ * @summary Test that Phi users are revisited when InlineTypeNode is pushed through.
+ * @bug 8386067
+ * @enablePreview
+ * @run driver ${test.main.class}
+ * @run main/othervm -XX:+IgnoreUnrecognizedVMOptions -XX:+UnlockDiagnosticVMOptions
+ *                   -XX:-TieredCompilation -Xbatch
+ *                   -XX:VerifyIterativeGVN=1110 -XX:+StressIGVN
```
TobiHartmann, line 34: Not setting a fixed seed here because this always reproduces after a low number of iterations.

TobiHartmann:
Found some more issues, converting back to draft.

TobiHartmann:
I modified `PhaseIterGVN::replace_in_uses` to handle self-references. This is needed for `PushInlineTypeDown` and should be fine with the other, existing usages (during allocation projection replacement, projections can't reference themselves and initializing stores can also not be self-referential). Tests are running.

test/hotspot/jtreg/compiler/valhalla/inlinetypes/TestPushThroughPhiNotification.java
```diff
@@ -0,0 +1,64 @@
...
+ */
+
+package compiler.valhalla.inlinetypes;
+
+/*
+ * @test
+ * @summary Test that Phi users are revisited when InlineTypeNode is pushed through.
+ * @bug 8386067
+ * @enablePreview
+ * @run driver ${test.main.class}
```
chhagedorn: Should be `main` to allow to run with passed in flags:

```suggestion
 * @run main ${test.main.class}
```

test/hotspot/jtreg/compiler/valhalla/inlinetypes/TestPushThroughPhiNotification.java
```diff
@@ -0,0 +1,64 @@
...
+                rhs = box(i);
+            }
+        }
+        // Phi with (deep) CmpL user from acmp implementation that needs to
+        // be re-visited by IGVN when InlineTypeNode is pushed through.
+        return lhs == rhs;
+    }
+
+    public static void main(String[] args) {
+        for (int i = 0; i < 50_000; i++) {
```
chhagedorn: Do you really need that many iterations?

src/hotspot/share/opto/phaseX.cpp
```diff
@@ -2488,13 +2488,12 @@ void PhaseIterGVN::subsume_node( Node *old, Node *nn ) {
 
 void PhaseIterGVN::replace_in_uses(Node* n, Node* m) {
   assert(n != nullptr, "sanity");
+  add_users_to_worklist(n);
```
chhagedorn, line 2492: This is probably also beneficial for existing uses of `replace_in_uses()`.

chhagedorn (commented):
Otherwise, the fix looks good to me, thanks!

src/hotspot/share/opto/phaseX.cpp
TobiHartmann: Yes, I think so too.

test/hotspot/jtreg/compiler/valhalla/inlinetypes/TestPushThroughPhiNotification.java
TobiHartmann: Good point, I reduced it to 5k and verified that it still reproduces the issue.

TobiHartmann:
Thanks for the review Christian! I addressed your comments and also added a comment to `PhaseIterGVN::replace_in_uses`.

chhagedorn (approved):
Thanks for the update!

TobiHartmann:
Thanks again!

PR #2598, opened 2026-06-30
8386160: [lworld] Post jdk-27+25 compiler/valhalla/inlinetypes/TestIntrinsics.java#id4 assertion

merykitty:
Hi,

It turns out this issue has been fixed by @marc-chevalier. However, after thinking about this issue, I find that the pattern we use `cast_to_exactness(false)->as_instance_type()` to obtain the `Type` of an oop that subtypes a klass pointer is flawed. This is because in the `Type` lattice, `cast_to_exactness` works in the `ciType` dimension, while we also need to handle the `_refined_type` dimension when trying to perform that action.

To be specific, in this issue, we have a klass pointer of unrefined arrays that subtype `Object[]`, `cast_to_exactness` finds that this `TypeKlassPtr` is already inexact, and return it as is. This is problematic, because there is no instance for an unrefined array klass, and `as_instance_type` is confused. @marc-chevalier solves the issue by walking around this issue, and assume when the klass pointer is unrefined, we really mean obtaining the instance that subtypes the pointer, not just those with the same type.

This PR does not actually solve the issue, I just refactored it, so the use sites work in a more rigourous manner, we can fix the implementation after Valhalla being merged.

Testing:

---------

---------
### Progress

### Issue
 * [JDK-8386160](https://bugs.openjdk.org/browse/JDK-8386160): [lworld] Post jdk-27+25 compiler/valhalla/inlinetypes/TestIntrinsics.java#id4 assertion (**Bug** - P3)

### Reviewers
 * [Tobias Hartmann](https://openjdk.org/census#thartmann) (@TobiHartmann - Committer) ⚠️ Review applies to [fcbdc303](https://git.openjdk.org/valhalla/pull/2598/files/fcbdc3032c8c9b5b1868b9de6ce96b18d037e847)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2598/head:pull/2598` \
`$ git checkout pull/2598`

`$ git checkout pull/2598` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2598/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2598`

`$ git pr show -t 2598`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2598.diff">https://git.openjdk.org/valhalla/pull/2598.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2598#issuecomment-4841295180)
</details>

### DIFF
--- a/src/hotspot/share/opto/compile.cpp
+++ b/src/hotspot/share/opto/compile.cpp
@@ -2411,7 +2411,7 @@ void Compile::adjust_flat_array_access_aliases(PhaseIterGVN& igvn) {
           ciArrayKlass* klass = klass_type->exact_klass()->as_array_klass();
           assert(klass->is_flat_array_klass(), "must be a flat array");
           ciInlineKlass* elem_klass = klass->element_klass()->as_inline_klass();
-          const TypeAryPtr* oop_type = klass_type->as_instance_type()->is_aryptr();
+          const TypeAryPtr* oop_type = klass_type->as_exact_instance_type()->is_aryptr();
           assert(oop_type->klass_is_exact(), "must be an exact klass");
 
           Node* base = alloc->in(TypeFunc::Memory);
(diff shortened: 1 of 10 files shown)


TobiHartmann (approved):
This looks reasonable to me.

> This PR does not actually solve the issue

Do you mean that there is an issue left or that the workaround that's currently implemented should be replaced by something more robust? Did you file a follow-up for this yet?

merykitty:
Thanks for your review. The implementation seems not cause an issue, but the idea behind it is shaky, I have created [JDK-8387493](https://bugs.openjdk.org/browse/JDK-8387493).

TobiHartmann:
Makes sense, thanks for the clarification!

merykitty:
Thanks a lot for your review.
/integrate

PR #2601, opened 2026-06-30
8387484: [lworld] javac should warn on use of value classes

lahodaj:
Consider source files like this:
```
$ cat /tmp/ValueClass.java 
value class ValueClass {}
$ cat /tmp/TestValue.java 
public class TestValue {
    ValueClass c;
}
```

When compiled like this:
```
$ javac --enable-preview --release 28 -Xlint:preview /tmp/ValueClass.java 
/tmp/ValueClass.java:1: warning: [preview] value classes are a preview feature and may be removed in a future release.
value class ValueClass {}
^
1 warning
$ javac --enable-preview --release 28 -Xlint:preview -classpath /tmp /tmp/TestValue.java 
warning: [preview] class file for /tmp/ValueClass.class uses preview features of Java SE 28.
1 warning
```

Note that while the `TestValue.class` is marked as preview, there's no warning on the use of `ValueClass`. This is not aligned with the JEP 12 wording:

> When compiling with preview features enabled, any source code reference to (i) a class or interface declared using a preview language feature, ..., causes a preview warning.

Currently javac marks the classfile preview using an explicit test in the `ClassWriter`, but that does not seem quite right. This PR proposes to mark value classes as `Preview.declaredUsingPreviewFeature == true`, which should automatically produce a warning, and mark the appropriate classfile as preview.

For the example above:
```
$ javac --enable-preview --release 28 -Xlint:preview -classpath /tmp /tmp/TestValue.java 
warning: [preview] class file for /tmp/ValueClass.class uses preview features of Java SE 28.
/tmp/TestValue.java:2: warning: [preview] class ValueClass is declared using a preview feature, which may be removed in a future release.
    ValueClass c;
    ^
2 warnings
```

Note this will also report the warning for `Integer`, when preview is enabled:
```
$ cat /tmp/Test.java 
public class Test {
    Integer i;
}
$ javac /tmp/Test.java
$ javac --enable-preview --release 28 -Xlint:preview /tmp/Test.java
warning: [preview] class file for /@/java.base/java/lang/Integer.sig uses preview features of Java SE 28.
/tmp/Test.java:2: warning: [preview] class Integer is declared using a preview feature, which may be removed in a future release.
    Integer i;
    ^
2 warnings
```

---------

---------
### Progress

### Issue
 * [JDK-8387484](https://bugs.openjdk.org/browse/JDK-8387484): [lworld] javac should warn on use of value classes (**Bug** - P4)

### Reviewers
 * [Maurizio Cimadamore](https://openjdk.org/census#mcimadamore) (@mcimadamore - Committer) ⚠️ Review applies to [d5313afa](https://git.openjdk.org/valhalla/pull/2601/files/d5313afa42fc6744bb10265a964cdbddcb7e623e)
 * [Chen Liang](https://openjdk.org/census#liach) (@liach - Committer) ⚠️ Review applies to [d5313afa](https://git.openjdk.org/valhalla/pull/2601/files/d5313afa42fc6744bb10265a964cdbddcb7e623e)
 * [Vicente Romero](https://openjdk.org/census#vromero) (@vicente-romero-oracle - Committer) ⚠️ Review applies to [d5313afa](https://git.openjdk.org/valhalla/pull/2601/files/d5313afa42fc6744bb10265a964cdbddcb7e623e)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2601/head:pull/2601` \
`$ git checkout pull/2601`

`$ git checkout pull/2601` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2601/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2601`

`$ git pr show -t 2601`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2601.diff">https://git.openjdk.org/valhalla/pull/2601.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2601#issuecomment-4843739504)
</details>

### DIFF
--- a/src/jdk.compiler/share/classes/com/sun/tools/javac/code/Preview.java
+++ b/src/jdk.compiler/share/classes/com/sun/tools/javac/code/Preview.java
@@ -255,7 +255,7 @@ public class Preview {
      * @return true iff sym has been declared using a preview language feature
      */
     public boolean declaredUsingPreviewFeature(Symbol sym) {
-        return false;
+        return sym.isValueClass();
     }
 
     public void checkSourceLevel(DiagnosticPosition pos, Feature feature) {
--- a/src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/ClassWriter.java
+++ b/src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/ClassWriter.java
@@ -1002,9 +1002,6 @@ public class ClassWriter extends ClassFile {
         Type fldType = v.erasure(types);
         if (fldType.requiresLoadableDescriptors(v.owner)) {
             poolWriter.enterLoadableDescriptorsClass(fldType.tsym);
-            if (preview.isPreview(Source.Feature.VALUE_CLASSES)) {
-                preview.markUsesPreview(null);
-            }
         }
         int acountIdx = beginAttrs();
         int acount = 0;
@@ -1035,17 +1032,11 @@ public class ClassWriter extends ClassFile {
         for (Type t : mtype.getParameterTypes()) {
             if (t.requiresLoadableDescriptors(m.owner)) {
                 poolWriter.enterLoadableDescriptorsClass(t.tsym);
-                if (preview.isPreview(Source.Feature.VALUE_CLASSES)) {
-                    preview.markUsesPreview(null);
-                }
             }
         }
         Type returnType = mtype.getReturnType();
         if (returnType.requiresLoadableDescriptors(m.owner)) {
             poolWriter.enterLoadableDescriptorsClass(returnType.tsym);
-            if (preview.isPreview(Source.Feature.VALUE_CLASSES)) {
-                preview.markUsesPreview(null);
-            }
         }
         int acountIdx = beginAttrs();
         int acount = 0;
--- a/test/langtools/tools/javac/patterns/DominationWithPP.out
+++ b/test/langtools/tools/javac/patterns/DominationWithPP.out
@@ -11,4 +11,6 @@ Domination.java:193:18: compiler.err.pattern.dominated
 Domination.java:202:18: compiler.err.pattern.dominated
 Domination.java:211:18: compiler.err.pattern.dominated
 Domination.java:228:18: compiler.err.pattern.dominated
+- compiler.note.preview.filename: Domination.java, DEFAULT
+- compiler.note.preview.recompile
 13 errors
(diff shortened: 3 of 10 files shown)


mcimadamore (approved):
Nice improvement!

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Preview.java
```diff
@@ -255,7 +255,7 @@ public Error disabledError(JavaFileObject classfile, int majorVersion) {
      * @return true iff sym has been declared using a preview language feature
      */
     public boolean declaredUsingPreviewFeature(Symbol sym) {
-        return false;
+        return sym.isValueClass();
```
liach, line 258: Should this include records too given how its semantics are changed? Maybe in another RFE?

liach (approved):

vicente-romero-oracle (approved):
lgtm

src/jdk.compiler/share/classes/com/sun/tools/javac/code/Preview.java
lahodaj: I am not completely certain, but I think this should not include records.

My understanding is that for records, only the implementation of the record's constructor changes, and that this is transparent(?) for the users of the record. I.e. the change not in the declaration/signatures, but in the implementation. If a class, e.g. `PM`, uses a preview feature inside an implementation only (i.e. not in a declaration), we don't taint another class, e.g. `Use`, that only uses `PM`.

For example:
```
$ cat PM.java 
public class PM {
    private boolean test(Object o) {
        return o instanceof byte b;
    }
}
$ cat Use.java 
public class Use {
    PM pm;
}
$ javac --enable-preview --source 28 -Xlint:preview PM.java Use.java
PM.java:3: warning: [preview] primitive patterns are a preview feature and may be removed in a future release.
        return o instanceof byte b;
                            ^
1 warning
```

There's no warning for the use of `PM` in `Use`, and `Use.class` is not marked a preview classfile.

Even though `PM` uses a preview feature (primitive patterns), it only does so inside an implementation, and so the use of the preview feature is invisible to the use site, leading to no warning.

My understanding (possibly wrong) is that the change in record constructors is similar to the use of pattern matching in the implementation of `PM`.

liach:
I think you can go ahead and integrate if `tools/javac` tests are clear.

PR #2602, opened 2026-06-30
8387488: [lworld] C2: acmp expansion fails with assert(replace != nullptr) failed: must succeed

TobiHartmann:
When slightly modifying the test that I added for [JDK-8386067](https://bugs.openjdk.org/browse/JDK-8386067), I hit an assert in `CallStaticJavaNode::replace_is_substitutable` because expansion of the substitutability call failed and `emit_substitutability_check` returned `nullptr` although `can_emit_substitutability_check` returned `true`.

In the new test, the problematic field is `ObjectHolder::obj` which, although it is of type `Object`, is always `exact` and therefore can't be a value object and does not require a substitutability test. `InlineTypeNode::can_emit_substitutability_check` correctly returns `true`:
https://github.com/openjdk/valhalla/blob/2ae8e47da016bb4d7718e7923c964c2b91c638e6/src/hotspot/share/opto/inlinetypenode.cpp#L611-L614

However, during recursive expansion, nullable value object fields are null-checked and then compared field-by-field:
https://github.com/openjdk/valhalla/blob/2ae8e47da016bb4d7718e7923c964c2b91c638e6/src/hotspot/share/opto/inlinetypenode.cpp#L910

The code unconditionally rebuilt those fields with `InlineTypeNode::make_from_oop` after the null and type checks because there's a `CastPP` in-between:
https://github.com/openjdk/valhalla/blob/2ae8e47da016bb4d7718e7923c964c2b91c638e6/src/hotspot/share/opto/inlinetypenode.cpp#L955-L960

 If the field was already scalarized, this discarded the existing `InlineTypeNode` and reloaded fields from the oop, losing precise information such as exact Object field Phis. The later Object field comparison then looks non-exact and `emit_substitutability_check_pointer` returns `nullptr`:
https://github.com/openjdk/valhalla/blob/2ae8e47da016bb4d7718e7923c964c2b91c638e6/src/hotspot/share/opto/inlinetypenode.cpp#L722-L724

The fix is to preserve already-scalarized recursive field operands when their inline klass matches the comparison klass, and only checkcast/reload operands that are not scalarized. With the fix, exact type information is preserved and we emit a pointer comparison for the exact object fields:
https://github.com/openjdk/valhalla/blob/2ae8e47da016bb4d7718e7923c964c2b91c638e6/src/hotspot/share/opto/inlinetypenode.cpp#L700-L702

The IR Framework test verifies this successful expansion.

Note: The null and type checks are not immediately folded because by GVN because we need to set `set_delay_transform` here:
https://github.com/openjdk/valhalla/blob/2ae8e47da016bb4d7718e7923c964c2b91c638e6/src/hotspot/share/opto/callnode.cpp#L1412-L1416

Thanks,
Tobias

---------

---------
### Progress

### Issue
 * [JDK-8387488](https://bugs.openjdk.org/browse/JDK-8387488): [lworld] C2: acmp expansion fails with assert(replace != nullptr) failed: must succeed (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2602/head:pull/2602` \
`$ git checkout pull/2602`

`$ git checkout pull/2602` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2602/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2602`

`$ git pr show -t 2602`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2602.diff">https://git.openjdk.org/valhalla/pull/2602.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2602#issuecomment-4844556105)
</details>

### DIFF
--- a/src/hotspot/share/opto/inlinetypenode.cpp
+++ b/src/hotspot/share/opto/inlinetypenode.cpp
@@ -953,6 +953,8 @@ Node* InlineTypeNode::emit_substitutability_check(GraphKit* kit, Node* lhs, Node
       Node* vk_klass = igvn.makecon(vk_klass_type);
 
       // Both must be vk
+      // No need to null-check/checkcast/reload an already scalarized inline type:
+      // its fields are already available directly from the InlineTypeNode.
       if (cur_lhs_inline != nullptr && cur_lhs_inline->inline_klass() == vk) {
         cur_lhs_field = cur_lhs_inline;
       } else {
--- a/src/hotspot/share/opto/inlinetypenode.cpp
+++ b/src/hotspot/share/opto/inlinetypenode.cpp
@@ -953,8 +953,7 @@ Node* InlineTypeNode::emit_substitutability_check(GraphKit* kit, Node* lhs, Node
       Node* vk_klass = igvn.makecon(vk_klass_type);
 
       // Both must be vk
-      // No need to null-check/checkcast/reload an already scalarized inline type:
-      // its fields are already available directly from the InlineTypeNode.
+      // InlineTypeNodes need no (new) field loads, so we can use the uncasted value below.
       if (cur_lhs_inline != nullptr && cur_lhs_inline->inline_klass() == vk) {
         cur_lhs_field = cur_lhs_inline;
       } else {
--- a/src/hotspot/share/opto/inlinetypenode.cpp
+++ b/src/hotspot/share/opto/inlinetypenode.cpp
@@ -865,6 +865,8 @@ Node* InlineTypeNode::emit_substitutability_check(GraphKit* kit, Node* lhs, Node
 
       Node* cur_lhs_field = cur_lhs->field_value(field_idx);
       Node* cur_rhs_field = cur_rhs->field_value(field_idx);
+      InlineTypeNode* cur_lhs_inline = cur_lhs_field->isa_InlineType();
+      InlineTypeNode* cur_rhs_inline = cur_rhs_field->isa_InlineType();
 
       Node* preprocess = emit_substitutability_check_pointer(kit, result, cur_lhs_field, cur_rhs_field);
       if (kit->stopped()) {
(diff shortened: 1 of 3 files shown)


test/hotspot/jtreg/compiler/valhalla/inlinetypes/TestOptimizePtrCompare.java
```diff
@@ -164,6 +164,7 @@ static MyValue2 notInlined5(MyValue2 v) {
         return v;
     }
 
+    // TODO 8376254: C1 bails out if the type of the nullable flat field is uninitialized
```
TobiHartmann, line 167: Just found this missing TODO when browsing code.

src/hotspot/share/opto/inlinetypenode.cpp
```diff
@@ -951,21 +953,29 @@ Node* InlineTypeNode::emit_substitutability_check(GraphKit* kit, Node* lhs, Node
       Node* vk_klass = igvn.makecon(vk_klass_type);
 
       // Both must be vk
-      Node* not_vk = kit->top();
-      cur_lhs_field = kit->gen_checkcast(cur_lhs_field, vk_klass, &not_vk);
-      if (!not_vk->is_top()) {
-        region->add_req(not_vk);
-        result->add_req(igvn.intcon(0));
+      if (cur_lhs_inline != nullptr && cur_lhs_inline->inline_klass() == vk) {
```
merykitty, line 957: You can add a comment that it is not an issue if `cur_lhs_inline` is nullable and `cur_lhs_field` is null-checked, it is null-checked so that we can load the fields, and an `InlineTypeNode` does that already.

test/hotspot/jtreg/compiler/valhalla/inlinetypes/TestLWorld.java
```diff
@@ -5479,4 +5479,36 @@ public static void testBadCastToInlineKlass_verifier() {
             // expected
         }
     }
+
+    // TODO 8376254: C1 bails out if the type of the nullable flat field is uninitialized
+    static ObjectHolderHolder tmp = new ObjectHolderHolder();
```
merykitty: It would probably be better to name this more specific to its type name so as not to have name collision.

marc-chevalier (commented):
Haven't found anything crazy, but with low confidence. Happy @merykitty was faster than me at reviewing this.

TobiHartmann:
Thanks for the reviews Quan Anh and Marc! @merykitty I added a comment and adjusted the field name in the test. Please let me know what you think.

src/hotspot/share/opto/inlinetypenode.cpp
merykitty: It's not quite what I want to say. The null check is still present, the thing is that we ignored the null-checked value (the `CastPP`) and just use the raw value. This is because normally, we need a non-null `CastPP` to load from the object. That is unnecessary for an `InlineTypeNode` because we don't perform the loads.

TobiHartmann: Okay, here's a new attempt :slightly_smiling_face:

merykitty (approved):
Thanks a lot, it looks good to me.

TobiHartmann:
Thanks for the review Quan Anh!

PR #2604, opened 2026-06-30
8387540: [lworld] ForceEarlyReturnVoid should fail with OPAQUE_FRAME if target thread's top frame is value class constructor

AlanBateman:
JVMTI ForceEarlyReturnVoid can be used to return early from a constructor or class initializer. Early return from a constructor with strictly-initialized instance fields is problematic, as is early return from a class initializer when there are strictly-initialized static fields. The proposal for the initial integration is to keep it simple and just disallow ForceEarlyReturnVoid to return early from a value class constructor.

---------

---------
### Progress

### Issue
 * [JDK-8387540](https://bugs.openjdk.org/browse/JDK-8387540): [lworld] ForceEarlyReturnVoid should fail with OPAQUE_FRAME if target thread's top frame is value class constructor (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2604/head:pull/2604` \
`$ git checkout pull/2604`

`$ git checkout pull/2604` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2604/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2604`

`$ git pr show -t 2604`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2604.diff">https://git.openjdk.org/valhalla/pull/2604.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2604#issuecomment-4847227544)
</details>

### DIFF
--- a/src/hotspot/share/prims/jvmti.xml
+++ b/src/hotspot/share/prims/jvmti.xml
@@ -3377,14 +3377,14 @@ err = (*jvmti)-&gt;Deallocate(jvmti, stack_info);
       <synopsis>Force Early Return - Void</synopsis>
       <description>
         This function can be used to return from a method with a return type
-        of <code>void</code>, a constructor or a class initializer.
+        of <code>void</code>.
         A constructor is a special method named <code>&lt;init&gt;</code> with
         a return type of <code>void</code>. A class initializer is a special
-        method named named <code>&lt;clinit&gt;</code> with  a return type of
+        method named named <code>&lt;clinit&gt;</code> with a return type of
         <code>void</code>.
         <p/>
-        This function can not be used to return early from the constructor
-        of a value class.
+        When preview features are enabled, this function can not be used to return
+        early from the constructor of a value class.
       </description>
       <origin>new</origin>
       <capabilities>
--- a/src/hotspot/share/prims/jvmti.xml
+++ b/src/hotspot/share/prims/jvmti.xml
@@ -3401,8 +3401,8 @@ err = (*jvmti)-&gt;Deallocate(jvmti, stack_info);
       <errors>
         <error id="JVMTI_ERROR_OPAQUE_FRAME">
           The implementation is unable to force the current frame to return
-          (e.g. current frame is executing a native method or the current frame is the
-          constructor for a value class)
+          (e.g. current frame is executing a native method or if preview features
+           are enabled, the current frame is the constructor for a value class)
         </error>
         <error id="JVMTI_ERROR_TYPE_MISMATCH">
           The called method has a result type.
--- a/src/hotspot/share/prims/jvmti.xml
+++ b/src/hotspot/share/prims/jvmti.xml
@@ -3376,11 +3376,15 @@ err = (*jvmti)-&gt;Deallocate(jvmti, stack_info);
     <function id="ForceEarlyReturnVoid" num="86" since="1.1">
       <synopsis>Force Early Return - Void</synopsis>
       <description>
-        This function can be used to return from a method with no result type.
-        That is, the called method must be declared <code>void</code>.
+        This function can be used to return from a method with a return type
+        of <code>void</code>, a constructor or a class initializer.
+        A constructor is a special method named <code>&lt;init&gt;</code> with
+        a return type of <code>void</code>. A class initializer is a special
+        method named named <code>&lt;clinit&gt;</code> with  a return type of
+        <code>void</code>.
         <p/>
-        An implementation may also allow this function be used to return early
-        from a class initializer or constructor.
+        This function can not be used to return early from the constructor
+        of a value class.
       </description>
       <origin>new</origin>
       <capabilities>
--- a/src/hotspot/share/prims/jvmti.xml
+++ b/src/hotspot/share/prims/jvmti.xml
@@ -3378,6 +3378,9 @@ err = (*jvmti)-&gt;Deallocate(jvmti, stack_info);
       <description>
         This function can be used to return from a method with no result type.
         That is, the called method must be declared <code>void</code>.
+        <p/>
+        An implementation may also allow this function be used to return early
+        from a class initializer or constructor.
       </description>
       <origin>new</origin>
       <capabilities>
@@ -3394,7 +3397,8 @@ err = (*jvmti)-&gt;Deallocate(jvmti, stack_info);
       <errors>
         <error id="JVMTI_ERROR_OPAQUE_FRAME">
           The implementation is unable to force the current frame to return
-          (e.g. current frame is executing a native method).
+          (e.g. current frame is executing a native method or the current frame is the
+          constructor for a value class)
         </error>
         <error id="JVMTI_ERROR_TYPE_MISMATCH">
           The called method has a result type.
--- a/src/java.se/share/data/jdwp/jdwp.spec
+++ b/src/java.se/share/data/jdwp/jdwp.spec
@@ -2153,7 +2153,8 @@ JDWP "Java(tm) Debug Wire Protocol"
             (Error THREAD_NOT_SUSPENDED)
             (Error OPAQUE_FRAME      "Unable to force the current frame to return "
                                      "(e.g. the current frame is executing a native method or "
-                                     "the current frame is the constructor for a value class).")
+                                     "if preview features are enabled, the current frame is "
+                                     "the constructor for a value class).")
             (Error NO_MORE_FRAMES)
             (Error NOT_IMPLEMENTED)
             (Error TYPE_MISMATCH   "Value is not an appropriate type for the "
--- a/src/java.se/share/data/jdwp/jdwp.spec
+++ b/src/java.se/share/data/jdwp/jdwp.spec
@@ -2129,6 +2129,9 @@ JDWP "Java(tm) Debug Wire Protocol"
         "language method. Forcing return on a thread with only one "
         "frame on the stack causes the thread to exit when resumed. "
         "<p>"
+        "When preview features are enabled, the method can not be the "
+        "constructor of a value class."
+        "<p>"
         "For void methods, the value must be a void value. "
         "For methods that return primitive values, the value's type must "
         "match the return type exactly.  For object values, there must be a "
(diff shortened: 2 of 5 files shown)


src/hotspot/share/prims/jvmti.xml
```diff
@@ -3394,7 +3397,8 @@ err = (*jvmti)-&gt;Deallocate(jvmti, stack_info);
       <errors>
         <error id="JVMTI_ERROR_OPAQUE_FRAME">
           The implementation is unable to force the current frame to return
-          (e.g. current frame is executing a native method).
+          (e.g. current frame is executing a native method or the current frame is the
+          constructor for a value class)
```
sspitsyn: It is nice that the value class initializers are not impacted, so there is no need to return `OPAQUE_FRAME` for value class initializer frames.

test/hotspot/jtreg/serviceability/jvmti/ForceEarlyReturn/ForceEarlyReturnValueClass.java
```diff
@@ -0,0 +1,148 @@
...
+        class TestCase {
+            static volatile boolean started;
+            static volatile boolean stop;
+            static volatile boolean finished;
+            static value class ValueClass {
+                ValueClass() {
+                    TestCase.started = true;
+                    while (!TestCase.stop) {}
+                    TestCase.finished = true;
+                }
```
sspitsyn, line 56: Nit: We may need one more test case for a helper method invoked from a value object constructor. My understanding is that such methods can not initialize the value object fields.

sspitsyn (commented):
The fix looks good in general. I've posted one nit with a suggestion which is not critical. Thank you for making this update!

src/hotspot/share/prims/jvmti.xml
AlanBateman: The more general restriction should probably be that ForceEarlyReturnVoid should fail for constructors if the class has strictly-initialized instance fields, and fail for class initializers if the class has strictly-initialized static fields. We can look at that later.

sspitsyn: Thanks. I wonder if we need to say that it only applies when preview features are enabled.
For instance, there is this phrase in the `ObjectFree` event spec:
"When preview features are enabled, this event does not support value object allocations."

sspitsyn (approved):

test/hotspot/jtreg/serviceability/jvmti/ForceEarlyReturn/ForceEarlyReturnValueClass.java
AlanBateman: The instance fields of a value class are strictly-initialized fields so have to be initialized before super or leaking  "this" by invoking an instance method. It would be feasible to have the test exercise ForceEarlyReturn to return early from something that the constructor invokes after super. I can a further test where there is a value class constructor on the stack that is not the top frame.

src/hotspot/share/prims/jvmti.xml
```diff
@@ -3376,8 +3376,15 @@ err = (*jvmti)-&gt;Deallocate(jvmti, stack_info);
     <function id="ForceEarlyReturnVoid" num="86" since="1.1">
       <synopsis>Force Early Return - Void</synopsis>
       <description>
-        This function can be used to return from a method with no result type.
-        That is, the called method must be declared <code>void</code>.
+        This function can be used to return from a method with a return type
+        of <code>void</code>, a constructor or a class initializer.
```
dholmes-ora: I thought class initializers were implicitly disallowed because the caller you are returning to must be a Java frame?

src/hotspot/share/prims/jvmti.xml
```diff
@@ -3376,8 +3376,15 @@ err = (*jvmti)-&gt;Deallocate(jvmti, stack_info);
...
+        A constructor is a special method named <code>&lt;init&gt;</code> with
+        a return type of <code>void</code>. A class initializer is a special
+        method named named <code>&lt;clinit&gt;</code> with  a return type of
+        <code>void</code>.
+        <p/>
+        This function can not be used to return early from the constructor
+        of a value class.
```
dholmes-ora: FWIW I don't think you can define sensible semantics for early return from a constructor with non-value types. What does it mean for final fields for example, if you can re-execute a constructor?

AlanBateman: ForceEarlyReturnVoid can be used to return early from a constructor. A crazy thing to do as it leaves things uninitialized or in an inconsistent state. It means that some final fields may have their default value.

All we're doing here is taming it for value class constructors as the instance fields are strictly-initialized and it would be even more crazy and problematic to allow early return before the fields have been initialized. It would be possible to make this more finer grain, meaning don't allow return before super but that could be more long term if needed.

src/hotspot/share/prims/jvmti.xml
AlanBateman: Okay, I can make that clearer.

src/hotspot/share/prims/jvmti.xml
AlanBateman: Are you thinking of PopFrame, which I don't think has ever allowed a pop of the top-frame when in a class initializer?

src/hotspot/share/prims/jvmti.xml
```diff
@@ -3376,8 +3376,15 @@ err = (*jvmti)-&gt;Deallocate(jvmti, stack_info);
...
       <synopsis>Force Early Return - Void</synopsis>
       <description>
-        This function can be used to return from a method with no result type.
-        That is, the called method must be declared <code>void</code>.
+        This function can be used to return from a method with a return type
+        of <code>void</code>.
+        A constructor is a special method named <code>&lt;init&gt;</code> with
+        a return type of <code>void</code>. A class initializer is a special
+        method named named <code>&lt;clinit&gt;</code> with a return type of
+        <code>void</code>.
```
plummercj, line 3384: I'm not sure of the purpose of calling out what constructors and class initializers are.

src/hotspot/share/prims/jvmti.xml
```diff
@@ -3394,7 +3401,8 @@ err = (*jvmti)-&gt;Deallocate(jvmti, stack_info);
       <errors>
         <error id="JVMTI_ERROR_OPAQUE_FRAME">
           The implementation is unable to force the current frame to return
-          (e.g. current frame is executing a native method).
+          (e.g. current frame is executing a native method or the current frame is the
+          constructor for a value class)
```
plummercj: I think you still need to mention "when preview mode is enabled" here. I know you already did in the description, but seems to me it should be repeated here.

src/java.se/share/data/jdwp/jdwp.spec
```diff
@@ -2149,7 +2152,8 @@ JDWP "Java(tm) Debug Wire Protocol"
             (Error INVALID_OBJECT    "Thread or value is not a known ID.")
             (Error THREAD_NOT_SUSPENDED)
             (Error OPAQUE_FRAME      "Unable to force the current frame to return "
-                                     "(e.g. the current frame is executing a native method).")
+                                     "(e.g. the current frame is executing a native method or "
+                                     "the current frame is the constructor for a value class).")
```
plummercj: Mention "when in preview mode" here also? Also, what about JDI ObjectReference.forceEarlyReturn()?

src/hotspot/share/prims/jvmti.xml
AlanBateman: This is setup for the sentence that follows. Without it, it would be too much of a surprise to say that the function cannot be used to return early from the constructor of a value class.

src/hotspot/share/prims/jvmti.xml
AlanBateman: Okay although I don't think it is strictly needed as it's already in the description (as you noted).

src/hotspot/share/prims/jvmti.xml
plummercj: I had considered that. There is no mention of class initializer below. Also, I would think that these terms are already generally understood and don't need clarification here.

AlanBateman: It's technical debt that ForceEarlyReturnVoid didn't clearly document that it could be used to force early return from constructors or static initializers. There was discussion on this in the CSR wondering about this.

In time, the restriction will need to be extended to class initializers when the class has strictly-initialized static fields.

sspitsyn (approved):
Nice spec tweaks and new test cases. Looks good.

src/hotspot/share/prims/jvmti.xml
sspitsyn: I guess, David is thinking of `PopFrame` which does not allow to return to a non-Java method.
However, it seems that the ForceEarlyReturn<Type> functions do not have this restriction. I do not remember the exact details, but it seems there is no problem to early return to non-Java methods in general.

dholmes-ora: No, I thought it had been mentioned by someone else in our internal discussions that the caller had to be a Java frame and so class-initializers did not have to have any special handling.

dholmes-ora: Though I will say that it seems very inconsistent that two very similar operations - terminate the current frame- would have different constraints in this regard.

AlanBateman: ForceEarlyReturn is simpler in that it just needs to arm the suspended target VM to return early when resumed. 

PopFrame is more complicated as it involves removing the current activation and restoring state to that the method can be re-invoked. Surprisingly, the spec doesn't prohibit PopFrame when the target thread is suspended in a class initializer. It fails with OPAQUE_FRAME of course because the caller is the VM. Maybe some other VM might allow it but it's seems an insane thing to do in the first place (the developer using a debugger wants to re-run the class initializer??). Maybe some future work in this area could have the spec prohibit it.

AlanBateman:
> Nice spec tweaks and new test cases. Looks good.

Thanks, I think we should just integrate this and move on.

src/hotspot/share/prims/jvmti.xml
dholmes-ora: The only difference I see between them is that force returns so we proceed after the call, and pop returns such that we redo the call. Both terminate the current activation immediately.

I thought PopFrame implicitly prohibited class initializers because both frames must be Java frames? Though I suppose it is an implementation detail that class initializers have a bunch of native frames before them.

Anyway more work needed on this aspect of the spec going forward.

PR #2607, opened 2026-06-30
8387552: [lworld] Add new capability to enable value objects support for JVMTI VMObjectAlloc and SampledObjectAlloc

sspitsyn:
This update is for Valhalla pre-integration. It introduces new JVMTI capability `can_support_value_objects`.
The variable `_can_support_value_objects_count` is introduced for optimization. It follows the pattern of the `can_support_virtual_threads` capability.

Additionally, this update includes test fixes:
 - test/hotspot/jtreg/serviceability/jvmti/HeapMonitor/libHeapMonitorTest.cpp:
   - Removed the `if (!jni->HasIdentity(object))` check  because it is not needed anymore as the JVMTI capability `can_support_value_objects` is no acquired by the test
 
 - These two tests are updated to provide both positive and negative coverage for new capability:
    test/hotspot/jtreg/serviceability/jvmti/valhalla/VMObjectAllocValue
    test/hotspot/jtreg/serviceability/jvmti/valhalla/SampledObjectAllocValue

Testing:
- Ran updated tests locally
- Submitted mach5 tiers 1-6

---------

---------
### Progress

### Issue
 * [JDK-8387552](https://bugs.openjdk.org/browse/JDK-8387552): [lworld] Add new capability to enable value objects support for JVMTI VMObjectAlloc and SampledObjectAlloc (**Enhancement** - P4)

### Reviewers
 * [Leonid Mesnik](https://openjdk.org/census#lmesnik) (@lmesnik - Committer) ⚠️ Review applies to [ee6e61d7](https://git.openjdk.org/valhalla/pull/2607/files/ee6e61d7f5f9651804fb4bd959f7883d7919d8cc)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2607/head:pull/2607` \
`$ git checkout pull/2607`

`$ git checkout pull/2607` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2607/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2607`

`$ git pr show -t 2607`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2607.diff">https://git.openjdk.org/valhalla/pull/2607.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2607#issuecomment-4848711048)
</details>

### DIFF
--- a/src/hotspot/share/prims/jvmti.xml
+++ b/src/hotspot/share/prims/jvmti.xml
@@ -10714,7 +10714,7 @@ myInit() {
         <description>
           Can support value objects.
           When preview features are enabled and this capability is enabled, then
-          the following two events are enabled for value object allocations:
+          events are sent for value objects when the following two events are enabled:
           <eventlink id="VMObjectAlloc"></eventlink>,
           <eventlink id="SampledObjectAlloc"></eventlink>.
         </description>
@@ -14024,7 +14024,7 @@ myInit() {
       for some or all of their methods can send this event.
       <p/>
       When preview features are enabled and the capability <code>can_support_value_objects</code>
-      is enabled, this event is sent for the allocated objects that have no identity.
+      is enabled, this event is sent for the value object allocations.
       <code>null</code> is passed for the <paramlink id="object"></paramlink> parameter in such cases.
       <p/>
       Note that the <internallink
--- a/src/hotspot/share/prims/jvmti.xml
+++ b/src/hotspot/share/prims/jvmti.xml
@@ -10710,6 +10710,15 @@ myInit() {
           <eventlink id="VirtualThreadEnd"></eventlink>.
         </description>
       </capabilityfield>
+      <capabilityfield id="can_support_value_objects" since="28">
+        <description>
+          Can support value objects.
+          When preview features are enabled and this capability is enabled, then
+          the following two events are enabled for value object allocations:
+          <eventlink id="VMObjectAlloc"></eventlink>,
+          <eventlink id="SampledObjectAlloc"></eventlink>.
+        </description>
+      </capabilityfield>
     </capabilitiestypedef>
 
     <function id="GetPotentialCapabilities" jkernel="yes" phase="onload" num="140">
@@ -14013,9 +14022,11 @@ myInit() {
       VM. These methods should send this event.
       Virtual machines which are incapable of bytecode instrumentation
       for some or all of their methods can send this event.
-      When preview features are enabled and the allocated object has no identity,
-      <code>null</code> is passed for the <paramlink id="object"></paramlink> parameter.
-
+      <p/>
+      When preview features are enabled and the capability <code>can_support_value_objects</code>
+      is enabled, this event is sent for the allocated objects that have no identity.
+      <code>null</code> is passed for the <paramlink id="object"></paramlink> parameter in such cases.
+      <p/>
       Note that the <internallink
       id="SampledObjectAlloc">SampledObjectAlloc</internallink>
       event is triggered on all Java object allocations, including those
@@ -14041,6 +14052,7 @@ myInit() {
     <origin>new</origin>
     <capabilities>
       <required id="can_generate_vm_object_alloc_events"></required>
+      <capability id="can_support_value_objects"></capability>
     </capabilities>
     <parameters>
       <param id="jni_env">
@@ -14061,7 +14073,8 @@ myInit() {
         <jobject/>
           <description>
             JNI local reference to the object that was allocated.
-            Null when preview features are enabled and the allocated object has no identity.
+            Null when preview features are enabled, the capability <code>can_support_value_objects</code>
+            is enabled and the allocated object has no identity.
           </description>
       </param>
       <param id="object_klass">
@@ -14108,13 +14121,15 @@ myInit() {
       In conjunction with weak references and the function
       <functionlink id="GetStackTrace"></functionlink>, a user can track which objects were allocated from which
       stack trace, and which are still live during the execution of the program.
-      When preview features are enabled and the allocated object has no identity,
-      <code>null</code> is passed for the <paramlink id="object"></paramlink> parameter.
-      Note that weak references support identity objects only.
+      <p/>
+      When preview features are enabled and the capability <code>can_support_value_objects</code>
+      is enabled, this event is sent for the allocated objects that have no identity.
+      <code>null</code> is passed for the <paramlink id="object"></paramlink> parameter in such cases.
     </description>
     <origin>new</origin>
     <capabilities>
       <required id="can_generate_sampled_object_alloc_events"></required>
+      <capability id="can_support_value_objects"></capability>
     </capabilities>
     <parameters>
       <param id="jni_env">
@@ -14135,7 +14150,8 @@ myInit() {
         <jobject/>
         <description>
           JNI local reference to the object that was allocated.
-          Null when preview features are enabled and the allocated object has no identity.
+          Null when preview features are enabled, the capability <code>can_support_value_objects</code>
+          is enabled and the allocated object has no identity.
         </description>
       </param>
       <param id="object_klass">
@@ -15502,6 +15518,12 @@ typedef void (JNICALL *jvmtiEventVMInit)
   </change>
   <change date="10 January 2025" version="25.0.0">
       Add new function ClearAllFramePops. Needed to speedup debugger single stepping.
+  </change>
+  <change date="30 June 2026" version="28.0.0">
+      Support for value objects (Preview):
+      Add new capability: can_support_value_objects which enables
+      the following events for value objects:
+      JVMTI_EVENT_VM_OBJECT_ALLOC, JVMTI_EVENT_SAMPLED_OBJECT_ALLOC.
   </change>
 </changehistory>
 
--- a/src/hotspot/share/prims/jvmti.xml
+++ b/src/hotspot/share/prims/jvmti.xml
@@ -14074,7 +14074,7 @@ myInit() {
           <description>
             JNI local reference to the object that was allocated.
             Null when preview features are enabled, the capability <code>can_support_value_objects</code>
-            is enabled and the allocated object has no identity.
+            is enabled, and the allocated object has no identity.
           </description>
       </param>
       <param id="object_klass">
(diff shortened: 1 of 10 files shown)


test/hotspot/jtreg/serviceability/jvmti/HeapMonitor/libHeapMonitorTest.cpp
```diff
@@ -529,11 +529,6 @@ static void event_storage_add(EventStorage* storage,
   jint count;
   jvmtiError err;
 
-  if (!jni->HasIdentity(object)) {
```
lmesnik, line 532: I wonder if it makes sense to update HepMontior tests to work with value classes events enabled? 
Not needed to implement in this fix, just file RFE for this.

src/hotspot/share/prims/jvmtiExport.cpp
```diff
@@ -2967,17 +2968,22 @@ void JvmtiExport::post_vm_object_alloc(JavaThread *thread, oop object) {
...
   EVT_TRIG_TRACE(JVMTI_EVENT_VM_OBJECT_ALLOC, ("[%s] Trg vm object alloc triggered",
                       JvmtiTrace::safe_get_thread_name(thread)));
   JvmtiEnvIterator it;
   for (JvmtiEnv* env = it.first(); env != nullptr; env = it.next(env)) {
-    if (env->is_enabled(JVMTI_EVENT_VM_OBJECT_ALLOC)) {
+    if (env->is_enabled(JVMTI_EVENT_VM_OBJECT_ALLOC) &&
+        (!is_inline || env->get_capabilities()->can_support_value_objects != 0)) {
       EVT_TRACE(JVMTI_EVENT_VM_OBJECT_ALLOC, ("[%s] Evt vmobject alloc sent %s",
                                          JvmtiTrace::safe_get_thread_name(thread),
-                                         object==nullptr? "null" : object->klass()->external_name()));
```
lmesnik, line 2980: Seems the 2980 line was incorrect even without the fix.
The object can't be null. See line 2964.

src/hotspot/share/prims/jvmtiExport.cpp
```diff
@@ -3004,19 +3010,24 @@ void JvmtiExport::post_sampled_object_alloc(JavaThread *thread, oop object) {
...
+    return;
+  }
 
   EVT_TRIG_TRACE(JVMTI_EVENT_SAMPLED_OBJECT_ALLOC,
                  ("[%s] Trg sampled object alloc triggered",
                   JvmtiTrace::safe_get_thread_name(thread)));
   JvmtiEnvThreadStateIterator it(state);
   for (JvmtiEnvThreadState* ets = it.first(); ets != nullptr; ets = it.next(ets)) {
-    if (ets->is_enabled(JVMTI_EVENT_SAMPLED_OBJECT_ALLOC)) {
+    JvmtiEnv *env = ets->get_env();
```
lmesnik, line 3023: Not related to you fix just general comment for next clean up.
It seems that the 'ets' is needed only to get 'env'.

So all loop might be rewritten similar to same in the
`void JvmtiExport::post_vm_object_alloc(JavaThread *thread, oop object) {`
to be

`for (JvmtiEnv* env = it.first(); env != nullptr; env = it.next(env)) {`

lmesnik (approved):
I reviewed the implementation and tests only, assuming that CSR will be approved.

The only general comment is that implementation doesn't check the '--enable-preview' so agent with enabled 'can_support_value_objects' is not going to fail.

Do we need a separate test for this? 

Also, there is a general lack of coverage of this behaviour for multi environments/agents.

Might be needed to to file RFE to review './vmTestbase/nsk/jvmti/scenarios/multienv' and see if they should be updated or write new tests.

There are few comments inline.

sspitsyn:
> The only general comment is that implementation doesn't check the '--enable-preview' so agent with enabled 'can_support_value_objects' is not going to fail. Do we need a separate test for this?

Thank you for looking at this change! This does not look as a real problem.
The runtime event paths already make the relevant behavioral decision from the actually allocated object:
  const bool is_inline = object->is_inline();

> Also, there is a general lack of coverage of this behaviour for multi environments/agents.
> Might be needed to to file RFE to review './vmTestbase/nsk/jvmti/scenarios/multienv' and see if they should be updated or write new tests.

Yes. We have a lack of coverage in this area. I hope, it is not a stopper for this PR. :)

test/hotspot/jtreg/serviceability/jvmti/HeapMonitor/libHeapMonitorTest.cpp
sspitsyn: Thank you for the suggestion. I'm not sure about this, but maybe. This test is already big and does many checks.

src/hotspot/share/prims/jvmtiExport.cpp
sspitsyn: Thanks. Right. It has caught my eyes, so I decided to simplify this.

src/hotspot/share/prims/jvmtiExport.cpp
sspitsyn: Thank you for the check. There is the line 3024 which needs the `ets`:
`     if (ets->is_enabled(JVMTI_EVENT_SAMPLED_OBJECT_ALLOC) &&`

src/hotspot/share/prims/jvmti.xml
```diff
@@ -10710,6 +10710,15 @@ myInit() {
           <eventlink id="VirtualThreadEnd"></eventlink>.
         </description>
       </capabilityfield>
+      <capabilityfield id="can_support_value_objects" since="28">
+        <description>
+          Can support value objects.
+          When preview features are enabled and this capability is enabled, then
+          the following two events are enabled for value object allocations:
+          <eventlink id="VMObjectAlloc"></eventlink>,
```
AlanBateman, line 10718: "then the following two events are enabled for value object allocations".  This reads a bit like enabling can_support_value_objects will enable these two events. I think it would be better to say something like "events are sent for value objects when the following two events are enabled".

src/hotspot/share/prims/jvmti.xml
```diff
@@ -14013,9 +14022,11 @@ myInit() {
       VM. These methods should send this event.
       Virtual machines which are incapable of bytecode instrumentation
       for some or all of their methods can send this event.
-      When preview features are enabled and the allocated object has no identity,
-      <code>null</code> is passed for the <paramlink id="object"></paramlink> parameter.
-
+      <p/>
+      When preview features are enabled and the capability <code>can_support_value_objects</code>
+      is enabled, this event is sent for the allocated objects that have no identity.
+      <code>null</code> is passed for the <paramlink id="object"></paramlink> parameter in such cases.
```
AlanBateman, line 14028: As the capability is "can_support_value_objects" then it may be simpler to just say that it sent for value objects when that capability is enabled, otherwise you end up with both "value objects" and "no identity" in the same sentence.

src/hotspot/share/prims/jvmtiExport.cpp
lmesnik: I wonder if it could be done through 'env' like
`env->is_enabled(JVMTI_EVENT_VM_OBJECT_ALLOC)`
?

sspitsyn: This event can be controlled at the thread level, so we need to check it is enabled at both thread and env levels.

src/hotspot/share/prims/jvmti.xml
sspitsyn: Thank you for the catch and suggestion. Fixed now.

src/hotspot/share/prims/jvmti.xml
sspitsyn: Agreed, thanks. Fixed now.

src/hotspot/share/prims/jvmti.xml
```diff
@@ -14061,7 +14073,8 @@ myInit() {
         <jobject/>
           <description>
             JNI local reference to the object that was allocated.
-            Null when preview features are enabled and the allocated object has no identity.
+            Null when preview features are enabled, the capability <code>can_support_value_objects</code>
+            is enabled and the allocated object has no identity.
```
plummercj:
```suggestion
            is enabled, and the allocated object has no identity.
```

src/hotspot/share/prims/jvmti.xml
```diff
@@ -14135,7 +14150,8 @@ myInit() {
         <jobject/>
         <description>
           JNI local reference to the object that was allocated.
-          Null when preview features are enabled and the allocated object has no identity.
+          Null when preview features are enabled, the capability <code>can_support_value_objects</code>
+          is enabled and the allocated object has no identity.
```
plummercj:
```suggestion
          is enabled, and the allocated object has no identity.
```

plummercj (commented):
It's a little unclear to me how this all works when not in preview mode. Does the jvmti client still get to enable can_support_value_objects, but it has no affect.

AlanBateman:
> It's a little unclear to me how this all works when not in preview mode. Does the jvmti client still get to enable can_support_value_objects, but it has no affect.

I think this came up with can_support_virtual_threads capability when virtual threads were in preview. I think we concluded that it's benign as it needs both the capability and preview features be enabled.

src/hotspot/share/prims/jvmti.xml
sspitsyn: Thanks. Fixed now.

src/hotspot/share/prims/jvmti.xml
sspitsyn: Thanks. Fixed now.

plummercj (approved):

sspitsyn:
Chris and Leonid, thank you for review!

src/hotspot/share/prims/jvmti.xml
AlanBateman: One other thing here: "can_support_value_objects" is a preview API. When we added can_support_virtual_threads in JDK 19 we added this to its description:

```
  <b> can_support_virtual_threads is a preview API of the Java platform. </b>
  <i>Preview features may be removed in a future release, or upgraded to
  permanent features of the Java platform.</i>
```

PR #2611, opened 2026-07-01
8387629: [lworld] Don't use apiNote for some recent JDI spec updates

plummercj:
There are 3 places in StackFrame that use apiNote for a section that should actually be considered being part of the spec. apiNote should be removed. Also, clarify that that we are talking about the target VM running in preview mode, not the JDI code. 

---------

---------
### Progress

### Issue
 * [JDK-8387629](https://bugs.openjdk.org/browse/JDK-8387629): [lworld] Don't use apiNote for some recent JDI spec updates (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2611/head:pull/2611` \
`$ git checkout pull/2611`

`$ git checkout pull/2611` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2611/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2611`

`$ git pr show -t 2611`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2611.diff">https://git.openjdk.org/valhalla/pull/2611.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2611#issuecomment-4859457076)
</details>


sspitsyn (approved):
This looks good.

sspitsyn:
I guess, the JBS issue has to be targeted to the `repo-valhalla`.

src/jdk.jdi/share/classes/com/sun/jdi/StackFrame.java
```diff
@@ -97,10 +97,9 @@ public interface StackFrame extends Mirror, Locatable {
      * The {@link ObjectReference} for 'this' is only available for
      * non-native instance methods.
      *
-     * @apiNote
      * <div class="preview-block">
      *     <div class="preview-comment">
-     *         When preview features are enabled,
+     *         When preview features are enabled for the target VM,
```
AlanBateman: Dropping `@apiNote` is good as this is normative test (not an API note).

For the preview-block text then I'm wondering about the case where the debugger/JDI-tool is running on JDK 28 and the target VM is JDK 35. Value classes and objects should be be permanent feature by then so there is value classes and objects without needing to enable preview. Having the back end and front end be different arises a lot in IDEs.

I don't think it's possible to future proof completely but maybe it could say something like "If the target VM supports value classes and objects (link to JEP 401), ..." and add a note that the feature requires enabling preview features in JDK 28?  If this starts your screaming then ignore my comment, it might be okay to keep it simple for now.

plummercj: I see your point, but it does start to get clumsy, and as you point out, completely future proofing is even more of a challenge, if even possible. With your suggestion, the "preview" reference never goes away. Even in JDK 35 with value objects being a permanent feature, we'd need to have wording indicating that preview features may need to be enabled on the target VM. Do we really want that?

AlanBateman: For reference, here are examples from JDK 19 when virtual threads were in preview.

https://docs.oracle.com/en/java/javase/19/docs/api/jdk.jdi/com/sun/jdi/ThreadReference.html
https://docs.oracle.com/en/java/javase/19/docs/api/jdk.jdi/com/sun/jdi/StackFrame.html

(search for "virtual thread" on these pages)

It avoided prefixing the clarifications with "When preview are enabled ..." and instead spoke about the feature in the target VM. This worked because the feature, and the behavior specified, didn't change in second preview or when it became a permanent feature in JDK 21.

plummercj: I've udpated this to be handled the same way as virtual threads was handled in JDK 19. There are no preview references (virtual threads only did that for new APIs, and we have no new APIs here). The spec simply says "when the object is a value object..." similar to how for JDK 19 it says "when the thread is a virtual thread".

AlanBateman (approved):

AlanBateman (changes requested):
Looking at it again, I think we should restore the "When preview features are enabled" text for the updated JDWP commands to remind developers that this is behavior for a preview feature. It's possible this could change in the next preview or when the feature becomes permanent.

sspitsyn (approved):

plummercj:
Thank you for the reviews Alan and Serguei!

/integrate

PR #2612, opened 2026-07-01
8387350: [lworld] ObjectOutputStream.writeObject() doesn't throw InvalidClassException for Serializable instance with strict init fields

liach:
Make sure object-writing serialization fails for "strict in serializable super". We currently already fail when they are deserialized.

---------

---------
### Progress

### Issue
 * [JDK-8387350](https://bugs.openjdk.org/browse/JDK-8387350): [lworld] ObjectOutputStream.writeObject() doesn't throw InvalidClassException for Serializable instance with strict init fields (**Bug** - P2)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2612/head:pull/2612` \
`$ git checkout pull/2612`

`$ git checkout pull/2612` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2612/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2612`

`$ git pr show -t 2612`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2612.diff">https://git.openjdk.org/valhalla/pull/2612.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2612#issuecomment-4860868860)
</details>

### DIFF
--- a/src/java.base/share/classes/java/io/ObjectStreamClass.java
+++ b/src/java.base/share/classes/java/io/ObjectStreamClass.java
@@ -123,7 +123,9 @@ public final class ObjectStreamClass implements Serializable {
     /** true if represents record type */
     private boolean isRecord;
     /** true if represented class cannot use allocate-and-fill deserialization,
-     * due to value class or strict field initialization restrictions. */
+     * due to value class or strict field initialization restrictions.
+     * Such a class either has a deserializer or has both serialize/deserialize
+     * exceptions once initialized. */
     private boolean requiresDeserializer;
     /** true if represented class implements Serializable */
     private boolean serializable;
@@ -350,13 +352,20 @@ public final class ObjectStreamClass implements Serializable {
         isProxy = Proxy.isProxyClass(cl);
         isEnum = Enum.class.isAssignableFrom(cl);
         isRecord = cl.isRecord();
-        requiresDeserializer = cl.isValue() || ValueClass.hasStrictInstanceField(cl);
         serializable = Serializable.class.isAssignableFrom(cl);
         externalizable = Externalizable.class.isAssignableFrom(cl);
+
+        // Note: AVC with no strict instance field like java.lang.Number is allowed;
+        // non-serializable superclasses are allowed because their constructors are called
+        requiresDeserializer = serializable && (ValueClass.isConcreteValueClass(cl) || ValueClass.hasStrictInstanceField(cl));
 
         Class<?> superCl = cl.getSuperclass();
         superDesc = (superCl != null) ? lookup(superCl, false) : null;
         localDesc = this;
+
+        if (superDesc != null) {
+            requiresDeserializer |= superDesc.requiresDeserializer;
+        }
 
         if (serializable) {
             if (isEnum) {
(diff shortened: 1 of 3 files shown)


src/java.base/share/classes/java/io/ObjectStreamClass.java
```diff
@@ -350,14 +352,21 @@ private ObjectStreamClass(final Class<?> cl) {
         isProxy = Proxy.isProxyClass(cl);
         isEnum = Enum.class.isAssignableFrom(cl);
         isRecord = cl.isRecord();
-        requiresDeserializer = cl.isValue() || ValueClass.hasStrictInstanceField(cl);
         serializable = Serializable.class.isAssignableFrom(cl);
         externalizable = Externalizable.class.isAssignableFrom(cl);
 
+        // Note: AVC with no strict instance field like java.lang.Number is allowed;
+        // non-serializable superclasses are allowed because their constructors are called
+        requiresDeserializer = serializable && (ValueClass.isConcreteValueClass(cl) || ValueClass.hasStrictInstanceField(cl));
```
AlanBateman, line 365: This looks the right test.  You might want to expand "AVC" comment as anyone reading this might not know it means abstract value class.

src/java.base/share/classes/java/io/ObjectStreamClass.java
```diff
@@ -379,14 +388,15 @@ private ObjectStreamClass(final Class<?> cl) {
...
+                    // Concrete value classes and classes with strict instance
+                    // fields must not breach their integrity with the serializable
+                    // constructor. Make sure they fail also upon serialization
+                    // in addition to deserialization if they don't have a
+                    // correct internal @Deserializer
+                    deserializer = findDeserializer(cl, fields);
                     if (deserializer == null) {
                         serializeEx = deserializeEx = new ExceptionInfo(cl.getName(),
-                                                                        "cannot serialize value class");
+                                "cannot serialize due to concrete value class or strict instance fields");
```
AlanBateman: Having "concrete" in the exception message might be confusing, I think it would be okay to reduce it to "value class" here.

ExE-Boss: Maybe non‑`abstract`?

Since `Serializable` fieldless `abstract` value classes like `java.lang​.Number` are ok.

jaikiran:
I don't have familiarity with this code, but having looked at it the last few days, this change looks reasonable to me. I haven't had a chance to run my own experiments, but does this change address any issue that might have arisen if the strictly initialized instance field belonged to some parent (identity) class that was not an immediate parent of the object instance being serialized/deserialized?

For example, the following class hierarchy:

```
class IdentityGrandParent implements Serializable {
   // strictly initialized instance field
   int x;
}

class IdentityParent extends IdentityGrandParent {
   // no instance fields
}

class Identity extends IdentityParent {
      // no instance fields
}
```
and then serialize/deserialize and instance of `Identity` class:
```
ObjectOutputStream.writeObject(new Identity());
```
Perhaps the `ValueSerializationTest` can be enhanced to include this case?

test/jdk/java/io/Serializable/valueObjects/ValueSerializationTest.java
```diff
@@ -240,8 +269,6 @@ static Stream<Arguments> classes() {
     @MethodSource("classes")
     void testDeser(Class<?> valueClass, byte flags, Class<? extends Exception> expectedException)
```
jaikiran: Now that the incoming `Class` is not always a value class, I think we should rename this parameter to just `cls` or something else. At the same time, the comment on the test method would need an update too, maybe:

> A byte stream is generated containing a reference to the given class  ...

src/java.base/share/classes/java/io/ObjectStreamClass.java
liach: Since value classes are either abstract or final, we can use "final value class".

test/jdk/java/io/Serializable/valueObjects/ValueSerializationTest.java
liach: Done.

liach:
Pushed an update addresssing comments.

For @jaikiran's concerns, I introduced `StrictInTwiceSuper` which has the same hierarchy as `Identity` in his example, where the strict declaration is in the two-level superclass from the class being serialized.

AlanBateman (approved):

jaikiran (approved):
Thank you for the update. This looks good to me.

TobiHartmann (approved):
Rubber stamped.

liach:
Thanks for the reviews, Alan and Jaikiran, and thanks Tobias for rubber stamping before the freeze!

/integrate

AlanBateman:
@liach java.io.InvalidClassException has a list of the cases where the exception is thrown. It lists value classes and it could potentially list classes with strictly-initialized fields too. It does include "Other conditions given in the JOSS" so not strictly needed but could be looked at.

liach:
I added an entry for strictly-initialized fields in #2631.

PR #2616, opened 2026-07-02
8387661: [lworld] move generate_return_value_stub() to SharedRuntime

rwestrel:
This addresses Andrew's comment:

https://github.com/openjdk/jdk/pull/31122#discussion_r3499128872

One of the 2 stubs are no longer used. I removed references to it.

---------

---------
### Progress

### Issue
 * [JDK-8387661](https://bugs.openjdk.org/browse/JDK-8387661): [lworld] move generate_return_value_stub() to SharedRuntime (**Enhancement** - P4)

### Reviewers
 * [Andrew Dinn](https://openjdk.org/census#adinn) (@adinn - no project role)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2616/head:pull/2616` \
`$ git checkout pull/2616`

`$ git checkout pull/2616` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2616/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2616`

`$ git pr show -t 2616`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2616.diff">https://git.openjdk.org/valhalla/pull/2616.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2616#issuecomment-4865329282)
</details>


src/hotspot/cpu/x86/sharedRuntime_x86_64.cpp
```diff
@@ -3877,6 +3877,171 @@ BufferedInlineTypeBlob* SharedRuntime::generate_buffered_inline_type_adapter(con
   return BufferedInlineTypeBlob::create(&buffer, pack_fields_off, pack_fields_jobject_off, unpack_fields_off);
 }
 
+// Call here from the interpreter or compiled code to either load
```
adinn: Same as with aarch64: I think this comment needs to be updated to omit reference to the load operation which is no longer used.

src/hotspot/cpu/x86/sharedRuntime_x86_64.cpp
```diff
@@ -3877,6 +3877,171 @@ BufferedInlineTypeBlob* SharedRuntime::generate_buffered_inline_type_adapter(con
...
+// multiple returned values from the inline type instance being
+// returned to registers or to store returned values to a newly
+// allocated inline type instance.
+// Register is a class, but it would be assigned numerical value.
+// "0" is assigned for xmm0. Thus we need to ignore -Wnonnull.
+PRAGMA_DIAG_PUSH
+PRAGMA_NONNULL_IGNORED
+RuntimeStub* SharedRuntime::generate_return_value_stub(address destination, bool has_res) {
```
adinn: Same as with aarch64: This only gets called with `has_res = true` so the parameter should be dropped and the code below for that case hard-wired.

src/hotspot/cpu/x86/sharedRuntime_x86_64.cpp
```diff
@@ -3877,6 +3877,171 @@ BufferedInlineTypeBlob* SharedRuntime::generate_buffered_inline_type_adapter(con
...
+  __ movptr(rax, Address(rsp, rax_off * BytesPerInt));
+
+  __ addptr(rsp, frame_size_in_bytes-8);
+
+  // check for pending exceptions
+  Label pending;
+  __ cmpptr(Address(r15_thread, Thread::pending_exception_offset()), (int32_t)NULL_WORD);
+  __ jcc(Assembler::notEqual, pending);
+
+  if (has_res) {
```
adinn: Same as with aarch64: The `if` here is not needed. The `if` body can simply be inlined.

src/hotspot/cpu/aarch64/sharedRuntime_aarch64.cpp
```diff
@@ -3181,6 +3181,155 @@ RuntimeStub* SharedRuntime::generate_throw_exception(StubId id, address runtime_
   return stub;
 }
 
+// Call here from the interpreter or compiled code to either load
+// multiple returned values from the inline type instance being
+// returned to registers or to store returned values to a newly
+// allocated inline type instance.
+RuntimeStub* SharedRuntime::generate_return_value_stub(address destination, bool has_res) {
```
adinn: This only gets called with `has_res = true` so the parameter should be dropped and the code below for that case hard-wired.

src/hotspot/cpu/aarch64/sharedRuntime_aarch64.cpp
```diff
@@ -3181,6 +3181,155 @@ RuntimeStub* SharedRuntime::generate_throw_exception(StubId id, address runtime_
...
+  __ ldpd(j_farg5, j_farg4, Address(__ post(sp, 2 * wordSize)));
+  __ ldpd(j_farg3, j_farg2, Address(__ post(sp, 2 * wordSize)));
+  __ ldpd(j_farg1, j_farg0, Address(__ post(sp, 2 * wordSize)));
+
+  // check for pending exceptions
+  Label pending;
+  __ ldr(rscratch1, Address(rthread, in_bytes(Thread::pending_exception_offset())));
+  __ cbnz(rscratch1, pending);
+
+  if (has_res) {
```
adinn: The if here is not needed. The if body can simply be inlined.

src/hotspot/cpu/aarch64/sharedRuntime_aarch64.cpp
```diff
@@ -3181,6 +3181,155 @@ RuntimeStub* SharedRuntime::generate_throw_exception(StubId id, address runtime_
   return stub;
 }
 
+// Call here from the interpreter or compiled code to either load
```
adinn: This comment need to be updated to reflect the fact that the load stub no longer exists.

adinn (changes requested):
This is all good. There is still one small tweak needed.

The list of stubs generated into the cache depends on the setting of `InlineTypeReturnedAsFields`. So, this flag setting needs to be recorded in the AOT cache when it is dumped and checked for consistency when the cache is loaded. That can be done by adding an entry to macro `AOTCODECACHE_CONFIGS_GENERIC_DO` in `src/hotspot/shared/code/aotCodecache.hpp`
```
#define AOTCODECACHE_CONFIGS_GENERIC_DO(do_var, do_fun)                 \
  do_var(int,   AllocateInstancePrefetchLines)          /* stubs and nmethods */ \
  . . .
  do_var(boolean,   InlineTypeReturnedAsFields)         /* stubs */              \
  . . .
```

Also, just checking: flag `InlineTypeReturnedAsFields` should not currently be set on any arch other than x86 or aarch64 because only those two generate the blob. Is there startup argument checking logic to detect an invalid setting? If not then can that lead to a crash on other arches should someone set it on the command line?

adinn:
n.b. the GHA failures are because the generator method for a `Shared` stub is called from generic code. So, it needs to exist in all arches even if it does not actually generate a stub.

TobiHartmann:
Testing on our side all passed. I can re-run once @adinn's requested changes are in.

rwestrel:
> Also, just checking: flag InlineTypeReturnedAsFields should not currently be set on any arch other than x86 or aarch64 because only those two generate the blob. Is there startup argument checking logic to detect an invalid setting? If not then can that lead to a crash on other arches should someone set it on the command line?

I see code that does that in arguments.cpp
Can you confirm @TobiHartmann ?

rwestrel:
> This is all good. There is still one small tweak needed.

Thanks for the careful review. I think I've addressed all you comments.

TobiHartmann:
> > Also, just checking: flag InlineTypeReturnedAsFields should not currently be set on any arch other than x86 or aarch64 because only those two generate the blob. Is there startup argument checking logic to detect an invalid setting? If not then can that lead to a crash on other arches should someone set it on the command line?
> I see code that does that in arguments.cpp
> Can you confirm @TobiHartmann ?

Here's the relevant code for PPC added with [JDK-8386638](https://bugs.openjdk.org/browse/JDK-8386638):
https://github.com/openjdk/valhalla/blob/e16732cc12cff297859b3a58f48237201c6be771/src/hotspot/cpu/ppc/vm_version_ppc.cpp#L345-L352

The same exists on other platforms Valhalla has been ported to.

I'm re-running testing.

TobiHartmann:
Testing passed.

adinn:
The current changes look fine to me.

I don't know why we are seeing a JUnit test failure on windows-x64. I cannot see how it might relate to this change but I don't know what the test is doing so someone else needs to look at it.

adinn (approved):

rwestrel:
Thanks for your help with this @adinn 
The failure doesn't seem related indeed.
@TobiHartmann should I integrate?

TobiHartmann:
@rwestrel Yes, ship it! :slightly_smiling_face:

adinn:
@rwestrel @TobiHartmann
I am afraid I am neither a Valhalla committer nor a reviewer. Someone from the Valhalla team needs to sign this off.

TobiHartmann (approved):

PR #2625, opened 2026-07-03
8387673: [lworld] C2: clone() intrinsic does not throw CloneNotSupportedException for value classes not implementing Cloneable

chhagedorn:
The `clone()` intrinsic directly returns a value object because it has no identity - the clone is trivially a no-op:
https://github.com/openjdk/valhalla/blob/2af7c2350b22ed03850b6b6c146b622ca758f441/src/hotspot/share/opto/library_call.cpp#L6041-L6045

But this is incorrect if the value class does not implement `Cloneable`. In this case, we must throw `CloneNotSupportedException` which is currently missing.

The fix adds a check whether the value class implements `Cloneable`. If it does, we execute the existing code no-op path. Otherwise, we could either do a call to the runtime `clone()` implementation, as done for the existing slow path of this intrinsic, or we simply emit an uncommon trap. I've decided to use the latter approach since it's more straight forward and rather an edge-case situation. I also included a `too_many_traps()` check to avoid repeated recompilations.

Thanks,
Christian
 
---------

---------
### Progress

### Issue
 * [JDK-8387673](https://bugs.openjdk.org/browse/JDK-8387673): [lworld] C2: clone() intrinsic does not throw CloneNotSupportedException for value classes not implementing Cloneable (**Bug** - P2)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2625/head:pull/2625` \
`$ git checkout pull/2625`

`$ git checkout pull/2625` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2625/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2625`

`$ git pr show -t 2625`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2625.diff">https://git.openjdk.org/valhalla/pull/2625.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2625#issuecomment-4876420195)
</details>

### DIFF
--- a/src/hotspot/share/ci/ciInlineKlass.cpp
+++ b/src/hotspot/share/ci/ciInlineKlass.cpp
@@ -62,6 +62,10 @@ bool ciInlineKlass::is_empty() {
   return nof_declared_nonstatic_fields() == 0;
 }
 
+bool ciInlineKlass::is_cloneable() const {
+  GUARDED_VM_ENTRY(return get_InlineKlass()->is_cloneable();)
+}
+
 int ciInlineKlass::inline_arg_length() const {
   VM_ENTRY_MARK;
   return get_InlineKlass()->extended_sig()->length();
--- a/src/hotspot/share/ci/ciInlineKlass.hpp
+++ b/src/hotspot/share/ci/ciInlineKlass.hpp
@@ -66,6 +66,7 @@ public:
   bool can_be_returned_as_fields() const;
 
   bool is_empty();
+  bool is_cloneable() const;
   int inline_arg_length() const;
   int inline_arg_slots() const;
   bool contains_oops() const;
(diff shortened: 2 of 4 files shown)


TobiHartmann (approved):
That looks good to me! As we discussed offline, please double-check that things still work when the `InlineTypeNode` is only found after parse time.

chhagedorn:
Thanks Tobias for your review! I've added some more tests to exercise the cases when we don't know that we are dealing with a value object during parsing:
- `testCloneAbstract`: We never know that it is a value object.
- `testCloneAbstractRef/Val`: We only know after parsing that we have a ref or value object.

TobiHartmann (approved):
Perfect, thanks for adding more tests!

chhagedorn:
Thanks @TobiHartmann!

/integrate

PR #2627, opened 2026-07-03
8387726: [lworld] JVMTI/JDWP/JDI docs cleanup

AlanBateman:
JVMTI GetLocalObject/GetLocalInstance: Right now, the behavior when preview features is enabled is buried in the description of the value_ptr parameter. Add text to the function descriptions so that it is more discoverable.

JDWP: I realize there has been some flip-flop on this but I think we need the "When preview features are enabled" text and remove any possibility of developers not understanding that the feature is in preview.

JDI: The first use of "value object" or "value class" in the updated methods now link to Class::isValue. This will ensure that the "PREVIEW" superscript appears in the usage. It might be that we come up with a better way to communicate that the behavior is for value objects which are in preview for this release.

JDI StackFrame.getValue/getValues: JDI has always filtered out "this" from the local variables so I think the only way for JDI tools to get a local variable for an object under construction is if the constructor invokes a method with "this" as a parameter (this-escape). Instance fields in value objects are strictly-initialized final fields so any leaking of "this" will be after the object has been initialized. I think we can drop the text added to these two methods. If this get a "this" for a value object under construction then it will already be initialized, and setValue will fail as the fields are final. We need to keep the text in StackFrame.thisObject, which is tweaked slightly to make it a bit clearer. Future work might be needed here to have ObjectReference.getValue fail if the field has not been initialized.

---------

---------
### Progress

### Issue
 * [JDK-8387726](https://bugs.openjdk.org/browse/JDK-8387726): [lworld] JVMTI/JDWP/JDI docs cleanup (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2627/head:pull/2627` \
`$ git checkout pull/2627`

`$ git checkout pull/2627` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2627/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2627`

`$ git pr show -t 2627`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2627.diff">https://git.openjdk.org/valhalla/pull/2627.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2627#issuecomment-4880915500)
</details>


src/jdk.jdi/share/classes/com/sun/jdi/ThreadReference.java
```diff
@@ -454,9 +454,10 @@ List<MonitorInfo> ownedMonitorsAndFrames()
      * Events, such as MethodExit, are generated as they would be in
      * a normal return.
      * <p>
-     * The called method must be a non-native Java programming
-     * language method. Forcing return on a thread with only one
-     * frame on the stack causes the thread to exit when resumed.
+     * The called method must not be a native method or the constructor of
+     * a {@linkplain Class#isValue() value class}. Forcing early return
+     * on a thread with only one frame on the stack causes the thread to
+     * exit when resumed.
```
plummercj, line 460: Do you want this update in the JDWP spec also?

sspitsyn:
I briefly looked at the update. The JVMTI changes look okay to me. Will do one more pass later.

src/jdk.jdi/share/classes/com/sun/jdi/ThreadReference.java
AlanBateman: > Do you want this update in the JDWP spec also?

The JDWP ThreadReference.ForceEarlyReturn command has "When preview features are enabled, the method can not be the constructor of a value class". It also has a sentence for native methods and a sentence about the one frame case (all pre-existing). But maybe you mean something else?

sspitsyn (approved):
I've done one more pass through the changes. The update is good.

sspitsyn:
> JDI StackFrame.getValue/getValues: JDI has always filtered out "this" from the local variables...

It is a surprise for me, thanks. Then it is a great simplification for the JDI StackFrame.getValue/getValues spec.

I see it in this JDI code:
```
    private List<LocalVariable> getVariables1() throws AbsentInformationException {
    . . .
       for (int i=0; i<count; i++) {
            JDWP.Method.VariableTableWithGeneric.SlotInfo si = vartab.slots[i];

            /*
             * Skip "this*" entries because they are never real
             * variables from the JLS perspective.
             */
            if (!si.name.startsWith("this$") && !si.name.equals("this")) {           <== !!!
                Location scopeStart = new LocationImpl(virtualMachine(),
                                                       this, si.codeIndex);
            . . .
```

src/jdk.jdi/share/classes/com/sun/jdi/ThreadReference.java
plummercj: Ok. I forgot you already made that change in #2604: https://github.com/openjdk/valhalla/pull/2604/changes#diff-570959dfa27b575bc2b0913024fd3580716e9d151b2513bce7413d2fb933dc86

plummercj (approved):

PR #2628, opened 2026-07-03
8386861: [lworld] emit_opSubstitutabilityCheck speedup

PeyaPeyaPeyang:
When `left == right` is known at code generation time, the operands are necessarily the same oop, so we can jump directly to `L_oops_equal`. 
This patch changes both x86 and aarch65 to do that.

I also considered emitting `equal_result` and returning early in that case.
I decided not to do that because `LIR_OpSubstitutabilityCheck::emit_code()` still appends the slow-path stub after `emit_opSubstitutabilityCheck()` returns.
An early return here would skip binding `op->stub()->continuation()`, leaving the appended stub with an unbound continuation label. 

---------

---------
### Progress

### Issue
 * [JDK-8386861](https://bugs.openjdk.org/browse/JDK-8386861): [lworld] emit_opSubstitutabilityCheck speedup (**Enhancement** - P4)

### Reviewers
 * [Martin Doerr](https://openjdk.org/census#mdoerr) (@TheRealMDoerr - no project role)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2628/head:pull/2628` \
`$ git checkout pull/2628`

`$ git checkout pull/2628` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2628/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2628`

`$ git pr show -t 2628`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2628.diff">https://git.openjdk.org/valhalla/pull/2628.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2628#issuecomment-4879216207)
</details>

### DIFF
--- a/src/hotspot/cpu/x86/c1_LIRAssembler_x86.cpp
+++ b/src/hotspot/cpu/x86/c1_LIRAssembler_x86.cpp
@@ -1594,10 +1594,6 @@ void LIR_Assembler::emit_opSubstitutabilityCheck(LIR_OpSubstitutabilityCheck* op
 
   Register left  = op->left()->as_register();
   Register right = op->right()->as_register();
-  if (left == right) {
-    move(op->equal_result(), op->result_opr());
-    return;
-  }
 
   __ cmpptr(left, right);
   __ jcc(Assembler::equal, L_oops_equal);
@@ -1632,10 +1628,8 @@ void LIR_Assembler::emit_opSubstitutabilityCheck(LIR_OpSubstitutabilityCheck* op
   } else {
     Register tmp1 = op->tmp1()->as_register();
     Register tmp2 = op->tmp2()->as_register();
-    if (left != right) {
-      __ cmp_klasses_from_objects(left, right, tmp1, tmp2);
-      __ jcc(Assembler::equal, *op->stub()->entry()); // same klass -> do slow check
-    }
+    __ cmp_klasses_from_objects(left, right, tmp1, tmp2);
+    __ jcc(Assembler::equal, *op->stub()->entry()); // same klass -> do slow check
     // fall through to L_oops_not_equal
   }
 
--- a/src/hotspot/cpu/x86/c1_LIRAssembler_x86.cpp
+++ b/src/hotspot/cpu/x86/c1_LIRAssembler_x86.cpp
@@ -1594,13 +1594,13 @@ void LIR_Assembler::emit_opSubstitutabilityCheck(LIR_OpSubstitutabilityCheck* op
 
   Register left  = op->left()->as_register();
   Register right = op->right()->as_register();
-
   if (left == right) {
-    __ jmp(L_oops_equal);
-  } else {
-    __ cmpptr(left, right);
-    __ jcc(Assembler::equal, L_oops_equal);
+    move(op->equal_result(), op->result_opr());
+    return;
   }
+
+  __ cmpptr(left, right);
+  __ jcc(Assembler::equal, L_oops_equal);
 
   // (1) Null check -- if one of the operands is null, the other must not be null (because
   //     the two references are not equal), so they are not substitutable,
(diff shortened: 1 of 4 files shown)


src/hotspot/cpu/x86/c1_LIRAssembler_x86.cpp
```diff
@@ -1595,8 +1595,12 @@ void LIR_Assembler::emit_opSubstitutabilityCheck(LIR_OpSubstitutabilityCheck* op
   Register left  = op->left()->as_register();
   Register right = op->right()->as_register();
 
-  __ cmpptr(left, right);
-  __ jcc(Assembler::equal, L_oops_equal);
+  if (left == right) {
```
chhagedorn: Thanks for picking this up. I have a general question about this patch: Couldn't we already check whether `left.result()` and `right.result()` refer to the same LIR operand earlier in `LIRGenerator::substitutability_check_common()` before creating `LIR_OpSubstitutabilityCheck`?
https://github.com/openjdk/valhalla/blob/a918cc9f5f8dcc35f4f73ed19b439a033fb39883/src/hotspot/share/c1/c1_LIRGenerator.cpp#L3379-L3381

TheRealMDoerr:
Why not
```
  if (left == right) {
    move(op->equal_result(), op->result_opr());
    return;
  }
```
?

PeyaPeyaPeyang:
Thanks @chhagedorn and @TheRealMDoerr for pointing that out. You are absolutely right. 
I initially avoided handling this there 
because I thought the change might become larger, but after looking into again,
it turns out to be quite small.
I’ll update the patch. Thanks again.

src/hotspot/share/c1/c1_LIRGenerator.cpp
```diff
@@ -3363,6 +3363,11 @@ void LIRGenerator::substitutability_check(If* x, LIRItem& left, LIRItem& right)
 void LIRGenerator::substitutability_check_common(Value left_val, Value right_val, LIRItem& left, LIRItem& right,
                                                  LIR_Opr equal_result, LIR_Opr not_equal_result, LIR_Opr result,
                                                  CodeEmitInfo* info) {
+  if (left.result() == right.result()) {
+    __ move(equal_result, result);
+    return;
+  }
```
TheRealMDoerr, line 3369: Good idea. Are the platform specific parts still needed after this?

PeyaPeyaPeyang: You are right, they are not needed any more after this, I think.
Thanks for pointing out..!

This is not a large change, so I'll clean up the
platform-specific codes with this PR as well.

TheRealMDoerr (approved):
LGTM. Thanks for improving it!

PeyaPeyaPeyang:
Hello @chhagedorn

I fear I was rather too hasty and nearly integrated this.
I've applied your suggestion and moved this to `LIRGenerator::substitutability_check_common()`. 
Could you kindly take another look?

Many thanks!

chhagedorn (approved):
Thanks for the update! That looks good to me, too. Let me give this a spin in our testing.

Just to let you know: For the mainline pre-integration testing, we are temporarily holding off on further integrations. We still need to decide whether follow-up changes should go into a separate Valhalla branch or be integrated into mainline afterward. We will get back to you later this week with further guidance.

PeyaPeyaPeyang:
@chhagedorn

Thanks for the review and the heads-up!
I'll hold off integrating until I hear back from you.

chhagedorn:
For non-blocking issues like this, we should hold off from integration and then re-base the PR to mainline after Valhalla integration.

Btw, testing looked good but we should rerun testing once we are in mainline.

PeyaPeyaPeyang:
Hello again @chhagedorn,

I have now rebased the change onto the JDK mainline and opened a PR as [openjdk/jdk#32149](https://github.com/openjdk/jdk/pull/32149).
Could you please re-review and approve the mainline PR?

PR #2629, opened 2026-07-06
8387405: [lworld] is_always_flat_in_array should be always false

marc-chevalier:
Nothing is flat! Well... sometimes, things are. But not arrays! Ok, arrays can be flat... but you can never be sure an array is flat from its unrefined type (and current JVM flags). Indeed, the "new" `ValueClass::newReferenceArray`: this allows to bypass `ObjArrayKlass::array_layout_selection(Klass*, ArrayProperties)` and get an array of references from any value class, regardless of the available layouts, thus fooling `InlineKlass::is_always_flat_in_array()`. In general, we can ensure the flatness of an array in only few situations: if we see the allocation, or after having tested the flatness with a runtime check. Thus `InlineKlass::is_always_flat_in_array()` doesn't really make sense... And consequently, it is not correct for `TypePtr::compute_flat_in_array(ciInstanceKlass* instance_klass, bool is_exact)` to ever return `Flat`. Yet, we still need the constructor `Flat` in `FlatInArray`. For instance, it is used in `TypeInstPtr::cast_to_flat_in_array()` which is indirectly used (through `load_from_unknown_flat_array`) in the "flat array" branch of `Parse::array_load(BasicType bt)`:

https://github.com/openjdk/valhalla/blob/a918cc9f5f8dcc35f4f73ed19b439a033fb39883/src/hotspot/share/opto/parse2.cpp#L121-L143

But here, casting to flat is fine since we are under a runtime check that enforces that the array is indeed flat.

Fortunately, I don't think the current usage could have lead to an error: we only use it to compare types: if one is sure to be flat but not the other one, the types can't be the same. It's fine, even if the "sure to be flat" criterion is buggy. Would work with a bit anything. Like "if a type starts with the letter 'A', but the other by the letter 'B', they can't be the same". If we had any tracking of that, it would allow to (very weirdly) prove type inequality even with imprecise types...

Thanks,
Marc

---------

---------
### Progress

### Issue
 * [JDK-8387405](https://bugs.openjdk.org/browse/JDK-8387405): [lworld] is_always_flat_in_array should be always false (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2629/head:pull/2629` \
`$ git checkout pull/2629`

`$ git checkout pull/2629` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2629/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2629`

`$ git pr show -t 2629`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2629.diff">https://git.openjdk.org/valhalla/pull/2629.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2629#issuecomment-4890597076)
</details>

### DIFF
--- a/src/hotspot/share/ci/ciInlineKlass.cpp
+++ b/src/hotspot/share/ci/ciInlineKlass.cpp
@@ -40,11 +40,6 @@ bool ciInlineKlass::maybe_flat_in_array() const {
   GUARDED_VM_ENTRY(return to_InlineKlass()->maybe_flat_in_array();)
 }
 
-// Are arrays containing an instance of this value class always flat?
-bool ciInlineKlass::is_always_flat_in_array() const {
-  GUARDED_VM_ENTRY(return to_InlineKlass()->is_always_flat_in_array();)
-}
-
 // Can this inline type be passed as multiple values?
 bool ciInlineKlass::can_be_passed_as_fields() const {
   GUARDED_VM_ENTRY(return to_InlineKlass()->can_be_passed_as_fields();)
--- a/src/hotspot/share/ci/ciInlineKlass.hpp
+++ b/src/hotspot/share/ci/ciInlineKlass.hpp
@@ -57,7 +57,6 @@ public:
   int payload_offset() const;
 
   bool maybe_flat_in_array() const override;
-  bool is_always_flat_in_array() const;
 
   // Scalarized calling convention support: pass/return this inline type as its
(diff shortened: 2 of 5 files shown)


TobiHartmann (approved):
That looks good to me!

marc-chevalier:
/integrate

Thanks @TobiHartmann!

PR #2631, opened 2026-07-06
8387749: [lworld] ValueBased.html should be reevaluated for its specification of value class constructors

liach:
`ValueBased` talks about constructor avoidance but does not specify the reason. It's better to include the reason because it's not straightforward from the fact that there's no binary incompatibility.

Also `InvalidClassException` benefits from descriptions about strictly-initialized fields.

---------

---------
### Progress

### Issue
 * [JDK-8387749](https://bugs.openjdk.org/browse/JDK-8387749): [lworld] ValueBased.html should be reevaluated for its specification of value class constructors (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2631/head:pull/2631` \
`$ git checkout pull/2631`

`$ git checkout pull/2631` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2631/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2631`

`$ git pr show -t 2631`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2631.diff">https://git.openjdk.org/valhalla/pull/2631.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2631#issuecomment-4898626235)
</details>


liach:
Hi @jaikiran, can you help review this?

src/java.base/share/classes/java/lang/doc-files/ValueBased.html
```diff
@@ -36,8 +36,8 @@ <h1 id="ValueBased">{@index "Value-based Classes"}</h1>
 as {@code value class} and preview features are enabled.
 A value-based class has the following properties:
 <ul>
-    <li>the class declares only final instance fields (though these may contain references
-        to mutable objects);</li>
+    <li>all instance fields declared in the class and all its superclasses are
```
viktorklang-ora:
```suggestion
    <li>all instance fields declared in the class and all of its superclasses are
```

src/java.base/share/classes/java/lang/doc-files/ValueBased.html
```diff
@@ -47,14 +47,15 @@ <h1 id="ValueBased">{@index "Value-based Classes"}</h1>
...
+    <li>the class does not declare, or discourages use of, accessible constructors
+        (because each constructor call promises a unique identity);</li>
     <li>the class does not provide any instance creation mechanism that promises
         a unique identity on each method call&mdash;in particular, any factory
         method's contract must allow for the possibility that if two independently-produced
         instances are equal according to {@code equals()}, they may also be
         equal according to {@code ==};</li>
     <li>the class is final, and extends either {@code Object} or a hierarchy of
-        abstract value classes.</li>
+        abstract classes.</li>
```
viktorklang-ora, line 60: I was under the impression that supertypes of value calsses must be either abstract value classes or java.lang.Object?

liach: Yes. But notice a "value-based" class can exist when preview features are disabled, in which case, all the superclasses of the VBC can only be abstract without being value (a preview-only concept)

viktorklang-ora: This list starts with the statement:

```
The compiler and runtime enforce the value based properties below if it is declared
as {@code value class} and preview features are enabled.
A value-based class has the following properties:
```

So I think we can presume that the list is "under the circumstance of preview being enabled". In which case we should be able to say "abstract value class"?

liach: All these text were initially added by @RogerRiggs, but I personally think this is an exaggeration of the enforcement because the enforcements are not complete, I think.

I would still describe VBC in the context of non-preview. Maybe @dansmithcode could comment on this.

viktorklang-ora: Yeah, @dansmithcode should be able to guide the verbiage here. 👍

src/java.base/share/classes/java/lang/doc-files/ValueBased.html
```diff
@@ -36,8 +36,8 @@ <h1 id="ValueBased">{@index "Value-based Classes"}</h1>
 as {@code value class} and preview features are enabled.
```
dansmithcode: I made this comment somewhere else, but it may have been dropped:

Identity classes can be "value-based". It doesn't really make sense to talk about a "value-based value class"—that's just a value class. Everything in this file is directed towards identity classes.

Suggest deleting this sentence.

src/java.base/share/classes/java/lang/doc-files/ValueBased.html
dansmithcode: See above—yeah, I think trying to make these rules correct for value classes is a misapplication of this concept.

src/java.base/share/classes/java/lang/doc-files/ValueBased.html
liach: I agree. Did some local revisions; hope you can review the updated version.

src/java.base/share/classes/java/lang/doc-files/ValueBased.html
```diff
@@ -32,12 +32,16 @@ <h1 id="ValueBased">{@index "Value-based Classes"}</h1>
 
 Some classes, such as {@code java.lang.Integer} and
 {@code java.time.LocalDate}, are <em>value-based</em>.
-The compiler and runtime enforce the value based properties below if it is declared
-as {@code value class} and preview features are enabled.
-A value-based class has the following properties:
+In a future release, they may become value classes (JLS
+{@jls value-objects-8.1.1.5}).
+The compiler and runtime enforce some properties for value-based classes
```
dansmithcode: If it's appropriate to link to JEPs from javadoc:

```
The compiler and runtime provide <a href="https://openjdk.org/jeps/390">warnings</a>
related to the misuse of value-based classes.
```

(Not really "enforcement", and I don't think we need to say anything more about migration.)

src/java.base/share/classes/java/lang/doc-files/ValueBased.html
```diff
@@ -66,11 +71,14 @@ <h1 id="ValueBased">{@index "Value-based Classes"}</h1>
     because the programmer cannot guarantee exclusive ownership of the
     associated monitor.</p>
 
-<p>Identity-related behavior of value-based classes may change when implemented as a value class.
-</p>
+<p>When preview features are enabled, some value-based classes become value classes.
+    Instances of these classes become value objects, which have different identity-related
+    behaviors compared to identity objects, as described in {@link Object}:</p>
```
dansmithcode: I don't want to treat Object as the Source of Truth here, maybe just delete "as desribed in Object:" and say "including:".

src/java.base/share/classes/java/lang/doc-files/ValueBased.html
```diff
@@ -66,11 +71,14 @@ <h1 id="ValueBased">{@index "Value-based Classes"}</h1>
...
 <ul>
-    <li>The class may choose to allocate/cache instances differently.
-    <li>The use of the value class for synchronization or with
+    <li>Previously distinguishable instances, such as those created from distinct
+        constructor invocations (See JLS {@jls value-objects-15.9.4}), may be no
+        longer distinguishable via {@code ==}.
+    <li>The use of the value objects for synchronization or with
     {@linkplain java.lang.ref.Reference object references} result in {@link IdentityException}.
```
dansmithcode: s/result/results/

src/java.base/share/classes/java/lang/doc-files/ValueBased.html
dansmithcode: If this spec text is supposed to be independent of implementation details like javac and HotSpot warnings, then I just wouldn't mention those features at all. (I think maybe that's the case? This is part of the Java SE specification, after all...)

src/java.base/share/classes/java/lang/doc-files/ValueBased.html
jddarcy: Shouldn't the specification of `java.lang.Object` be the source of truth for the semantics of `Object`?

dansmithcode: Sure, but not necessarily the authoritative catalog of all "different identity-related behaviors compared to identity objects". Some of that is language semantics, some of that is the behavior of particular APIs that interact with value objects, etc.

src/java.base/share/classes/java/lang/doc-files/ValueBased.html
liach: Sure, dropping this section. The JEP 390 warnings are really implementation-specific warnings instead of parts of Java SE, where this documentation is.

src/java.base/share/classes/java/lang/doc-files/ValueBased.html
liach: Wanted to link to `Object` for its preview block, guess that's not necessary. Dropping that section.

dlsmith-dev (approved):

liach:
Thanks for the reviews! FYI @jaikiran even though you wishes to review, we need this for CSR's APIDiff so I will go ahead and integrate.

/integrate

PR #2633, opened 2026-07-08
8387875: [lworld] Provide value-class-preview.md file

liach:
This is some extremely basic overview about value class migrations.

1. Build: I intentionally skipped the streamlined build system installed by David Beaumont. Instead, I stressed the custom handling used by Value Objects JEP. (The "Preview Classes in Interim `javac`" section might go to general documentation later.)

2. Testing: I started from the motivations/goals and describe how those goals are accomplished.
3. Wrapper class caches: Some simple overview.

There are definitely more topics to cover; feel free to raise.

Paging @jddarcy and @viktorklang-ora for proofreading.

---------

---------
### Progress

### Issue
 * [JDK-8387875](https://bugs.openjdk.org/browse/JDK-8387875): [lworld] Provide value-class-preview.md file (**Enhancement** - P2)

### Reviewers
 * [Viktor Klang](https://openjdk.org/census#vklang) (@viktorklang-ora - no project role) ⚠️ Review applies to [e9bbd532](https://git.openjdk.org/valhalla/pull/2633/files/e9bbd5321c36dd9eceaf61c1e6345cf14815b918)
 * [Dan Smith](https://openjdk.org/census#dlsmith) (@dansmithcode - Committer) ⚠️ Review applies to [e9bbd532](https://git.openjdk.org/valhalla/pull/2633/files/e9bbd5321c36dd9eceaf61c1e6345cf14815b918)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2633/head:pull/2633` \
`$ git checkout pull/2633`

`$ git checkout pull/2633` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2633/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2633`

`$ git pr show -t 2633`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2633.diff">https://git.openjdk.org/valhalla/pull/2633.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2633#issuecomment-4937338594)
</details>


doc/value-class-preview.md
```diff
@@ -0,0 +1,82 @@
+% Migration of JDK Classes to Value Classes
+
+## Introduction
+
+The "Value Objects" JEP introduces value objects and migrates suitable classes
+to value classes. This means that when preview features are enabled, a different
+classpath is available to Java programs. To accomplish this, the JDK has a
+custom build and deployment process to allow access to different classpaths.
```
jddarcy: I think this intro should explicit words to the effect of "there are separate class files for migrated value classes with preview enabled and with preview enabled. First new source files are generated from the original files to be migrated and then these new source files are compiled into separate class files are put into a distinct class path, etc."

(The salient point here for core libs maintainers include there is a new source-based transform and, as explained later in the note, understanding the limited functionality of the transform. Other technically possible ways to do this transform would include class file re-writing, possibly at startup, etc.)

doc/value-class-preview.md
```diff
@@ -0,0 +1,82 @@
...
+
+For the Value Objects JEP, for each module, the preview source files are in
+`$(SUPPORT_OUTPUTDIR)/gensrc-valueclasses/<module>/` directory. However, only
+the module `java.base` actually makes use of this directory, created by
+[`GensrcValueClasses.gmk`](../make/modules/java.base/gensrc/GensrcValueClasses.gmk).
+This processor simply looks for occurrences of `/*value*/ class` and
+`/*value*/ record` in the hardcoded list of source files, and generate preview
+source files where these occurrences are replaced by `value class` and
+`value record`, respectively.
+
```
jddarcy, line 27: So, with the current system, consequently, any needed testing of "is preview enabled" must be done at runtime rather than via a compile-time macro?
(The other little code gen systems used in the JDK have a more expressive set of primitives. In my estimation, the simple system here is good for the purpose it is being used for, but it is helpful to document it here.)

jddarcy:
And once the text of the note if finalized, links to it in `//` comments should be added to the wrapper classes and the `java.time` classes.

doc/value-class-preview.md
viktorklang-ora: I think Joe makes a good point here

doc/value-class-preview.md
```diff
@@ -0,0 +1,82 @@
...
+The Value Objects JEP does not intend to introduce value classes that are not
```
viktorklang-ora: Too many negations in this sentence

doc/value-class-preview.md
```diff
@@ -0,0 +1,82 @@
...
+migrated from a class that is available when preview features are disabled, nor
+does it intend to introduce migrations from modules other than `java.base`. The
+preview source file setup would require substantial changes to accomplish these
+goals.
+
+### Preview Classes in Interim `javac`
+
+The preview classpath files are not available to the interim `javac`, most
+commonly used in `SetupJavaCompilation` call. For each module that ships
+preview-specific binaries, the following javac flag is required for a
```
viktorklang-ora: What's the definition of "preview-specific binaries" here? If we mean "value object-capable binaries" I think we should be clear about that.

viktorklang-ora:
Thanks for taking this on, @liach!

doc/value-class-preview.md
liach: Added a somewhat better overview for this.

doc/value-class-preview.md
liach: Yes. I don't understand how a compile-time macro can work - we have one runtime that must support both preview-on and preview-off runs.

doc/value-class-preview.md
liach: Done, converted this to a list of prerequisites.

doc/value-class-preview.md
liach: Now defined at top about how "preview-specific" stuff overrides regular stuff.

liach:
Updated. Please help proofread the latest version, thanks!

doc/value-class-preview.md
```diff
@@ -4,45 +4,78 @@
...
-goals.
+1. A hardcoded list of files are selected for generation.
+
+2. For each file, search for any occurrences of `/*value*/ class` or
+   `/*value*/ record`, and replace with `value class` or `value record`.
+
+3. The replaced, preview-specific source files are located in
+   `support/gensrc-valueclasses/java.base/`.
+
+4. The general preview source to binary pipeline recognizes this
```
viktorklang-ora: *build pipeline*

doc/value-class-preview.md
```diff
@@ -4,45 +4,78 @@
...
+   ```
+   --patch-module <module>=$(SUPPORT_OUTPUTDIR)/preview/<module>
+   ```
+
+   See [`BuildMicroBenchmarks.gmk`](../make/test/BuildMicrobenchmark.gmk) for an example.
+
+### Non-Goals
 
-### Preview Classes in Interim `javac`
+The Value Objects JEP only plans to introduce a value class that:
```
viktorklang-ora: >The Value Objects JEP only plans to introduce a value class that:

This doesn't read right to me, perhaps we mean something more akin to "The Value Objects JEP only migrates a select few platform types that:"

viktorklang-ora:
After commentsI had some minor comments, and after addressing those I think this is ready from my point-of-view. I'll let @jddarcy decide if anything more is necessary.

doc/value-class-preview.md
jddarcy: In the space of generate-different-source-code-from-a-template solutions, it would be possible to have code conditionally emitted for one case but not the other. The solution used for JEP 401 currently just uncomments the `value` modifier.

liach: We tried to use annotation processors and discovered that it just introduced unnecessary hassles and complexities beyond what is required for JEP 401. I have described the JEP 401 process here; we can tweak this document if we do tweak this pipeline in the future.

jddarcy: I understand some of the history for the transitory use of annotation processors, but the readers of this document might have not that history (and might not care about it either ;-) ). 

Given that the general shape of the solution is "we have generating preview-specific versions of source files which get compiled into preview-specific class files," the current mechanism we are choosing to do the generation has very limited functionality, which is sufficient for the purposes at hand. However, that mechanism is less capable than several analagous mechanisms elsewhere in the JDK build.

doc/value-class-preview.md
```diff
@@ -0,0 +1,120 @@
+% Migration of JDK Classes to Value Classes
+
+## Introduction
+
+The "Value Objects" JEP introduces value objects and migrates suitable classes
```
jddarcy: Rewording suggestion to avoid the need to say "JEP"; something like: "Value objects and value classes are a preview feature that migrates selected platform classes to value classes. This mean when ..."

doc/value-class-preview.md
```diff
@@ -0,0 +1,120 @@
...
+to value classes. This means that when preview features are enabled, a different
+classpath is available to Java programs.
```
jddarcy: Nit: is it a _class_path or a _module_path or does that detail not matter here?

doc/value-class-preview.md
```diff
@@ -0,0 +1,120 @@
...
+6. The interim `javac` used by the build system cannot pickup the
+   preview-specific overrides; they must be supplied explicitly with this
+   javac flag for every single module where overrides are significant:
+
+   ```
+   --patch-module <module>=$(SUPPORT_OUTPUTDIR)/preview/<module>
+   ```
+
+   See [`BuildMicroBenchmarks.gmk`](../make/test/BuildMicrobenchmark.gmk) for an example.
+
```
jddarcy, line 73: I recommend explicitly stating up front "Different class files are used at runtime when preview is enabled."

I believe that is a new-with-JEP-401 implementation technique for JEPs.

doc/value-class-preview.md
```diff
@@ -0,0 +1,120 @@
+% Migration of JDK Classes to Value Classes
+
+## Introduction
+
+Value Objects feature introduces value objects and migrates suitable classes to
```
dholmes-ora: Either "The Value Objects feature introduces ..." or "Value Objects is a feature that introduces ..."

doc/value-class-preview.md
```diff
@@ -0,0 +1,120 @@
...
+value classes. This means that when preview features are enabled, different
+class files are used for the migrated classes in the Java class library.
+
+To accomplish this, the JDK uses *preview-specific* files in `META-INF/preview`,
+which overrides files of the same name in the regular classpath. For example,
```
dholmes-ora: "classpath"? Isn't this the modulepath these days?

liach: I use the term "classpath" for where the resources like class files are discovered. I don't know if we ever use the term "modulepath".

jddarcy: Both the `java` and `javac` commands have "module path" options, along with "class path" options.

liach: Is there a better term to refer to directories or locations where such resources? Maybe something like "overrides files of the same name in the regular locations"?

viktorklang-ora: "To accomplish this, the JDK uses *preview-specific* files in `META-INF/preview` that will be used instead of the original file when running in preview mode"?

liach: Done. Please review again.

doc/value-class-preview.md
```diff
@@ -0,0 +1,120 @@
...
+`META-INF/preview`.
+
+## The Build Process
+
+### The Custom Handling for the Value Objects JEP
+
+The Value Objects JEP requires a few select classes in the `java.base` module
+to become value classes when preview features are enabled.
+
+The build of `java.base` module first create the source code of those value
```
viktorklang-ora:
```suggestion
The build of `java.base` module first creates the source code of those value
```

doc/value-class-preview.md
```diff
@@ -0,0 +1,120 @@
...
+1. A hardcoded list of files are selected for generation.
+
+2. For each file, search for any occurrences of `/*value*/ class` or
+   `/*value*/ record`, and replace with `value class` or `value record`.
+
+3. The replaced, preview-specific source files are located in
+   `support/gensrc-valueclasses/java.base/`.
+
+4. The general preview source to binary build pipeline recognizes this
+   directory as where `java.base` places its preview-specific source files.
```
viktorklang-ora: @liach This text is sort of ambiguous about whether the original files still remain in place, I think it should be spelled out that the original remains in place, and that the value-capable version of the source is a transformed copy.

viktorklang-ora (changes requested):
This is LGTM after considering my latest comments.

doc/value-class-preview.md
```diff
@@ -0,0 +1,124 @@
+% Migration of JDK Classes to Value Classes
+
+## Introduction
+
+The Value Objects feature introduces value objects and migrates suitable classes
+to value classes. This means that when preview features are enabled, different
+class files are used for the migrated classes in the Java class library.
+
+To accomplish this, the JDK uses *preview-specific* files in `META-INF/preview`,
```
jddarcy: To clarify,
"To accomplish this, a built JDK uses..."

That is, this section is discussing the artifact people run, not the source structure, etc.

jddarcy:
Other than the small edit to the text, I think the only remaining task is to add cross references in non-javadoc comments in the affected files, the wrapper classes, etc. to this information.

liach:
@dansmithcode Do you think a one liner in each of the migrated class like:

```
// See doc/value-class-preview.md for how value classes are declared in the Java Platform APIs.
```

would be the best way forward? Feel free to suggest exact wording too.

doc/value-class-preview.md
```diff
@@ -0,0 +1,124 @@
...
+   to compile these source files into class files.
+
+3. The class files from each of these tasks reside in `support/preview/<module>`
+   for each module.
+
+4. These preview-specific class files and other resources are copied to the
+   `META-INF/preview` directory of the regular output directory.
+
+5. At run-time, jimage will pick up the preview-specific overrides from
+   `META-INF/preview`  when preview features are enabled.
```
viktorklang-ora:
```suggestion
5. At run-time, jimage will pick up the preview-specific overrides from
   `META-INF/preview` only when preview features are enabled.
```

doc/value-class-preview.md
```diff
@@ -0,0 +1,124 @@
...
+
+6. The interim `javac` used by the build system cannot pickup the
+   preview-specific overrides; they must be supplied explicitly with this
```
viktorklang-ora:
```suggestion
   preview-specific overrides; they must be supplied explicitly with the following
```

viktorklang-ora (approved):
Made some suggestions for edits, but this looks good to me.

doc/value-class-preview.md
```diff
@@ -0,0 +1,124 @@
...
+   to compile these source files into class files.
+
+3. The class files from each of these tasks reside in `support/preview/<module>`
+   for each module.
+
+4. These preview-specific class files and other resources are copied to the
+   `META-INF/preview` directory of the regular output directory.
+
+5. At run-time, jimage will pick up the preview-specific overrides from
+   `META-INF/preview` only when features are enabled.
```
dansmithcode: when _preview_ features

doc/value-class-preview.md
```diff
@@ -0,0 +1,124 @@
...
+1. In the `java.base` module.
+
+   There's no plan to migrate other classes in other modules.
+
+2. Migrated from existing classes.
+
+   These classes are available as identity classes when preview features are
+   disabled. There's no plan to introduce completely new value classes.
+
+Support for other value classes require significant changes to the build system.
```
dansmithcode: _would_ require

doc/value-class-preview.md
```diff
@@ -0,0 +1,124 @@
...
+   The jtreg [VALUE_CLASS_PLUGIN](testing.html#VALUE_CLASS_PLUGIN) allows tests
+   to migrate their own classes to value classes when running with the plugin.
+
+## Wrapper Class Caches
+
+Currently, wrapper class caches are retained even when preview features are
+enabled. They are created with `ValueClass.newReferenceArray`, which creates
+a non-flat array storing object references, allowing a value object wrapping
+a primitive value to be available quickly without redundant allocations and
+associated performance regressions. Their existence have no semantic meaning,
```
dansmithcode: _has_ no semantic meaning

dlsmith-dev (approved):

liach:
Thanks for the reviews! I will integrate to allow building the latest apidiff against last mainline merge.

liach:
Hmm, forgot to run `make update-build-docs` for the last steps... Will fix later I guess (

PR #2634, opened 2026-07-09
8387997: [lworld] Add JVMTI general spec clarification about value objects

sspitsyn:
This update is for `repo-valhalla` pre-integration. It adds a special intro section to the `jvmti.xml` with a general clarification related to value objects. Also, it fixes a couple of small typos.

Testing: N/A

---------

---------
### Progress

### Issue
 * [JDK-8387997](https://bugs.openjdk.org/browse/JDK-8387997): [lworld] Add JVMTI general spec clarification about value objects (**Enhancement** - P2)

### Reviewers

### Contributors
 * David Holmes `<dholmes@openjdk.org>`

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2634/head:pull/2634` \
`$ git checkout pull/2634`

`$ git checkout pull/2634` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2634/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2634`

`$ git pr show -t 2634`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2634.diff">https://git.openjdk.org/valhalla/pull/2634.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2634#issuecomment-4921402765)
</details>

### DIFF
--- a/src/hotspot/share/prims/jvmti.xml
+++ b/src/hotspot/share/prims/jvmti.xml
@@ -925,6 +925,18 @@ Agent_OnUnload_L(JavaVM *vm)</example>
     and/or its affiliates, in the U.S. and other countries.
   </intro>
 
+  <intro id="identityAndValueObjects" label="Identity and Value Objects">
+    When preview features are enabled, JVMTI may observe both identity objects and value objects.
+    JVMTI treats both kinds of objects in the same manner as JNI. Agents can use the JNI
+    <externallink id="jni/functions.html#hasidentity"><code>HasIdentity</code></externallink>
+    function to determine whether a JNI object reference denotes an object with identity.
+    JNI functions whose semantics depend on identity, including
+    <externallink id="jni/functions.html#issameobject"><code>IsSameObject</code></externallink>
+    and
+    <externallink id="jni/functions.html#newweakglobalref"><code>NewWeakGlobalRef</code></externallink>,
+    specify the JNI behavior for value objects. Any additional JVMTI-specific behavior
+    is specified by the affected JVMTI function or event.
+  </intro>
 
 <functionsection label="Functions">
   <intro id="jvmtiEnvAccess" label="Accessing Functions">
@@ -3444,7 +3456,7 @@ err = (*jvmti)-&gt;Deallocate(jvmti, stack_info);
       event is sent for tagged objects when the garbage collector frees the object.
       When preview features are enabled, the Object Free event is only sent for tagged
       <externallink id="jni/functions.html#hasidentity">identity objects</externallink>.
-      <b>The event is not sent for tagged values objects.</b>
+      <b>The event is not sent for tagged value objects.</b>
     </intro>
 
     <intro id="heapCallbacks" label="Heap Callback Functions">
@@ -14211,7 +14223,7 @@ myInit() {
       <p/>
       When preview features are enabled, the Object Free event is only sent for tagged
       <externallink id="jni/functions.html#hasidentity">identity objects</externallink>.
-      <b>The event is not sent for tagged values objects.</b>
+      <b>The event is not sent for tagged value objects.</b>
       <p/>
       The event handler must not use JNI functions and
       must not use <jvmti/> functions except those which
--- a/src/hotspot/share/prims/jvmti.xml
+++ b/src/hotspot/share/prims/jvmti.xml
@@ -926,16 +926,10 @@ Agent_OnUnload_L(JavaVM *vm)</example>
   </intro>
 
   <intro id="identityAndValueObjects" label="Identity and Value Objects">
-    When preview features are enabled, JVMTI may observe both identity and value objects.
-    JVMTI treats both kinds of objects in the same manner as JNI. Agents can use the JNI
-    <externallink id="jni/functions.html#hasidentity"><code>HasIdentity</code></externallink>
-    function to determine whether a JNI object reference denotes an object with identity.
-    JNI functions whose semantics depend on identity, including
-    <externallink id="jni/functions.html#issameobject"><code>IsSameObject</code></externallink>
-    and
-    <externallink id="jni/functions.html#newweakglobalref"><code>NewWeakGlobalRef</code></externallink>,
-    specify the JNI behavior for value objects. Any additional JVMTI-specific behavior
-    is specified by the affected JVMTI functions or events.
+    If preview features are enabled then some Java objects will have identity and others
+    (instances of value classes) will not. <jvmti/> makes no distinction between these cases
+    in general, but some specific functions will behave differently if invoked upon
+    a Java object that does not have identity.
   </intro>
 
 <functionsection label="Functions">
--- a/src/hotspot/share/prims/jvmti.xml
+++ b/src/hotspot/share/prims/jvmti.xml
@@ -929,7 +929,19 @@ Agent_OnUnload_L(JavaVM *vm)</example>
     If preview features are enabled then some Java objects will have identity and others
     (instances of value classes) will not. <jvmti/> makes no distinction between these cases
     in general, but some specific functions will behave differently if invoked upon
-    a Java object that does not have identity.
+    a Java object that does not have identity. Such differences in behavior are described
+    by the individual functions and events where they apply.
+    <p/>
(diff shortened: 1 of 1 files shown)


src/hotspot/share/prims/jvmti.xml
```diff
@@ -925,6 +925,18 @@ Agent_OnUnload_L(JavaVM *vm)</example>
     and/or its affiliates, in the U.S. and other countries.
   </intro>
 
+  <intro id="identityAndValueObjects" label="Identity and Value Objects">
+    When preview features are enabled, JVMTI may observe both identity and value objects.
```
dholmes-ora: JVMTI should be `<jvmti/> `. I see a couple of other misuses of `JVMTI` have also crept in through recent changes.

src/hotspot/share/prims/jvmti.xml
```diff
@@ -925,6 +925,18 @@ Agent_OnUnload_L(JavaVM *vm)</example>
...
+    JVMTI treats both kinds of objects in the same manner as JNI. Agents can use the JNI
+    <externallink id="jni/functions.html#hasidentity"><code>HasIdentity</code></externallink>
+    function to determine whether a JNI object reference denotes an object with identity.
+    JNI functions whose semantics depend on identity, including
+    <externallink id="jni/functions.html#issameobject"><code>IsSameObject</code></externallink>
+    and
+    <externallink id="jni/functions.html#newweakglobalref"><code>NewWeakGlobalRef</code></externallink>,
+    specify the JNI behavior for value objects. Any additional JVMTI-specific behavior
```
dholmes-ora: I don't think we need to say all this. It really isn't for the JVMTI spec to state where the JNI spec may treat objects with/without identity differently, and individual JVMTI methods that behave differently already state that you can use `HasIdentity` to query that. I would keep this very succinct as I did for the JNI spec itself:
> If preview features are enabled then some Java objects will have identity and others (instances of value classes) will not. The JNI makes no distinction between these cases in the general programming model, but some specific functions will behave differently if invoked upon a Java object that does not have identity.

So for JVMTI then something like:
> If preview features are enabled then some Java objects will have identity and others (instances of value classes) will not. `<jvmti/>` makes no distinction between these cases in general, but some specific functions will behave differently if invoked upon a Java object that does not have identity.

sspitsyn: Thanks, David. Updated as you suggested.

sspitsyn: Joe is also suggesting to summarize the value objects impact on JVMTI functions and events. I'm thinking if the following would be good to add:
```
diff --git a/src/hotspot/share/prims/jvmti.xml b/src/hotspot/share/prims/jvmti.xml
index cbdecfc1cb1..33202ac6b5d 100644
--- a/src/hotspot/share/prims/jvmti.xml
+++ b/src/hotspot/share/prims/jvmti.xml
@@ -930,6 +930,16 @@ Agent_OnUnload_L(JavaVM *vm)</example>
     (instances of value classes) will not. <jvmti/> makes no distinction between these cases
     in general, but some specific functions will behave differently if invoked upon
     a Java object that does not have identity.
+    <p/>
+    Value objects can be tagged. Tags for value objects follow value equality
+    semantics, and tagged value objects are kept alive while tagged. The
+    <eventlink id="ObjectFree"/> event is not sent for tagged value objects.
+    <p/>
+    Value objects also affect allocation events, including
+    <eventlink id="VMObjectAlloc"/> and <eventlink id="SampledObjectAlloc"/>,
+    and functions related to value object constructor frames, including
+    <functionlink id="GetLocalInstance"/>, <functionlink id="GetLocalObject"/>,
+    and <functionlink id="ForceEarlyReturnVoid"/>.
   </intro>

 <functionsection label="Functions">
```

dholmes-ora: If there was a "what's new" section then I would agree, but I don't think it is appropriate to add this here. The places where value objects affect JVM TI should be described clearly in the JEP and in any associated release note IMO.

sspitsyn: Agreed, thanks.

sspitsyn:
I've made another update according to the offline agreement. There is also a minor tweak in the "Object Tags" section.

src/hotspot/share/prims/jvmti.xml
```diff
@@ -14211,7 +14229,7 @@ myInit() {
       <p/>
       When preview features are enabled, the Object Free event is only sent for tagged
       <externallink id="jni/functions.html#hasidentity">identity objects</externallink>.
-      <b>The event is not sent for tagged values objects.</b>
+      <b>The event is not sent for tagged value objects.</b>
```
dholmes-ora, line 14231:
```suggestion
      <b>The event is not sent for tagged value objects because they are kept alive while tagged.</b>
```
To be consistent with earlier change.

dholmes-ora (approved):
Looks good. One suggested update.

sspitsyn:
David, thank you for review!

src/hotspot/share/prims/jvmti.xml
sspitsyn: We have discussed with you that this is confusing as the lifecycle of value objects is not defined. My move is in opposite direction. It is to remove this: "because they are kept alive while tagged."

src/hotspot/share/prims/jvmti.xml
sspitsyn: Thanks. Fixed.

dholmes-ora (approved):
Update looks good. Thanks

dholmes-ora:
Aside: you should add JVMTI to the issue synopsis. Thanks

sspitsyn:
> Aside: you should add JVMTI to the issue synopsis. Thanks

Fixed now.

plummercj (approved):
Looks good.

src/hotspot/share/prims/jvmti.xml
```diff
@@ -925,6 +925,23 @@ Agent_OnUnload_L(JavaVM *vm)</example>
...
   </intro>
 
+  <intro id="identityAndValueObjects" label="Identity and Value Objects">
+    If preview features are enabled then some Java objects will have identity and others
+    (instances of value classes) will not. <jvmti/> makes no distinction between these cases
+    in general, but some specific functions will behave differently if invoked upon
+    a Java object that does not have identity. Such differences in behavior are described
+    by the individual functions and events where they apply.
+    <p/>
+    In summary, tags for value objects use value equality semantics, and
```
AlanBateman, line 935: I think this is tricky to word as tags are (jlong) values. While more verbose, I wonder if it should be something like "Object tags are `jlong` values that a JVMTI agent can associate with an object. JVMTI allows a value object to be tagged despite the object having no identity. Consequently, the tag is a `jong` value associated with the object's fields rather than its identity.".  Having it link to the "Object Tags" section would help too. Alternatively, we could beef up the "Object Tags" section so that this overview/intro section just links without doing into any details except to highlight that value objects can be tagged but ObjectFree events are only sent for tagged identity objects.

sspitsyn: Good suggestions, thanks. I'll try to update as suggested. I like the alternative suggestion as well but it can be a little bit more tricky.

sspitsyn: @AlanBateman There was a request to finalize the SVC CSR now. So, I'll integrate this now as is, and we can incorporate your suggestion later,

sspitsyn:
David and Chris, thank you for review!

src/hotspot/share/prims/jvmti.xml
AlanBateman: Okay, and I assume you'll create a JBS issue for the follow-up.

sspitsyn: Filed this issue:
  [8388540](https://bugs.openjdk.org/browse/JDK-8388540) [lworld] Need minor tweak in JVMTI intro section on value objects

Will publish a PR soon.

PR #2639, opened 2026-07-10
8388113: [lworld] JDI and JDWP spec update: ObjectReference summary, Value class clarification, strict field initialization clarification

plummercj:
The following spec edits were made (no code changes):
- a brief value object summary in the ObjectReference overview section
- A clarification on the difference between "value classes/objects" and the JDI Value class
- Specifying where JDI ignores strict field initialization rules.

The JBS issue contains screenshots of the rendered relevant javadoc.

---------

---------
### Progress

### Issue
 * [JDK-8388113](https://bugs.openjdk.org/browse/JDK-8388113): [lworld] JDI and JDWP spec update: ObjectReference summary, Value class clarification, strict field initialization clarification (**Bug** - P2)

### Reviewers
 * [Joe Darcy](https://openjdk.org/census#darcy) (@jddarcy - Committer) ⚠️ Review applies to [54eb553e](https://git.openjdk.org/valhalla/pull/2639/files/54eb553e0747ba9f7571fac02aa69c1760302514)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2639/head:pull/2639` \
`$ git checkout pull/2639`

`$ git checkout pull/2639` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2639/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2639`

`$ git pr show -t 2639`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2639.diff">https://git.openjdk.org/valhalla/pull/2639.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2639#issuecomment-4941057417)
</details>

### DIFF
--- a/src/jdk.jdi/share/classes/com/sun/jdi/ObjectReference.java
+++ b/src/jdk.jdi/share/classes/com/sun/jdi/ObjectReference.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 1998, 2024, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1998, 2026, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
--- a/src/jdk.jdi/share/classes/com/sun/jdi/ObjectReference.java
+++ b/src/jdk.jdi/share/classes/com/sun/jdi/ObjectReference.java
@@ -55,31 +55,15 @@ import com.sun.jdi.event.VMDisconnectEvent;
  * <div class="preview-block">
  *      <div class="preview-comment">
  * <h2><a id=valueObjects>Value Objects</a></h2>
- * When preview features are enabled, JDI supports value objects and classes
- * in a manner consistent with <cite>The Java Language Specification</cite> and
- * <cite>The Java Virtual Machine Specification</cite>, with one exception noted
- * below. In particular, two value objects are considered to be the same if both
- * refer to statewise-equivalent value objects (x == y has the value {@code true}).
- * {@link ObjectReference#equals} has always been defined to return true if
- * applying the "==" operator on the mirrored objects in the target VM
- * evaluates to true. That has not changed with value objects. See Sections
- * {@jls value-objects-8.1.1.5 Value Classes} and
- * {@jls value-objects-15.21.3 Object Equality Operators} of
- * <cite>The Java Language Specification</cite>, and Section
- * {@jvms value-objects-6.5 if_acmp_cond} of the
- * <cite>The Java Virtual Machine Specification</cite>.
+ *If preview features are enabled, JDI supports value objects and classes.
+ * However, the support does in some cases deviate from identity object
+ * support in behavior or expectations as noted below:
  * <p>
- * Where JDI does differ from the JLS is in allowing the user to obtain an
- * ObjectReference to a value object under construction (while in the "larval"
- * state). For example, the JDI user could fetch 'this' after hitting a
- * breakpoint in the value object contructor. This could lead to confusing
- * behavior for the JDI user. In particular, the value object may change after
- * the ObjectReference is obtained, even though value objects are considered
- * to be immutable. Because of this, if 'this' is fetched from a value object
- * constructor, an ObjectReference representing a statewise-equivalent "snapshot"
- * is returned instead, and this ObjectReference will not reflect changes to the
- * value object that happen later during construction. See
- * {@link StackFrame#thisObject}.
+ * If an ObjectReference is obtained for a value object under construction,
+ * it will be for a <em>snapshot</em> of the value object at that point in time.
+ * Any further changes to the initialization state of the value object will not
+ * be reflected in this ObjectReference. A new ObjectReference would need to
+ * be obtained to see the updated state. See {@link StackFrame#thisObject}.
  *      </div>
  * </div>
  * @author Robert Field
@@ -116,6 +100,12 @@ public interface ObjectReference extends Value {
      * The Field must be valid for this ObjectReference;
      * that is, it must be from
      * the mirrored object's class or a superclass of that class.
+     * <div class="preview-block">
+     *      <div class="preview-comment">
+     * If preview features are enabled, this method does not prevent a
+     * strictly-initialized field from being read before it has been initialized.
+     *      </div>
+     * </div>
      *
      * @param sig the field containing the requested value
      * @return the {@link Value} of the instance field.
@@ -129,6 +119,12 @@ public interface ObjectReference extends Value {
      * The Fields must be valid for this ObjectReference;
      * that is, they must be from
      * the mirrored object's class or a superclass of that class.
+     * <div class="preview-block">
+     *      <div class="preview-comment">
+     * If preview features are enabled, this method does not prevent a
+     * strictly-initialized field from being read before it has been initialized.
+     *      </div>
+     * </div>
      *
      * @param fields a list of {@link Field} objects containing the
      * requested values.
--- a/src/jdk.jdi/share/classes/com/sun/jdi/ObjectReference.java
+++ b/src/jdk.jdi/share/classes/com/sun/jdi/ObjectReference.java
@@ -65,7 +65,7 @@ import com.sun.jdi.event.VMDisconnectEvent;
  * evaluates to true. That has not changed with value objects. See Sections
  * {@jls value-objects-8.1.1.5 Value Classes} and
  * {@jls value-objects-15.21.3 Object Equality Operators} of
- * <cite>The Java Language Specification</cite>, and Section 
+ * <cite>The Java Language Specification</cite>, and Section
  * {@jvms value-objects-6.5 if_acmp_cond} of the
  * <cite>The Java Virtual Machine Specification</cite>.
  * <p>
--- a/src/jdk.jdi/share/classes/com/sun/jdi/ObjectReference.java
+++ b/src/jdk.jdi/share/classes/com/sun/jdi/ObjectReference.java
@@ -55,7 +55,7 @@ import com.sun.jdi.event.VMDisconnectEvent;
  * <div class="preview-block">
  *      <div class="preview-comment">
  * <h2><a id=valueObjects>Value Objects</a></h2>
- *If preview features are enabled, JDI supports value objects and classes.
+ * If preview features are enabled, JDI supports value objects and classes.
  * However, the support does in some cases deviate from identity object
  * support in behavior or expectations as noted below:
  * <p>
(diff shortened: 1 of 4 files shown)


jddarcy (approved):

src/jdk.jdi/share/classes/com/sun/jdi/ObjectReference.java
```diff
@@ -52,7 +52,36 @@
...
+ * {@jls value-objects-15.21.3 Object Equality Operators} of
+ * <cite>The Java Language Specification</cite>, and Section
+ * {@jvms value-objects-6.5 if_acmp_cond} of the
+ * <cite>The Java Virtual Machine Specification</cite>.
+ * <p>
+ * Where JDI does differ from the JLS is in allowing the user to obtain an
+ * ObjectReference to a value object under construction (while in the "larval"
+ * state). For example, the JDI user could fetch 'this' after hitting a
+ * breakpoint in the value object contructor. This could lead to confusing
+ * behavior for the JDI user. In particlar, the value object may change after
```
JnRouvignac: "in particular"

plummercj: Fixed. thank you.

src/jdk.jdi/share/classes/com/sun/jdi/ObjectReference.java
```diff
@@ -52,7 +52,36 @@
...
+ * {@jvms value-objects-6.5 if_acmp_cond} of the
+ * <cite>The Java Virtual Machine Specification</cite>.
+ * <p>
+ * Where JDI does differ from the JLS is in allowing the user to obtain an
+ * ObjectReference to a value object under construction (while in the "larval"
+ * state). For example, the JDI user could fetch 'this' after hitting a
+ * breakpoint in the value object contructor. This could lead to confusing
+ * behavior for the JDI user. In particular, the value object may change after
+ * the ObjectReference is obtained, even though value objects are considered
+ * to be immutable. Because of this, if 'this' is fetched from a value object
```
dholmes-ora:
```suggestion
 * to be immutable. Consequently, if 'this' is fetched from a value object
```

src/jdk.jdi/share/classes/com/sun/jdi/ObjectReference.java
```diff
@@ -52,7 +52,36 @@
  * takes <code>ObjectReference</code> as parameter may throw
  * {@link ObjectCollectedException} if the mirrored object has been
  * garbage collected.
- *
+ * <div class="preview-block">
+ *      <div class="preview-comment">
+ * <h2><a id=valueObjects>Value Objects</a></h2>
+ * When preview features are enabled, JDI supports value objects and classes
```
dholmes-ora: This whole paragraph seems excessive when there is nothing special to say about value objects here. You shouldn't have to say that JDI follows JLS and JVMS as that is a given, except where otherwise noted.

src/jdk.jdi/share/classes/com/sun/jdi/ObjectReference.java
```diff
@@ -52,7 +52,36 @@
...
+ * in a manner consistent with <cite>The Java Language Specification</cite> and
+ * <cite>The Java Virtual Machine Specification</cite>, with one exception noted
+ * below. In particular, two value objects are considered to be the same if both
+ * refer to statewise-equivalent value objects (x == y has the value {@code true}).
```
dholmes-ora: Did you mean "two ObjectReferences are considered to be the same if ..." ?

src/jdk.jdi/share/classes/com/sun/jdi/ObjectReference.java
```diff
@@ -52,7 +52,36 @@
...
+ * {@link ObjectReference#equals} has always been defined to return true if
+ * applying the "==" operator on the mirrored objects in the target VM
+ * evaluates to true. That has not changed with value objects. See Sections
+ * {@jls value-objects-8.1.1.5 Value Classes} and
+ * {@jls value-objects-15.21.3 Object Equality Operators} of
+ * <cite>The Java Language Specification</cite>, and Section
+ * {@jvms value-objects-6.5 if_acmp_cond} of the
+ * <cite>The Java Virtual Machine Specification</cite>.
+ * <p>
+ * Where JDI does differ from the JLS is in allowing the user to obtain an
```
dholmes-ora: I don't think this conversational style is suitable for this kind of API spec. The spec should just say what happens, not what it differs from. I would phrase more succinctly as:
> If an ObjectReference is obtained for a value object under construction, then it is a snapshot of the value object at that point in time. Any further changes to the initialization state of the value object will not be reflected in the initial ObjectReference.

dholmes-ora (commented):
If there is something special about the interaction of ObjectReferences and value objects then that should simply be stated at the appropriate location. This reads like a "what's new" section for value objects but I think it says far too much and is not something you would want to keep in this form once value objects are out of preview.

src/jdk.jdi/share/classes/com/sun/jdi/ObjectReference.java
plummercj: It's not a given. JDI does not implement the Java language nor a Java Virtual Machine, so it is not beholden to either spec. References to the JLS and JVMS (which already exist in the JDI javadoc) are used in places to aid in specifying JDI behavior.

src/jdk.jdi/share/classes/com/sun/jdi/ObjectReference.java
plummercj: I was referring to debuggee side value objects in this case. ObjectReferences are identity objects so two ObjectReferences are only "the same" if they are literally the same heap objects, just like with any other identity object. As noted below, ObjectReferences are considered to be equal (Object.equals()) if "==" of the target VM objects evaluates to true.

src/jdk.jdi/share/classes/com/sun/jdi/ObjectReference.java
plummercj: Ok. I'll work on this.

src/jdk.jdi/share/classes/com/sun/jdi/ObjectReference.java
plummercj: I assume this edit is no longer relevant after your earlier suggested rewrite of this section.

src/jdk.jdi/share/classes/com/sun/jdi/ObjectReference.java
dholmes-ora: Then I don't understand the original sentence - it makes it sound like there are additional value objects. Did you just mean to state what the value object equality (edit: not identity) check is? i.e.
> In particular, two value objects are considered to be the same if they are state-wise equivalent (`x == y` has the value true)

src/jdk.jdi/share/classes/com/sun/jdi/ObjectReference.java
dholmes-ora: Not sure what "beholden to the spec" actually means in this context, but to say JDI value object support is "consistent with JLS/JVMS" really doesn't tell you anything IMO. The main point here is simply that you can obtain an `ObjectReference` for a value object, just like any identity object. The only, very important, difference, is that it is a snapshot of the value object at that point in time. And whilst the value object is immutable, different ObjectReferences obtained during construction could capture different initalization states.

plummercj: Your summary is correct. I was just trying to provide more context and understanding. I can take a stab at a more succinct version that makes no spec references.

src/jdk.jdi/share/classes/com/sun/jdi/ObjectReference.java
plummercj: Yes, if by "identity" check you mean what is done for "==".

src/jdk.jdi/share/classes/com/sun/jdi/ObjectReference.java
plummercj: ...but keep in mind that I'm also trying to satisfy Joe's requests:

>Additional comment: for JDI, I recommend including a short discussion of objects of value classes somewhere in the API; this would supplement the links to Class.isValue already in the patch.

And

> my request is posed from the perspective of a reader of the JDI docs who is trying to find out how JDI works with respect to value classes. Value classes differ on == semantics, no locking, etc. So besides the point changes in various methods reflecting those updates, I think a brief paragraph (which could potentially be linked to from elsewhere in the JDI javadoc) explaining those points would be useful.

I don't want this to end up being reduced to just restating what is in StackFrame::thisObject(), which seems to be where your suggestions are headed.

src/jdk.jdi/share/classes/com/sun/jdi/ObjectReference.java
plummercj: After re-reviewing you comments, it seem the only thing you really want in this section is to say "JDI supports value objects" and then mention the constructor warning you've written above. The latter isn't really adding any value since it is already explained in ObjectRefernece.thisObject(), so then this section boils down to "JDI supports value objects".

src/jdk.jdi/share/classes/com/sun/jdi/ObjectReference.java
dholmes-ora: I am somewhat at odds with Joe's requests. I understand where they are coming from but they are based on the "what's new" view of spec reading. When value objects are new it is understandable to ask "where are all the places where value objects need special handling". But to me that is where the JEP and/or release notes come into play. The spec's do not have these "what's new" sections. If they have overview sections then we can certainly add some mention of value objects (as done for JNI and JVMTI) but in minimal ways. Obviously it is necessary to explain what an ObjectReference means for a value object, but that is all that is needed within the ObjectReference spec.

src/jdk.jdi/share/classes/com/sun/jdi/ObjectReference.java
dholmes-ora: Sorry typo: s/identity/equality/

src/jdk.jdi/share/classes/com/sun/jdi/ObjectReference.java
dholmes-ora: No, I want ObjectReference to explain how an ObjectReference that mirrors a value object differs from one that mirrors an identity object. If that duplicates what is said in `StackFrame.thisObject()` then maybe `thisObject` is not where it should be being described. Or maybe it is worth noting in both places. But it certainly has to be noted in ObjectReference IMO.

dholmes-ora (approved):
This all seems fine to me. Thanks

dholmes-ora (approved):

sspitsyn (approved):
This looks good.

dlsmith-dev (approved):

plummercj (commented):
Thank you everyone for the reviews!

/integrate

AlanBateman:
I see this has already been integrated but it is problematic to use "If preview features are enabled" in the JDI spec without clarifying that it means preview features in the target VM are enabled. Whether preview features are enabled or not in the current VM is not really relevant. It gets more complicated when the JDK version used for the front-end (JDI) is different to the JDK version in the back-end (target VM), and more complicated again once JDI is "upgraded" or copied into the JDK used by the IDE/debugger.

It's a similar discussion to that in JDK-8387726 which is why we decided that the simplest is for StackFrame.thisObject to link to Class.isValue instead of "If preview features are enabled".

PR #2641, opened 2026-07-14
8388193: [lworld] C2's ClearArray keeps zero fill value alive in register

TobiHartmann:
Valhalla extended `ClearArrayNode` with a value input to support nonzero initialization of flattened/null-restricted arrays. Ordinary array clears still supplied a `ConL(0)` and on x86 the matcher treated that value as an input constrained to `RAX`. This extended the live range causing increasing spills and a performance regression with the DaCapo23-biojava-large benchmark.

The zero-fill rules now match `immL0`, only use `RAX` as a local temporary and materialize zero in `MacroAssembler::clear_mem` like [mainline code does](https://github.com/openjdk/jdk/blob/master/src/hotspot/cpu/x86/macroAssembler_x86.cpp#L6039-L6041).  The fix also renames `word_copy_only` to `requires_word_fill`, improved comments and predicates and fixes some of the outdated opto assembly printing.

With this fix, we are back to baseline performance (from 47 ms/op  to 35 ms/op).

Thanks,
Tobias

---------

---------
### Progress

### Issue
 * [JDK-8388193](https://bugs.openjdk.org/browse/JDK-8388193): [lworld] C2's ClearArray keeps zero fill value alive in register (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2641/head:pull/2641` \
`$ git checkout pull/2641`

`$ git checkout pull/2641` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2641/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2641`

`$ git pr show -t 2641`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2641.diff">https://git.openjdk.org/valhalla/pull/2641.diff</a>

</details>


src/hotspot/cpu/x86/macroAssembler_x86.cpp
```diff
@@ -6506,11 +6506,11 @@ void MacroAssembler::remove_frame(int initial_framesize, bool needs_stack_repair
 
 #ifdef COMPILER2
 
-// clear memory of size 'cnt' qwords, starting at 'base' using XMM/YMM/ZMM registers
+// Fill memory with 'val', for 'cnt' qwords starting at 'base', using XMM/YMM/ZMM registers.
 void MacroAssembler::xmm_clear_mem(Register base, Register cnt, Register val, XMMRegister xtmp, KRegister mask) {
```
chhagedorn: Should we also update `xmm_clear_mem` to `xmm_fill_mem`?

We should also do these comment updates in the header file.

src/hotspot/cpu/x86/macroAssembler_x86.cpp
```diff
@@ -6506,11 +6506,11 @@ void MacroAssembler::remove_frame(int initial_framesize, bool needs_stack_repair
...
   // cnt - number of qwords (8-byte words).
   // base - start address, qword aligned.
-  Label L_zero_64_bytes, L_loop, L_sloop, L_tail, L_end;
+  Label L_fill_64_bytes, L_loop, L_sloop, L_tail, L_end;
```
chhagedorn, line 6514: There are still some references in the `.ad` file to `L_zero_64_bytes` that should also be updated to `L_fill_64_bytes`.

src/hotspot/cpu/x86/x86.ad
```diff
@@ -14976,31 +14985,29 @@ instruct rep_stos_evex_word_copy(rcx_RegL cnt, rdi_RegP base, legRegD tmp, kReg
...
-       $$emit$$"vpxor   ymm0,ymm0,ymm0\n\t"
+    if (UseXMMForObjInit) {
+       $$emit$$"movdq   $tmp, $val\n\t"
+       $$emit$$"punpcklqdq $tmp, $tmp\n\t"
+       $$emit$$"vinserti128_high $tmp, $tmp\n\t"
        $$emit$$"jmpq    L_zero_64_bytes\n\t"
        $$emit$$"# L_loop:\t# 64-byte LOOP\n\t"
-       $$emit$$"vmovdqu ymm0,(rax)\n\t"
-       $$emit$$"vmovdqu ymm0,0x20(rax)\n\t"
+       $$emit$$"vmovdqu $tmp,(rax)\n\t"
```
chhagedorn: There are some other hardcoded uses of `ymm0`, for example in `rep_stos_evex()`. Do they also need to be updated to `$tmp`?

chhagedorn (commented):
The fix looks good, thanks! And good clean-ups, I have some small comments following up on that.

TobiHartmann:
Thanks for your thorough review Christian! I did another pass over all the affected code. I renamed "zero" and "clear" to "fill" where appropriate, replaced hardcoded registers in the opto assembly printing by the correct placeholders and also fixed some of the other comments. Your comments should be all addressed now.

chhagedorn (approved):
Looks good, thanks for the update!

TobiHartmann:
Thanks for the review Christian!

TobiHartmann:
Re-opened for mainline: https://github.com/openjdk/jdk/pull/32124

PR #2642, opened 2026-07-14
8252185: [lworld] Improve performance of identityHashCode for value objects

marc-chevalier:
It is conceptually very similar to acmp: it has two parts.

# Static Expansion

If the operand of `identityHashCode` is known at compile-time, we can basically inline the implementation of `ValueObjectMethods.valueObjectHashCode`. There are a few points worth noting.
1. The seed of the hash is the hash of the mirror class object. That object is not a value object, but an identity class. We look in the header of the said class whether it was cached already, which is very likely. Otherwise, we give up the static expansion: it is not worth replacing a call with a call.
2. Oops are hell. We don't expand when oops are involved.
3. The runtime implementation performs some unsafe gets to get all pieces of a segment in a simple, greedy way. Which means that in the case we have a value class made of 2 `int`, the likely outcome is that it would get both at once with a single call to `getLong`. This is fine, we can do that, but we need to mark the access as mismatch and unsafe (but aligned, since the acmp maps are smartly done). But in the case we have a class made of a single `int`, it would be unfortunate to mark the `getInt` access as mismatch since it prevents some optimizations. So if we are getting exactly a field, we detect it, and we mark the load as non-mismatch (match?).

# Fast Path

This supports only objects with a simple shape: one segment of data, being 1, 2, 4 or 8-byte long, which is enough to cover many migrated classes (but not dates for instance). This feature is morally very similar to the acmp fast path. It is controlled by the diagnostic flag `UseHashcodeFastPath`. Alike acmp, we need to sabotage the fast path in case we can do a static expansion later.

I can only suggest you take a look at `inlineKlass.hpp` on how that works.

# Benchmarking

Microbenchmarking is rather unsurprising:
- haven't changed:
  - `null` was and is still fast (1-2ns)
  - hashcode cached in the header was and is still fast (2ns)
  - identity objects without cached hashcode are not too slow, and still aren't (20-25ns)
  - known at compile time (60-70ns):
    - value objects with oops
  - unknown at compile time (60-70ns):
    - value objects with sizes that are not 1, 2, 4 or 8
    - value objects with oops
- got better (60-70ns -> 2-7ns):
  - value objects with a type known at compile time, without oops
  - value objects with a type unknown at compile time, without oops, of size 1, 2, 4 or 8.

<details>
<summary>Full results of <tt>valhalla.hash.FastPath</tt></summary>

Before:
```
Benchmark                                      Mode  Cnt   Score   Error  Units
FastPath.big_mix                               avgt   15  49.680 ± 0.628  ns/op
FastPath.heterogeneous_small                   avgt   15  48.337 ± 0.865  ns/op
FastPath.heterogeneous_too_big                 avgt   15  57.891 ± 0.652  ns/op
FastPath.heterogeneous_with_obj                avgt   15  22.411 ± 0.323  ns/op
FastPath.homogeneous_byte                      avgt   15  56.707 ± 1.043  ns/op
FastPath.homogeneous_byte_static               avgt   15  57.202 ± 1.102  ns/op
FastPath.homogeneous_empty                     avgt   15  56.501 ± 1.734  ns/op
FastPath.homogeneous_empty_static              avgt   15  54.758 ± 1.700  ns/op
FastPath.homogeneous_int                       avgt   15  56.281 ± 1.197  ns/op
FastPath.homogeneous_int_int                   avgt   15  57.415 ± 1.268  ns/op
FastPath.homogeneous_int_int_static            avgt   15  56.711 ± 1.197  ns/op
FastPath.homogeneous_int_static                avgt   15  57.024 ± 1.262  ns/op
FastPath.homogeneous_long                      avgt   15  57.192 ± 1.170  ns/op
FastPath.homogeneous_long_static               avgt   15  56.798 ± 1.375  ns/op
FastPath.homogeneous_null                      avgt   15   1.371 ± 0.040  ns/op
FastPath.homogeneous_short                     avgt   15  58.196 ± 1.259  ns/op
FastPath.homogeneous_short_static              avgt   15  58.724 ± 1.449  ns/op
FastPath.homogeneous_too_big_long_int          avgt   15  61.896 ± 2.272  ns/op
FastPath.homogeneous_too_big_long_int_static   avgt   15  59.864 ± 1.394  ns/op
FastPath.homogeneous_too_big_long_long         avgt   15  59.559 ± 1.188  ns/op
FastPath.homogeneous_too_big_long_long_static  avgt   15  59.124 ± 1.078  ns/op
FastPath.homogeneous_weird_size                avgt   15  60.049 ± 1.678  ns/op
FastPath.homogeneous_weird_size_static         avgt   15  59.820 ± 1.373  ns/op
FastPath.homogeneous_with_obj_array            avgt   15  22.113 ± 0.757  ns/op
FastPath.homogeneous_with_obj_string           avgt   15  23.866 ± 0.444  ns/op
FastPath.homogeneous_with_oop                  avgt   15  71.719 ± 1.863  ns/op
FastPath.homogeneous_with_oop_static           avgt   15  71.977 ± 1.539  ns/op
FastPath.pre_hashed                            avgt   15   1.906 ± 0.054  ns/op
```

After:
```
Benchmark                                      Mode  Cnt   Score   Error  Units
FastPath.big_mix                               avgt   15  27.101 ± 0.519  ns/op
FastPath.heterogeneous_small                   avgt   15   5.175 ± 0.104  ns/op
FastPath.heterogeneous_too_big                 avgt   15  62.371 ± 1.041  ns/op
FastPath.heterogeneous_with_obj                avgt   15  23.229 ± 0.408  ns/op
FastPath.homogeneous_byte                      avgt   15   5.878 ± 0.071  ns/op
FastPath.homogeneous_byte_static               avgt   15   2.407 ± 0.054  ns/op
FastPath.homogeneous_empty                     avgt   15   3.459 ± 0.082  ns/op
FastPath.homogeneous_empty_static              avgt   15   2.156 ± 0.047  ns/op
FastPath.homogeneous_int                       avgt   15   5.916 ± 0.044  ns/op
FastPath.homogeneous_int_int                   avgt   15   7.530 ± 0.076  ns/op
FastPath.homogeneous_int_int_static            avgt   15   7.120 ± 0.109  ns/op
FastPath.homogeneous_int_static                avgt   15   2.117 ± 0.035  ns/op
FastPath.homogeneous_long                      avgt   15   3.668 ± 0.072  ns/op
FastPath.homogeneous_long_static               avgt   15   2.211 ± 0.055  ns/op
FastPath.homogeneous_null                      avgt   15   1.415 ± 0.042  ns/op
FastPath.homogeneous_short                     avgt   15   5.832 ± 0.058  ns/op
FastPath.homogeneous_short_static              avgt   15   2.099 ± 0.025  ns/op
FastPath.homogeneous_too_big_long_int          avgt   15  62.184 ± 1.775  ns/op
FastPath.homogeneous_too_big_long_int_static   avgt   15   2.536 ± 0.033  ns/op
FastPath.homogeneous_too_big_long_long         avgt   15  60.679 ± 1.968  ns/op
FastPath.homogeneous_too_big_long_long_static  avgt   15   2.894 ± 0.103  ns/op
FastPath.homogeneous_weird_size                avgt   15  60.444 ± 1.753  ns/op
FastPath.homogeneous_weird_size_static         avgt   15   2.410 ± 0.039  ns/op
FastPath.homogeneous_with_obj_array            avgt   15  22.019 ± 0.359  ns/op
FastPath.homogeneous_with_obj_string           avgt   15  24.011 ± 0.332  ns/op
FastPath.homogeneous_with_oop                  avgt   15  73.037 ± 1.287  ns/op
FastPath.homogeneous_with_oop_static           avgt   15  74.006 ± 2.115  ns/op
FastPath.pre_hashed                            avgt   15   1.888 ± 0.063  ns/op
```
</details>

On real benchmark, improvements are noticeable only on hash-intensive cases.

# Future Work

We can also add profiling and speculate on the type of the argument, as we do for acmp. That is left as an exercise for the reader.

The current implementation of both the static expansion and the fast path are not caching the result in the header. There are a few reasons for that:
1. It is not clear it is worth it.
2. It is non-trivial to do only when there is already a buffer, to avoid keeping an allocation around just for that. This will be easier in the future, with a macro-based hashcode, where the expansion would do the storing only if something else kept the buffer alive (or already exists). With that, we also would need not to remove the fast path after static expansion.
3. It is non-trivial to store something in the header while being race-safe. It probably needs cooperation of the backend to do something alike `JVM_IHashCode`.

Thanks,
Marc

---------

---------
### Progress

### Issue
 * [JDK-8252185](https://bugs.openjdk.org/browse/JDK-8252185): [lworld] Improve performance of identityHashCode for value objects (**Enhancement** - P2)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2642/head:pull/2642` \
`$ git checkout pull/2642`

`$ git checkout pull/2642` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2642/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2642`

`$ git pr show -t 2642`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2642.diff">https://git.openjdk.org/valhalla/pull/2642.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/oops/inlineKlass.hpp
+++ b/src/hotspot/share/oops/inlineKlass.hpp
@@ -114,6 +114,13 @@ class InlineKlass: public InstanceKlass {
     int _fast_acmp_offset;    // if < 0, fast acmp doesn't apply
     int64_t _fast_acmp_mask;  // can be 0 for empty value classes
 
+    int _fast_hashcode_offset;   // if < 0, fast hashcode doesn't apply
+#ifdef VM_LITTLE_ENDIAN
+    int _fast_hashcode_shift;
+#else
+    int64_t _fast_hashcode_mask;
+#endif
+
     Members();
 
     void print_on(outputStream* st) const;
@@ -211,6 +218,17 @@ class InlineKlass: public InstanceKlass {
 
   int64_t fast_acmp_mask() const                              { return members()._fast_acmp_mask; }
   void set_fast_acmp_mask(int64_t mask)                       { members()._fast_acmp_mask = mask; }
+
+  int fast_hashcode_offset() const                                { return members()._fast_hashcode_offset; }
+  void set_fast_hashcode_offset(int offset)                       { members()._fast_hashcode_offset = offset; }
+
+#ifdef VM_LITTLE_ENDIAN
+  int fast_hashcode_shift() const                              { return members()._fast_hashcode_shift; }
+  void set_fast_hashcode_shift(int shift)                       { members()._fast_hashcode_shift = shift; }
+#else
+  int64_t fast_hashcode_mask() const                              { return members()._fast_hashcode_mask; }
+  void set_fast_hashcode_mask(int64_t mask)                       { members()._fast_hashcode_mask = mask; }
+#endif
 
   bool supports_nullable_layouts() const {
     return has_nullable_non_atomic_layout() || has_nullable_atomic_layout();
@@ -334,6 +352,19 @@ class InlineKlass: public InstanceKlass {
   static ByteSize fast_acmp_mask_offset() {
     return byte_offset_of(Members, _fast_acmp_mask);
   }
+
+  static ByteSize fast_hashcode_offset_offset() {
+    return byte_offset_of(Members, _fast_hashcode_offset);
+  }
+#ifdef VM_LITTLE_ENDIAN
+  static ByteSize fast_hashcode_shift_offset() {
+    return byte_offset_of(Members, _fast_hashcode_shift);
+  }
+#else
+  static ByteSize fast_hashcode_mask_offset() {
+    return byte_offset_of(Members, _fast_hashcode_mask);
+  }
+#endif
 
   oop null_reset_value() const;
   void set_null_reset_value(oop val);
--- a/src/hotspot/share/oops/inlineKlass.hpp
+++ b/src/hotspot/share/oops/inlineKlass.hpp
@@ -114,9 +114,9 @@ class InlineKlass: public InstanceKlass {
     int _fast_acmp_offset;    // if < 0, fast acmp doesn't apply
     int64_t _fast_acmp_mask;  // can be 0 for empty value classes
 
-    // When we can't intrinsify the identityHashCode, we can still aboid the Java call at runtim if the value object is nice
+    // When we can't intrinsify the identityHashCode, we can still avoid the Java call at runtime if the value object is nice
     // enough. This fast path basically implements the method ValueObjectMethods::valueObjectHashCode in a special case. This
-    // special case is when there is at most one no-oop segment in the acmp maps, that this segment (if it exsits) is 1, 2 4
+    // special case is when there is at most one no-oop segment in the acmp maps, that this segment (if it exists) is 1, 2, 4
     // or 8 byte long, and there is no oop in the acmp maps. Basically, valueObjectHashCode makes 0 or 1 iteration of the big
     // outer loop, and one iteration of one of the inner loops. The fast path loads a long at the given offset, isolates the
     // numeric value we are interested in, and does the arithmetic.
@@ -127,11 +127,11 @@ class InlineKlass: public InstanceKlass {
     // 3. the object has one segment: we set _fast_hashcode_offset according to where we should load.
     //
     // Alike for the acmp fast path, we must not load further than the object, and we use the same trick as for acmp, and we
-    // load possibly some part of the header. The cases 2. and 3. cannot collide since loading loading at offset 0 would read
-    // only the header, and no payload.
+    // load possibly some part of the header. The cases 2. and 3. cannot collide since loading at offset 0 would read only the
+    // header, and no payload.
     //
     // But unlike acmp, we need the actual arithmetic value, and resetting irrelevant bits is not correct. To do that, we need
-    // to have a different logic wrt. endianness. Moreover, the fast path needs to handle differently a when the segment is 8-byte
+    // to have a different logic wrt. endianness. Moreover, the fast path needs to handle differently when the segment is 8-byte
     // long, just as valueObjectHashCode does, while the arithmetic for segments of size 1, 2 or 4 is the same. This is also known
     // by a endianness-dependent test.
     int _fast_hashcode_offset;   // if < 0, fast hashcode doesn't apply
@@ -162,7 +162,7 @@ class InlineKlass: public InstanceKlass {
     // This field is storing the mask. Since we keep 1, 2, 4 or 8 bytes, the legal values of _fast_hashcode_mask
     // are 0xff ff ff ff ff ff ff ff, 0x00 00 00 00 ff ff ff ff, 0x00 00 00 00 00 00 ff ff or 0x00 00 00 00 00 00 00 ff
     //
-    // The fast path is aware we are loading a long if the shift is (long)-1.
+    // The fast path is aware we are loading a long if the mask is (long)-1.
     // Value is not specified (and does not matter) if _fast_hashcode_offset <= 0
     int64_t _fast_hashcode_mask;
 #endif
--- a/src/hotspot/share/oops/inlineKlass.hpp
+++ b/src/hotspot/share/oops/inlineKlass.hpp
@@ -264,15 +264,15 @@ class InlineKlass: public InstanceKlass {
   int64_t fast_acmp_mask() const                              { return members()._fast_acmp_mask; }
   void set_fast_acmp_mask(int64_t mask)                       { members()._fast_acmp_mask = mask; }
 
-  int fast_hashcode_offset() const                                { return members()._fast_hashcode_offset; }
-  void set_fast_hashcode_offset(int offset)                       { members()._fast_hashcode_offset = offset; }
+  int fast_hashcode_offset() const                            { return members()._fast_hashcode_offset; }
+  void set_fast_hashcode_offset(int offset)                   { members()._fast_hashcode_offset = offset; }
 
 #ifdef VM_LITTLE_ENDIAN
-  int fast_hashcode_shift() const                              { return members()._fast_hashcode_shift; }
-  void set_fast_hashcode_shift(int shift)                       { members()._fast_hashcode_shift = shift; }
+  int fast_hashcode_shift() const                             { return members()._fast_hashcode_shift; }
+  void set_fast_hashcode_shift(int shift)                     { members()._fast_hashcode_shift = shift; }
 #else
-  int64_t fast_hashcode_mask() const                              { return members()._fast_hashcode_mask; }
-  void set_fast_hashcode_mask(int64_t mask)                       { members()._fast_hashcode_mask = mask; }
+  int64_t fast_hashcode_mask() const                          { return members()._fast_hashcode_mask; }
+  void set_fast_hashcode_mask(int64_t mask)                   { members()._fast_hashcode_mask = mask; }
 #endif
 
   bool supports_nullable_layouts() const {
--- a/src/hotspot/share/oops/inlineKlass.hpp
+++ b/src/hotspot/share/oops/inlineKlass.hpp
@@ -105,7 +105,7 @@ class InlineKlass: public InstanceKlass {
     //
     // This doesn't always apply, for instance, if there are oops among the fields, we shouldn't carelessly load and compare:
     // the GC might move the object in between.
-    // To signal this fast path cannot be done on this current class, simply put 0 in _fast_acmp_mask.
+    // To signal this fast path cannot be done on this current class, simply put -1 in _fast_acmp_offset.
     //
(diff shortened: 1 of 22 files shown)


src/hotspot/share/oops/inlineKlass.hpp
```diff
@@ -114,6 +114,58 @@ class InlineKlass: public InstanceKlass {
...
+    // 0x01 02 03 04 HH HH HH HH, where HH are header bytes. To get the integer value, we can simply do a unsigned shift right,
+    // by 4 bytes (32 bits) in this case.
+    // This field is saying by how much we need to shift. Since we keep 1, 2, 4 or 8 bytes, the legal values of _fast_hashcode_shift
+    // are 8 * (8 - (1, 2, 4, 8)) = 8 * (7, 6, 4, 0) = 56, 48, 32, 0.
+    //
+    // The fast path is aware we are loading a long if the shift is 0.
+    // Value is not specified (and does not matter) if _fast_hashcode_offset <= 0
+    int _fast_hashcode_shift;
+#else
+    // In little endian, the memory layout, with a 4-byte segment whose value (as returned by getInt) would be 0x01 02 03 04. The
```
liach:
```suggestion
    // In big endian, the memory layout, with a 4-byte segment whose value (as returned by getInt) would be 0x01 02 03 04. The
```

Also "load to far" -> "load too far" both above and below

src/hotspot/share/opto/inlinetypenode.cpp
```diff
@@ -994,6 +994,95 @@ Node* InlineTypeNode::emit_substitutability_check(GraphKit* kit, Node* lhs, Node
...
+  Node* result = kit->intcon(checked_cast<jint>(klass_hash));
+  for (int i = 0; i < number_of_nonoop_entries; i++) {
+    AcmpMapSegment segment = vk->get_nonoop_segment_of_acmp_map(i);
+    int offset = segment._offset;
+    int size = segment._size;
+    int nlong = size / 8;
+    for (int j = 0; j < nlong; j++) {
+      Node* la = make_load(offset, TypeLong::LONG, T_LONG);
+      result = kit->AddI(kit->MulI(thirty_one, result), kit->ConvL2I(la));
+      result = kit->AddI(kit->MulI(thirty_one, result), kit->ConvL2I(kit->URShiftL(la, kit->intcon(32))));
```
liach, line 1057: I assume this must be kept in parity with ValueObjectMethods - a bit risky especially if we don't have comments pointing maintainers. Also feel free to use whatever computation that is the most friendly to an optimizing compiler; in core libs people are talking about replacing this `*31+` with `Long.hashCode`.

src/hotspot/share/opto/library_call.cpp
```diff
@@ -5531,48 +5605,98 @@ bool LibraryCallKit::inline_native_hashcode(bool is_virtual, bool is_static) {
...
+#else
+              // *(obj + offset) & mask
+              Node* mask_addr = off_heap_plus_addr(members, in_bytes(InlineKlass::fast_hashcode_mask_offset()));
+              Node* mask = make_load(control(), mask_addr, TypeLong::LONG, T_LONG, MemNode::unordered);
+              Node* obj_extracted = AndL(obj_payload, mask);
+              Node* is_long_payload_bol = BoolCmpI(mask, BoolTest::eq, longcon(-1));
+#endif
+              IfNode* iff_is_long_payload = create_and_map_if(control(), is_long_payload_bol, PROB_FAIR, COUNT_UNKNOWN);
+
+              Node* result_int = AddI(MulI(intcon(31), result_empty), ConvL2I(obj_extracted));
```
liach, line 5682: Same long handling comment here.

src/hotspot/share/opto/inlinetypenode.cpp
marc-chevalier: Indeed, if somebody changes one side, it is possible they won't know about the other places the computation happens and not update there. Yet, I don't think there is much risk: dedicated tests will immediately fail, so I don't think that with any reasonable testing we might realistically end up with really diverging implementations.

Moreover, I don't have strong opinion on how hash should work, and I'm not too unhappy with how it is now. I suspect core libs folks might have stronger opinion, or want to discuss options... If they decide to change, I can help and adapt that, but I don't feel like sneaking a change in how hash code work here.

src/hotspot/share/opto/inlinetypenode.cpp
```diff
@@ -994,6 +994,95 @@ Node* InlineTypeNode::emit_substitutability_check(GraphKit* kit, Node* lhs, Node
   return result;
 }
 
+// Check if a substitutability check between 'lhs' and 'rhs' can be implemented in IR.
```
SirYwell: The first sentence does not apply here

src/hotspot/share/oops/inlineKlass.hpp
```diff
@@ -114,6 +114,59 @@ class InlineKlass: public InstanceKlass {
     int _fast_acmp_offset;    // if < 0, fast acmp doesn't apply
     int64_t _fast_acmp_mask;  // can be 0 for empty value classes
 
+    // When we can't intrinsify the identityHashCode, we can still aboid the Java call at runtim if the value object is nice
```
ExE-Boss:
```suggestion
    // When we can't intrinsify the identityHashCode, we can still avoid the Java call at runtime if the value object is nice
```

src/hotspot/share/oops/inlineKlass.hpp
```diff
@@ -114,6 +114,59 @@ class InlineKlass: public InstanceKlass {
...
+    // enough. This fast path basically implements the method ValueObjectMethods::valueObjectHashCode in a special case. This
+    // special case is when there is at most one no-oop segment in the acmp maps, that this segment (if it exsits) is 1, 2 4
```
ExE-Boss:
```suggestion
    // special case is when there is at most one no-oop segment in the acmp maps, that this segment (if it exists) is 1, 2, 4,
```

src/hotspot/share/oops/inlineKlass.hpp
```diff
@@ -114,6 +114,59 @@ class InlineKlass: public InstanceKlass {
...
+    // outer loop, and one iteration of one of the inner loops. The fast path loads a long at the given offset, isolates the
+    // numeric value we are interested in, and does the arithmetic.
+    //
+    // There are cases:
+    // 1. hashcode fast path doesn't apply: we set _fast_hashcode_offset < 0
+    // 2. the object has no segments (i.e. it is empty): we set _fast_hashcode_offset = 0
+    // 3. the object has one segment: we set _fast_hashcode_offset according to where we should load.
+    //
+    // Alike for the acmp fast path, we must not load further than the object, and we use the same trick as for acmp, and we
+    // load possibly some part of the header. The cases 2. and 3. cannot collide since loading loading at offset 0 would read
```
ExE-Boss:
```suggestion
    // load possibly some part of the header. The cases 2. and 3. cannot collide since loading at offset 0 would read
```

src/hotspot/share/oops/inlineKlass.hpp
```diff
@@ -114,6 +114,59 @@ class InlineKlass: public InstanceKlass {
...
+    // only the header, and no payload.
+    //
+    // But unlike acmp, we need the actual arithmetic value, and resetting irrelevant bits is not correct. To do that, we need
+    // to have a different logic wrt. endianness. Moreover, the fast path needs to handle differently a when the segment is 8-byte
```
ExE-Boss:
```suggestion
    // to have a different logic wrt. endianness. Moreover, the fast path needs to handle differently when the segment is 8-byte
```

src/hotspot/share/oops/inlineKlass.hpp
```diff
@@ -114,6 +114,59 @@ class InlineKlass: public InstanceKlass {
...
+    //                v- start of payload
+    // ....header.... | 01 02 03 04
+    //    \___________|___________/
+    // Not to load too far, we load at offset "start of payload" - 4, so, we get some header bytes, and we get the long value
+    // 0xHH HH HH HH 01 02 03 04, where HH are header bytes. To get the integer value, we can simply filter out the upper bits, with
+    // the mask 0x00 00 00 00 ff ff ff ff in this case.
+    // This field is storing the mask. Since we keep 1, 2, 4 or 8 bytes, the legal values of _fast_hashcode_mask
+    // are 0xff ff ff ff ff ff ff ff, 0x00 00 00 00 ff ff ff ff, 0x00 00 00 00 00 00 ff ff or 0x00 00 00 00 00 00 00 ff
+    //
+    // The fast path is aware we are loading a long if the shift is (long)-1.
```
ExE-Boss:
```suggestion
    // The fast path is aware we are loading a long if the mask is (long)-1.
```

marc-chevalier:
Moved to mainline: https://github.com/openjdk/jdk/pull/32144

PR #2644, opened 2026-07-15
8388256: [lworld] Clone of reference array is broken in C2

TobiHartmann:
The slow path of C2's clone intrinsic calls `OptoRuntime::new_array_C` which reconstructs the array klass from its element klass and properties through this path:

https://github.com/openjdk/valhalla/blob/0b66ce5e7725fc4c120532694a764c8195f91361/src/hotspot/share/memory/oopFactory.cpp#L113-L116

->

https://github.com/openjdk/valhalla/blob/0b66ce5e7725fc4c120532694a764c8195f91361/src/hotspot/share/oops/objArrayKlass.cpp#L239-L241

For the array klass of a reference array, created through `ValueClass.newReferenceArray`, with nullable + atomic properties, a `FlatArrayKlass` is selected. As a result, we clone from a reference array into a flat array and things are going south.

The fix is to select based on the klass, not the properties. In the test, I use `-XX:-UseTLAB` to trigger the slow path reliably.

Thanks,
Tobias

---------

---------
### Progress

### Issue
 * [JDK-8388256](https://bugs.openjdk.org/browse/JDK-8388256): [lworld] Clone of reference array is broken in C2 (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2644/head:pull/2644` \
`$ git checkout pull/2644`

`$ git checkout pull/2644` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2644/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2644`

`$ git pr show -t 2644`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2644.diff">https://git.openjdk.org/valhalla/pull/2644.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/oops/objArrayKlass.cpp
+++ b/src/hotspot/share/oops/objArrayKlass.cpp
@@ -237,14 +237,18 @@ ObjArrayKlass* ObjArrayKlass::allocate_klass_from_description(ArrayDescription a
 }
 
 objArrayOop ObjArrayKlass::allocate_instance(int length, ArrayProperties props, TRAPS) {
-  check_array_allocation_length(length, arrayOopDesc::max_array_length(T_OBJECT), CHECK_NULL);
   ObjArrayKlass* ak = klass_with_properties(props, CHECK_NULL);
-  switch (ak->kind()) {
+  return ak->allocate_instance(length, CHECK_NULL);
+}
+
+objArrayOop ObjArrayKlass::allocate_instance(int length, TRAPS) {
+  check_array_allocation_length(length, arrayOopDesc::max_array_length(T_OBJECT), CHECK_NULL);
+  switch (kind()) {
     case Klass::RefArrayKlassKind:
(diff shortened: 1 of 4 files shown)


marc-chevalier (approved):
That makes sense. The selection from properties caused problems like that before. I'm not sure how to make it less likely to happen. Maybe the selection logic should expect more than just the properties (like a layout kind), to force the developer to provide it of "invalid" or something when there is no previous kind to honor. Just throwing an idea...

TobiHartmann:
Thanks for the review Marc! Maybe @fparain has some more input on this.

fparain (approved):
Changes look good to me.

TobiHartmann:
Thanks Fred!

TobiHartmann:
Re-opened for mainline: https://github.com/openjdk/jdk/pull/32125

PR #2648, opened 2026-07-16
8388317: [lworld] LibraryCallKit::inline_unsafe_flat_access asserts on null literal

TobiHartmann:
We hit an assert in `value_type->inline_klass()` in the `Unsafe::putFlatValue` C2 intrinsic when `null` is passed because `TypePtr::NULL_PTR` is not a `TypeInstPtr`. I fixed the assert to account for that.

While writing the test, I noticed that `TEST31_VT_FLATTENED` is always `false` because `test31_vt` is of type `MyValue1` which is way too large to be flat because nullable implies atomic. I think this is a regression from when we migrated from Q-types. I made the field null-free to make sure it can be flattened. I then noticed that we are missing a null-check in the intrinsic and also fixed that.

Thanks,
Tobias

---------

---------
### Progress

### Issue
 * [JDK-8388317](https://bugs.openjdk.org/browse/JDK-8388317): [lworld] LibraryCallKit::inline_unsafe_flat_access asserts on null literal (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2648/head:pull/2648` \
`$ git checkout pull/2648`

`$ git checkout pull/2648` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2648/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2648`

`$ git pr show -t 2648`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2648.diff">https://git.openjdk.org/valhalla/pull/2648.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/opto/library_call.cpp
+++ b/src/hotspot/share/opto/library_call.cpp
@@ -2782,13 +2782,17 @@ bool LibraryCallKit::inline_unsafe_flat_access(bool is_store, AccessKind kind) {
       value = new_value;
     }
 
-    assert(value_type->inline_klass() == value_klass, "value is of type %s while valueType is %s", value_type->inline_klass()->name()->as_utf8(), value_klass->name()->as_utf8());
+    assert(value_type == TypePtr::NULL_PTR || value_type->inline_klass() == value_klass,
+           "value is of type %s while value klass is %s", value_type->inline_klass()->name()->as_utf8(), value_klass->name()->as_utf8());
     if (layout == LayoutKind::REFERENCE) {
       const TypePtr* ptr_type = (decorators & C2_MISMATCHED) != 0 ? TypeRawPtr::BOTTOM : _gvn.type(ptr)->is_ptr();
       access_store_at(base, ptr, ptr_type, value, value_type, T_OBJECT, decorators);
     } else {
       bool atomic = LayoutKindHelper::is_atomic_flat(layout);
       bool null_free = !LayoutKindHelper::is_nullable_flat(layout);
+      if (null_free) {
+        null_check(value);
+      }
       value->as_InlineType()->store_flat(this, base, ptr, atomic, immutable_memory, null_free, decorators);
     }
 
--- a/test/hotspot/jtreg/compiler/valhalla/inlinetypes/TestIntrinsics.java
+++ b/test/hotspot/jtreg/compiler/valhalla/inlinetypes/TestIntrinsics.java
@@ -149,6 +149,7 @@ public class TestIntrinsics {
 
     public TestIntrinsics() {
         test24_vt = MyValue1.createWithFieldsInline(rI, rL);
+        test31_vt = MyValue1.createDefaultInline();
         super();
     }
 
@@ -658,6 +659,7 @@ public class TestIntrinsics {
         Asserts.assertEQ(v.v1, res);
     }
 
+    @NullRestricted
     MyValue1 test31_vt;
     private static final long TEST31_VT_OFFSET;
     private static final boolean TEST31_VT_FLATTENED;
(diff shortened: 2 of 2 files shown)


src/hotspot/share/opto/library_call.cpp
```diff
@@ -2782,13 +2782,17 @@ bool LibraryCallKit::inline_unsafe_flat_access(bool is_store, AccessKind kind) {
       value = new_value;
     }
 
-    assert(value_type->inline_klass() == value_klass, "value is of type %s while valueType is %s", value_type->inline_klass()->name()->as_utf8(), value_klass->name()->as_utf8());
+    assert(value_type == TypePtr::NULL_PTR || value_type->inline_klass() == value_klass,
+           "value is of type %s while value klass is %s", value_type->inline_klass()->name()->as_utf8(), value_klass->name()->as_utf8());
```
chhagedorn, line 2786: Should we add a `ResourceMark` for the `as_utf8()` calls?

chhagedorn (approved):
Looks good to me, thanks!

src/hotspot/share/opto/library_call.cpp
merykitty: Normally, we are aborting anyway, so `ResourceMark` is unnecessary. The macro is expanded so that the arguments are only evaluated if the condition is false.

src/hotspot/share/opto/library_call.cpp
```diff
@@ -2782,13 +2782,17 @@ bool LibraryCallKit::inline_unsafe_flat_access(bool is_store, AccessKind kind) {
...
     if (layout == LayoutKind::REFERENCE) {
       const TypePtr* ptr_type = (decorators & C2_MISMATCHED) != 0 ? TypeRawPtr::BOTTOM : _gvn.type(ptr)->is_ptr();
       access_store_at(base, ptr, ptr_type, value, value_type, T_OBJECT, decorators);
     } else {
       bool atomic = LayoutKindHelper::is_atomic_flat(layout);
       bool null_free = !LayoutKindHelper::is_nullable_flat(layout);
+      if (null_free) {
```
merykitty, line 2793: That's reasonable, but normally, `Unsafe` accesses require the user to know what they are doing, so `must_be_not_null` is a more reasonable choice?

src/hotspot/share/opto/library_call.cpp
chhagedorn: You're right, it's not needed then, thanks!

src/hotspot/share/opto/library_call.cpp
TobiHartmann: The unsafe implementation in the runtime also implements such a check, that's why I added it here. See:

https://github.com/openjdk/valhalla/blob/53ef1bfa0951e5cc1eeb064026eb0d73329e7c5e/src/hotspot/share/prims/unsafe.cpp#L402
->
https://github.com/openjdk/valhalla/blob/53ef1bfa0951e5cc1eeb064026eb0d73329e7c5e/src/hotspot/share/oops/valuePayload.inline.hpp#L573-L577

It's based on the layout that is passed:
https://github.com/openjdk/valhalla/blob/53ef1bfa0951e5cc1eeb064026eb0d73329e7c5e/src/hotspot/share/oops/valuePayload.inline.hpp#L444-L446

src/hotspot/share/opto/library_call.cpp
TobiHartmann: I agree with Quan-Anh, I don't think it's needed because if the assert condition fails, we exit them VM anyway.

TobiHartmann:
Thanks for your reviews Quan-Anh and Christian!

src/hotspot/share/opto/library_call.cpp
fparain: Quan-Anh is right, users of Unsafe are supposed to know what they are doing.
The javadoc of Unsafe.putFlatValue says:
```
     * @throws RuntimeException No defined exceptions are thrown, not even
     *         {@link NullPointerException}
```
The runtime version of this method throws an NPE because we factorized all write operations of flat values into a single set of methods, and the null check is required for regular writes (putfield, aastore).
That being said, having the C2 intrinsic behaving like the runtime version is not a bad thing.
I wonder if the javadoc should be updated, not to guarantee that the null check will be performed, but to warn users it could happen. But null-restriction is going to be a post JEP-401 feature, so the doc update could be delayed until we  have the final model.

TobiHartmann:
Re-opened for mainline: https://github.com/openjdk/jdk/pull/32126

src/hotspot/share/opto/library_call.cpp
TobiHartmann: Right, I don't have a strong opinion on this but I also think that C2 should ideally behave like the interpreter here.

PR #2649, opened 2026-07-16
8388400: [lworld] Incorrect null check in jdk.internal.jrtfs.ExplodedImage$PathNode constructor

jaikiran:
Can I please get a review of this change which reorders a null check in `ExplodedImage`? While at it, the `isClosable` is made `final` in `JrtFileSystem` too.

---------

---------
### Progress

### Integration blocker
&nbsp;⚠️ Title mismatch between PR and JBS for issue [JDK-8388400](https://bugs.openjdk.org/browse/JDK-8388400)

### Issue
 * [JDK-8388400](https://bugs.openjdk.org/browse/JDK-8388400): Incorrect null check in jdk.internal.jrtfs.ExplodedImage$PathNode constructor (**Bug** - P4) ⚠️ Title mismatch between PR and JBS.

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2649/head:pull/2649` \
`$ git checkout pull/2649`

`$ git checkout pull/2649` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2649/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2649`

`$ git pr show -t 2649`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2649.diff">https://git.openjdk.org/valhalla/pull/2649.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2649#issuecomment-4993813290)
</details>

### DIFF
--- a/src/java.base/share/classes/jdk/internal/jrtfs/ExplodedImage.java
+++ b/src/java.base/share/classes/jdk/internal/jrtfs/ExplodedImage.java
@@ -120,9 +120,9 @@ class ExplodedImage extends SystemImage {
          * Used for each module-named directory that are leafs of /packages/...
          */
         private PathNode(String name, PathNode link) {
-            super(name, link.getFileAttributes());
+            super(name, Objects.requireNonNull(link).getFileAttributes());
             this.file = null;
-            this.link = Objects.requireNonNull(link);
+            this.link = link;
             this.directories = null;
             this.childNames = null;
         }
--- a/src/java.base/share/classes/jdk/internal/jrtfs/ExplodedImage.java
+++ b/src/java.base/share/classes/jdk/internal/jrtfs/ExplodedImage.java
@@ -120,7 +120,7 @@ class ExplodedImage extends SystemImage {
          * Used for each module-named directory that are leafs of /packages/...
          */
         private PathNode(String name, PathNode link) {
-            super(name, Objects.requireNonNull(link).getFileAttributes());
+            super(name, link.getFileAttributes());
             this.file = null;
             this.link = link;
             this.directories = null;
--- a/src/java.base/share/classes/jdk/internal/jrtfs/JrtFileSystem.java
+++ b/src/java.base/share/classes/jdk/internal/jrtfs/JrtFileSystem.java
@@ -78,7 +78,7 @@ class JrtFileSystem extends FileSystem {
     private final JrtFileSystemProvider provider;
     private final JrtPath rootPath = new JrtPath(this, "/");
     private volatile boolean isOpen;
-    private volatile boolean isClosable;
+    private final boolean isClosable;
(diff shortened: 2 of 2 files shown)


src/java.base/share/classes/jdk/internal/jrtfs/ExplodedImage.java
```diff
@@ -120,9 +120,9 @@ private PathNode(String name, List<Path> directories, BasicFileAttributes attrs)
          * Used for each module-named directory that are leafs of /packages/...
          */
         private PathNode(String name, PathNode link) {
-            super(name, link.getFileAttributes());
+            super(name, Objects.requireNonNull(link).getFileAttributes());
             this.file = null;
-            this.link = Objects.requireNonNull(link);
```
liach, line 125: Look at the other constructors: Notice they have `requireNonNull` as an intention for what kind of node is constructed and the different state contracts of different kinds of nodes.

jaikiran: I don't follow what you mean. The change proposed in this PR continues to have the explicit `requireNonNull` check for `link` parameter.

liach: For consistency, for the other constructors that do requireNonNull, I think we can move those rNN to be ahead of all the field assignments (including ones that write null). And we can completely drop rNN here.

AlanBateman:
@jaikiran Did you run into an issue/exception, just curious if this was drive-by/review fix or if it actually happened with an exploded build.

jaikiran:
Hello Alan, this came in as a review suggestion in the JEP-401 PR review https://github.com/openjdk/jdk/pull/31123#discussion_r3578999129

AlanBateman:
> Hello Alan, this came in as a review suggestion in the JEP-401 PR review [openjdk/jdk#31123 (comment)](https://github.com/openjdk/jdk/pull/31123#discussion_r3578999129)

Okay, so we don't have a bug in the jrtfs implementation that causes this NPE, instead it's that the Objects.requireNonNull is redundant.  What you have is okay, alternatively just remove the Objects.requireNonNull because link.getFileAttributes() will give a helpful NPE if we have a bug that causes it to be called with a null link.

jaikiran:
> alternatively just remove the Objects.requireNonNull because link.getFileAttributes() will give a helpful NPE if we have a bug that causes it to be called with a null link.

Done. I've updated this PR with this suggestion.

AlanBateman (approved):

jaikiran:
This has been addressed through https://github.com/openjdk/jdk/pull/32145.

PR #2651, opened 2026-07-17
8388177: [lworld] Remove "statewise equivalent" terminology from APIs

dlsmith-dev:
Some final polishing to align javadocs with the JEP.

- Main goal is to remove usage of "statewise equivalent", as noted in the title
- A little bit of general editing in the Object intro, `equals` method, and `clone` method
- Corrected "the finalize method of a value ~class~ _object_ will never be invoked"

---------

---------
### Progress

### Issue
 * [JDK-8388177](https://bugs.openjdk.org/browse/JDK-8388177): [lworld] Remove "statewise equivalent" terminology from APIs (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2651/head:pull/2651` \
`$ git checkout pull/2651`

`$ git checkout pull/2651` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2651/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2651`

`$ git pr show -t 2651`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2651.diff">https://git.openjdk.org/valhalla/pull/2651.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2651#issuecomment-4998502037)
</details>

### DIFF
--- a/src/java.base/share/classes/java/lang/Object.java
+++ b/src/java.base/share/classes/java/lang/Object.java
@@ -44,11 +44,10 @@ import jdk.internal.vm.annotation.IntrinsicCandidate;
  *          When preview features are disabled, all classes are identity classes and
  *          all objects are identity objects.
  *          <p>
- *          It is not possible to synchronize on a value object. Applying a {@code
- *          synchronize} statement to a value object causes an {@link IdentityException}
- *          to be thrown.
+ *          It is not possible to synchronize on a value object. An attempt to {@code
+ *          synchronize} on a value object causes {@link IdentityException} to be thrown.
  *          <p>
- *          The {@link #finalize()} method of a value object will never be invoked by
+ *          The {@link #finalize()} method of a value class will never be invoked by
  *          the garbage collector.
  *          <p>
  *          A {@linkplain java.lang.ref.Reference Reference object} can only refer to an
--- a/src/java.base/share/classes/java/lang/Object.java
+++ b/src/java.base/share/classes/java/lang/Object.java
@@ -41,16 +41,17 @@ import jdk.internal.vm.annotation.IntrinsicCandidate;
  *          other objects, including arrays, are <em>identity objects</em>. See
  *          The Java Language Specification {@jls value-objects-8.1.1.5 Value Classes}.
  *          <p>
- *          All classes are identity classes and all objects are identity objects when
- *          preview features are disabled.
+ *          When preview features are disabled, all classes are identity classes and
+ *          all objects are identity objects.
  *          <p>
- *          It is not possible to synchronize on a value object. An attempt to {@code
- *          synchronize} on a value object causes {@link IdentityException} to be thrown.
+ *          It is not possible to synchronize on a value object. Applying a {@code
+ *          synchronize} statement to a value object causes an {@link IdentityException}
+ *          to be thrown.
  *          <p>
- *          The {@link #finalize()} method of a value class will never be invoked by
+ *          The {@link #finalize()} method of a value object will never be invoked by
  *          the garbage collector.
  *          <p>
- *          A {@linkplain java.lang.ref.Reference Reference Object} can only refer to an
+ *          A {@linkplain java.lang.ref.Reference Reference object} can only refer to an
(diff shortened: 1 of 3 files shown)


src/java.base/share/classes/java/lang/Object.java
```diff
@@ -199,15 +200,18 @@ public boolean equals(Object obj) {
     /**
      * Creates and returns a copy of this object.  The precise meaning
      * of "copy" may depend on the class of the object. The general
-     * intent is that, for any identity object {@code x}, the expression:
+     * intent is that, for any object {@code x}, the expression:
```
dansmithcode, line 202: I started here: identity object clones should usually be !=, value object clones should usually be ==.

But then I read "It should be rare for new classes to implement the Cloneable interface", and I realized that _most_ value classes that implement `Cloneable` will do so because they need to clone a nested identity object. In that scenario, it will still be the case the `x.clone() != x`.

So I adjusted to _acknowledge_ the possibility of `==` value object clones without presenting it as typical.

src/java.base/share/classes/java/util/IdentityHashMap.java
```diff
@@ -50,9 +50,9 @@
  * </b>
  * <div class="preview-block">
  *      <div class="preview-comment">
- *          When preview features are enabled, keys and values may be identity objects,
- *          value objects or null.
- *          For value classes, two value objects are state-wise equivalent if they are instances
+ *          When preview features are enabled, keys and values may be
+ *          {@linkplain java.util.Objects#hasIdentity value objects}.
+ *          Two value object keys are {@code == } if they are instances
```
liach:
```suggestion
 *          Two value object keys are {@code ==} if they are instances
```

src/java.base/share/classes/java/lang/Object.java
```diff
@@ -41,16 +41,17 @@
...
- *          All classes are identity classes and all objects are identity objects when
- *          preview features are disabled.
+ *          When preview features are disabled, all classes are identity classes and
+ *          all objects are identity objects.
  *          <p>
- *          It is not possible to synchronize on a value object. An attempt to {@code
- *          synchronize} on a value object causes {@link IdentityException} to be thrown.
+ *          It is not possible to synchronize on a value object. Applying a {@code
+ *          synchronize} statement to a value object causes an {@link IdentityException}
+ *          to be thrown.
```
dholmes-ora: I think the original is preferable as the normal terminology is about "synchronizing on objects". I don't recall "applying a statement" to be common terminology with regards to statements.

src/java.base/share/classes/java/lang/Object.java
```diff
@@ -41,16 +41,17 @@
...
  *          <p>
- *          The {@link #finalize()} method of a value class will never be invoked by
+ *          The {@link #finalize()} method of a value object will never be invoked by
```
jaikiran: Hello Dan, I think changing this from value class to value object will be confusing. The Object.finalize() method specification (and implementation) continues to state:

> The {@code finalize} method of a value class is never directly invoked by the garbage collector. This includes the case where an abstract value class declares a {@code finalize} method and the class is extended by an identity class;

So it's not just value objects whose finalize don't get invoked but even identity objects whose super class is a value class with the finalize method.

dansmithcode: Ah, thanks, I noted that in the comments of JDK-8388161. I'll revert this, let's handle it there.

src/java.base/share/classes/java/util/IdentityHashMap.java
dansmithcode: Don't know why they were being written that way, but you're right, it renders better without the extra space. I'll fix them all.

src/java.base/share/classes/java/lang/Object.java
dansmithcode: ok, that's fine

liach (approved):
lgtm, thanks for the updates!

PR #2652, opened 2026-07-17
8388361: [lworld] C2 hits node limit during scalarization in safepoints

TobiHartmann:
During scalarization of an inline type in a large number of safepoints, we hit the node limit. This only happens in extreme cases where the test generates a large number of live locals and safepoints + `-XX:+DeoptimizeALot` which marks every local live at every safepoint. In the regression test, we have 60 safepoints, each one with 100 live Integers, and end up creating 6000 new `SafePointScalarObjectNode`.

The fix is to check the node budget and bail out from compilation if it's exceeded. Alternatively, we could fall back to `use_oop = true` if `allow_oop` is set and the inline type is allocated. However, I don't think that's worth it for this edge case.

Thanks,
Tobias

---------

---------
### Progress

### Issue
 * [JDK-8388361](https://bugs.openjdk.org/browse/JDK-8388361): [lworld] C2 hits node limit during scalarization in safepoints (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2652/head:pull/2652` \
`$ git checkout pull/2652`

`$ git checkout pull/2652` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2652/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2652`

`$ git pr show -t 2652`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2652.diff">https://git.openjdk.org/valhalla/pull/2652.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/opto/compile.cpp
+++ b/src/hotspot/share/opto/compile.cpp
@@ -2130,7 +2130,7 @@ void Compile::process_inline_types(PhaseIterGVN &igvn, bool remove) {
   for (int i = _inline_type_nodes.length()-1; i >= 0; i--) {
     InlineTypeNode* vt = _inline_type_nodes.at(i)->as_InlineType();
     if (!vt->make_scalar_in_safepoints(&igvn)) {
-      record_failure("Out of nodes during scalarization");
+      record_failure("out of nodes during scalarization");
       return;
     }
     igvn.record_for_igvn(vt);
--- a/src/hotspot/share/opto/compile.cpp
+++ b/src/hotspot/share/opto/compile.cpp
@@ -2129,7 +2129,10 @@ void Compile::process_inline_types(PhaseIterGVN &igvn, bool remove) {
   set_scalarize_in_safepoints(true);
   for (int i = _inline_type_nodes.length()-1; i >= 0; i--) {
     InlineTypeNode* vt = _inline_type_nodes.at(i)->as_InlineType();
-    vt->make_scalar_in_safepoints(&igvn);
+    if (!vt->make_scalar_in_safepoints(&igvn)) {
+      record_failure("Out of nodes during scalarization");
+      return;
+    }
     igvn.record_for_igvn(vt);
   }
   if (remove) {
@@ -3106,6 +3109,7 @@ void Compile::Optimize() {
 
   // Process inline type nodes now that all inlining is over
   process_inline_types(igvn);
+  if (failing()) return;
 
   adjust_flat_array_access_aliases(igvn);
 
@@ -3292,6 +3296,7 @@ void Compile::Optimize() {
   // Process inline types before macro expansion. Otherwise, we will not be able to
   // remove unused allocations because it cannot match the expanded allocation.
   process_inline_types(igvn);
+  if (failing()) return;
 
   {
     TracePhase tp(_t_macroExpand);
@@ -3325,6 +3330,7 @@ void Compile::Optimize() {
   // Process inline type nodes again and remove them. From here
   // on we don't need to keep track of field values anymore.
   process_inline_types(igvn, /* remove= */ true);
+  if (failing()) return;
 
   {
     TracePhase tp(_t_barrierExpand);
--- a/src/hotspot/share/opto/inlinetypenode.hpp
+++ b/src/hotspot/share/opto/inlinetypenode.hpp
@@ -59,7 +59,7 @@ private:
   // Get the klass defining the field layout of the inline type
   ciInlineKlass* inline_klass() const { return type()->inline_klass(); }
 
-  void make_scalar_in_safepoint(PhaseIterGVN* igvn, Unique_Node_List& worklist, SafePointNode* sfpt) const;
+  bool make_scalar_in_safepoint(PhaseIterGVN* igvn, Unique_Node_List& worklist, SafePointNode* sfpt) const;
   uint add_fields_to_safepoint(Unique_Node_List& worklist, SafePointNode* sfpt) const;
 
   // Checks if the inline type is loaded from memory and if so returns the oop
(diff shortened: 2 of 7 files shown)


src/hotspot/share/opto/compile.cpp
```diff
@@ -2129,7 +2129,10 @@ void Compile::process_inline_types(PhaseIterGVN &igvn, bool remove) {
   set_scalarize_in_safepoints(true);
   for (int i = _inline_type_nodes.length()-1; i >= 0; i--) {
     InlineTypeNode* vt = _inline_type_nodes.at(i)->as_InlineType();
-    vt->make_scalar_in_safepoints(&igvn);
+    if (!vt->make_scalar_in_safepoints(&igvn)) {
+      record_failure("Out of nodes during scalarization");
```
marc-chevalier: nit: do we put caps for these error messsages?

src/hotspot/share/opto/inlinetypenode.hpp
```diff
@@ -122,9 +122,9 @@ class InlineTypeNode : public TypeNode {
   uint          field_index(int offset) const;
 
   // Replace InlineTypeNodes in debug info at safepoints with SafePointScalarObjectNodes
-  void make_scalar_in_safepoints(PhaseIterGVN* igvn, bool allow_oop = true);
+  bool make_scalar_in_safepoints(PhaseIterGVN* igvn, bool allow_oop = true);
```
marc-chevalier: It's a bit unfortunate we don't use stuff like `[[nodiscard]] ` to easily catch everywhere we need to change.

test/hotspot/jtreg/compiler/valhalla/inlinetypes/TestSafepointScalarizationNodeLimit.java
```diff
@@ -0,0 +1,129 @@
...
+            i60 = x + 60, i61 = x + 61, i62 = x + 62, i63 = x + 63, i64 = x + 64,
+            i65 = x + 65, i66 = x + 66, i67 = x + 67, i68 = x + 68, i69 = x + 69,
+            i70 = x + 70, i71 = x + 71, i72 = x + 72, i73 = x + 73, i74 = x + 74,
+            i75 = x + 75, i76 = x + 76, i77 = x + 77, i78 = x + 78, i79 = x + 79,
+            i80 = x + 80, i81 = x + 81, i82 = x + 82, i83 = x + 83, i84 = x + 84,
+            i85 = x + 85, i86 = x + 86, i87 = x + 87, i88 = x + 88, i89 = x + 89,
+            i90 = x + 90, i91 = x + 91, i92 = x + 92, i93 = x + 93, i94 = x + 94,
+            i95 = x + 95, i96 = x + 96, i97 = x + 97, i98 = x + 98, i99 = x + 99;
+
+        // 30 x 30 = 90 safepoints, each one with 100 live Integers
```
marc-chevalier: 30 * 30 = 90 ? Almost!

test/hotspot/jtreg/compiler/valhalla/inlinetypes/TestSafepointScalarizationNodeLimit.java
```diff
@@ -0,0 +1,129 @@
...
+            i60 = x + 60, i61 = x + 61, i62 = x + 62, i63 = x + 63, i64 = x + 64,
+            i65 = x + 65, i66 = x + 66, i67 = x + 67, i68 = x + 68, i69 = x + 69,
+            i70 = x + 70, i71 = x + 71, i72 = x + 72, i73 = x + 73, i74 = x + 74,
+            i75 = x + 75, i76 = x + 76, i77 = x + 77, i78 = x + 78, i79 = x + 79,
+            i80 = x + 80, i81 = x + 81, i82 = x + 82, i83 = x + 83, i84 = x + 84,
+            i85 = x + 85, i86 = x + 86, i87 = x + 87, i88 = x + 88, i89 = x + 89,
+            i90 = x + 90, i91 = x + 91, i92 = x + 92, i93 = x + 93, i94 = x + 94,
+            i95 = x + 95, i96 = x + 96, i97 = x + 97, i98 = x + 98, i99 = x + 99;
+
+        // 20 x 30 = 60 safepoints, each one with 100 live Integers
```
marc-chevalier: Also surprising arithmetic.

marc-chevalier (commented):
Otherwise, the fix looks like the right thing to do, and it does it right.

src/hotspot/share/opto/compile.cpp
TobiHartmann: I think you are right, other places use lowercase. I'll update it accordingly.

src/hotspot/share/opto/inlinetypenode.hpp
TobiHartmann: Right, I agree.

test/hotspot/jtreg/compiler/valhalla/inlinetypes/TestSafepointScalarizationNodeLimit.java
TobiHartmann: Embarrassing :slightly_frowning_face: Fixed.

test/hotspot/jtreg/compiler/valhalla/inlinetypes/TestSafepointScalarizationNodeLimit.java
TobiHartmann: Thanks, fixed as well.

TobiHartmann:
Thanks for your review Marc! I addressed all your comments.

src/hotspot/share/opto/inlinetypenode.hpp
marc-chevalier: Don't we use C++17? I guess we could? Or anything says we shouldn't? I often refrain, but maybe there is no basis for that?

TobiHartmann: Oh and we even already use it in HotSpot code! I added it.

marc-chevalier: OH! Guess who is going on a nodiscard frenzy...

fparain (approved):
Changes make sense to me.
I agree that bailing out is the right decision at this point, it can be revisited if we receive reports of this policy causing issues in production environment.

TobiHartmann:
Thanks for the review Fred!

TobiHartmann:
Re-opened for mainline: https://github.com/openjdk/jdk/pull/32127

PR #2653, opened 2026-07-17
8386769: [lworld] C2 parts of JDK-8350865 - Part. 2

marc-chevalier:
# PhaseIterGVN::add_users_of_use_to_worklist / PhaseCCP::push_cast

https://github.com/openjdk/valhalla/blob/53ef1bfa0951e5cc1eeb064026eb0d73329e7c5e/src/hotspot/share/opto/phaseX.cpp#L2819
https://github.com/openjdk/valhalla/blob/53ef1bfa0951e5cc1eeb064026eb0d73329e7c5e/src/hotspot/share/opto/phaseX.cpp#L3236

I think that is needed, it's not trivial to make fail without it, but reading code, it must be useful.

# ciTypeFlow::StateVector::do_aload

https://github.com/openjdk/valhalla/blob/53ef1bfa0951e5cc1eeb064026eb0d73329e7c5e/src/hotspot/share/ci/ciTypeFlow.cpp#L603

I've did that, I think.

# ciReplay.cpp

https://github.com/openjdk/valhalla/blob/53ef1bfa0951e5cc1eeb064026eb0d73329e7c5e/src/hotspot/share/ci/ciReplay.cpp#L517

For that, I just check objArray, not only refArray. Getting from a objArray does the dispatch. It shouldn't be harder than that.

https://github.com/openjdk/valhalla/blob/53ef1bfa0951e5cc1eeb064026eb0d73329e7c5e/src/hotspot/share/ci/ciReplay.cpp#L1143-L1145

This one is less direct. So now, I dump the flatness of the array as `ref` or `flat`, and the nullability (`nullable` or `null-free`). After that, if the array is flat, I add atomicity (`atomic` or `non-atomic`). On parsing, I get all these pieces into a an `ArrayProperties` given to the `oopFactory::new_*Array`.

# TestIntrinsics.java

https://github.com/openjdk/valhalla/blob/53ef1bfa0951e5cc1eeb064026eb0d73329e7c5e/test/hotspot/jtreg/compiler/valhalla/inlinetypes/TestIntrinsics.java#L2038-L2040
https://github.com/openjdk/valhalla/blob/53ef1bfa0951e5cc1eeb064026eb0d73329e7c5e/test/hotspot/jtreg/compiler/valhalla/inlinetypes/TestIntrinsics.java#L2056-L2058

That was already done, it seems:

https://github.com/openjdk/valhalla/blob/53ef1bfa0951e5cc1eeb064026eb0d73329e7c5e/src/hotspot/share/opto/library_call.cpp#L4907-L4910

# TestFlatInArraysFolding.java

https://github.com/openjdk/valhalla/blob/53ef1bfa0951e5cc1eeb064026eb0d73329e7c5e/test/hotspot/jtreg/compiler/valhalla/inlinetypes/TestFlatInArraysFolding.java#L81
https://github.com/openjdk/valhalla/blob/53ef1bfa0951e5cc1eeb064026eb0d73329e7c5e/test/hotspot/jtreg/compiler/valhalla/inlinetypes/TestFlatInArraysFolding.java#L168
https://github.com/openjdk/valhalla/blob/53ef1bfa0951e5cc1eeb064026eb0d73329e7c5e/test/hotspot/jtreg/compiler/valhalla/inlinetypes/TestFlatInArraysFolding.java#L285

Since the important part is that classes are more of less flattable on a setting, I've tried to replace `FlatArrayElementMaxSize` with `FlatArrayElementMaxOops`.

# TestBasicFunctionality.java

https://github.com/openjdk/valhalla/blob/53ef1bfa0951e5cc1eeb064026eb0d73329e7c5e/test/hotspot/jtreg/compiler/valhalla/inlinetypes/TestBasicFunctionality.java#L578-L581

It looks like it works now, but it should be 3, not 2. I've commented why, and I think it's behaving as it should.

# TestLWorld.java

https://github.com/openjdk/valhalla/blob/53ef1bfa0951e5cc1eeb064026eb0d73329e7c5e/test/hotspot/jtreg/compiler/valhalla/inlinetypes/TestLWorld.java#L4646

I've quadruplicated the test for different kind of factories. I've also made sure the flatness hypotheses are still matched with nullable arrays, so I needed to shrink the classes.

# SubTypeCheckNode::sub

https://github.com/openjdk/valhalla/blob/53ef1bfa0951e5cc1eeb064026eb0d73329e7c5e/src/hotspot/share/opto/subtypenode.cpp#L50

I think it's good actually. We are already using that inside `Compile::static_subtype_check`, called a bit lower, but we just need a bit of preparation. We could put that in a method of the type system, but for which benefit? I suggest we keep it as it is for now

# Parse::array_store

We could, but we didn't do. It would only matter when the array element has no null-free layout (atomic or non-atomic), but the array is not provably non-flat, which means, there is a nullable layout. That seems unlikely at the very least, or rather impossible in the current state of things.

<details>
<summary>The experimental patch, for record</summary>

```diff
diff --git a/src/hotspot/share/opto/parse2.cpp b/src/hotspot/share/opto/parse2.cpp
index 1b0b02c53ee..7511237a28f 100644
--- a/src/hotspot/share/opto/parse2.cpp
+++ b/src/hotspot/share/opto/parse2.cpp
@@ -253,24 +253,41 @@ void Parse::array_store(BasicType bt) {
       // Array might be a flat array, emit runtime checks (for null, a simple inline_array_null_guard is sufficient).
       assert(UseArrayFlattening && !not_flat && elemtype->is_oopptr()->can_be_inline_type() &&
              (!array_type->klass_is_exact() || array_type->is_flat()), "array can't be a flat array");
-      // TODO 8350865 Depending on the available layouts, we can avoid this check in below flat/not-flat branches. Also the safe_for_replace arg is now always true.
-      array = inline_array_null_guard(array, stored_value_casted, 3, true);
+      // If the array cannot be flat and null-free, we don't need to do inline_array_null_guard
+      // in the flat branch under. In this case, the flat branch is thus a bit faster, but the
+      // non-flat branch always need to be protected. If the array can be both flat and null-free,
+      // we just do the inline_array_null_guard guard the flat/non-flat branching so we don't
+      // duplicate it.
+      bool can_be_flat_and_null_free =
+           !elemtype->is_inlinetypeptr()  // Can't tell statically whether it's possible, so we must soundly assume so
+        || elemtype->inline_klass()->has_null_free_atomic_layout()
+        || elemtype->inline_klass()->has_null_free_non_atomic_layout();
+      if (can_be_flat_and_null_free) {
+        array = inline_array_null_guard(array, stored_value_casted, 3, true);
+        array_type = _gvn.type(array)->is_aryptr();
+      }
       // Reload array type which could have been updated by inline_array_null_guard().
-      array_type = _gvn.type(array)->is_aryptr();
       IdealKit ideal(this);
       ideal.if_then(flat_array_test(array, /* flat = */ false)); {
         // Non-flat array
         if (!array_type->is_flat()) {
           sync_kit(ideal);
-          assert(array_type->is_not_flat() || ideal.ctrl()->in(0)->as_If()->is_flat_array_check(&_gvn), "Should be found");
+          Node* array_ = array;
+          const TypeAryPtr* array_type_ = array_type;
+          if (!can_be_flat_and_null_free) {
+            array_ = inline_array_null_guard(array_, stored_value_casted, 3, false);
+            array_type_ = _gvn.type(array_)->is_aryptr();
+          }
+          assert(array_type_->is_not_flat() || ideal.ctrl()->in(0)->as_If()->is_flat_array_check(&_gvn), "Should be found");
           inc_sp(3);
-          access_store_at(array, adr, adr_type, stored_value_casted, elemtype, bt, MO_UNORDERED | IN_HEAP | IS_ARRAY, false);
+          access_store_at(array_, adr, adr_type, stored_value_casted, elemtype, bt, MO_UNORDERED | IN_HEAP | IS_ARRAY, false);
           dec_sp(3);
           ideal.sync_kit(this);
         }
       } ideal.else_(); {
         // Flat array
         sync_kit(ideal);
+        // Either array can be null-free and flat, and inline_array_null_guard is done above, or it cannot, and we don't need to perform this check.
         if (!array_type->is_not_flat()) {
           // Try to determine the inline klass type of the stored value
           ciInlineKlass* vk = nullptr;
```
</details>

At least, we can remove the last parametert of `inline_array_null_guard` since we are not doing this change.

# GraphKit::null_free_atomic_array_test

https://github.com/openjdk/valhalla/blob/53ef1bfa0951e5cc1eeb064026eb0d73329e7c5e/src/hotspot/share/opto/graphKit.cpp#L3985

It's a comment from a time with less testing. Let's not add one more

# LibraryCallKit::inline_unsafe_flat_access

This seems to be a bigger piece. I've filed [JDK-8388444](https://bugs.openjdk.org/browse/JDK-8388444) for now.

Thanks,
Marc

---------

---------
### Progress

### Issue
 * [JDK-8386769](https://bugs.openjdk.org/browse/JDK-8386769): [lworld] C2 parts of JDK-8350865 - Part. 2 (**Sub-task** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2653/head:pull/2653` \
`$ git checkout pull/2653`

`$ git checkout pull/2653` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2653/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2653`

`$ git pr show -t 2653`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2653.diff">https://git.openjdk.org/valhalla/pull/2653.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/ci/ciInstanceKlass.cpp
+++ b/src/hotspot/share/ci/ciInstanceKlass.cpp
@@ -863,8 +863,21 @@ void StaticFieldPrinter::do_field_helper(fieldDescriptor* fd, oop mirror, bool i
       break;
     }
     case T_ARRAY:  // fall-through
-    case T_OBJECT:
-      if (!fd->is_null_free_inline_type()) {
+    case T_OBJECT: {
+      InstanceKlass* k = nullptr;
+      ResetNoHandleMark rnhm;
+      Thread* THREAD = Thread::current();
+      SignatureStream ss(fd->signature(), false);
+      Symbol* name = ss.as_symbol();
+      assert(!HAS_PENDING_EXCEPTION, "can resolve klass?");
+      InstanceKlass* holder = fd->field_holder();
+      if (field_type == T_OBJECT) {
+        k = SystemDictionary::find_instance_klass(THREAD, name, Handle(THREAD, holder->class_loader()));
+        // handling of null free inline type
+        guarantee(k != nullptr && !HAS_PENDING_EXCEPTION, "can resolve klass?");
+      }
+
+      if (k == nullptr || !k->is_inline_klass()) {
         _out->print("%s ", fd->signature()->as_quoted_ascii());
         oop value =  mirror->obj_field_acquire(fd->offset());
         if (value == nullptr) {
@@ -886,6 +899,23 @@ void StaticFieldPrinter::do_field_helper(fieldDescriptor* fd, oop mirror, bool i
           _out->print("%d", a->length());
           if (value->is_objArray()) {
             objArrayOop oa = (objArrayOop)value;
+            if (value->is_flatArray()) {
+              FlatArrayKlass* klass = ((flatArrayOop)oa)->klass();
+              LayoutKind lk = klass->layout_kind();
+              _out->print(" flat");
+              if (LayoutKindHelper::is_atomic_flat(lk)) {
+                _out->print(" atomic");
+              } else {
+                _out->print(" non-atomic");
+              }
+              if (LayoutKindHelper::is_nullable_flat(lk)) {
+                _out->print(" nullable");
+              } else {
+                _out->print(" null-free");
+              }
+            } else {
+              _out->print(" ref");
+            }
             const char* klass_name  = value->klass()->name()->as_quoted_ascii();
             _out->print(" %s", klass_name);
           }
@@ -894,16 +924,6 @@ void StaticFieldPrinter::do_field_helper(fieldDescriptor* fd, oop mirror, bool i
         }
         break;
       } else {
-        // handling of null free inline type
-        ResetNoHandleMark rnhm;
-        Thread* THREAD = Thread::current();
-        SignatureStream ss(fd->signature(), false);
-        Symbol* name = ss.as_symbol();
-        assert(!HAS_PENDING_EXCEPTION, "can resolve klass?");
-        InstanceKlass* holder = fd->field_holder();
-        InstanceKlass* k = SystemDictionary::find_instance_klass(THREAD, name,
-                                                                 Handle(THREAD, holder->class_loader()));
-        guarantee(k != nullptr && !HAS_PENDING_EXCEPTION, "can resolve klass?");
         InlineKlass* vk = InlineKlass::cast(k);
         oop obj;
         if (is_flat) {
@@ -916,6 +936,7 @@ void StaticFieldPrinter::do_field_helper(fieldDescriptor* fd, oop mirror, bool i
         vk->do_nonstatic_fields(&print_field);
         break;
       }
+    }
     default:
       ShouldNotReachHere();
   }
--- a/src/hotspot/share/ci/ciInstanceKlass.cpp
+++ b/src/hotspot/share/ci/ciInstanceKlass.cpp
@@ -863,21 +863,8 @@ void StaticFieldPrinter::do_field_helper(fieldDescriptor* fd, oop mirror, bool i
       break;
     }
     case T_ARRAY:  // fall-through
-    case T_OBJECT: {
-      InstanceKlass* k = nullptr;
-      ResetNoHandleMark rnhm;
-      Thread* THREAD = Thread::current();
-      SignatureStream ss(fd->signature(), false);
-      Symbol* name = ss.as_symbol();
-      assert(!HAS_PENDING_EXCEPTION, "can resolve klass?");
-      InstanceKlass* holder = fd->field_holder();
-      if (field_type == T_OBJECT) {
-        k = SystemDictionary::find_instance_klass(THREAD, name, Handle(THREAD, holder->class_loader()));
-        // handling of null free inline type
-        guarantee(k != nullptr && !HAS_PENDING_EXCEPTION, "can resolve klass?");
-      }
-
-      if (k == nullptr || !k->is_inline_klass()) {
+    case T_OBJECT:
+      if (!fd->is_null_free_inline_type()) {
         _out->print("%s ", fd->signature()->as_quoted_ascii());
         oop value =  mirror->obj_field_acquire(fd->offset());
         if (value == nullptr) {
@@ -899,28 +886,6 @@ void StaticFieldPrinter::do_field_helper(fieldDescriptor* fd, oop mirror, bool i
           _out->print("%d", a->length());
           if (value->is_objArray()) {
             objArrayOop oa = (objArrayOop)value;
-            if (value->is_flatArray()) {
-              FlatArrayKlass* klass = ((flatArrayOop)oa)->klass();
-              LayoutKind lk = klass->layout_kind();
-              _out->print(" flat");
-              if (LayoutKindHelper::is_nullable_flat(lk)) {
-                _out->print(" nullable");
-              } else {
-                _out->print(" null-free");
-              }
-              if (LayoutKindHelper::is_atomic_flat(lk)) {
-                _out->print(" atomic");
-              } else {
-                _out->print(" non-atomic");
-              }
-            } else {
-              _out->print(" ref");
-              if (oa->klass()->is_null_free_array_klass()) {
-                _out->print(" null-free");
-              } else {
-                _out->print(" nullable");
-              }
-            }
             const char* klass_name  = value->klass()->name()->as_quoted_ascii();
             _out->print(" %s", klass_name);
           }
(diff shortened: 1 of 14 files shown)


src/hotspot/share/ci/ciInstanceKlass.cpp
```diff
@@ -863,8 +863,21 @@ void StaticFieldPrinter::do_field_helper(fieldDescriptor* fd, oop mirror, bool i
...
+      InstanceKlass* k = nullptr;
+      ResetNoHandleMark rnhm;
+      Thread* THREAD = Thread::current();
+      SignatureStream ss(fd->signature(), false);
+      Symbol* name = ss.as_symbol();
+      assert(!HAS_PENDING_EXCEPTION, "can resolve klass?");
+      InstanceKlass* holder = fd->field_holder();
+      if (field_type == T_OBJECT) {
+        k = SystemDictionary::find_instance_klass(THREAD, name, Handle(THREAD, holder->class_loader()));
+        // handling of null free inline type
```
TobiHartmann: But now we do a lookup for the type of each field, not only null-free ones, right? And wouldn't it be expected for the lookup to fail, for example, for static fields of uninitialized type?

src/hotspot/share/ci/ciInstanceKlass.cpp
```diff
@@ -886,6 +899,28 @@ void StaticFieldPrinter::do_field_helper(fieldDescriptor* fd, oop mirror, bool i
           _out->print("%d", a->length());
           if (value->is_objArray()) {
             objArrayOop oa = (objArrayOop)value;
+            if (value->is_flatArray()) {
+              FlatArrayKlass* klass = ((flatArrayOop)oa)->klass();
+              LayoutKind lk = klass->layout_kind();
+              _out->print(" flat");
```
TobiHartmann: I think these new properties require a `REPLAY_VERSION` bump.

src/hotspot/share/ci/ciInstanceKlass.cpp
TobiHartmann, line 915: Do we correctly handle `obj == nullptr` here if the field is null? Before, we would have taken the `if (!fd->is_null_free_inline_type())` branch above.

src/hotspot/share/ci/ciReplay.cpp
```diff
@@ -1234,17 +1247,22 @@ class CompileReplay : public StackObj {
...
-      bool res = process_staticfield_reference(field_signature, java_mirror, &fd, CHECK);
-      if (!res)  {
-        report_error("unhandled staticfield");
+      Klass* kelem = nullptr;
+      if (fd.field_type() == T_OBJECT) {
+        kelem = resolve_klass(field_signature, CHECK);
+      }
+      if (kelem != nullptr && kelem->is_inline_klass()) {
+        InlineKlass* vk = InlineKlass::cast(kelem);
+        oop value = vk->allocate_instance(CHECK);
```
TobiHartmann: What if the field is nullable and the value is null?

src/hotspot/share/ci/ciTypeFlow.cpp
```diff
@@ -600,15 +600,18 @@ void ciTypeFlow::StateVector::do_aload(ciBytecodeStream* str) {
...
   if (!element_klass->is_loaded() && element_klass->is_instance_klass()) {
     Untested("unloaded array element class in ciTypeFlow");
     trap(str, element_klass,
          Deoptimization::make_trap_request
          (Deoptimization::Reason_unloaded,
           Deoptimization::Action_reinterpret));
   } else {
-    push_object(element_klass);
+    ciType* maybe_null_free_element_klass = element_klass;
+    if (array_klass->is_null_free()) {
```
TobiHartmann: But this is checking `ciType::is_null_free` which is only ever true for `ciWrapper::is_null_free` and this is used for value objects, not for arrays? And even if the array would never be null, this does not imply that the elements are never null, right?

TobiHartmann: Maybe this should be something like `array_klass->is_refined() && array_klass->is_elem_null_free()`

TobiHartmann (commented):
Thanks a lot for working on this Marc! The changes look good except for ciReplay and ciTypeFlow. I added some comments there.

marc-chevalier:
The replay part will be done separately, it's a bigger fish than expected. Thus, it has been removed from here.

TobiHartmann (approved):
That looks good to me. Thanks a lot for working on this Marc!

marc-chevalier:
Moved to mainline: https://github.com/openjdk/jdk/pull/32146

PR #2660, opened 2026-07-21
8388544: [lworld] Bad dominance assert in C2 when late-inlining through MH calls

TobiHartmann:
When late-inlining a method handle call that returns a value object in scalarized form, we need to ..

https://github.com/openjdk/valhalla/blob/1c3301cb96d779d06bb6e2391c35ce729bc28189/src/hotspot/share/opto/callGenerator.cpp#L779-L788

We then update the oop input of the returned InlineTypeNode to the newly allocated buffer:
https://github.com/openjdk/valhalla/blob/1c3301cb96d779d06bb6e2391c35ce729bc28189/src/hotspot/share/opto/callGenerator.cpp#L873-L874

This is incorrect since the returned InlineTypeNode is shared and can be used further up in the graph. The new test demonstrates where this will lead to a def-after-use dominance violation.

The fix is to clone the InlineTypeNode if required before updating the oop input, similar to what we do already in normal buffering:
https://github.com/openjdk/valhalla/blob/1c3301cb96d779d06bb6e2391c35ce729bc28189/src/hotspot/share/opto/inlinetypenode.cpp#L1071-L1073

Thanks,
Tobias

---------

---------
### Progress

### Issue
 * [JDK-8388544](https://bugs.openjdk.org/browse/JDK-8388544): [lworld] Bad dominance assert in C2 when late-inlining through MH calls (**Bug** - P3)

### Reviewers
 * [Quan Anh Mai](https://openjdk.org/census#qamai) (@merykitty - Committer) ⚠️ Review applies to [d99c704e](https://git.openjdk.org/valhalla/pull/2660/files/d99c704efadadb7c09c17b927523b4032ee0a207)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2660/head:pull/2660` \
`$ git checkout pull/2660`

`$ git checkout pull/2660` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2660/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2660`

`$ git pr show -t 2660`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2660.diff">https://git.openjdk.org/valhalla/pull/2660.diff</a>

</details>


marc-chevalier (commented):
Looks convincing to me, but I'm not sure enough of my judgement. Why does it happen only with method handles? Morally, I don't see what is unique to method handles in the description of the problem.

TobiHartmann:
Thanks for the review Marc!

> Why does it happen only with method handles? Morally, I don't see what is unique to method handles in the description of the problem.

All of this code is specific to method handle inlining:
https://github.com/openjdk/valhalla/blob/db17ab61eafb2881a5e9769300a5bd40f415ef09/src/hotspot/share/opto/callGenerator.cpp#L777-L781

For a normal call, the caller and the callee agree on the return convention. If the callee returns a value object as fields, the caller also expects those fields. Any conversion to Object happens separately through the normal buffering code.

With method handles, C2 inlines right *through* the `invokeBasic/linkTo*` layer into the actual target, i.e. basically two levels of inlining at once. The MH call can have a signature returning Object, while the target returns a value object in scalarized form. The late-inlining code therefore has to associate the returned InlineTypeNode with a (per-allocated) buffer expected by the original caller.

For example:
```
value class Point {
    int x, y;
    ...
}

static Point make() {
    return new Point(1, 2);
}

static final MethodHandle MH = LOOKUP.findStatic(MyTest.class, "make", methodType(Point.class)).asType(methodType(Object.class));

static Object test() throws Throwable {
    return MH.invokeExact();
}
```

Before inlining, the MH call returns `Object` and its users expect an oop. After resolving and inlining `make`, its result can be `Point` in scalarized form.

Below are some more code pointers for future reference.

For `invokeBasic`, C2 extracts the real target from the constant `MethodHandle` via `get_vmtarget()`:
https://github.com/openjdk/valhalla/blob/db17ab61eafb2881a5e9769300a5bd40f415ef09/src/hotspot/share/opto/callGenerator.cpp#L1217-L1238

For `linkTo*`, it extracts the target from the trailing constant `MemberName`:

https://github.com/openjdk/valhalla/blob/db17ab61eafb2881a5e9769300a5bd40f415ef09/src/hotspot/share/opto/callGenerator.cpp#L1249-L1264

We then asks `C->call_generator(target, ..., jvms, ...)` for a generator for that target:
https://github.com/openjdk/valhalla/blob/db17ab61eafb2881a5e9769300a5bd40f415ef09/src/hotspot/share/opto/callGenerator.cpp#L1323-L1328

merykitty (approved):
Thanks, it looks reasonable to me.

TobiHartmann:
Thanks for the review Quan Anh!

TobiHartmann:
Re-opened for mainline: https://github.com/openjdk/jdk/pull/32129

PR #2661, opened 2026-07-21
8388318: [lworld] jdk.internal.value.Deserializer annotation should not be honored on application classes

liach:
The Deserializer documentation states that it only applies to the boot class path:

https://github.com/openjdk/valhalla/blob/1c3301cb96d779d06bb6e2391c35ce729bc28189/src/java.base/share/classes/jdk/internal/value/Deserializer.java#L58-L59

There is an implementation defect right now that this annotation is scanned from all sources.

Updated `ValueSerializationTest` to add positive and negative tests for a value class with `@Deserializer` both on and off the boot class path.

In the test updates I noticed `SerializedObjectCombo` is broken that it never generates value classes somehow; tracked in [JDK-8388597](https://bugs.openjdk.org/browse/JDK-8388597).

---------

---------
### Progress

### Issue
 * [JDK-8388318](https://bugs.openjdk.org/browse/JDK-8388318): [lworld] jdk.internal.value.Deserializer annotation should not be honored on application classes (**Bug** - P2)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2661/head:pull/2661` \
`$ git checkout pull/2661`

`$ git checkout pull/2661` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2661/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2661`

`$ git pr show -t 2661`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2661.diff">https://git.openjdk.org/valhalla/pull/2661.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2661#issuecomment-5036778644)
</details>

### DIFF
--- a/src/java.base/share/classes/java/io/ObjectStreamClass.java
+++ b/src/java.base/share/classes/java/io/ObjectStreamClass.java
@@ -1362,6 +1362,10 @@ public final class ObjectStreamClass implements Serializable {
      */
     private static Executable findDeserializer(Class<?> clazz,
                                                ObjectStreamField[] fields) {
+        if (clazz.getClassLoader() != null) {
+            // Only for boot loader classes
+            return null;
+        }
         return Stream.concat(
                 Arrays.stream(clazz.getDeclaredMethods()).filter(m -> Modifier.isStatic(m.getModifiers())),
                 Arrays.stream(clazz.getDeclaredConstructors()))
--- a/test/jdk/java/io/Serializable/valueObjects/SimpleValueGraphs.java
+++ b/test/jdk/java/io/Serializable/valueObjects/SimpleValueGraphs.java
@@ -28,7 +28,7 @@
  * @enablePreview
  * @modules java.base/jdk.internal
  * @modules java.base/jdk.internal.value
- * @run junit/othervm SimpleValueGraphs
+ * @run junit/bootclasspath/othervm SimpleValueGraphs
  */
 
 import java.io.ByteArrayInputStream;
--- a/test/jdk/java/io/Serializable/valueObjects/ValueSerializationTest.java
+++ b/test/jdk/java/io/Serializable/valueObjects/ValueSerializationTest.java
@@ -23,6 +23,7 @@
 
 /*
  * @test
+ * @bug 8388318
  * @summary Test serialization of value classes
  * @enablePreview
  * @modules java.base/jdk.internal.value
@@ -33,7 +34,8 @@
  *          generate classfiles with STRICT_INIT access flags for its annotated fields
  * @run driver jdk.test.lib.helpers.StrictProcessor
  *             ValueSerializationTest$IdentityStrictPoint
- * @run junit ${test.main.class}
+ * @run junit/othervm -DdeserializerOnBootclasspath=false ${test.main.class}
+ * @run junit/bootclasspath/othervm -DdeserializerOnBootclasspath=true ${test.main.class}
  */
 
 import java.io.ByteArrayInputStream;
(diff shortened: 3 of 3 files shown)


liach:
Ran Oracle's tier 1-6, no related or new failures.

liach:
In offline discussion, @jaikiran argues for restriction of this to just java.lang wrapper classes; if we do apply such restrictions, I would prefer to hardcode in serialization instead of keeping the annotation and its scanning. Meanwhile @AlanBateman argues for keeping the boot restriction for now, with the ability to expand to platform loader in case other jdk modules need this migration tool.

jaikiran:
The change here proposes to ignore `@Deserializer` annotation from classes that aren't loaded by the boot classloader. I was in agreement with that change until I saw the test in this PR. The test illustrates a scenario where, even after the proposed fix in this PR, if some application-provided class with a `@Deserializer` annotation is placed on the bootclasspath, then the JDK's legacy deserializer will continue to honour that annotation. I think we shouldn't allow that. So what I proposed was something like the following (untested) change which very specifically only allows this annotation on the primitive wrapper classes:

```diff
diff --git a/src/java.base/share/classes/java/io/ObjectStreamClass.java b/src/java.base/share/classes/java/io/ObjectStreamClass.java
@@ -1362,6 +1362,21 @@ private ObjectStreamClass getVariantFor(Class<?> cl)
      */
     private static Executable findDeserializer(Class<?> clazz,
                                                ObjectStreamField[] fields) {
+        // The JDK allows the @Deserializer annotation to be present only on
+        // some specific classes. Currently it's only allowed on the primitive
+        // wrapper classes. We ignore @Deserializer on everything else.
+        if (clazz != Integer.class
+                && clazz != Long.class
+                && clazz != Short.class
+                && clazz != Double.class
+                && clazz != Float.class
+                && clazz != Byte.class
+                && clazz != Boolean.class
+                && clazz != Character.class) {
+            // not a primitive wrapper class, thus no @Deserializer applicable
+            return null;
+        }
+        // class is eligible for having a @Deserializer, find if it has any
         return Stream.concat(
                 Arrays.stream(clazz.getDeclaredMethods()).filter(m -> Modifier.isStatic(m.getModifiers())),
                 Arrays.stream(clazz.getDeclaredConstructors()))

```
I think this will allow us to not think of scenarios where someone sets up a class on the bootclasspath with a `@Deserializer` annotation.

> if we do apply such restrictions, I would prefer to hardcode in serialization instead of keeping the annotation and its scanning.

That's going to be a much bigger change and I think we should avoid it. Having the current `@Deserializer` infrastructure in place is good. It's just that we want to restrict its usage to a very select few classes.

AlanBateman:
> Meanwhile @AlanBateman argues for keeping the boot restriction for now, with the ability to expand to platform loader in case other jdk modules need this migration tool.

I only pointing out that some of the standard/JDK modules are mapped to the platform class loader.

Just to Jai's concern, restricting it to a named module mapped to the boot loader will remove the concern with -Xbootclasspath/a and the agent equivalent.

liach:
I don't think the extra restrictions are necessary for this patch to solve our immediate problems for JEP 401 integration. I have created https://bugs.openjdk.org/browse/JDK-8388784 to track the future restrictions.

jaikiran:
> Just to Jai's concern, restricting it to a named module mapped to the boot loader will remove the concern with -Xbootclasspath/a and the agent equivalent.

That sounds good to me.

dlsmith-dev (approved):
I like getting this in now and considering further restrictions via JDK-8388784.

(FWIW, if you're patching the bootclasspath, you're in hacked JDK territory, and unspecified behavior can occur. I'm not sure of the details, but I think it makes sense to apply whatever standard we use to enforce "internal only" for other JDK annotations. No need to go further. But let's sort that out later.)

AlanBateman (approved):
Okay for now.

jaikiran (approved):

liach:
Thanks for your reviews to unblock JEP 401!

/integrate

PR #2664, opened 2026-07-23
8388848: [lworld] ConnectionGraph::process_call_arguments does not correctly handle wide arguments

TobiHartmann:
As described by https://github.com/openjdk/valhalla/pull/2079, EA processes non-inlined call arguments using BCEA information to add argument and return edges to the ConnectionGraph based on their escape state. `DomainIterator` maps scalarized `CallNode` arguments back to their corresponding Java signature arguments and return projections.

With mixed scalarized calling conventions, that code associates returned arguments with incorrect return projections and triggers an assertion. The problem is that the code confuses JVMS slot indices with Java signature indices. The JVMS contains an additional `Type::HALF` slot for `long` and `double`, while `Method::is_scalarized_arg()` expects Java signature indices.

The new test disables preloading so that `Unloaded` is still unloaded when the callee methods are linked, while the already-loaded `Integer` arguments are scalarized. In the `(long/double, Integer, Unloaded, Integer)` signature, `Unloaded` is
  signature argument 2 but JVMS slot 3. The old code therefore queried signature argument 3 - the scalarized trailing Integer - and incorrectly treated the buffered `Unloaded` argument as scalarized. The preceding scalarized `Integer` supplied a stale field offset, causing an incompatible return projection and the assertion.

The fix tracks Java signature indices separately. I also tried to add a few comments to explain the purpose of the iterator fields.

@rwestrel Since the original `DomainIterator` code is from [JDK-8377480](https://bugs.openjdk.org/browse/JDK-8377480), could you please take a look?

Thanks,
Tobias

---------

---------
### Progress

### Issue
 * [JDK-8388848](https://bugs.openjdk.org/browse/JDK-8388848): [lworld] ConnectionGraph::process_call_arguments does not correctly handle wide arguments (**Bug** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2664/head:pull/2664` \
`$ git checkout pull/2664`

`$ git checkout pull/2664` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2664/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2664`

`$ git pr show -t 2664`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2664.diff">https://git.openjdk.org/valhalla/pull/2664.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/opto/escape.cpp
+++ b/src/hotspot/share/opto/escape.cpp
@@ -2155,13 +2155,13 @@ bool ConnectionGraph::add_final_edges_unsafe_access(Node* n, uint opcode) {
 // (Object, Object), (MyValue, int), (MyValue, float), (int, int)
 class DomainIterator : public StackObj {
 private:
-  const TypeTuple* _domain;               // Domain of the Java signature
+  const TypeTuple* _domain;               // Domain of the JVM signature
   const TypeTuple* _domain_cc;            // Domain of the scalarized calling convention
   const GrowableArray<SigEntry>* _sig_cc; // Entries of the scalarized calling convention
 
-  uint _i_domain;        // Java signature domain index (long/double take two slots)
+  uint _i_domain;        // JVM signature domain index (long/double take two slots)
   uint _i_domain_cc;     // Scalarized calling convention domain index
-  int _i_arg;            // Java signature argument index (long/double take one argument)
+  int _i_arg;            // JVM signature argument index (long/double take one argument)
   int _i_sig_cc;         // Scalarized calling convention SigEntry index
   uint _depth;           // Scalarized value nesting depth
   uint _first_field_pos; // Scalarized calling convention domain index of the first non-hidden value field
--- a/src/hotspot/share/opto/escape.cpp
+++ b/src/hotspot/share/opto/escape.cpp
@@ -2152,19 +2152,28 @@ bool ConnectionGraph::add_final_edges_unsafe_access(Node* n, uint opcode) {
 // }
 // void m(Object o, MyValue v, int i)
 // produces the pairs:
-// (Object, Object), (Myvalue, int), (MyValue, float), (int, int)
+// (Object, Object), (MyValue, int), (MyValue, float), (int, int)
 class DomainIterator : public StackObj {
 private:
-  const TypeTuple* _domain;
-  const TypeTuple* _domain_cc;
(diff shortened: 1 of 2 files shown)


src/hotspot/share/opto/escape.cpp
```diff
@@ -2152,19 +2152,28 @@ bool ConnectionGraph::add_final_edges_unsafe_access(Node* n, uint opcode) {
...
+  const TypeTuple* _domain_cc;            // Domain of the scalarized calling convention
+  const GrowableArray<SigEntry>* _sig_cc; // Entries of the scalarized calling convention
 
-  uint _i_domain;
-  uint _i_domain_cc;
-  int _i_sig_cc;
-  uint _depth;
-  uint _first_field_pos;
-  const bool _is_static;
+  uint _i_domain;        // Java signature domain index (long/double take two slots)
```
merykitty: This should be JVM signature instead.

src/hotspot/share/opto/escape.cpp
```diff
@@ -2152,19 +2152,28 @@ bool ConnectionGraph::add_final_edges_unsafe_access(Node* n, uint opcode) {
...
+  uint _i_domain_cc;     // Scalarized calling convention domain index
```
merykitty, line 2163: How many slots does a `long` take here?

src/hotspot/share/opto/escape.cpp
TobiHartmann: Right, fixed!

src/hotspot/share/opto/escape.cpp
TobiHartmann: A `long` takes two slots (also mentioned in the comment above):
https://github.com/openjdk/valhalla/blob/f181286389fad995be1e71de60f30d14eb1c9122/src/hotspot/share/utilities/globalDefinitions.cpp#L259

TobiHartmann:
Thanks for looking at this Quan Anh!

src/hotspot/share/opto/escape.cpp
TobiHartmann: The new test therefore also covers long and double.

TobiHartmann:
Re-opened for mainline: https://github.com/openjdk/jdk/pull/32131

PR #2670, opened 2026-07-27
8389089: [lworld] C1 should not initialize value class when accessing flat field or array

TobiHartmann:
C1 eagerly allocates a buffer when loading a flat field or array element. If the value is null and its value class is still uninitialized, this allocation incorrectly triggers class initialization, potentially with side effects via the static initializer that can be observed as incorrect result (good catch by a fuzzer generated test, credits to @lepestock). See the new regression test for more details.

I first tried to fix this with a simple compilation bailout. However, if the field / array only ever contains null, we'll stay in the interpreter forever, which could lead to a performance issue if the null load/store is on the hot path.

So for atomic flat loads with an uninitialized value class, C1 now loads the complete payload and checks its null marker without allocating. If null, we just return null. If non-null, we deoptimize and re-compile once the interpreter took care of class initialization.

Thanks,
Tobias

---------

---------
### Progress

### Issue
 * [JDK-8389089](https://bugs.openjdk.org/browse/JDK-8389089): [lworld] C1 should not initialize value class when accessing flat field or array (**Bug** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2670/head:pull/2670` \
`$ git checkout pull/2670`

`$ git checkout pull/2670` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2670/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2670`

`$ git pr show -t 2670`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2670.diff">https://git.openjdk.org/valhalla/pull/2670.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/valhalla/pull/2670#issuecomment-5092146078)
</details>

### DIFF
--- a/src/hotspot/share/c1/c1_GraphBuilder.cpp
+++ b/src/hotspot/share/c1/c1_GraphBuilder.cpp
@@ -1107,16 +1107,20 @@ void GraphBuilder::load_indexed(BasicType type) {
       set_pending_load_indexed(dli);
       return; // Nothing else to do for now
     } else {
-      NewInstance* buffer = new NewInstance(elem_klass, state_before, false, true);
-      buffer->set_null_free(true);
-      _memory->new_instance(buffer);
-      result = append_split(buffer);
       load_indexed = new LoadIndexed(array, index, length, type, state_before);
-      load_indexed->set_buffer(buffer);
-      // The LoadIndexed node will initialize this instance by copying from
-      // the flat field.  Ensure these stores are visible before any
-      // subsequent store that publishes this reference.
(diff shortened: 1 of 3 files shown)


TobiHartmann:
Accidentally opened the PR and now converted back to draft. This will only be integrated into mainline once Valhalla is in.

TobiHartmann:
Re-opened for mainline: https://github.com/openjdk/jdk/pull/32133

PR #2673, opened 2026-07-28
8389234: [lworld] Bad cast for delayed access through abstract value class

TobiHartmann:
C1 delays some flat field and array reads to fuse subsequent getfield operations into a single load. The code currently assumes that the next field’s holder is a `ciInlineKlass` and uses its `payload_offset()`. However, if the field holder is an abstract value class, it is represented as `ciInstanceKlass`, triggering an assert.

The fix restricts the optimization to fields whose holder is a concrete inline class. A more complete alternative would track the concrete flattened receiver class in the delayed-access state and calculate offsets relative to that layout. I quickly prototyped this but it's much more complicated and I'm therefore leaving this to future work.

Thanks,
Tobias

---------

---------
### Progress

### Issue
 * [JDK-8389234](https://bugs.openjdk.org/browse/JDK-8389234): [lworld] Bad cast for delayed access through abstract value class (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2673/head:pull/2673` \
`$ git checkout pull/2673`

`$ git checkout pull/2673` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2673/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2673`

`$ git pr show -t 2673`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2673.diff">https://git.openjdk.org/valhalla/pull/2673.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/c1/c1_GraphBuilder.cpp
+++ b/src/hotspot/share/c1/c1_GraphBuilder.cpp
@@ -1092,12 +1092,15 @@ void GraphBuilder::load_indexed(BasicType type) {
       bool is_null_free = array_klass->is_elem_null_free();
       bool will_link;
       ciField* next_field = s.get_field(will_link);
-      bool next_needs_patching = !next_field->holder()->is_initialized() ||
+      ciInstanceKlass* next_holder = next_field->holder();
+      bool next_needs_patching = !next_holder->is_initialized() ||
                                  !next_field->will_link(method(), Bytecodes::_getfield) ||
                                  PatchALot;
       bool needs_atomic_access = array_klass->is_elem_atomic();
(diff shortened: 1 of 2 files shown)


merykitty (approved):
I would assume we just need to compute `next_holder` using the type of the current tos, but it seems more complicated than that.

TobiHartmann:
Thanks for the review Quan Anh! Yes, I think it's more complicated than that.

TobiHartmann:
Re-opened for mainline: https://github.com/openjdk/jdk/pull/32135

PR #2675, opened 2026-07-28
8389214: [lworld] StoreFlatNode with TOP value input not correctly handled by IGVN

TobiHartmann:
Late inlining of `TestDeadStoreFlat::box` exposes an unconditional throw causing the returned `Integer` `InlineTypeNode` to be replaced by `TOP`. Depending on the IGVN worklist order, it might then process the dead but still reachable `StoreFlatNode` with now `TOP` value input. Calling `value()` from `StoreFlatNode::Value` then asserts because it's not an `InlineTypeNode`:
https://github.com/openjdk/valhalla/blob/f181286389fad995be1e71de60f30d14eb1c9122/src/hotspot/share/opto/inlinetypenode.hpp#L229

The fix is to return `TOP` if the value input is `TOP`. I also added an assert to check the invariant that the input is always an `InlineTypeNode` otherwise.

Thanks,
Tobias

---------

---------
### Progress

### Issue
 * [JDK-8389214](https://bugs.openjdk.org/browse/JDK-8389214): [lworld] StoreFlatNode with TOP value input not correctly handled by IGVN (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2675/head:pull/2675` \
`$ git checkout pull/2675`

`$ git checkout pull/2675` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2675/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2675`

`$ git pr show -t 2675`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2675.diff">https://git.openjdk.org/valhalla/pull/2675.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/opto/inlinetypenode.cpp
+++ b/src/hotspot/share/opto/inlinetypenode.cpp
@@ -2287,10 +2287,12 @@ const Type* LoadFlatNode::Value(PhaseGVN* phase) const {
 }
 
 const Type* StoreFlatNode::Value(PhaseGVN* phase) const {
+  Node* val = in(TypeFunc::Parms + 2);
   if (phase->type(in(TypeFunc::Control)) == Type::TOP || phase->type(in(TypeFunc::Memory)) == Type::TOP ||
-      phase->type(base()) == Type::TOP || phase->type(ptr()) == Type::TOP || phase->type(value()) == Type::TOP) {
+      phase->type(base()) == Type::TOP || phase->type(ptr()) == Type::TOP || phase->type(val) == Type::TOP) {
     return Type::TOP;
   }
+  assert(val->is_InlineType(), "must be InlineTypeNode: %s", val->Name());
   return bottom_type();
 }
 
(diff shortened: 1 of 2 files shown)


lepestock:
The original gentest does not appear to fail on windows-x64-debug, while it does fail on the other debug platforms. The proposed fix and the accompanying test are not OS-specific, which is concerning. Probably worth double-checking.

marc-chevalier (approved):
Good.

TobiHartmann:
> The original gentest does not appear to fail on windows-x64-debug, while it does fail on the other debug platforms. The proposed fix and the accompanying test are not OS-specific, which is concerning. Probably worth double-checking.

@lepestock This is not surprising and most likely due to different scheduling / timing on Windows which affects profiling etc. which in turn affects inlining / graph shape etc. We often see this with C2 bugs like this.

TobiHartmann:
Thanks for the review Marc!

TobiHartmann:
Re-opened for mainline: https://github.com/openjdk/jdk/pull/32137

PR #2677, opened 2026-07-29
8388709: [lworld] replay parts of JDK-8350865

marc-chevalier:
This addresses a few TODOs left in Valhalla... But it also makes replay file printing nicer (less random spaces, empty lines...), and slightly nicer to parse. Which is a good news because now, we have more Java test for replay files! The new test is trying to be smart: after dumping the original replay file, it replays the compilation and dumps a second replay file for the same method. Then, these files are parsed, sanity-checked, and compared. Textual comparison wouldn't work since replay files contain unstable data, such as addresses.

This was motivated by the fixes included in this PR: it's easy to misparse the replay file, but then the reconstituted internal state will not be the desired one. And it is reasonably likely that the second replay file will show that. It doesn't protect very well against misdumping tho. But for that, since we know what to expect and we now can parse replay files on the Java side, we can check whether some lines exist and contain what we expect. It would really be quite unlucky if the interesting lines are consistently the same, and what we expect, while parsing and/or printing are wrong.

I do not aim here at providing a total sanity check for replay files, it probably can be extended. Just targeted pieces that I deemed risky in this PR, and to give example of what can be done. Otoh, comparison of replay files is probably closer from the most precise possible.

Thanks,
Marc

---------

---------
### Progress

### Issue
 * [JDK-8388709](https://bugs.openjdk.org/browse/JDK-8388709): [lworld] replay parts of JDK-8350865 (**Sub-task** - P4)

### Reviewers
 * [Tobias Hartmann](https://openjdk.org/census#thartmann) (@TobiHartmann - Committer) ⚠️ Review applies to [adc93493](https://git.openjdk.org/valhalla/pull/2677/files/adc9349302af526d1b1c052bff5f86f710d5ff67)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2677/head:pull/2677` \
`$ git checkout pull/2677`

`$ git checkout pull/2677` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2677/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2677`

`$ git pr show -t 2677`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2677.diff">https://git.openjdk.org/valhalla/pull/2677.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/ci/ciInstanceKlass.cpp
+++ b/src/hotspot/share/ci/ciInstanceKlass.cpp
@@ -863,8 +863,21 @@ void StaticFieldPrinter::do_field_helper(fieldDescriptor* fd, oop mirror, bool i
       break;
     }
     case T_ARRAY:  // fall-through
-    case T_OBJECT:
-      if (!fd->is_null_free_inline_type()) {
+    case T_OBJECT: {
+      InstanceKlass* k = nullptr;
+      ResetNoHandleMark rnhm;
+      Thread* THREAD = Thread::current();
+      SignatureStream ss(fd->signature(), false);
+      Symbol* name = ss.as_symbol();
+      assert(!HAS_PENDING_EXCEPTION, "can resolve klass?");
+      InstanceKlass* holder = fd->field_holder();
+      if (field_type == T_OBJECT) {
+        k = SystemDictionary::find_instance_klass(THREAD, name, Handle(THREAD, holder->class_loader()));
+        // handling of null free inline type
+        guarantee(k != nullptr && !HAS_PENDING_EXCEPTION, "can resolve klass?");
+      }
+
+      if (k == nullptr || !k->is_inline_klass()) {
         _out->print("%s ", fd->signature()->as_quoted_ascii());
         oop value =  mirror->obj_field_acquire(fd->offset());
         if (value == nullptr) {
@@ -886,6 +899,23 @@ void StaticFieldPrinter::do_field_helper(fieldDescriptor* fd, oop mirror, bool i
           _out->print("%d", a->length());
           if (value->is_objArray()) {
             objArrayOop oa = (objArrayOop)value;
+            if (value->is_flatArray()) {
+              FlatArrayKlass* klass = ((flatArrayOop)oa)->klass();
+              LayoutKind lk = klass->layout_kind();
+              _out->print(" flat");
+              if (LayoutKindHelper::is_atomic_flat(lk)) {
+                _out->print(" atomic");
+              } else {
+                _out->print(" non-atomic");
+              }
+              if (LayoutKindHelper::is_nullable_flat(lk)) {
+                _out->print(" nullable");
+              } else {
+                _out->print(" null-free");
+              }
+            } else {
+              _out->print(" ref");
+            }
             const char* klass_name  = value->klass()->name()->as_quoted_ascii();
             _out->print(" %s", klass_name);
           }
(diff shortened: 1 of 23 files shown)


TobiHartmann (approved):
Very nice! This looks good to me.

Maybe @dean-long could take a look as well since he worked on ci-replay before.

src/hotspot/share/ci/ciInstanceKlass.cpp
```diff
@@ -929,7 +951,7 @@ void ciInstanceKlass::dump_replay_instanceKlass(outputStream* out, InstanceKlass
   if (ik->is_hidden()) {
     const char *name = CURRENT_ENV->dyno_name(ik);
     if (name != nullptr) {
-      out->print_cr("instanceKlass %s # %s", name, ik->name()->as_quoted_ascii());
+      out->print_cr("instanceKlass%s # %s", name, ik->name()->as_quoted_ascii());
```
dean-long: I'm confused by this change.

marc-chevalier: It is not important change, but `dyno_name` always starts with a space inserted here
https://github.com/openjdk/valhalla/blob/12e87d09246d366cc74cb409e4c914351256e98b/src/hotspot/share/ci/ciEnv.cpp#L231
called from
https://github.com/openjdk/valhalla/blob/12e87d09246d366cc74cb409e4c914351256e98b/src/hotspot/share/ci/ciEnv.cpp#L1539-L1542
or
https://github.com/openjdk/valhalla/blob/12e87d09246d366cc74cb409e4c914351256e98b/src/hotspot/share/ci/ciEnv.cpp#L1560

Without this change, we get in the replay file
```
instanceKlass  @cpi ...
```
with a double space, which triggers my formatting perfectionist brain.

dean-long: OK, but I'd rather fix the extra space than have the print code assume it exists.

marc-chevalier: I can revert that. Fixing it in https://github.com/openjdk/valhalla/blob/12e87d09246d366cc74cb409e4c914351256e98b/src/hotspot/share/ci/ciEnv.cpp#L225-L234 is more risky: a lot of code seems to expect the space before as well. For instance, if I understand correctly, I don't think this would work correctly without the space before
https://github.com/openjdk/valhalla/blob/12e87d09246d366cc74cb409e4c914351256e98b/src/hotspot/share/ci/ciEnv.cpp#L1333-L1339

There are 21 usages of `RecordLocation`, including the `@bci` and `@cpi` ones, and the example just above, and most seem not correct without the leading space (all but the `@bci` and `@cpi`). So I think it is fine to assume the space exists. But again, it's not a big deal in this PR, and I can just revert this piece.

marc-chevalier: I could also make another ctor of `RecordLocation` that will not put the leading space to be used only with `@cpi` and `@bci`. Or not put a space when `_dyno_name` is empty maybe. Might be the best option actually.

marc-chevalier: I've pushed this solution. Let me know what you think. I've also fixed the comment above `RecordLocation` (that (even more) wrong before).

marc-chevalier:
Moved to https://github.com/openjdk/jdk/pull/32147

PR #2680, opened 2026-07-29
8389341: [lworld] C2 Class.isAssignableFrom intrinsic returns false for identical array Class arguments

TobiHartmann:
Since [JDK-8366705](https://bugs.openjdk.org/browse/JDK-8366705), an object array Java type is represented by an **_unrefined_** array Klass. Actual array objects use a **_refined_** array Klass describing their concrete layout/properties. All refined variants share the same Java Class mirror, which refers to the unrefined Klass.

C2’s `SubTypeCheckNode` normally compares a refined sub-Klass loaded from an object header against an unrefined Java-level super-Klass. For fast pointer comparisons, it converts an exact array superclass to its refined form. This is correct for callers such as instanceof, casts, and arraycopy:
https://github.com/openjdk/valhalla/blob/f181286389fad995be1e71de60f30d14eb1c9122/src/hotspot/share/opto/graphKit.cpp#L3052-L3057

The `Class.isAssignableFrom` intrinsic is different though: both klasses are loaded from Class mirrors and are therefore unrefined. Above code then nevertheless refines the super-klass for the direct comparison while leaving the sub-klass unrefined. The pointers differed even when both represented the same Java array type, causing the intrinsic to incorrectly return false.

The intrinsic now compares the two unrefined klasses loaded from the mirror before invoking the general subtype helper. Identical Java types return true directly, while unequal types continue through the existing hierarchy-based subtype check.

Thanks,
Tobias

---------

---------
### Progress

### Issue
 * [JDK-8389341](https://bugs.openjdk.org/browse/JDK-8389341): [lworld] C2 Class.isAssignableFrom intrinsic returns false for identical array Class arguments (**Bug** - P2)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2680/head:pull/2680` \
`$ git checkout pull/2680`

`$ git checkout pull/2680` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2680/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2680`

`$ git pr show -t 2680`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2680.diff">https://git.openjdk.org/valhalla/pull/2680.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/opto/library_call.cpp
+++ b/src/hotspot/share/opto/library_call.cpp
@@ -4686,6 +4686,7 @@ bool LibraryCallKit::inline_native_subtype_check() {
                                 // {P,P} & superc!=subc => false
     _prim_same_path,            // {P,P} & superc==subc => true
     _prim_1_path,               // {N,P} => false
+    _ref_same_path,             // {N,N} & superk==subk => true
     _ref_subtype_path,          // {N,N} & subtype check wins => true
     _both_ref_path,             // {N,N} & subtype check loses => false
     PATH_LIMIT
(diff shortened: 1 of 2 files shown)


TobiHartmann:
Re-opened for mainline: https://github.com/openjdk/jdk/pull/32138

PR #2685, opened 2026-07-30
8389391: [lworld] Return buffer allocation for late-inlined MH calls should not initialize the class

TobiHartmann:
During late MethodHandle inlining, C2 preallocates a return buffer when the inlined target returns a scalarized value object but the caller expects an oop. This allocation can initialize the value class even when the target only returns null and the buffer is never used.

Similar to https://github.com/openjdk/valhalla/pull/2670, let's just assert that the result is null if the value class is uninitialized and deopt if we observe a non-null return value.

Thanks,
Tobias

---------

---------
### Progress

### Issue
 * [JDK-8389391](https://bugs.openjdk.org/browse/JDK-8389391): [lworld] Return buffer allocation for late-inlined MH calls should not initialize the class (**Bug** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2685/head:pull/2685` \
`$ git checkout pull/2685`

`$ git checkout pull/2685` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2685/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2685`

`$ git pr show -t 2685`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2685.diff">https://git.openjdk.org/valhalla/pull/2685.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/opto/callGenerator.cpp
+++ b/src/hotspot/share/opto/callGenerator.cpp
@@ -27,6 +27,7 @@
 #include "ci/ciMemberName.hpp"
 #include "ci/ciMethodHandle.hpp"
 #include "ci/ciObjArray.hpp"
+#include "ci/ciStreams.hpp"
 #include "classfile/javaClasses.hpp"
 #include "compiler/compileLog.hpp"
 #include "oops/accessDecorators.hpp"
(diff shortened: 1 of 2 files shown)


TobiHartmann:
Re-opened for mainline: https://github.com/openjdk/jdk/pull/32140

PR #2687, opened 2026-07-30
8389130: [lworld] Compiled frame walking is slow due to scalarized calling convention handling

marc-chevalier:
The problem is that Valhalla made `frame::sender` a tiny bit bigger, but it is enough for gcc to decide not to inline it anymore. And then, functions such as `void java_lang_Throwable::fill_in_stack_trace(Handle throwable, const methodHandle& method, TRAPS)` have a call, instead of having it inlined all the way to `intptr_t* frame::repair_sender_sp(intptr_t* sender_sp, intptr_t** saved_fp_addr) const`. While there are a lot of paths, and its job is a bit subtle `frame frame::sender_for_compiled_frame(RegisterMap* map) const` is mostly doing rather basic arithmetic: a call can easily create some slowdown, Especially in cases where walking the stack is on a hot path.

So, to address this, I propose a 2-part solution: trying to shrink the size of `sender`, and marking it as `ALWAYS_INLINE`. To shrink the size, I cache whether the method needs stack repairs in its flags rather than recomputing it, since it is a constant. It doesn't seem to make it as small as the mainline version, but quite closer. This change is not satisfactory by itself because, even if it were enough to reenable inlining (which it is not), we are so clone from the inlining threshold that the least change in the future could make it bad again. Since we know we are not far from inlining, force inlining the method cannot be too bad.

It is to note that because it is about codesize and inlining problem, the performance issue appear without `--enable-preview`. Trying to branch on that just makes the code bigger and the problem worse.

Valhalla:
```
Time (mean ± σ):      9.429 s ±  0.098 s    [User: 9.274 s, System: 0.171 s]
Range (min … max):    9.246 s …  9.548 s    10 runs
```
Mainline:
```
Time (mean ± σ):      7.957 s ±  0.082 s    [User: 7.817 s, System: 0.152 s]
Range (min … max):    7.778 s …  8.051 s    10 runs
```
This PR:
```
Time (mean ± σ):      8.107 s ±  0.077 s    [User: 7.943 s, System: 0.183 s]
Range (min … max):    7.967 s …  8.218 s    10 runs
```
We can see the changes for the performances of mainline and this PR overlap. There is still a gap, but of 0.15s, that is about 2% instead of 18.5%. It seems the rest of the slowdown is mostly happening during startup.

For comparison, here are the timing for the given test, with only the flag fix, without forcing inlining:
```
Time (mean ± σ):      9.326 s ±  0.075 s    [User: 9.164 s, System: 0.179 s]
Range (min … max):    9.188 s …  9.422 s    10 runs
```
That is only marginally (but consistently) better than Valhalla's current state, but clearly not good enough.

Thanks,
Marc

---------

---------
### Progress

### Issue
 * [JDK-8389130](https://bugs.openjdk.org/browse/JDK-8389130): [lworld] Compiled frame walking is slow due to scalarized calling convention handling (**Bug** - P3)

### Reviewers
 * [Tobias Hartmann](https://openjdk.org/census#thartmann) (@TobiHartmann - Committer) ⚠️ Review applies to [838c5283](https://git.openjdk.org/valhalla/pull/2687/files/838c5283bb904fdb0ae033df79bcfff56cc4c2ef)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/valhalla.git pull/2687/head:pull/2687` \
`$ git checkout pull/2687`

`$ git checkout pull/2687` \
`$ git pull https://git.openjdk.org/valhalla.git pull/2687/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2687`

`$ git pr show -t 2687`

</details>

<a href="https://git.openjdk.org/valhalla/pull/2687.diff">https://git.openjdk.org/valhalla/pull/2687.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/code/nmethod.hpp
+++ b/src/hotspot/share/code/nmethod.hpp
@@ -259,17 +259,19 @@ public:
 
     enum : uint8_t {
       UNSAFE_ACCESS = 1 << 0,
-      WIDE_VECTORS  = 1 << 1,
-      MONITORS      = 1 << 2,
-      SCOPED_ACCESS = 1 << 3
+      WIDE_VECTORS = 1 << 1,
+      MONITORS = 1 << 2,
+      SCOPED_ACCESS = 1 << 3,
+      NEEDS_STACK_REPAIR = 1 << 4,
     };
 
     Flags() : _bits(0) {}
-    Flags(bool has_unsafe_access, bool has_wide_vectors, bool has_monitors, bool has_scoped_access) :
+    Flags(bool has_unsafe_access, bool has_wide_vectors, bool has_monitors, bool has_scoped_access, bool needs_stack_repair) :
       _bits((has_unsafe_access ? UNSAFE_ACCESS : 0) |
-            (has_wide_vectors  ? WIDE_VECTORS  : 0) |
-            (has_monitors      ? MONITORS      : 0) |
-            (has_scoped_access ? SCOPED_ACCESS : 0))
+            (has_wide_vectors ? WIDE_VECTORS : 0) |
+            (has_monitors ? MONITORS : 0) |
+            (has_scoped_access ? SCOPED_ACCESS : 0) |
+            (needs_stack_repair ? NEEDS_STACK_REPAIR : 0))
     {}
 
     // May fault due to unsafe access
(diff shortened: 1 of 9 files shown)


src/hotspot/share/code/nmethod.hpp
```diff
@@ -283,6 +285,9 @@ class nmethod : public CodeBlob {
 
     // Used by shared scope closure (scopedMemoryAccess.cpp)
     bool has_scoped_access() const { return (_bits & SCOPED_ACCESS) != 0; }
+
+    // Stack has been extended and needs reair. See comment in MacroAssembler::remove_frame
```
TobiHartmann:
```suggestion
    // Stack has been extended and needs repair. See comment in MacroAssembler::remove_frame
```

TobiHartmann (approved):
Looks good to me. Someone from runtime should take a look at this as well though.

marc-chevalier:
Moved to: https://github.com/openjdk/jdk/pull/32148

